PR #62, opened 2016-02-03
Importer fixed to process watch point reports correctly.

mrakitin:
Hello Rob, Paul,

Here is a fixed parser for a watch point report. I will continue to implement reformatting of the parser/importer.

Thanks,
Maksim


robnagler:
Would you add a test case for this?

mrakitin:
Yes, sure. I'll post a message here when it's done.

PR #86, opened 2016-02-24
New SRW beamline scripts importer, LCLS and textbook interference examples

mrakitin:
Hi Rob & Paul,

Please review the code to be merged. Here is a short overview of the changes:
- New SRW virtual beamline scripts importer implemented. This functionality involves changes in `sirepo/server.py`, `sirepo/importer.py` and the updated `srwlib.py` file - please grab my file from https://github.com/mrakitin/srw_sirepo_utils/blob/master/parser/srwlib.py.
- Updated LCLS examples (`sirepo/package_data/srw_examples/lcls-sxr-simplified.json` and `sirepo/package_data/srw_examples/lcls-sxr.json`).
- Two new textbook examples for Young's double slit experiment in visual light (`sirepo/package_data/srw_examples/lens-slit-obstacle.json`, `sirepo/package_data/srw_examples/slit-obstacle.json`, `sirepo/package_data/static/json/srw-examples.json`), they are not finalized and under review at the moment, updates are possible.

Please let me know if you have any questions or concerns regarding the suggested changes.

Regards,
Maksim Rakitin


sirepo/server.py
```diff
@@ -294,12 +294,24 @@ def app_find_by_name(simulation_type, application_mode, simulation_name):
...
     try:
-        srwblParam, appParam, el, pp = sirepo.sirepo_parser.sirepo_parser(f.read())
-        args = sirepo.sirepo_parser.list2dict(srwblParam)
-        v = sirepo.sirepo_parser.Struct(**args)
-        args = sirepo.sirepo_parser.list2dict(appParam)
-        app = sirepo.sirepo_parser.Struct(**args)
-        data = sirepo.sirepo_parser.parsed_dict(v, app, el, pp)
+        from werkzeug import secure_filename
+        filename = secure_filename(f.filename)
+        imported_file_path = os.path.join('./', filename)
```
robnagler: We shouldn't store relative paths. Look at 
[app_upload_file](https://github.com/radiasoft/sirepo/blob/master/sirepo/server.py#L552). 
This code needs to be factored out. Possibly need to have an "imports" directory to avoid collisions. Use py.local and pkio.mkdir_parent_only() so that you know the directory exists.

Also, we don't want to grab just any old name. They could be rather long. We should generate the saved name, and store the actual name in a comment in the upload or the generated json.

sirepo/server.py
```diff
@@ -294,12 +294,24 @@ def app_find_by_name(simulation_type, application_mode, simulation_name):
...
+        try:
+            f.save(imported_file_path)
+        except:
+            raise Exception('File <{}> cannot be saved.'.format(filename))
```
robnagler: Do not catch the exception here. Rather just let it fall through to the outer exception. You'll have the arguments to generate a proper error message.

sirepo/server.py
```diff
@@ -294,12 +294,24 @@ def app_find_by_name(simulation_type, application_mode, simulation_name):
...
+
+        # Here we provide path to mirror_1d.dat .dat file and use it for each element requiring profile as placeholder:
+        lib_dir = str(_simulation_lib_dir('srw'))
```
robnagler: Do not hardwire strings. This should either be passed in.

This brings up the point of the file name. This should be in an "importer" (just like template) directory and it should be named "srw_python.py", because it imports python only.

sirepo/server.py
```diff
@@ -294,12 +294,24 @@ def app_find_by_name(simulation_type, application_mode, simulation_name):
...
+        # Perform initialization of the object and get "v" parameters:
+        o = sirepo.importer.SRWParser(imported_file_path, lib_dir=lib_dir)
+
+        # Main SRW calculation and conversion to JSON:
+        o.to_json()
+        data = o.json_content
+
     except Exception as e:
         return flask.jsonify({
             'error': 'File import failed: {}'.format(e)
```
robnagler: This is a potential security hole. Consider this python file:

```python
raise Exception(os.popen('id; cat /etc/passwd').read())
```

This would allow an attacker to do all kinds of nefarious things and get the output. While we can't prevent the code from doing bad things, we can prevent people from probing output. We should understand the types of errors we get, parse them, and then return them to the user. You can dump the entire exception to the error log so that we can help people debug the problems if the software doesn't understand them.

mrakitin:
Hi Rob,

I tried to implement all your suggestions. Please have a look at https://github.com/mrakitin/sirepo/commit/daeed355ad80ac2133fe1a815ac776c0df01dc3d when you have time and let me know if any other places should be fixed.

Thanks,
Maksim

PR #130, opened 2016-03-22
Minor README touchups

nsipplswezey:
After going through the readme on a fresh machine, these are some recommendations.

Minor stuff that likely doesn't come up on machines already configured for development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,23 +3,24 @@
 # THIS IS ALPHA SOFTWARE
 
 Sirepo is a scientific application framework, currently for particle accelator codes.
-The codes run inside a Python web server.
+Sirepo runs inside a Python web server.
 
-Learn more at sireop.com.
+Learn more at sireop.com(coming soon).
 
 Documentation: http://rssynergia.readthedocs.org/en/latest/
 
 ### Vagrant Install
 
-You can start Sirepo with Vagrant. First create a Vagrantfile:
+You can start Sirepo with [Vagrant](https://www.vagrantup.com/).
+
+First create a Vagrantfile:
 
 ```bash
-cat > Vagrantfile <<'EOF'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,23 +3,24 @@
...
 ### Vagrant Install
 
-You can start Sirepo with Vagrant. First create a Vagrantfile:
+You can start Sirepo with [Vagrant](https://www.vagrantup.com/).
+
+First create a Vagrantfile:
 
 ```bash
-cat > Vagrantfile <<'EOF'
+//Vagrantfile
```
robnagler: Thanks, Nick! I'm not familiar with `//Vagrantfile` syntax. How does it work?

nsipplswezey: Actually, the syntax doesn't work!
Just tested it and it throws a syntax error. It was intended to be a comment.

Vagrant files are actually parsed as ruby, so it should be:

# Vagrantfile

It's totally minor, but in how-to step-by-step documentation, I include the file name at the top of the snippet as a comment. That way you can easily write sequential snippets that jump between files following the logical flow of the thinking and instructions. It's usefully concise. Totally minor in this case, but it's a reflex now.

It's something I picked up while drafting some blog content in objective-C where every class has a .h and .m file, and coherent step-by-step iterative development in one file usually requires changes in the other.

Nick

robnagler: Ah, I thought it was some bash-thing I didn't know about.

The former code (which I'll reinstate) actually creates the file if you put it into the shell. 

The confusing part may be that I don't begin the line with `$` to indicate it is a shell command. The reason I do that is that it makes it hard to copy and paste the text.

nsipplswezey: Good point.

Nick

PR #707, opened 2017-03-28
pkconfig formatted values going away

robnagler:
I think this fix is fine, but it could break cookies so I want to make sure I get it right. I want to get rid of [formatted values](https://github.com/radiasoft/pykern/commit/3b9be7133ea7b2ae884b41507e481e7dcacc1707#diff-6a7e51d1079ccbc0ababe499d109774a) in pkconfig. I wanted to be able to parse a value as JSON, and that conflicts with the formatting syntax. I don't think it is that valuable of a feature. And is easy enough to work around.


mrakitin:
Looks good to me.

PR #748, opened 2017-05-20
Touched all animationArgs so needs review

robnagler:
I ran all the apps. animationArgs is now parsed in template_common.parse_animation_args(). Interface isn't ideal, but would be better if model driven off the schema. Didn't have time for more changes.

### DIFF
--- a/sirepo/package_data/static/html/elegant-visualization.html
+++ b/sirepo/package_data/static/html/elegant-visualization.html
@@ -64,7 +64,7 @@
     </div>
     <div class="clearfix"></div>
     <div data-ng-repeat="item in visualization.outputFiles" class="col-md-6 col-xl-4">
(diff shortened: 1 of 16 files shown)


mrakitin:
What is the advantage of the new implementation?

robnagler:
The animationArgs was an unversioned message encapsulation. Now it has a version so we can write code that [supports changes to the args](https://github.com/radiasoft/sirepo/commit/df4bc1ee214c6ac8d3dceac887aa0a9900b24c4b#diff-eafbb1a15d5b0b44d9b7995cb3b450c3R277) without breaking clients.

I do think the animationArgs should be in the schema, but I didn't have time for that. 

Another issue is that the format for the encapsulation should change to be a bit more robust. We a lot of different separation characters:

* underscore '_' for animationArgs
* dash '-' for modelKeys
* asterisk '*' for frame args

By encapsulating the underscore in two places, it will be easier to change, which I am guessing will happen at some point.

mrakitin:
Thanks for the explanation Rob.

Issue #744, opened 2017-05-17
rename apps: Warp-PBA and Warp-VND


robnagler:
@bruhwiler writes:
I suggest the present Warp app should be renamed to "Warp-PBA", with or without the dash.
I suggest the Fete app should be renamed to "Warp-VND" with or without the dash.

robnagler:
On alpha

PR #761, opened 2017-05-31
Fix (bigger half) of #744: rename Warp to Warp PBA

robnagler:
This is a rather large change so I would like people to try it out on their dev systems. Make sure you have tested a Warp example with some values before checking out this change. Then run:

```
sirepo db upgrade
sirepo service http
```

The upgrade is idempotent. It renames any warp directories to warppba. If there are no warp directories, does nothing. You must run the upgrade with all servers (sirepo and celery) stopped.

The fixup_old_data renames simulationType: warp to simulationType: warppba. It does not rename the examples. Those have been renamed (Laser Pulse and Electron Beam) for new users. 

There is a permanent redirect for /warp to /warppba. This can cause confusion if you go back to your old code after testing this branch. I think permanent is correct. We probably have some links floating out there.

I didn't fix the client to handle the rename seamlessly. A refresh of the browser window should work (except perhaps in Safari, haven't tried).  We have very few warp users so I think this is a reasonable compromise.


sirepo/pkcli/db.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def upgrade():
+    """Upgrade the database"""
+    from pykern import pkio
+    from sirepo import simulation_db
+    from sirepo import server
+
+    server.init()
+    to_rename = []
```
mrakitin: Is this list used anywhere?

robnagler: Nope. Good catch. :)

I'll leave it in, because it's harmless, and the upgrade is ephemeral.

moellep:
I tried it with my warp simulations and they were renamed correctly. I didn't see any problems when running the app with the new sim name on Firefox or Safari.

robnagler:
If Alpha is stable, I'd like to push that to Beta, run the purge, and then merge this branch. I can do this tomorrow. OK?

mrakitin:
It looks like the previous sets of updates didn't have an incremented version at some point. This caused empty pages with header for existing users. It looks like this update will fix the problem.

mrakitin:
As for the alpha/beta update, it's fine for me. Let us know when alpha is ready for testing. I'll give it a try.

Issue #815, opened 2017-07-19
fete: conductor must intersect grid


moellep:
A conductor element will be ignored by the solver if it doesn't intersect the grid. If a conductor is moved or dropped and doesn't intersect the grid, show an error message such as “Objects that do not intersect the grid will be ignored by the solver.”  or “Please move the object so that it intersects the grid."

PR #823, opened 2017-08-01
Fix for issue #815

mkeilman:
Conductor shapes now become empty with a red dashed outline to indicate that they will not be seen by the solver. Mouseover also provides a message to that effect.  Any change to the geometry (moving, resizing, changing grid or conductor properties) is reflected in the next call to drawShapes().

Since the grid's z values in the model do not usually correspond to the vertical lines drawn in the view, whether they intersect a conductor can be visually unclear. This does not appear to matter in practical use, but I invite discussion on the matter...


sirepo/package_data/static/js/warpvnd.js
```diff
@@ -571,6 +570,44 @@ SIREPO.app.directive('conductorGrid', function(appState, panelState, plotting) {
...
+            function doesShapeCrossGridLine(shape) {
+                var numX = appState.models.simulationGrid.num_x;  // number of vertical cells
+                var halfChannel = toMicron(appState.models.simulationGrid.channel_width/2.0);
+                var cellHeight = toMicron(appState.models.simulationGrid.channel_width / numX);  // height of one cell
+                var numZ = appState.models.simulationGrid.num_z;  // number of horizontal cells
+                var cellWidth = toMicron(appState.models.simulationGrid.plate_spacing / numZ);  // width of one cell
+                if( cellHeight == 0 || cellWidth == 0 ) {  // pathological?
+                    return true;
+                }
+                else {
```
robnagler: I know this gives you the heebie geebies, but the return true means you don't need the else. This means you can remove the else clauses, shortening the function and removing indentation, which makes the code more readable imiho.

sirepo/package_data/static/js/warpvnd.js
```diff
@@ -562,7 +561,7 @@ SIREPO.app.directive('conductorGrid', function(appState, panelState, plotting) {
                     .enter().append('rect')
                     .on('dblclick', editPosition)
                     .attr('class', function(d) {
-                        return d.voltage > 0 ? 'warpvnd-shape warpvnd-shape-voltage' : 'warpvnd-shape';
+                        return !doesShapeCrossGridLine(d) ? 'warpvnd-shape warpvnd-shape-noncrossing' : (d.voltage > 0 ? 'warpvnd-shape warpvnd-shape-voltage' : 'warpvnd-shape');
```
robnagler: Long tertiary ops should be split over multiple lines.

sirepo/package_data/static/js/warpvnd.js
```diff
@@ -571,6 +570,44 @@ SIREPO.app.directive('conductorGrid', function(appState, panelState, plotting) {
...
+                var numX = appState.models.simulationGrid.num_x;  // number of vertical cells
+                var halfChannel = toMicron(appState.models.simulationGrid.channel_width/2.0);
+                var cellHeight = toMicron(appState.models.simulationGrid.channel_width / numX);  // height of one cell
+                var numZ = appState.models.simulationGrid.num_z;  // number of horizontal cells
+                var cellWidth = toMicron(appState.models.simulationGrid.plate_spacing / numZ);  // width of one cell
+                if( cellHeight == 0 || cellWidth == 0 ) {  // pathological?
+                    return true;
+                }
+                else {
+                    if( shape.height >= cellHeight || shape.width >= cellWidth ) {  // shape always crosses grid line if big enough
```
robnagler: Looks good to me. Didn't check the math. I'll leave that to @moellep. I'll just comment on the insubstantial stuff. :)

mkeilman: Cool, I'll add your suggestions

moellep:
I recently discovered the d3.classed() method which can simplify assigning classes, so instead of this:
```
  .attr('class', function(d) {
    return !doesShapeCrossGridLine(d) 
      ?  'warpvnd-shape warpvnd-shape-noncrossing' 
      : (d.voltage > 0 ? 'warpvnd-shape warpvnd-shape-voltage' : 'warpvnd-shape');
  })
```

You can have this:
```
  .attr('class', 'warpvnd-shape')
  .classed('warpvnd-shape-noncrossing', function(d) {
    return !  doesShapeCrossGridLine(d);
  })
  .classed('warpvnd-shape-voltage', function(d) {
    return d.voltage > 0;
  })
```

The d3.classed() turns on a css class based on a boolean expression. That allows the static css classes to be set with attr() and dynamic ones with classed(). It is a little easier to read, and the static class 'warpvnd-shape' doesn't need to get included in the other values.

moellep:
Looks good - if you want to apply @robnagler suggestions and resubmit the request, I'll merge it.

Issue #809, opened 2017-07-19
Don't allow slash (/) in filename


robnagler:
Any other magic file chars shouldn't be allowed.

robnagler:
I did a quick check of names on beta. I don't think there any '/' or other specials like &, :, \, etc. so we can restrict these to avoid unnecessarily complexity in downloaded file names.

mkeilman:
I've been working on this and was wondering: should we keep offending characters from ever appearing in the input field? Also, I presume we need to inform the user of the restrictions somehow.

moellep:
You could add a new field type "FileName" to the fieldEditor directive (in sirepo-components.js) and add a custom validator, similar to the stringToNumber directive used by Float and Integer types. The validator would return undefined if the value has invalid characters such as '/' and that will cause the field to highlight the error. Experiment with the stringToNumber directive to get an idea.

mkeilman:
Good, I have been taking that approach

PR #841, opened 2017-08-10
Fix for issue #809

mkeilman:
This leverages the existing field validation with a "SafePath" type - each application schema would require an update (only warpvnd is included here). Also displays a warning message when the user enters invalid text, details of which merit further discussion

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -4,6 +4,20 @@ var srlog = SIREPO.srlog;
 var srdbg = SIREPO.srdbg;
 
 SIREPO.NUMBER_REGEXP = /^\s*(\-|\+)?(\d+|(\d*(\.\d*)))([eE][+-]?\d+)?\s*$/;
+SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS = [   // array of illegal filename characters
+    '\\\\',  // doubly escaped
+    '\/',
+    '\&',
+    ':',
+    '+',
+    '\?',
+    '\"',
+    '\'',
+    '<',
+    '>',
+    '\|'
+];
+SIREPO.SAFE_PATH_REGEXP_EXCLUDE = new RegExp("["+SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS.join('')+"]+", "");  // regex
 
 SIREPO.app.directive('advancedEditorPane', function(appState, $timeout) {
     return {
@@ -317,6 +331,10 @@ SIREPO.app.directive('fieldEditor', function(appState, panelState, requestSender
               '</div>',
               '<div data-ng-switch-when="String" data-ng-class="fieldClass">',
                 '<input data-ng-model="model[field]" class="form-control" required />',
+              '</div>',
+              '<div data-ng-switch-when="SafePath" data-ng-class="fieldClass">',
+                '<input data-safe-path="" data-ng-keypress="acceptChar($event)" data-ng-model="model[field]" class="form-control" required />',
+                '<div class="sr-input-warning" data-ng-show="showWarning">{{warningText}}</div>',
               '</div>',
               '<div data-ng-switch-when="InputFile" class="col-sm-7">',
                 '<div data-file-field="field" data-model="model" data-model-name="modelName" data-empty-selection-text="No File Selected"></div>',
@@ -884,6 +902,38 @@ SIREPO.app.directive('msieFontDisabledDetector', function(errorService, $interva
     };
 });
 
+SIREPO.app.directive('safePath', function() {
+    return {
+        restrict: 'A',
+        require: 'ngModel',
+        link: function(scope, element, attrs, ngModel) {
+
+            scope.warningText = scope.info[0] + ' must not include any of the following: ' +
+                SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS.join(' ').replace(/\\\\/, '\\');
+
+            scope.acceptChar = function(event) {
+
+                var isInvalidChar = SIREPO.SAFE_PATH_REGEXP_EXCLUDE.test(String.fromCharCode(event.keyCode));
+                if(isInvalidChar) {
+                    event.preventDefault();  // don't display the character in the text box
+                }
+                scope.showWarning = isInvalidChar;
+            }
+            ngModel.$validators.safe = function(value) {
+                var hasInvalidChars = SIREPO.SAFE_PATH_REGEXP_EXCLUDE.test(value);
+                scope.showWarning = hasInvalidChars;
+                return !hasInvalidChars;
+            }
+
+            scope.$on('cancelChanges', function(e, data) {  // broadcast by root scope
+                scope.showWarning = false;
+                scope.warningText = "";
+            });
+
+        },
+    };
+});
+
 SIREPO.app.directive('numberToString', function() {
     return {
         restrict: 'A',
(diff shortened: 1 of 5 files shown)


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4,6 +4,20 @@ var srlog = SIREPO.srlog;
 var srdbg = SIREPO.srdbg;
 
 SIREPO.NUMBER_REGEXP = /^\s*(\-|\+)?(\d+|(\d*(\.\d*)))([eE][+-]?\d+)?\s*$/;
+SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS = [   // array of illegal filename characters
```
robnagler: We generally don't do "side comments". In this particular case, the name is clear so the comment is unnecessary.

I would say "UNSAFE_PATH_CHARS" for a name. I understand the long name, but it's overdone in this case.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4,6 +4,20 @@ var srlog = SIREPO.srlog;
...
+    ':',
+    '+',
+    '\?',
+    '\"',
+    '\'',
+    '<',
+    '>',
+    '\|'
+];
+SIREPO.SAFE_PATH_REGEXP_EXCLUDE = new RegExp("["+SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS.join('')+"]+", "");  // regex
```
robnagler: Side comment is unnecessary. Put spaces around operators.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4,6 +4,20 @@ var srlog = SIREPO.srlog;
 var srdbg = SIREPO.srdbg;
 
 SIREPO.NUMBER_REGEXP = /^\s*(\-|\+)?(\d+|(\d*(\.\d*)))([eE][+-]?\d+)?\s*$/;
+SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS = [   // array of illegal filename characters
+    '\\\\',  // doubly escaped
```
robnagler: Side comment is unnecessary.

I would avoid too much escaping. It would be clearer to do this, and I'll demonstrate usage later:

```javascript
SIREPO.UNSAFE_PATH_CHARS = '\\/&:+?\'"<>|'.split('');
SIREPO.UNSAFE_PATH_WARN = must not include any of the following: '
    + SIREPO.UNSAFE_PATH_CHARS.join(' ');
SIREPO.UNSAFE_PATH_RE = new.RegExp('[\\' + SIREPO.UNSAFE_PATH_CHARS.join('\\') + ']');
```

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -884,6 +902,31 @@ SIREPO.app.directive('msieFontDisabledDetector', function(errorService, $interva
...
 });
 
+SIREPO.app.directive('safePath', function() {
+    return {
+        restrict: 'A',
+        require: 'ngModel',
+        link: function(scope, element, attrs, ngModel, $sce) {
+
+            scope.showWarning = false;
+            scope.warningText = scope.info[0] + ' must not include any of the following: ' +
```
robnagler: I computed this once. I don't think it wants to be computed every time so rather:

```javascript
scope.warningText = '';
```

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -884,6 +902,31 @@ SIREPO.app.directive('msieFontDisabledDetector', function(errorService, $interva
...
+            ngModel.$validators.safe = function(value) {
+                var hasInvalidChars = SIREPO.SAFE_PATH_REGEXP_EXCLUDE.test(value);
+                scope.showWarning = hasInvalidChars;
+                return !hasInvalidChars;
+            }
+
+            scope.$on('cancelChanges', function(e, data) {  // broadcast by root scope
+                scope.showWarning = false;
+                ngModel.$setViewValue('');
+                ngModel.$render();  // appears that invalid text will not be cleared unless we do this
```
robnagler: Put this comment above the line as it is important. I would consider encapsulating this, e.g.

```
function clearViewValue(model) {
    model.$setViewValue('');
    model.$render();
}
```

Seems like this would be a general problem we have when we cancel, even resetting all fields.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -884,6 +902,31 @@ SIREPO.app.directive('msieFontDisabledDetector', function(errorService, $interva
...
+        restrict: 'A',
+        require: 'ngModel',
+        link: function(scope, element, attrs, ngModel, $sce) {
+
+            scope.showWarning = false;
+            scope.warningText = scope.info[0] + ' must not include any of the following: ' +
+                SIREPO.SAFE_PATH_REGEXP_EXCLUDE_CHARS.join(' ').replace(/\\\\/, '\\');
+
+            ngModel.$validators.safe = function(value) {
+                var hasInvalidChars = SIREPO.SAFE_PATH_REGEXP_EXCLUDE.test(value);
```
robnagler: Unnecessary intermediate variable:

```
scope.showWarning = SIREPO.UNSAFE_PATH_RE.test(value);
if (scope.showWarning) {
    scope.warningText = scope.info[0] + SIREPO.UNSAFE_PATH_WARN;
}
return !scope.showWarning;

File: sirepo/template/template_common.py [hunk 7]
```diff
@@ -230,7 +230,7 @@ def validate_model(model_data, model_schema, enum_info):
                 'BeamList', 'MirrorFile', 'ImageFile', 'String', 'OptionalString', 'MagneticZipFile',
```
robnagler: This list should be alphabetical and line by line so it can be easily sorted. Otherwise, too easy to get duplicates and/or miss something. Probably would define as a constant.

sirepo/package_data/static/js/sirepo-components.js
mkeilman: I'll put it in the fieldEditor directive

sirepo/template/template_common.py
mkeilman: Should the list include the 'Float' and 'Integer' types, even though they get handled separately?

sirepo/package_data/static/js/sirepo-components.js
mkeilman: I'll take care of all this - I was pondering as I went whether they really need to be globals, since they are likely only to be used in one place, but the existing number regex was there so...

sirepo/template/template_common.py
robnagler: The list is for string types that need to be escaped. Others are parsed differently so should not be in this list.

sirepo/package_data/static/js/sirepo-components.js
robnagler: Always a tricky question. You could use closure variables to accomplish the same thing:

```javascript
SIREPO.app.directive('safePath', function() {
    var unsafe_path_chars = '\\/&:+?\'"<>|'.split('');
    ...
```

Ask @moellep what he prefers. :)

moellep:
I think it would be good to keep it private to the directive unless it
needs to get used by other modules. The NUMBER_REGEXP used to be private,
but was moved out when elegant.js also needed it.

moellep:
I think the list of types should be removed and replaced with

else:
  model_data[k] = _escape(value)

and also remove the "unknown field" exception. Each app is going to have
its own list of special types and the template_common.py shouldn't require
updating each time with the names.

If type type validation is useful, the template_common.py could be changed
to parse the schema for the app to identify the full list of types.

Issue #716, opened 2017-04-07
elegant: improve source page to include non-gaussian definition


moellep:
The elegant source page does not cover all the options in the bunched_beam command. The fields can be edited from the Control tab, but it would be nice to allow selecting and configuring non-gaussian beams from the source UI as well.

robnagler:
To be clear, we should be able to set all these parameters (Pages 1-4) from the Source page when you click on the pencil from the bunch?

<img width="926" alt="bunch screen" src="https://cloud.githubusercontent.com/assets/5495179/26125721/5ef3ad42-3a40-11e7-9d6a-8ea89f53197e.png">

moellep:
I don't think we want every parameter. But it would be nice to define a
uniform distribution beam, or to control the distribution of one axis. It
would be nice if the fields were displayed in a nice manner - for example,
rather than fields for  distribution_type "gaussian,gaussian,hard-edge"
with a distribution_cutoff "1,2,3", split them into a grid with x,y,z
columns (6 fields instead of 2).

See the bunched_beam description in the manual:
www.aps.anl.gov/asd/oag/manuals/elegant_latest/elegant.pdf

Also look at the bunched beam examples in the elegantExamples.tar.gz file.

moellep:
@mkeilman 
One improvement needed for #716 is to separate out some of the elegant
beam fields. For example, elegant specifies the X/Y/Z distributions in
one comma-separated field. So instead of typing in "gaussian,
gaussian, uniform-ellipse", it should be displayed inline as three
drop-down elements, each with an enum value ("gaussian", "hard-edge",
"uniform-ellipse", ...).

A similar change was made to the InputFileXY field type - elegant
expects the value as a string in the format "<filename>=<x>+<y>", so
the schema separates it into three fields, for example "waveform",
"waveformX" and "waveformY".

Below is the elegant bunched_beam page and you can see several
combined fields such as "Distribution Cuttoff" and Centroid.

<img width="934" alt="screen shot 2017-08-14 at 9 16 05 am" src="https://user-images.githubusercontent.com/6516910/29289235-e525830a-80f8-11e7-9ecc-2253475bd6e2.png">

moellep:
The new Bunch layout now allows setting many new parameters, including the x/y/z distributions, halton sequences and beam position. Below are some examples.

[...]

PR #849, opened 2017-08-17
For Issue #716 - Separated distribution type into 3 drop-down lists

mkeilman:
I worked with the elegant format of a comma-delimited string, so there is a certain amount of dis/re-assembly.  See also the validation in template_common.py 

Labels for the lists are tacked on to the end of the array in the schema

I did not change the pop-up text but it might be worth a look considering the UI change

### DIFF
--- a/sirepo/package_data/static/js/elegant.js
+++ b/sirepo/package_data/static/js/elegant.js
@@ -42,6 +42,9 @@ SIREPO.appFieldEditors = [
     '<div data-ng-switch-when="ValueList" data-ng-class="fieldClass">',
       '<select class="form-control" data-ng-model="model[field]" data-ng-options="item as item for item in model[\'valueList\'][field]"></select>',
     '</div>',
+    '<div data-ng-switch-when="DistributionType" data-ng-class="fieldClass">',
+        '<div data-distribution-type="" data-model="model" data-field="model[field]" data-info="info" data-type-list="enum[info[1]]"></div>',
+    '</div>',
 ].join('');
 SIREPO.appDownloadLinks = [
     '<li><a href data-ng-href="{{ dataFileURL(\'csv\') }}">CSV Data File</a></li>',
@@ -2469,6 +2472,35 @@ SIREPO.app.directive('inputFileXY', function() {
             $scope.fieldY = function() {
                 return $scope.field + 'Y';
             };
+        },
+    };
+});
+
+SIREPO.app.directive('distributionType', function() {
+    return {
+        restrict: 'A',
+        scope: {
+            model: '<',
+            field: '=',
+            info: '<',
+            typeList: '<',
+        },
+        template: [
+            '<div data-ng-repeat="defaultSelection in field.split(\',\') track by $index" style="display: inline-block" >',
+                '<span style="font-weight: 500">{{distLabels[$index] || \'Plane \' + $index}}: </span>',
+                '<select ',
+                    'style="width: 10em; margin-right: 1em; margin-bottom: 4px; ',
+                    'class="form-control" data-ng-model="distributions[$index]" data-ng-change="didChange(distributions[$index])"',
+                    'data-ng-options="item[0] as item[1] for item in typeList">',
+                '</select>',
+            '</div>'
+        ].join(''),
+        controller: function($scope) {
+            $scope.distributions = $scope.field.split(/,\s*/);
+            $scope.distLabels = ($scope.info[4] || '').split(/,\s*/);
+            $scope.didChange = function(item) {
+                $scope.field = $scope.distributions.join(', ');
+            }
         },
     };
 });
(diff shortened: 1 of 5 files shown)


sirepo/template/template_common.py
```diff
@@ -207,7 +207,10 @@ def validate_model(model_data, model_schema, enum_info):
             raise Exception('no value for field "{}" and no default value in schema'.format(k))
         if field_type in enum_info:
             if str(value) not in enum_info[field_type]:
-                raise Exception('invalid enum value: {} for {}'.format(value, k))
+                # Check a comma-delimited string against the enumeration
+                for item in re.split(",\s*", str(value)):
```
robnagler: When splitting, you want r'\s*,\s*': the "r" ensures no interpretation of the \, the single quotes are the string form we use by default, the \s* starting is looser.

sirepo/template/template_common.py
```diff
@@ -207,7 +207,10 @@ def validate_model(model_data, model_schema, enum_info):
...
+                    if item not in enum_info[field_type]:
+                        raise Exception('invalid enum value: {} for {}'.format(value, k))
```
robnagler: Exception() is not a good thing. At a minimum, use `assert` or raise AssertionError. Exception is generic so it can't be caught by type. Assertion indicates that you are throwing a program error, not a generic Exception (e.g. IOException subclasses Exception). 

As to the format statement, it should be of the form: "<value>:  <error>", e.g. 

```python
assert item in enum_info[field_type], \
    '{}: invalid enum "{}" value for field "{}"'.format(item, field_type, k)
```

sirepo/package_data/static/js/elegant.js
```diff
@@ -2473,6 +2476,35 @@ SIREPO.app.directive('inputFileXY', function() {
...
+        restrict: 'A',
+        scope: {
+            model: '<',
+            field: '=',
+            info: '<',
+            typeList: '<',
+        },
+        template: [
+            '<div data-ng-repeat="defaultSelection in field.split(\',\') track by $index" style="display: inline-block" >',
+                '<span style="font-weight: 500">{{distLabels[$index] || \'Plane \' + $index}}: </span>',
```
robnagler: This font-weight should be in CSS.

sirepo/package_data/static/js/elegant.js
```diff
@@ -2473,6 +2476,35 @@ SIREPO.app.directive('inputFileXY', function() {
...
+                '<select ',
+                    'style="width: 10em; margin-right: 1em; margin-bottom: 4px; ',
+                    'class="form-control" data-ng-model="distributions[$index]" data-ng-change="didChange(distributions[$index])"',
+                    'data-ng-options="item[0] as item[1] for item in typeList">',
+                '</select>',
+            '</div>'
+        ].join(''),
+        controller: function($scope) {
+            $scope.distributions = $scope.field.split(/,\s*/);
+            $scope.distLabels = ($scope.info[4] || '').split(/,\s*/);
```
robnagler: split /\s*,\s*/ for looser parsing

sirepo/package_data/static/js/elegant.js
```diff
@@ -2473,6 +2476,35 @@ SIREPO.app.directive('inputFileXY', function() {
...
+            model: '<',
+            field: '=',
+            info: '<',
+            typeList: '<',
+        },
+        template: [
+            '<div data-ng-repeat="defaultSelection in field.split(\',\') track by $index" style="display: inline-block" >',
+                '<span style="font-weight: 500">{{distLabels[$index] || \'Plane \' + $index}}: </span>',
+                '<select ',
+                    'style="width: 10em; margin-right: 1em; margin-bottom: 4px; ',
```
robnagler: This should be in CSS, too.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -935,25 +927,33 @@ SIREPO.app.directive('msieFontDisabledDetector', function(errorService, $interva
 });
 
 SIREPO.app.directive('safePath', function() {
+
+    var UNSAFE_PATH_CHARS = '\\/|&:+?\'"<>'.split('');
```
robnagler: these should be lower, camel case. Global constants are upper case, not locals.

mkeilman:
I have implemented the suggestions - I've fetched and merged and pushed so it should be good now...???

Regarding assert, would the preceding Exception also be a candidate for that?

robnagler:
Fix #716: Separate out DistributionType into independent fields

robnagler:
I don't think we should ever "raise Exception". `assert` should be used when we don't really know how to continue, and ValueError or something like it should be used if it is to be caught and parsed. I don't think we should refactor the code now, because we have quite a few raise Exceptions.

robnagler:
The build failed: https://travis-ci.org/radiasoft/sirepo/builds/266144476

```text
sirepo/package_data/static/js/elegant.js: line 2502, col 14, Missing semicolon.
sirepo/package_data/static/js/sirepo-components.js: line 394, col 14, Missing semicolon.
sirepo/package_data/static/js/sirepo-components.js: line 950, col 14, Missing semicolon.
```

I recommend running test.sh before checking in.

Issue #636, opened 2017-01-22
back button consistent


robnagler:
The back button should work on pages in the simulation library (and elsewhere). We should include the folder path in the anchor.

PR #858, opened 2017-08-31
Fix for Issue #636

mkeilman:
Main features here: new routing with a folder path parameter, file tree persistence, and recovery of the stored path from the URL


sirepo/package_data/static/js/sirepo.js
```diff
@@ -1907,13 +1963,35 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
         self.activeFolder = item;
         self.activeFolderPath = [];
         while (item) {
             self.activeFolderPath.unshift(item);
             item = item.parent;
         }
         appState.setActiveFolderPath(self.pathName(self.activeFolder));
+        appState.setActiveFolder(self.activeFolder);
+        if( prevPath !== self.pathName(self.activeFolder) ) {
+            var compoundPath = pathToCompoundPath(self.pathName(self.activeFolder));
```
robnagler: pathToCompoundPath can return '/' so it seems if you want to check if it is root folder first before calling compoundPath.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2147,4 +2225,47 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
         appState.saveQuietly('simulationFolder');
     });
     loadList();
+
+    // invoked in loadList() callback
+    function checkURLForFolder() {
+
+        var pathNoSectionRE = new RegExp('^\/' + activeSection.getActiveSection());
```
robnagler: This shouldn't be a regular expression. getActiveSection needs to be escaped if it is, but more importantly, it's literal match below.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2147,4 +2225,47 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
+                appState.setActiveFolderPath(canonicalPath);
+                appState.setActiveFolder(newFolder);
+                self.openItem(newFolder);
+            }
+            else {
+                requestSender.localRedirect('notFound');
+            }
+        }
+        else {
+            if( !appState.getActiveFolder() ) {
```
robnagler: Seems like this check needs to be first. The root case would eliminate the getActiveFolderPath check (|| '/') above.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2147,4 +2225,47 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
+                self.openItem(rootFolder());
+            }
+        }
+    }
+    function folderForPathInList (folderPath, folderList) {
+        for( var i = 0; i < folderList.length; ++i ) {
+            var item = folderList[i];
+            if( item.isFolder ) {
```
robnagler: short circuit

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2147,4 +2225,47 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
+                var lPath = self.pathName(item);
```
robnagler: I'm confused about compoundPath vs pathName. Seems like they are similar.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1907,13 +1963,35 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
+                var params = {
+                    ':folderPath?': compoundPath,
+                };
+                requestSender.localRedirect('simulationsFolder', params);
+            }
+        }
+    }
+
+    function pathToCompoundPath(path) {
+        return path.replace(/^\//,'').replace(/\//g, '_');
```
robnagler: I don't understand this replacement. '_' is a legitimate path character so it can't be used as a separator. These two paths are equal as compound paths: '/a_b' and '/a/b'. However, that's incorrect. You have to use a character that is not a valid path character as a separator. Not sure why compound path is needed at all.

moellep:
I think it may be worth separating out the folder functionality into a separate factory now. The appState factory's API grew by 2 private vars and 6 public methods. Making a new factory, like fileManager, would make it easier to follow the code.

sirepo/package_data/static/js/sirepo.js
mkeilman: Right, I meant to ask what to use.  I would advocate for something that does not get url encoded...how about "!" like in old-school email addresses?  Then just add it to the list of forbidden characters

sirepo/package_data/static/js/sirepo.js
```diff
@@ -132,6 +137,16 @@ SIREPO.app.factory('appState', function(errorService, requestSender, requestQueu
     var savedModelValues = {};
     var activeFolderPath = null;
 
+    var appActiveFolder = null;
```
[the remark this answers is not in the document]
mkeilman: For my own sanity so I could keep things straight while I was jumping around in the code, I'll revert it

sirepo/package_data/static/js/sirepo.js
robnagler: Let's not add more invalid chars. We already have a good enough list, e.g. could use ':'. 

I wonder about compoundpath, however. why is it necessary?

mkeilman: I think angular can't find the route if we use "/" -- but I may be misinterpreting your question. BTW I believe ":" will fail for a similar reason (it will be interpreted as a route param) but I'll try it

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1856,35 +1900,47 @@ SIREPO.app.controller('SimulationsController', function (appState, panelState, r
...
-                    self.openItem(showItem.parent);
-                }
-                else {
-                    self.openItem(rootFolder());
-                }
+                checkURLForFolder();
             });
     }
+    // compares simulationID only
+    function isSimInArray(sim, arr) {
```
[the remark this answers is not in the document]
mkeilman: I guess we don't even need the data objects in their entirety; might as well store only the sim id. indexOfSimInArray was there for deletion purposes (as referenced in your other comment) but I forgot to complete that task, bah. Will clean that up

sirepo/package_data/static/js/sirepo.js
mkeilman: Followup - colon is fine as it gets url encoded. I find the resulting URLs hard to parse visually, but barring user feedback to that effect it doesn't matter I suppose

sirepo/package_data/static/js/sirepo.js
mkeilman: They are identical except for the delimiter. I guess pathName is just used to depict an artificial file system on the UI so we could change its delimiter and eliminate compoundPath - but I'd want to confirm that's OK!

sirepo/package_data/static/js/sirepo.js
mkeilman: `            if( !item.isFolder ) {
                continue;
            }
...
`
?

robnagler: Yes, but not one line:

```
if (! item.isFolder) {
    continue;
}
```

Note the spacing, btw. We put a space after the keyword and the paren, no spaces between the paren and the next token, space after `!` and the value it is operating on so it is more visible.

sirepo/package_data/static/js/sirepo.js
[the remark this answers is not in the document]
robnagler: Yes, that's true. I think Paul's advice is the best: separate out in to a factory. The rest may fall out of that.

sirepo/package_data/static/js/sirepo.js
robnagler: Given the way angular works, we might be better off using a search (query) parameter:

```text
/simulations?p=/a/b/c
```

This will produce a readable URL. If angular had the concept of path info, we'd be fine, but it doesn't. Using "search" has the added benefit of not requiring two routes as empty searches work just fine.

robnagler:
@moellep I think you should review. I told @mkeilman we shouldn't spend more time on this unless there are bugs.

moellep:
I've merged in the code and added commit 8ffed6d to move the compoundPath code into fileManager. More of the code in simulationsController could also move to fileManager. I needed to do this so the "Simulations" link would return to the original folder when returning from a simulation to the list.

Issue #485, opened 2016-10-06
Avoid blank pages when switching apps


robnagler:
When you click on a simulation in the calculator, the page switch results in a blank page. If there's latency (we're in Trieste), it stays up for a couple of seconds.

moellep:
A good way to simulate this is Firefox's built-in Responsive Design Mode:

https://blog.nightly.mozilla.org/2016/11/07/simulate-slow-connections-with-the-network-throttling-tool/

I found I could reproduce the problem by selecting "Amazon Kindle Fire" along with "Regular 2G" settings. Rather than showing a blank page, a "page loading" or similar animation would be nice. It shouldn't show up if the app is loading quickly (wait a few seconds or something similar).

Possibly something similar to http://pathgather.github.io/please-wait/ would be useful.

mrakitin:
Good idea Paul.

mkeilman:
I've pushed an initial try on this to my branch; I don't think worth a pull request yet. Some notes:

I placed a script reference to a new file into the header of index.html, so it gets loaded early. I don't know whether it's better to put the code in directly to avoid a file fetch, or whether it needs to be "minified"

Other than an initial call in the html (so we don't wait for all the angular stuff to load), the show/hide lives in one spot in the nav controller. It has covered every case I've seen, but asynch has tricked me before...

FYI Chrome can also throttle network speed, under its developer tools

I took the liberty of whipping up a quick animated gif of the sirepo logo, just as a demo

moellep:
I like it - and it isn't much code which is good. One thing would be to get rid of the global values and instead put them on the SIREPO variable, something like this:

```
// sirepo-pageLoad.js
SIREPO.pageLoader = function() {
    var pageLoaderTimeout = null;
    var showPageLoader = function () {
        var pageLoader = document.getElementsByClassName('sr-page-load')[0];
        if (pageLoaderTimeout) {
            return;
        }
        console.log('starting timer to show page loader');
        pageLoaderTimeout = setTimeout(function () {
            pageLoader.style.display = 'block';
        }, 1000);
    };
    var hidePageLoader = function () {
        clearTimeout(pageLoaderTimeout);
        pageLoaderTimeout = null;
        var pageLoader = document.getElementsByClassName('sr-page-load')[0];
        pageLoader.style.display = 'none';
    };
    return {
        'show': showPageLoader,
        'hide': hidePageLoader,
    };
}();
```

Then the code in sirepo.js can be simplified to:
```
[...]

Issue #374, opened 2016-09-14
SRW - link for sharing a simulation


mrakitin:
We need to provide a button for sharing a simulation which will give a user the link. It will resolve the ambiguity how to share the simulation. Something like this can be used (from StackOverflow):
![share](https://cloud.githubusercontent.com/assets/13209176/18523035/7704626c-7a80-11e6-8417-d46270512205.png)

robnagler:
Link is there, need popup

PR #875, opened 2017-09-12
Fixes for Issues #485, #374

mkeilman:
Includes a new animated gif file for page loading. Also includes a bug fix for Issue #636 exposed by testing of slow connections (intermittent redirecting to 404)


moellep:
The code works well - just a few comments below.

In sirepo/package_data/static/html/index.html:

```
+      SIREPO.pageLoaderTimeout = null;
+      SIREPO.setPageLoaderVisible = function (doShow) {
+              var pageLoader = document.getElementsByClassName('sr-page-load')[0];
+              if (doShow) {
+                  if (self.pageLoaderTimeout) {
+                    return;
+                  }
+                  self.pageLoaderTimeout = setTimeout(
+                    function () {
+                      pageLoader.style.display = 'block';
+                    },
+                    1000
+                  );
+              }
+              else {
+                clearTimeout(self.pageLoaderTimeout);
+                self.pageLoaderTimeout = null;
+                pageLoader.style.display = 'none';
+              }
+          };

```

The `self.pageLoaderTimeout` should be `SIREPO.pageLoaderTimeout`. `self` actually refers to document.window unless there is a local var which overrides it. Our use of the same name `var self` in the angular code isn't great because of this.

Also, the formatting is a bit wonky - with an 8 char indent followed by 2 or 3 character indents.

You could get rid of the global SIREPO.pageLoaderTimeout if the code was encapsulated in an anonymous function. Something like this:

```
SIREPO.setPageLoaderVisible = function() {
    var pageLoaderTimeout = null;
    return  function (doShow) {
        // insert setPageLoaderVisible() code from above
    };
}();

```

Issue #213, opened 2016-06-15
SRW - autorun animated Flux report for fast cases


mrakitin:
Need to run the animated Flux report automatically without pressing the start button in the case when the calculation is fast, e.g., `Flux Computation Method` is set to `Use Approximate Method` (=`-1`).

For more details about fast/slow calculations please see #189.

moellep:
When "Use Approximate Method" is selected, the report should behave like a normal, short-running report. The start button and elapsed time should not be displayed and the report should auto-run. The simulationStatusPanel directive will need special code to handle this.

PR #885, opened 2017-09-19
Added autorun for Issue #213

mkeilman:
This hides the start/stop buttons and runs the flux simulation automatically if the approximate method is selected. The simulation remains persistent. Still requiring thought is proper handling of queue conflicts in the cancelation process

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1315,20 +1315,22 @@ SIREPO.app.factory('simulationQueue', function($rootScope, $interval, requestSen
         }
     };
 
-    self.cancelItem = function (qi) {
+    // TODO(mvk): handle possible queue state conflicts
+    self.cancelItem = function (qi, successCallback, errorCallback) {
         if (! qi) {
-            return;
+            return false;
         }
         qi.qMode = 'transient';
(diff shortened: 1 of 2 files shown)


moellep:
I've tried it out and it works well. I think we want to avoid re-running the approximate case each time the page is visited, but I understand that the fluxAnimation.method value may get out-of-sync with the value that was used to compute the output file on the server. I think I can fix that case by having the server return the fluxAnimation.method which was used when the client calls runStatus().

Stylistically, this code:

```
            $scope.cancelPersistentSimulation = function () {
                $scope.cancelSimulation(cancelSuccess, cancelError);
            };
            var cancelSuccess = function (data, status) {
                if( $scope.isApproximateMethod() ) {
                    $scope.runSimulation();
                }
            };
            var cancelError = function (data, status) {
                // ignore error
                cancelSuccess(data, status);
            };
```

could be written without the extra variables as this:

```
            $scope.cancelPersistentSimulation = function () {
                var cancelSuccess = function (data, status) {
                    if( $scope.isApproximateMethod() ) {
                        $scope.runSimulation();
                    }
                };
                // ignore error case
                $scope.cancelSimulation(cancelSuccess, cancelSuccess);
            };
```

I'll merge this change in and update the code to avoid auto-running if the server has a computed datafile which matches the selected fluxAnimation.method.

Issue #50, opened 2015-11-30
Setting watchpoint title to blank causes server error when saving


moellep:
If a beamline watchpoint title is cleared and the "Save Changes" is selected, the server will return a "server error" response for all reports.

mrakitin:
That's still an issue.

moellep:
A simple fix for this might be adding the code below to the beamlineBuilder directive.
```
            $scope.$on('beamline.changed', function(evt, name) {
                appState.models.beamline.forEach(function(m) {
                    if (m.type == 'watch') {
                        if (! m.title) {
                            // fix the title before the save is committed
                            appState.setModelDefaults(m, 'watch');
                            appState.saveQuietly('beamline');
                        }
                    }
                });
            });
```
It doesn't solve the general case. For example, a similar error can occur if any beamline item has an invalid field (such as 'x' in a Float field) and the beamline is saved.

A better fix would be to detect if one of the beamline elements is invalid and prevent "Save Changes" from being selected.

A test case would be: Open an empty beamline, drop an aperture and a watchpoint, select the aperture and enter 'x' in the Horizontal Size field (the Close button will appear disabled). Dismiss the aperture editor by clicking on the aperture icon and select the "Save Changes" button.

PR #890, opened 2017-09-21
Fixes for Issue #50

mkeilman:
This forbids saving the beamline with invalid data and highlights the offending item(s). I also added the beginnings of a general-purpose field validator in a new 'utilities' service


mkeilman:
FYI there's a bug with mirror elements that I'm tracking down now...

mkeilman:
The bug (mirrors always invalid) was due to the value of the "MirrorFile" field being blank, which is actually OK. I also see that crystals show as invalid, because a bunch of Float fields default to empty. I presume that is also OK, which is going to require some thought ("OptionalFloat" type?)

moellep:
Let's investigate tomorrow to see if the defaults are occurring on the server and if that can be pushed to the client. Otherwise, the OptionalFloat could be an option.

Issue #888, opened 2017-09-19
Discard changes feature should be available on all apps


moellep:
The "Discard changes" to example simulations which is available in the SRW app should be made available to all apps. Possibly the cog-wheel menu could be added as a general app feature.

moellep:
@mkeilman, in srw.js the appHeader has several features, such as "Discard Changes to Example" which should be moved to sirepo-components.js and made available to all the apps. Possibly, the cog-wheel dropdown could be moved out, with a ng-transclude portion which allows apps to add custom items to the menu. SRW, for example, would have custom menu items for "Simulation Documentation URL" and "Initial Wavefront SImulation Grid". "Open as a New Copy" and "Discard/Delete" could be available in all apps. The "Export JSON Data File" should be replaced with "Export as Zip" and use the same functionality as "Export as Zip" on the simulation list drop-down menu.

![screen shot 2017-09-21 at 10 49 15 am](https://user-images.githubusercontent.com/6516910/30708190-33519250-9ebb-11e7-94d0-cc090f15d657.png)

robnagler:
I think Simulation Documentation URL should be available to all apps, no?

mkeilman:
The schema for Simulation Documentation exists only for srw, so it would have to be added to the other apps (is there a common area for views/models?)

Is this the time to pull some of the other repeated stuff out of the apps and into sirepo-components? I'm thinking of the new sim/folder functions by way of example...

moellep:
Right now there is no common area for view/models - for example, each app schema has a copy of the "simulation" and "simulationFolder" models/views.
Yes, this would be a good time to also move any common code from the apps into sirepo-components.

robnagler:
See also #354

Issue #354, opened 2016-09-10
simulation => sim or s8n


robnagler:
The word "simulation" appears everywhere in the code and is too long. I propose to replace it with sim or s8n (like i18n, k8s, etc.).

moellep:
I prefer sim to s8n. So simulationId would become simId?

Paul

robnagler:
Agreed. I think we would have simId, simType, simSerial, etc. I am also
thinking we want to rename "simulation" to "simMeta" instead of "sim" and
move "models" up a level so that the datastruct would look like:

{
    version: 1,
    simMeta: {
       simId: xx,
       simType: srw,
       simSerial: 999,
       simFolder: '/SR Calculator',
       simDocUrl: "",
       simName: "Bending Magnet Radiation",
       simIsExample: true,
    },
}

I think the rest should go in another model eg

     simParams: {
          appMode: calculator,
          sourceType: 'm',
          ...
     }

simMeta would be defined completely for each simulation. You have to have
all those elements, kind of like a directory entry. simParams would be
something that's application specific.

Given that Javascript is so flat, I think it is import to label entities
like in C with a prefix. I think this also helps Python, because duck
typing is really a global name space (any object could potentially be
checked for a specific attribute name).

What do you think?

mrakitin:
I would vote for `sim` too. What does `s8m` stand for?

robnagler:
s8m was modeled after i18n (internationalization), l10n (localization), k8s
(kubernetes), etc. It's the new rage: s8m => s followed 8 characters then m

mrakitin:
Thanks Rob, it's clear now.

robnagler:
@mkeilman This should be something to do as we go along with refactorings.

robnagler:
Slowly migrating, no need for an issue.

PR #903, opened 2017-10-06
 Refactoring to address Issue #888 and other features common to all apps

mkeilman:
It's a big one. Notes:

I use a fair bit of multi-slot transclusion to allow different custom bits in different places.  It stands out because the angular requires it to use "element" style (as opposed to the "attribute" style we use everywhere else).  I put some usage comments in the apps that don't currently have custom content.

To refactor "new simulation", "new folder", & "documentation url", and to begin to address issue #354, I added "commonViews" and "commonModels" to schema-common.json. In particular "simulation" lives there - it holds the fields every simulation uses and the appState factory folds them in to the local definition (but will not replace them if they exist). Consequently I also removed "name" etc. from each app's schema.

Also, in srw is a fix to address inappropriate auto-running of simulations that do not involve the approximate method (introduced by fix to issue #213).


moellep:
It looks good. I have two minor UI items:

* the cog wheel menu item doesn't appear to have the padding it once did. Compare the screenshot below with the the version on beta:

screen shot 2017-10-09 at 11 43 07 am

* The custom SRW menu item is appear as a link (it underlines when you hover) rather than a menu item (highlights when you hover):

screen shot 2017-10-09 at 11 44 42 am

I think the copyCommonModels() in sirepo.js should be moved to the server in simulation_db.get_schema(). That server code is already doing some of the merging, and you should be able to translate the javascript into python and add it there. The pkcollections.mapping_merge() won't work in this case because it clobbers on collision, but you can do the merging manually as you are now.

robnagler:
I would recommend fixing mapping_merge. It was an example of doing the simple thing at the time. I don't think it would break anything to change it. Easy to check.

I did look at fixing mapping merge, but decided to export pkconfig.flatten_values recently for rsconf. There's some code that could be looked at. Lots of people have merge modules, but I think it would be better to own the dependency in mapping_merge.

mkeilman:
Latest push should correct the UI - I also simplified the settings menu transclusion (flattening out a couple levels of ```<ul>```s). BTW the current beta code squishes away that menu if the viewport is too skinny; that's taken care of here as well

Issue #910, opened 2017-10-11
Boolean toggle UI


moellep:
Boolean values are currently displayed as an enum with two values:

![screen shot 2017-10-11 at 10 57 01 am](https://user-images.githubusercontent.com/6516910/31454517-ef761bb0-ae72-11e7-8642-10d14d4985e7.png)

A toggle switch would be a better choice in this case. Below are a few example implementations (taken from issue #861)

- http://www.bootstraptoggle.com/
- https://www.w3schools.com/howto/howto_css_switch.asp

mrakitin:
Related to #936.

PR #928, opened 2017-10-16
Changes and new files to add toggle switches per Issue #910

mkeilman:
This adds the Bootstrap Toggle plugin and uses the existing Boolean enum to present toggle switches.  They are almost default in appearance but can be changed.  Notes:

For this issue I ignore elegant's "BooleanStringArray" and the SRW propagation parameters, which have their own thing going on.

I added some constants mapping to indexes into various view and model arrays, mostly so I could keep them straight. They have general utility, should they appeal...

Also in this push is a fix to a bug where clicking the simulation name doesn't do anything (the function got moved to a central location)


moellep:
The code and UI looks good. It does cause a memory leak which is common with jquery plugins. I'll check-in a follow up with the ways to reproduce it.

Issue #706, opened 2017-03-17
report colormap selection


moellep:
Add a report setting for selecting the colormap for intensities.

moellep:
Additionally, if the colormap selected is non-grayscale, then a colorbar should be shown which shows the color range.

bruhwiler:
I suggest that the default should be viridis

mrakitin:
Additional discussion was here - https://github.com/radiasoft/sirepo/commit/7a9ce3576b9a830b18775878e683fb8bb6308045. Probably we need to create a separate issue for general colormaps processing for all applications, not just Shadow3.

moellep:
As a first step, the colormap should be selectable on a per-simulation basis. Possibly from a "Report Preferences" link in the cog wheel. Apps would continue to have a default colormap, but the user could override it from the preferences, which could be stored in a model which is common across all apps.

mkeilman:
I'm wondering if setting the colormap per report is reasonable, as it doesn't pertain outside that context. Here it is as part of the cog wheel:
![colormapapp](https://user-images.githubusercontent.com/30352290/31914111-a1ab105a-b806-11e7-9cb3-b05a79531c7e.png)

I've been finding the multiple nested dropdown menus a bit hard to navigate...

Here it is as part of the report's model and view:
![colormapreport](https://user-images.githubusercontent.com/30352290/31914235-1cbea34c-b807-11e7-9493-bffc3bed9a0a.png)

In either case, is the displayed color range in line with what you were imagining?

bruhwiler:
Looking good, Mike.
I think setting the color map independently for each report is the way to go.
I suggest they default to viridis in most cases.
I would put grayscale and viridis at the top of the list.
I see no reason to limit the number of options unless each option requires additional work. Why not list all that are available...?

mkeilman:
Cool - btw it appears that is all of the color maps that sirepo knows about:
```
    var COLOR_MAP = {
        grayscale: ['#333', '#fff'],
        afmhot: colorsFromString('000000...ffffff'),
        coolwarm: colorsFromString('3b4cc0...b40426'),
        jet: colorsFromString('000080...800000'),
        viridis: colorsFromString('440154...fde725'),
    };
```
so I would need to know where to look to expand the list. I presume @moellep is the man.

viridis is set as the default - in addition, elegant currently uses afmhot, srw uses grayscale, and warpvnd uses coolwarm.

mrakitin:
[...]

PR #953, opened 2017-11-07
Added per-report color map for 3d plots, pursuant to Issue #706

mkeilman:
Notes:

Adding UI for simulation-wide defaults turns out to be a heavy task, as per @moellep - it requires a new mechanism to store data outside the simulation folder structure. That could prove useful for other things as well, but was outside the scope of this issue.

I added a 'commonEnums' object to schema-common.json, and put the various color map names in there. Might be a good home for Boolean etc.

The color map applied to a plot is the first one found in 1) the stored model data 2) the default for that report (in the schema) 3) the app-level default as stored in SIREPO.PLOTTING_COLOR_MAP (in the javascript) and 4) a global default SIREPO.DEFAULT_COLOR_MAP (which is currently 'viridis').

Since the reports for srw are pretty complex, I opted to put the color map menu in its own tab, labeled 'Other', for those.

### DIFF
--- a/sirepo/package_data/static/css/sirepo.css
+++ b/sirepo/package_data/static/css/sirepo.css
@@ -256,6 +256,49 @@
         box-shadow: 0 5px 15px rgba(0, 0, 0, .5);
     }
 
+    ul.sr-button-menu {
+        list-style: none;
+    }
+    li.sr-button-menu {
+        padding: 3px;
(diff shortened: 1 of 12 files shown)


moellep:
The changes look good - I've merge them and made a few minor changes for SRW.

PR #984, opened 2017-12-16
Sirepo home page overhaul

mkeilman:
As envisioned by @messamer et al. I expect tweaks will be plentiful.  A couple of notes:

As discussed, there is no way to reach the "light" page directly but it still exists (/light#/home). I applied the same branding changes.  Since they are legacy pages it might be better to back out the changes, but anyway take a look.

Only the SRW page has a video right now.

Buttons to launch codes that Sirepo does not actually contain are commented out.  Most have a Jupyter counterpart.  Genesis goes nowhere and does nothing...


bruhwiler:
I edited the text on all of the pages.
I pushed these changes to the repo, hopefully in a correct way.

For now, I've removed references to GENESIS and to FELs.
These can be restored when GENESIS is running on Jupyter again.

I think the layout/design is fine.
We should not put more time into that, until we have a professional opinion from a graphic artist.

robnagler:
Genesis is running on Jupyter still.  It is not working on the F27 upgrade, which is still in dev.

moellep:
I've released the new homepage to alpha now.

bruhwiler:
My changes the text went into a branch origin/mkeiman-master by user 'github user'.
I think I've fixed my user id in my local git configuration.

I've merged the above branch into master. Hopefully correctly.
After checking that my locally merged instance of master is working correctly, I deleted the branch.

The RadiaSoft notes (written I think by Pepe Barbe) were very helpful, https://github.com/radiasoft/devops/wiki/Git

bruhwiler:
I added FELs and Genesis back, since Genesis is still being supported.
I modified the Jupyter landing page.
I'll create a new issue for @messamer to review the text.

PR #1101, opened 2018-03-14
SRW: fix missing mirror profile file

mrakitin:
This PR fixed testing for existence of the mirror profile file.

Should we raise if the file is not found or just silently proceed with calculation?

### DIFF
--- a/sirepo/package_data/template/srw/parameters.py.jinja
+++ b/sirepo/package_data/template/srw/parameters.py.jinja
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+import os
 try:
     __IPYTHON__
     import sys
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -1626,12 +1626,14 @@ def _height_profile_element(item, propagation, height_profile_el_name, overwrite
     dat_file = str(simulation_db.simulation_lib_dir(SIM_TYPE).join(item['heightProfileFile']))
     dimension = find_height_profile_dimension(dat_file)
 
-    res = '{}ifn{} = "{}"\n'.format(shift, height_profile_el_name, item['heightProfileFile'])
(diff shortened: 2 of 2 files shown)


bruhwiler:
I think if a file is expected, but not found, then the calculation should stop and the user should be notified.

moellep:
This issue came up when mrakitin was demoing running sirepo SRW generated source in a jupyter notebook. The original SRW simulation required a mirror error file, and crashed the jupyter notebook when it was missing. I'll merge this pull request and change it so it shows a better error message in this case. It shouldn't ignore the error, or the user may wonder why the simulation results were different in the notebook compared with the original simulation.

mrakitin:
Thanks @bruhwiler and @moellep for your feedback. Yes, I agree the simulation should not proceed without the required files, but the error should be more user-friendly (ideally testing that file exists and raising should be done at SRW side).

moellep:
I've checked in a change now, and the generate code is similar to this:

>     ifnElMirror1 = "mirror_2d.dat"
>     if ifnElMirror1:
>         assert os.path.isfile(ifnElMirror1), "Missing input file mirror_2d.dat, required by Elliptical Cylinderx beamline element"
>         hProfDataElMirror1 = srwlib.srwl_uti_read_data_cols(ifnElMirror1, "\t")
>

mrakitin:
Great, thanks @moellep! For history: https://github.com/radiasoft/sirepo/commit/ec2d2a456744c815d841cc6a2f135a1c0f9a8995. I hope `Cylinderx` is a typo in your example, I don't see that extra `x` in the code.

moellep:
The "Elliptical Cylinderx" was the name of my beamline element in my test simulation. The generated code would include whatever title the beamline element was given.

mrakitin:
Great, I see, thanks!

Issue #1074, opened 2018-02-14
SRW: create direct links to partially-coherent report page


mrakitin:
It was discussed via emails. The idea is to have a direct link to a partially-coherent simulation, so that it can be shared. This issue should accompany #1070.

moellep:
Also, when sharing a simulation, the target simulation should go to the selected page in the link. For example, if a link to the SRW beamline is clicked and the simulation is copied between sessions, the beamline page should be shown, rather than the source page.

moellep:
Just to clarify - there are two parts to this issue:

- If you are on the SRW partially coherent tab on the Beamline page and bookmark the URL, using the bookmark should display the partially coherent tab, rather than the normal beamline reports.
- If you share a simulation link, the recipient should arrive at the same page. Currently, if you share a simulation link to the Beamline page, the new copy shows the Source page.

Issue #1060, opened 2018-02-01
X-coordinate not fully seen in Intensity reports if FWHM is displayed.


ochubar:
In some cases (at some browser window dimensions), there seem to be not enough room for X, Y, and FWHM values to be shown at the same time, so X-coordinate in partially "cut" from left. Number of displayed decimal digits could be reduced in all these values. Also, these numbers could be made "word-wrapped", so that the information would never be lost, at any browser window / intensity report dimensions.

Issue #1059, opened 2018-02-01
Plots with multiple axes


moellep:
Line plots should be improved to show different y axes on either side of the report, with mouse-over showing the exact values. I think the Strava app does a good job with this. Below is a plot with two different y scales:

<img width="933" alt="screen shot 2018-02-01 at 9 23 27 am" src="https://user-images.githubusercontent.com/6516910/35690035-1eafc63e-0732-11e8-94f1-921d004885a6.png">

When a third y scale is added, the right side is left blank:

<img width="949" alt="screen shot 2018-02-01 at 9 23 47 am" src="https://user-images.githubusercontent.com/6516910/35690105-3f773b72-0732-11e8-851d-410935fc38ed.png">

mrakitin:
I like that!

Issue #1024, opened 2018-01-17
elegant: how do I delete an element from a beamline in the drag/drop pane?


bruhwiler:
It seems like if you drag it away, and then release the mouse, it does get removed from the beamline. It's confusing.

PR #1153, opened 2018-04-10
Progress toward multi-axis plots (#1059, #1060), shareable URLs for SRW partial coherence (#1074)

mkeilman:
Notes:

#1074: The beamline URL now includes a parameter specifying the coherence and can be copied in the usual way.  A user opening the link will be directed to the correct tab.  Included is some route manipulation to reload the whole page only when necessary

#1059, #1060: Mostly a large refactoring effort to support multi-axis plots.  It still allows only but does include the popup report to display the focus point information.  The structure is as follows:

- Overseeing everything is the "plot" directive (2d, 3d,...), as before.  They create the focus points and "delegates" (see below)
- The overlay used for mouse zoom is now a directive that handles clicks etc.  The plot gives it a focus point and delegates via the scope.  A plot can have a number of overlays (3d uses two for example)
- New are "info view" directives of which the popup report is one.  The focus circle has been made one as well.  These display focus point data in different ways.  The plot gives them a focus point and a delegate via the scope
- Delegates have functions like showFocusPointInfo and hideFocusPointInfo that the info views are to override.  The overlay invokes these functions in response to user input
- Focus point creation and management now lives in the focusPointService
- Note that "focus points" now encapsulate data and state, but have been divorced from visual representation

This is a little complicated but should allow for easier management of future focus point visualization

Other details of interest:

- The popup report is available for 2d plots only
- 3d plots still display the focus point info along the top of the plot; however, the text now shrinks to fit the space to address #1060.  I added a little arrow at the end to more clearly indicate which of the two auxiliary plots the text refers to, and a close "x" to clear it
- The popup report is draggable but not resizable
- The plotAxis global function has been moved to the new layoutService ('plotting' being a little overstuffed)


mkeilman:
Commit  f71de50 also includes mouseover delete buttons for elegant beamline elements (Issue #1024)

mkeilman:
Commit 0a7849f completes #1017 and adds notes for simulations (not reports yet) (#963)

PR #1167, opened 2018-04-18
More Py3 fixed - replace basestring & convert a map object to list

mrakitin:
Motivation: extra `basestring`s from https://github.com/radiasoft/sirepo/commit/0a7849fb0e8c899a864908ac55d086c7a954dc2e#diff-40887cde02088c25476cf3326a42dbd3R267 after merge of #1162.

### DIFF
--- a/sirepo/importer.py
+++ b/sirepo/importer.py
@@ -20,7 +20,10 @@ def do_form(form):
     from sirepo import uri_router
     from sirepo import simulation_db
     import base64
-    import StringIO
+    try:
+        import StringIO
+    except:
+        from io import StringIO
 
     if not 'zip' in form:
         raise uri_router.NotFound('missing zip in form')
--- a/sirepo/importer.py
+++ b/sirepo/importer.py
@@ -20,14 +20,20 @@ def do_form(form):
     from sirepo import uri_router
     from sirepo import simulation_db
     import base64
+    '''
     try:
         import StringIO
     except:
-        from io import StringIO
+        import io as StringIO
+    import io
+    '''
+    from six import io
+
+    # from six.io import BytesIO
 
     if not 'zip' in form:
         raise uri_router.NotFound('missing zip in form')
-    data = read_zip(StringIO.StringIO(base64.decodestring(form['zip'])))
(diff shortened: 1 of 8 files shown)


mrakitin:
@robnagler, I fixed the issue for SRW/Py3 with map objects (see 8821726). Do we want to do it for other packages? There are the following occurrences of `map()`:
```py
(sirepo_py3.6) ✔ ~/src/radiasoft/sirepo [replace_basestring|✚ 1]
10:56 $ grep ' map(' * -r | grep '\.py:'
sirepo/template/srw.py:        zdata = map(lambda fn: template_common.zip_path_for_file(mmz.z, fn), mmz.dat_files)
sirepo/template/srw.py:    file_names_in_zip = map(lambda path: os.path.basename(path),  [f for f in zf.namelist() if not f.endswith('/') and f != index_file_name(zf)])
sirepo/template/sdds_util.py:        'values': map(lambda v: _safe_sdds_value(v), values),
sirepo/template/jspec.py:    return map(lambda v: v[0], sorted(res, key=lambda v: v[1]))
sirepo/template/elegant_lattice_parser.py:    models['rpnVariables'] = map(lambda x: { 'name': x, 'value': models['rpnVariables'][x] }, models['rpnVariables'].keys())
sirepo/template/rs4pi.py:    return map(lambda x: float(x), ar)
sirepo/template/rs4pi.py:    return map(lambda x: str(x), ar)
sirepo/pkcli/admin.py:            names = map(
```
The first two are fine since we are looping through them, but I'm not sure about the rest.

robnagler:
`map` is not pythonish. People recommend `for`. Also, you don't need to lambda a single function so:

```py
```

mrakitin:
Yes, I like list comprehension more, +1 for this. That update was after I finished active contributions to Sirepo, not sure why we have such PEP-8 non-compliant one-liners, which are hard to read and support. I'll change template_common and SRW's parts.

mrakitin:
Attn @kalebswartz7, I pushed a few more updates (19cd077) left in my local branch.

Issue #1154, opened 2018-04-11
plot hover-focus with arrow keys is confusing


moellep:
It is confusing that a plot requires hover-focus for the arrow keys to manipulate the selected point. If the plot can determine it has focus through the click event, it should keep the keyboard handler until a different plot/field is selected.

Issue #990, opened 2018-01-01
SRW:  built-in directories should be handled differently than user directories


bruhwiler:
It's confusing for new users.
Should they save new simulations inside the Examples/ directory?
If they do, are these then visible as examples to other users?

I think it would be cleaner if the 'special' directories could not be modified by users.
Maybe a different color could also be used...?
When a user creates a copy, they should be notified that it will be saved in the top-level or they could be forced to use one of their own subdirectories...

PR #1189, opened 2018-05-04
Focus point keyboard input and example folders

mkeilman:
Notes:

#1154: Keyboard manipulation of focus points no longer requires the cursor to be over the plot in question. This comes with other features:
- Multiple plots can be be active, though only one at a time can process input
- The plot currently processing input is highlighted
- Deactivating the the current plot (by mouse or ESC) automatically moves input to the next plot, if any
- Users can TAB between active plots to change which one moves its focus points
- When input fields (for numbers etc.) receive focus they take over receiving the keyboard input if any plots are active.  The plots regain processing after input fields lose focus

#990: Users can no longer create new or move existing simulations or folders inside example folders.  Note that:
- Example folders and simulations in icon view now have a "gear" decoration to suggest their status:

image image

- Example simulations cannot be moved
- Example simulations can still be copied, but the copy must live in a user folder (or root)
- Any items a user previously added to an example folder remain in place
- User-created items have a different color to distinguish them from example items
- Creation of both simulations and folders requires specifying a folder to contain them

Other items:
- Users could move folders into their own child folders, causing an infinite loop. That is no longer an option
- Fixed a bug where double-clicking on a particle plot (such as SRW brightness) threw an error

### DIFF
--- a/sirepo/package_data/static/css/sirepo.css
+++ b/sirepo/package_data/static/css/sirepo.css
@@ -420,6 +420,10 @@
     .sr-hide-report {
       opacity: 0;
     }
+    .sr-panel-active {
+        border-width: 2px;
+        border-color: green;
+    }
 
     /* plot styles */
     .sr-plot .axis text {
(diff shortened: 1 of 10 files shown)


mkeilman:
One other item here: users can now drag-select text in focus point popup reports

Issue #1243, opened 2018-06-30
Change enable_bluesky to bluesky_secret


robnagler:
To make the bluesky interface more secure, I suggest we add a secret shared between the bluesky and sirepo. It would be very easy to do. Just add a known string that's a secret that's shared between the two codes. There would be a blueskySignature on the request which would be verified by the shared secret: 

```python
h = hashlib.sha256()
h.update(cfg.blueksky_secret)
h.update(req.simulationType + req.simulationId)
if 'v1:' + h.hex_digest() == req.blueskySignature:
    ... valid message ...
werkzeug.exceptions.abort(404)
```

The client would be similar. Ideally, this code would be shared on the client and server. (BTW, is there any sharing done now?

robnagler:
I'll implement a backward compatible interface so that you can migrate whenever it is convenient. I do think this is an important security issue.

mrakitin:
Thanks @robnagler, please let me know the new usage pattern.

robnagler:
@mrakitin the basic usage is something like:

```python
req = pykern.pkcollections.Dict(simulationId=xyz, simulationType=srw)
sirepo.bluesky.auth_hash(req) # fills in fields authHash and authNonce
post(req)
```

See bluesky_test.py. You will need to set the variable $SIREPO_BLUESKY_AUTH_SECRET when the server starts and when the client imports sirepo.bluesky. If this variable is not set, will use deprecated variable $SIREPO_SERVER_ENABLE_BLUESKY.

Let me know if you have questions.

robnagler:
@mrakitin have you tested this?

mrakitin:
Will test it today. We still use the old approach. @kalebswartz7, can you try?

mrakitin:
@robnagler, from your example it looks like I have to have Sirepo installed on a client (where Bluesky is run), but we don't want this since we use Python 3.x installation (Bluesky supports Python 3.5+). Any ideas how to run it independently of Sirepo via the provided API?

robnagler:
Does sirepo.bluesky run on Python 3? It is simple. I could maybe move it into PyKern which should be fully Python 3 compatible.

robnagler:
There are two parts: auth_hash and auth_login. I think we might be able to move it to a module in pykern that does the computation with parameters. auth_hash doesn't need to depend on sirepo.util.

mrakitin:
[...]

PR #1276, opened 2018-07-20
API: return schema to an external caller

mrakitin:
@kalebswartz7 and I need schema to retrieve the units for the fields to move Sirepo motors via Bluesky. This PR adds the explained functionality.

### DIFF
--- a/sirepo/server.py
+++ b/sirepo/server.py
@@ -85,7 +85,9 @@ app.config.update(
 
 def api_blueskyAuth():
     from sirepo import bluesky
-    return _json_response(bluesky.auth_login(_json_input()))
+    req = _json_input()
+    sim_type = req['simulationType']
+    return _json_response(bluesky.auth_login(req, schema=simulation_db.get_schema(sim_type)))
 
 
 def api_copyNonSessionSimulation():
--- a/sirepo/bluesky.py
+++ b/sirepo/bluesky.py
@@ -85,7 +85,7 @@ def auth_hash(req, verify=False):
         )
 
 
-def auth_login(req):
+def auth_login(req, **kwargs):
     from sirepo import server
 
     if cfg.auth_secret:
@@ -104,6 +104,7 @@ def auth_login(req):
     return pkcollections.Dict(
         status='OK',
         data=simulation_db.open_json_file(sim_type, sid=sid),
+        **kwargs
     )
 
 


sirepo/server.py
```diff
@@ -85,7 +85,9 @@
 
 def api_blueskyAuth():
     from sirepo import bluesky
-    return _json_response(bluesky.auth_login(_json_input()))
+    req = _json_input()
+    sim_type = req['simulationType']
+    return _json_response(bluesky.auth_login(req, schema=simulation_db.get_schema(sim_type)))
```
robnagler: It would be better to move the response from bluesky (lines 104-108), and just do this in server.py:

```python
req = _json_input()
bluesky.auth_login(req)
return _json_response(
    status='OK',
    data=simulation_db.open_json_file(req.simulationType, sid=req.simulationId),
    schema=simulation_db.get_schema(req.simulationType)),
)
```

This would be less coupling.

Why is "status='OK'" used when the rest of sirepo use "state='ok'"?

mrakitin: Good idea, @robnagler! I'll update the code and test it.

Not sure about `status='OK'`, it was originally there even before your update https://github.com/radiasoft/sirepo/commit/85dfe66742ae14e92ea526a3b4304bce834387be. @moellep was the first who implemented that in https://github.com/mrakitin/sirepo/commit/267fa6f87dd0118e4de475a044b5b4f569cacd4e. Should it be changed to `ok`?

robnagler: Seems like we should share a common "ok". In one case it is state and another it is status. I'm sure it was just an oversight.

I think we should maybe add and argument to _json_response_ok() so that it can merge the state so we don't have to repeat "state='ok'" in the code.

mrakitin: I'm happy with whatever solution you propose regarding `OK/ok`. I'm not sure it should be a part of this PR though.

mrakitin: Btw, should it be `req.simulationType` or `req['simulationType']`? I remember there were some disagreements what we should use.

robnagler:
Implemented with 8d1f2b3914cf9622eaae6b0bf32e23e38e4e5972 including a test, _json_response_ok refactoring, and a test which validates https://github.com/radiasoft/sirepo/issues/1243#issuecomment-406686160 (including finding a bug).

mrakitin:
Ugh, just pushed an update to my branch, didn't notice that PR was closed, but not merged. Why don't you merge? I noticed that in the past. That's much more convenient to have PRs for features than scattered commits. You were welcome to push to my feature branch as it allows updates from maintainers.

Anyway, thanks. I'll check if it works.

robnagler:
It was faster for me to code it directly and merge. I'm sorry if it messed up your branch. Use git reflog to revert.

mrakitin:
No worries. The branch is OK, but it makes the PR "unaccepted" :( and we duplicated our efforts in the same/similar updates.

Issue #1292, opened 2018-08-01
Axes for 3d plots


mkeilman:
Splitting out some of the subtasks from #1260. Rudimentary axes exist but they are not useful for actual work yet

Tasks:
- [x] Dynamically adjust axes to match orientation and position
- [x] Add axis orientation widget
- [x] Recalculate domains when edges of the plot go out of frame
- [x] Refactor axis binding to remove semantic confusion

Issue #1260, opened 2018-07-17
3d particle report


moellep:
From @mkeilman pull request:

- vtk/Fete: I wanted to get something in place for folks to look at. Consequently I left in a lot of commented and far-from-final code. A few features so far:

- The number of particles can be selected just as in 2d. Also, the user can choose to connect every N points to speed up renders

- Users can toggle absorbed and reflected particle tracks, impact points, and conductors

- Zoom and pan are vtk default - double-click to get back to the starting position

- Tracks are colored according to the fieldReport

Tasks:
- [x] Axes #1292
- [x] Adjust spacing of conductors so faces don't bleed through
- [x] Add buttons to jump to straight-on camera positions in each direction
- [x] Allow selection of color map independent of field report
- [x] Research putting all point data into single actor (may make ```joinEvery``` moot)
- [x] Changes relevant to pull req comments

moellep:
I like the particle 3d report - a good start and using vtk as well.

The main issue is performance when the "particles to render" is high and "join every N points" is 1. In some cases it causes firefox or chrome to become unresponsive for a minute or so. We should  estimate a good render time based on the number of points and particle count and set a limit - for example if 100 particles are selected, then maybe the "join every N points" needs to have a lower bounds of 5.

![screen shot 2018-07-17 at 9 38 41 am](https://user-images.githubusercontent.com/6516910/42828105-351f81e8-89a5-11e8-986b-8012763efc85.png)

robnagler:
Very cool @mkeilman!!!

PR #1317, opened 2018-08-24
Vtk plotting, axis progress and refactoring (#1260, #1292)

mkeilman:
Notes:

- #1260: 
   - Created new ```sirepo-plotting-vtk.js``` file as a place to house all direct references to the vtk API.  App files like ```warpvnd.js``` should not need to know anything about vtk directly, including its internal coordinate system; conversely, the plotting does not need to know about the data and its coordinates.  A simple transform mediates the two.
   - Added buttons for predefined rotations
   - Added buttons to switch between "move mode" and "select mode".  Does nothing now but change mouse event handling, but the idea is to allow selection of objects and maybe drawing lines to find relative values of field or whatever between endpoints

- #1292: 
   - Added an axis widget in the corner (on this see more below)
   - Most significantly, d3 axes are now bound to the edges of the "scene" and move freely with it. The domain of the axes is dynamic as the edges get resized and dragged off-screen.  As the tick marks change, grids that match up to them in the 3d space also change (though will probably switch to lines since they don't track quite right).  Please note that all this is in a transitional state: each axis is handled separately and there is a lot of repeated code, but I have also begun the process of abstracting that into the vtkPlotting factory.  That's incomplete but will serve to indicate where my thinking is on it

In order to add the little axis widget, I was obliged to modify the vtk source (very minimally - the code was in place but not reachable). The source is publicly available at their [website](https://kitware.github.io/vtk-js/index.html) via npm. They also use npm to build and [webpack](https://webpack.js.org/) to package their stuff. After installing locally I took these steps:

- Edited ```Sources/Rendering/Core/index.js``` to add imports for the objects of interest (AxesActor and AnnotatedCubeActor)
- Ran ```npm run build:release``` to generate a minified javascript file
- Edited ```webpack.config.js``` to add ```devtool: 'source-map'``` to the ```module.exports``` object
- Ran ```npm run build:debug``` which in concert with the line above generated a .js map file.  This also makes a non-minified .js file.

All that *locally* -- I did not add it to any repository (yet?). The resulting files are part of this pull request


sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-                            width: zfs,
-                            height: zlabel.text().length * zfs
-                        }
-                    }
-                };
-                return g;
+        }
+        // vertical line
+        if(startX == endX) {
+            return {
```
robnagler: This code seems repetitive. I bet we do this other places.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
+                left: [bounds.left, -Infinity],
+                top: [startX, bounds.top],
+                right: [bounds.right, Infinity],
+                bottom: [endX, bounds.bottom]
             };
+        }
 
+        var m = (endY - startY) / (endX - startX);
+        return {
+            left: [bounds.left, startY + m * (bounds.left - startX)],
```
robnagler: the math here is repetitive. It relates to the above comment. Is there a way to modularize the tests in a loop. It seems we maybe want to be using a 2d matrix (dims) not an object for this. That would at least allow a loop.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-                var vpTop = worldCoord.getComputedViewportValue();
-                var vpTopY = vpTop[1] / 2.0;
-
-                var vpWidth = Math.abs(vpRightX - vpLeftX);
-                var vpHeight = Math.abs(vpTopY - vpBottomY);
-                var tanTheta = (vpBottom[1] - vpBottomIn[1]) / (vpBottom[0] - vpBottomIn[0]);
-                var theta = 180 * Math.atan(tanTheta) / Math.PI;
-                $scope.zAxisAngle = theta;
+    this.isPointWithinBounds = function (p, b) {
+        return p[0] >= b.left && p[0] <= b.right && p[1] >= b.top && p[1] <= b.bottom;
```
robnagler: I understand that "top" is bottom on web space, but the math is confusing as a result. 

Function params should speak for themselves: point, bounds. Inside a function, it's fine to use single letter variables.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-                        color: "red"
-                    },
-                    {
-                        x: vpBottom[0] / 2.0,
-                        y: vpBottom[1] / 2.0,
-                        color: "blue"
-                    }
-                ];
+    // Returns the point(s) that have the smallest (reverse == false) or largest value in the given dimension
+    this.extrema = function(pArr, dim, reverse) {
```
robnagler: reverse => wantReverse or isReversed. Since there are no types, it's hard to know what it is, because the value is just passed through.

pArr is points? I don't think we want hungarian notation so labelling it Arr is not clear to me, and adds a cognitive load when the word "Array" is only two characters longer.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
+        var sPArr = self.sortInDimension(pArr, dim, reverse);
+        if(! sPArr) {
```
robnagler: Why is this not an error? I don't understand how a sort could return nothing.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-                    },
-                    {
-                        x: vpBottom[0] / 2.0,
-                        y: vpBottom[1] / 2.0,
-                        color: "blue"
-                    }
-                ];
+    // Returns the point(s) that have the smallest (reverse == false) or largest value in the given dimension
+    this.extrema = function(pArr, dim, reverse) {
+        var sPArr = self.sortInDimension(pArr, dim, reverse);
```
robnagler: I'd just call it "res" or "result" not sPArr. It's just the value cascading through the function. It lets you know that when you say "result" you are operating on the return value right from the start, just performing different ops along the way.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-                        width: $scope.vtkCanvasGeometry().size.width,
-                        height: $scope.vtkCanvasGeometry().size.height
-                    }, select);
+    // Returns the members of an array of edges (point pairs) that
+    // contain any of the points in another array
+    this.edgesWithCorners = function(edgeArr, pArr) {
+        var edges = edgeArr.filter(function (edge) {
+            return edge.some(function (corner) {
+                return pArr.some(function (point) {
+                    return point[0] == corner[0] && point[1] == corner[1];
```
robnagler: I like naming things: isEqual? Maybe even have a point object so you can just call a function on it with an argument.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-
-            function select(selector) {
-                var e = d3.select($scope.element);
-                return selector ? e.select(selector) : e;
+    // Returns edges that fit entirely within the given bounds (if any)
+    this.edgesInsideBounds = function(edges, bounds) {
+        var clippedEnds = [];
+        for(var edge in edges) {
+            var p = edges[edge];
+            if(! self.isPointWithinBounds(p, bounds)) {
```
robnagler: in this case early exit is more code. Reverse the test and call clippedEnds.push(p); inside the if. Seems like this could be some kind for map() operation that just returns clippedEnds e.g. using values() on edges.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
             }
+            clippedEnds.push(p);
+        }
+        return clippedEnds;
+    };
 
+    // Sort (with optional reversal) the point array by the values in the given dimension;
+    // Array is cloned first so the original is unchanged
+    this.sortInDimension = function (pArr, dim, reverse) {
+        if(!pArr || !pArr.length || dim >= pArr[0].length ) {
```
robnagler: seems like this would be an error. I don't understand why null is a reasonable result of a sort. Indeed, if a null is ever returned, the code will bomb later, which is not fail fast. Less code here would make it more reliable, because an error would throw an exception.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3052,992 +3071,110 @@ SIREPO.app.directive('particle', function(plotting, layoutService, utilities) {
...
-            if (vtkService.vtk) {
-                vtkService.vtk().then(function(vtk) {
-                    plotting.vtkPlot(scope, element);
-                });
-            }
-        },
+    // returns the distance bewteen two points
+    this.dist = function(p1, p2) {
+        if(p1.length != p2.length) {
+            return -1;
```
robnagler: -1 for a distance? Shouldn't this be an error?

sirepo/package_data/static/js/sirepo-plotting.js
robnagler: Looking at the use of this:

```js

                var lowestCorners = plotUtilities.extrema(vpCorners, 1, true);
                var leftmostCorners = plotUtilities.extrema(vpCorners, 0, false);
                var highestCorners = plotUtilities.extrema(vpCorners, 1, false);
                var rightmostCorners = plotUtilities.extrema(vpCorners, 0, true);

                // for the z (out-in) axis,
                var bottomofleftmost = plotUtilities.extrema(leftmostCorners, 1, true);
                var leftofbottommost = plotUtilities.extrema(lowestCorners, 0, false);
                var bottomofrightmost = plotUtilities.extrema(rightmostCorners, 1, true);
                var rightofbottommost = plotUtilities.extrema(lowestCorners, 0, true);                var zLeft = [];  var zRight = [];  var index = 0;
                for(index in bottomofleftmost) {
                    zLeft.push(bottomofleftmost[index]);
                }
                for(index in leftofbottommost) {
                    zLeft.push(leftofbottommost[index]);
                }
                for(index in bottomofrightmost) {
                    zRight.push(bottomofrightmost[index]);
                }
                for(index in rightofbottommost) {
                    zRight.push(rightofbottommost[index]);
                }
```

This should be a loop or recursion and the dimensions should be an array. There's an impedance mismatch here between the APIs (extrema, sortInDimension) and the inputs and outputs. The granularity is wrong and the objects are non programmatic. bottomofleftmost has a lot of meaning, but only within the context of a matrix so the API should operate on a matrix and produce a matrix as a result.

moellep:
We were looking to put join the dynamic promises in loadModel(), but that didn't work out because some app dependencies are brought in on the simulations list, such as elegant's custom importer. So all the dependencies need to be available at app start.

PR #1319, opened 2018-08-28
Session with cookies. Removed beaker, added sirepo.beaker_compat to e…

moellep:
…xtract uid from old cookies. Separate server.set_session_user() from server.session_user(). Raise a forbidden error if session_user() is accessed outside of a valid cookie exchange.

### DIFF
--- a/sirepo/server.py
+++ b/sirepo/server.py
@@ -49,7 +49,7 @@ _RUN_STATES = ('pending', 'running')
 _SESSION_KEY_USER = 'uid'
 
 #: Identifies if the session has been returned by the client
-_SESSION_KEY_VALID = 'valid'
+_SESSION_KEY_COOKIE_SENTINEL = 'ok'
 
 #: Parsing errors from subprocess
 _SUBPROCESS_ERROR_RE = re.compile(r'(?:warning|exception|error): ([^\n]+?)(?:;|\n|$)', flags=re.IGNORECASE)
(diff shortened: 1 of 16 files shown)


sirepo/server.py
```diff
@@ -697,6 +689,18 @@ def all_uids():
     return oauth.all_uids(app)
 
 
+@app.after_request
```
robnagler: Don't use a decorator. I don't like binding a dynamic variable to a static module value. During init just call: app.after/before_request(flask_before_request), which just registers the function.

sirepo/server.py
```diff
@@ -697,6 +689,18 @@ def all_uids():
...
+def app_after_request(response):
+    if re.search(r'^(2|3)', str(response.status_code)):
```
robnagler: I think this is clearer:
if 200 <= response.status_code <  400:

sirepo/server.py
```diff
@@ -697,6 +689,18 @@ def all_uids():
...
+        flask.session[_SESSION_KEY_VALID] = 1
```
robnagler: You don't want to change the cookie if is already is valid. I think it's a bit confusing to use the word "valid", but rather something like "cookie sentinel". It's only used for the case when we don't have this value so the logic looks like:

flask.session.setdefault('_SESSION_KEY_COOKIE_SENTINEL', 1)

The sentinel can just be "ok" in the cookie itself.

sirepo/server.py
```diff
@@ -732,75 +738,35 @@ def javascript_redirect(redirect_uri):
...
     """
-    environ = kwargs.get('environ', None)
-    session = environ.get(_ENVIRON_KEY_BEAKER) if environ else flask.session
-    if args:
-        session[_SESSION_KEY_USER] = args[0]
-        _wsgi_app.set_log_user(args[0])
-    res = session.get(_SESSION_KEY_USER)
-    if not res and kwargs.get('checked', True):
-        raise KeyError(_SESSION_KEY_USER)
+    res = flask.session.get(_SESSION_KEY_USER)
```
robnagler: What about the case where we want to create a user. It needs to assert that the cookie sentinel is there.

sirepo/util.py
```diff
@@ -9,6 +9,11 @@
 import werkzeug.exceptions
 
 
+def raise_forbidden(fmt, *args, **kwargs):
```
robnagler: thanks for adding this

sirepo/server.py
moellep: The app.secret_key is using the existing config - cfg.beaker_session.secret. I re-used the beaker_session key and secret for the session cookie. That way existing config doesn't need to change.

sirepo/server.py
moellep: The existing code in simulation_db catches the KeyError and creates a user. That is why server.session_user() only raises KeyError if the cookie is valid. Otherwise it raises forbidden.

moellep:
I believe I fixed all the issues - please give the code another look. I also fixed a bug where the oauth state didn't transition from the beaker session into the new cookie session.

Issue #1299, opened 2018-08-01
dynamic load on optional javascript modules


moellep:
Apps should load the required libraries dynamically. For example, vtk.js should only load if required. sirepo-lattice.js should only be loaded for lattice apps (elegant and synergia).

Tasks:
- [x] lattice
- [x] beamline
- [x] d3
- [x] vtk
- [x] remove leading ```static/js``` from file references in schema, to be handled in python
- [x] determine which files must stay in ```index.html```

mkeilman:
Question for the group: Would it make sense to move all of the javascript from ```<script>``` tags in ```index.html``` to the dynamic model, except maybe the IE-specific ones?  I can experiment...

robnagler:
Sounds good to me. Generating HTML doesn't seem to have any particular value to me.

mkeilman:
The only files that seem to require living in the html are angular, jquery, cookieconsent, and sirepo, along with the  specific app's js file name (```elegant.js``` )  which is filled in...

robnagler:
Kind of strange. The app can't be dynamically loaded?

robnagler:
I would think that dynamically loading as many modules as possible would be better rather than having it in the html.

mkeilman:
Removed tasks for moving to ```schema.py```, as that has its own issue #1333

PR #1325, opened 2018-08-29
Dynamic modules (#1299)

mkeilman:
Notes:

This is currently isolated to a single case, loading ```sirepo-lattice.js``` for elegant and synergia, having removed the script tag from ```index.html```.  If it is suitable then it can be expanded to other items such as beamline, vtk, etc.


moellep:
I've simplified it a bit in 8c2404e 
The schema config is now just a list of javascript source paths. I removed the onreadystatechange because the browsers we support all have onload().

PR #1326, opened 2018-08-30
Added dynamic mod for beamline (#1299)

mkeilman:
Affects SRW and Shadow

### DIFF
--- a/sirepo/package_data/static/html/index.html
+++ b/sirepo/package_data/static/html/index.html
@@ -83,7 +83,7 @@
     <script src="/static/js/ext/jquery-2.2.4.min.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/bootstrap-3.3.7.min.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/ngDraggable.js{{ source_cache_key }}"></script>
-    <script src="/static/js/ext/angular-d3.js{{ source_cache_key }}"></script>
+    <!--<script src="/static/js/ext/angular-d3.js{{ source_cache_key }}"></script>-->
     <script src="/static/js/ext/angular-vtk.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/colorbar.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/Blob.js{{ source_cache_key }}"></script>
--- a/sirepo/package_data/static/html/index.html
+++ b/sirepo/package_data/static/html/index.html
@@ -92,7 +92,7 @@
     <script src="/static/js/ext/rgbcolor.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/StackBlur.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/canvg.js{{ source_cache_key }}"></script>
-    <script src="/static/js/ext/dom-to-image.min.js{{ source_cache_key }}"></script>
+    <!--<script src="/static/js/ext/dom-to-image.min.js{{ source_cache_key }}"></script>-->
     <script src="/static/js/ext/stacktrace-0.6.4.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/split-1.3.5.min.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/bootstrap-toggle.min.js{{ source_cache_key }}"></script>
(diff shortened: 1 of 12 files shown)


mkeilman:
For the d3 module, I removed all other references to d3 and the ```isPlottingReady``` stuff, but did not delete the ```angluar-d3.js``` file.  If it all works then that can go as well

sirepo/simulation_db.py
```diff
@@ -244,13 +244,16 @@ def get_schema(sim_type):
     schema['simulationType'] = sim_type
     _SCHEMA_CACHE[sim_type] = schema
 
-    # TODO (mvk): merge common modules into app modules
+    # TODO (mvk): improve merging common and local schema
```
robnagler: funny where we disagree on spaces: #TODO(mvk) is how to do this. I'll mention here that `if()` should be `if ()` in javascript. Spaces after statement keywords and no spaces after functions in Javascript.

sirepo/util.py
```diff
@@ -26,5 +26,9 @@ def merge_dicts(base, derived, depth = 1):
             derived[key] = base[key]
         merge_dicts(base[key], derived[key], depth-1)
 
+def merge_lists(base, derived):
+    for item in base:
+        derived.append(item)
```
robnagler: `derived.extend(base)` does what you want.

sirepo/package_data/static/html/index.html
```diff
@@ -83,24 +83,24 @@
     <script src="/static/js/ext/jquery-2.2.4.min.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/bootstrap-3.3.7.min.js{{ source_cache_key }}"></script>
     <script src="/static/js/ext/ngDraggable.js{{ source_cache_key }}"></script>
-    <script src="/static/js/ext/angular-d3.js{{ source_cache_key }}"></script>
-    <script src="/static/js/ext/angular-vtk.js{{ source_cache_key }}"></script>
+    <!--<script src="/static/js/ext/angular-d3.js{{ source_cache_key }}"></script>-->
```
robnagler: don't comment out, just remove

sirepo/package_data/static/html/index.html
```diff
@@ -83,24 +83,24 @@
     <script src="/static/js/ext/jquery-2.2.4.min.js{{ source_cache_key }}"></script>
```
robnagler: maybe this whole thing should be a for loop in jinja with the modules coming from python schema: staticModules. That way we can use them programmatically elsewhere:

```txt
{% for m  in static_js_files %}
<script src="{{ }}"></script>
{% endfor %}
```

The creation of static_js_files would be inside the python code schema. The load order could be there, too. This would allow more flexibility and could reuse "static" from uri_router or whatever.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -5,7 +5,7 @@ var srdbg = SIREPO.srdbg;
 SIREPO.PLOTTING_LINE_CSV_EVENT = 'plottingLineoutCSV';
 SIREPO.DEFAULT_COLOR_MAP = 'viridis';
 
-SIREPO.app.factory('plotting', function(appState, d3Service, frameCache, panelState, utilities, requestQueue, simulationQueue, $interval, $rootScope, $window) {
+SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilities, requestQueue, simulationQueue, $interval, $rootScope, $window) {
 
     var INITIAL_HEIGHT = 400;
     var MAX_PLOTS = 11;
```
robnagler: 11? I'm sure there's a reason for this, but why?

sirepo/package_data/static/json/schema-common.json
```diff
@@ -83,6 +83,9 @@
         "updateFolder": "/update-folder",
         "uploadFile": "/upload-file/<simulation_type>/<simulation_id>/<file_type>"
     },
+    "commonDynamicModules": [
+        "/static/js/ext/d3-3.5.9.min.js"
```
robnagler: /static/js should not be everywhere. It should be generated on the python side. commonDynamicModules is a funny name. What needs to happen is that schema-common needs to append or prefix the lists of dynamicModules in the python.

sirepo/package_data/static/js/sirepo-plotting.js
moellep: MAX_PLOTS is actually the max number of lines to show on a convergence plot. The constant should be renamed to MAX_CONVERGENCE_LINES or something similar.

Issue #1330, opened 2018-09-01
linspace => linearlySpacedArray


robnagler:
https://github.com/radiasoft/sirepo/blob/8575f6ab8e4f275dc89d9603c94bcaab6346a02f/sirepo/package_data/static/js/sirepo-plotting.js#L197

We should be vary careful about using abbreviations. If you look up linspace, you invariable see something like: 

> The linspace function generates linearly spaced vectors.

This is a superfluous comment if the function were named clearly.

Issue #1338, opened 2018-09-05
split out utilities for simulation runners from simulation_db


robnagler:
simulation_db should not be imported into runners, because this brings in cookie (among other things).

Maybe simulation_db should be split into sim_db, sim_json, sim_user?

robnagler:
Enough has been done for this issue.

Issue #1286, opened 2018-07-30
Switch field info in schema from arrays to objects


mkeilman:
Opening this for discussion at least - field info lives in arrays with positions mapped somewhat opaquely, and not uniquely, to function (```["Label", "Type", ...]```).  Clearer to use objects with aptly-named keys (```{label: "Label", type: "Type", ...}```).  This will require a tool to convert the schema files, then dealing with the exceptional cases and fixing up the related javascript.

Tasks:
- [x] Converter tool
- [ ] Cleanup of non-conforming cases
- [ ] Fix javascript

PR #1341, opened 2018-09-07
More dynamic (and now static) loading of files (#1299) plus #1330

mkeilman:
Notes:

- ```schema-common``` and the app schemae now list out "dynamicFiles" and "staticFiles" that refer only to the file name.  Static files are loaded by ```index.html``` with jinja loops, while dynamic files get loaded by sirepo.  Most ```<script>``` tags have been removed.  Consequently, each app's javascript is now referenced in its schema and dynamically loaded
- Files are grouped into "local" and "external" which map to directories
- Also in ```schema-common```,  things like "commonModels" have been moved under a single "common" section and named the same as in the app schema (so just "model").  The idea is to eventually loop over them to merge
- The schema fetch and validation in ```simulation_db``` use ``` pkcollections.Dict``` and dot syntax
- ```merge_dicts``` now allows unbounded merge depth and a ```merge_style``` keyword that lets values in arrays from the base dict be added to the corresponding arrays in the derived dict (this specifically to let us merge the dynamic file lists)

I have not yet split out the schema stuff into its own file

Also #1330 is in there

### DIFF
--- a/sirepo/package_data/static/json/warpvnd-schema.json
+++ b/sirepo/package_data/static/json/warpvnd-schema.json
@@ -1,11 +1,11 @@
 {
     "dynamicFiles": {
-        "external": {
+        "externalLibs": {
             "js": [
                 "vtk.min.js"
             ]
         },
-        "local": {
+        "sirepoLibs": {
             "js": [
                 "warpvnd.js"
             ]
--- a/sirepo/package_data/static/json/warpvnd-schema.json
+++ b/sirepo/package_data/static/json/warpvnd-schema.json
@@ -1,7 +1,16 @@
 {
-    "dynamicModules": [
-        "/static/js/ext/vtk.min.js"
-    ],
+    "dynamicFiles": {
+        "external": {
+            "js": [
+                "vtk.min.js"
+            ]
+        },
+        "local": {
+            "js": [
+                "warpvnd.js"
+            ]
+        }
+    },
     "localRoutes": {
         "source": {
             "config": {
--- a/sirepo/simulation_db.py
+++ b/sirepo/simulation_db.py
@@ -115,7 +115,7 @@ def app_version():
     """
     if pkconfig.channel_in('dev'):
         return datetime.datetime.utcnow().strftime('%Y%m%d.%H%M%S')
-    return SCHEMA_COMMON['version']
+    return SCHEMA_COMMON.version
 
 
 def celery_queue(data):
@@ -198,30 +198,30 @@ def fixup_old_data(data, force=False):
         bool: True if data changed
     """
     try:
-        if not force and 'version' in data and data['version'] == SCHEMA_COMMON['version']:
+        if not force and 'version' in data and data.version == SCHEMA_COMMON.version:
             return data, False
         try:
-            data.fixup_old_version = data['version']
-        except KeyError:
+            data.fixup_old_version = data.version
+        except AttributeError:
             data.fixup_old_version = _OLDEST_VERSION
-        data.version = SCHEMA_COMMON['version']
+        data.version = SCHEMA_COMMON.version
         if 'simulationType' not in data:
-            if 'sourceIntensityReport' in data['models']:
-                data['simulationType'] = 'srw'
-            elif 'fieldAnimation' in data['models']:
-                data['simulationType'] = 'warppba'
-            elif 'bunchSource' in data['models']:
-                data['simulationType'] = 'elegant'
+            if 'sourceIntensityReport' in data.models:
+                data.simulationType = 'srw'
+            elif 'fieldAnimation' in data.models:
+                data.simulationType = 'warppba'
+            elif 'bunchSource' in data.models:
+                data.simulationType = 'elegant'
             else:
                 pkdlog('simulationType: not found; data={}', data)
                 raise AssertionError('must have simulationType')
-        elif data['simulationType'] == 'warp':
-            data['simulationType'] = 'warppba'
-        elif data['simulationType'] == 'fete':
-            data['simulationType'] = 'warpvnd'
-        if 'simulationSerial' not in data['models']['simulation']:
-            data['models']['simulation']['simulationSerial'] = 0
-        sirepo.template.import_module(data['simulationType']).fixup_old_data(data)
+        elif data.simulationType == 'warp':
+            data.simulationType = 'warppba'
+        elif data.simulationType == 'fete':
+            data.simulationType = 'warpvnd'
+        if 'simulationSerial' not in data.models.simulation:
+            data.models.simulation.simulationSerial = 0
+        sirepo.template.import_module(data.simulationType).fixup_old_data(data)
         pkcollections.unchecked_del(data.models, 'simulationStatus')
         pkcollections.unchecked_del(data, 'fixup_old_version')
         return data, True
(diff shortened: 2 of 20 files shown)


sirepo/simulation_db.py
```diff
@@ -946,6 +974,27 @@ def _create_example_and_lib_files(simulation_type):
...
+        type (str): type of the file (css, js, etc.)
+
+    Returns:
+        [str]: combined list of local and external file paths
+    """
+    paths = []
+    for lf_prop in LOADED_FILE_PROPS:
+        paths.extend(
+            map(lambda file_name:
+                _pkg_relative_path_static(file_type + '/' + lf_prop.dir, file_name),
```
robnagler: I have updated 
[DesignHints](https://github.com/radiasoft/pykern/wiki/DesignHints#impedance-matches-data-structures-align-with-algorithms)
to talk about this specific example

sirepo/simulation_db.py
```diff
@@ -963,15 +1012,62 @@ def _init():
         nfs_tries=(10, int, 'How many times to poll in hack_nfs_write_status'),
         nfs_sleep=(0.5, float, 'Seconds sleep per hack_nfs_write_status poll'),
     )
+    props = pkcollections.Dict()
```
robnagler: This doesn't need to be here. It could be defined in the constant declaration. 
```py
LOADED_FILE_PROPS = [
    props=pkcollections.Dict(
        source='external',
        dir='ext',
...
```

The data structure is not the right "impedance". Rather, code

sirepo/simulation_db.py
```diff
@@ -1151,11 +1264,18 @@ def _validate_name_uniquify(data, starts_with):
...
+    Note that currently the values in enum arrays at the indices below are sometimes
+    used for other purposes, so we return rather than fail for non-numeric values
+
+    Args:
+        val: numeric value to validate
+        sch_field_info ([str]): field info array from schema
+    Raises:
+        AssertionError if val is out of schema-specified range
+    """
     if len(sch_field_info) <= 4:
```
robnagler: Magic numbers are a problem here What is "4"?

sirepo/simulation_db.py
```diff
@@ -1151,11 +1264,18 @@ def _validate_name_uniquify(data, starts_with):
...
 
 
-
-# Ensure the value of a numeric field falls within the supplied limits (if any)
-# Note that currently the values in enum arrays at the indices below are sometimes
-# used for other purposes, so we return for non-numeric values rather than fail
 def _validate_number(val, sch_field_info):
+    """Ensure the value of a numeric field falls within the supplied limits (if any)
+
+    Note that currently the values in enum arrays at the indices below are sometimes
```
robnagler: This comment should be at the place in the code, since it is not relevant to the description of the function as whole.

sirepo/simulation_db.py
```diff
@@ -1151,11 +1264,18 @@ def _validate_name_uniquify(data, starts_with):
...
+    used for other purposes, so we return rather than fail for non-numeric values
+
+    Args:
+        val: numeric value to validate
+        sch_field_info ([str]): field info array from schema
+    Raises:
```
robnagler: Raises is too much documentation for a private function.

sirepo/simulation_db.py
```diff
@@ -1151,11 +1264,18 @@ def _validate_name_uniquify(data, starts_with):
...
+        AssertionError if val is out of schema-specified range
+    """
     if len(sch_field_info) <= 4:
         return
     min = sch_field_info[4]
```
robnagler: unnecessary assignment

sirepo/simulation_db.py
```diff
@@ -1176,22 +1296,34 @@ def _validate_number(val, sch_field_info):
             raise AssertionError(util.err(sch_field_info, 'numeric value {} out of range', val))
```
robnagler: The code above should use assert statements:

```py
assert fv <= fmax, \
    util.err(field_info, 'numeric value {} out of range', val)
```

For some reason I can't comment on the code above in GitHub (only on changed lines or near the change?). I thin the if len(field_info) > 5 should be a short circuit exit.

I think the "float" conversions should be modularized so you can say something like:

```
_assert_field_info(val, field_info, 5)
```

I actually don't understand the modularization especially the code:

```py
            sch_field_info = sch_model[field_name]
            _validate_enum(val, sch_field_info, sch_enums)
            _validate_number(val, sch_field_info)
```

It seems to me that something is either an enum or a number, not both.

sirepo/simulation_db.py
```diff
@@ -202,7 +205,7 @@ def fixup_old_data(data, force=False):
         except KeyError:
             data.fixup_old_version = _OLDEST_VERSION
         data.version = SCHEMA_COMMON['version']
-        if not 'simulationType' in data:
+        if 'simulationType' not in data:
```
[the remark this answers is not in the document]
mkeilman: I actually got it from a PEP8 warning in PyCharm...I'll turn this one off

sirepo/simulation_db.py
robnagler: I wasn't clear about assert_field_info. The use of assert_field_info should be simpler in validate_number (which really should be assert_number):

```py
_assert_field_info(field_info, 4, v, float.__ge__, 'less than min')
_assert_field_info(field_info, 5, v, float.__le__, 'great than max')

def _assert_field(field_info, i, v, op, msg):
    try:
        m = float(field_info(i))
    except ValueError:
        return
    assert op(v, m), \
        util.err(v, 'value is {}={}', msg, m)
```

Perhaps this might be better called _assert_bound.

sirepo/simulation_db.py
mkeilman: This is due to the info being an array; min goes in index 3, etc.  That will be dealt with by #1286 (I built the tool already but have not taken the plunge on converting everything)

robnagler: If the magic number were _FIELD_INFO['min'] = 4 and _FIELD_INFO['max'] = 5, then you would be able to refer to those inside _assert_float/field_info (see other comment) programmatically. You could then simplify the function signature further.

sirepo/simulation_db.py
```diff
@@ -963,15 +1012,62 @@ def _init():
...
+
+    Args:
+        file_dir (str): sub-directory of package_data/static
+        file_name (str): name of the file.  If starts with '/', just return file_name
+
+    Returns:
+        str: full relative path of the file
+    """
+    pkdlog('getting static path for {}:{}', file_dir, file_name)
+    if file_name[0] == '/':
```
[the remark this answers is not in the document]
mkeilman: This was to handle the special case of calls to ```/user-state.js``` (actually a jinja template) that @moellep added as part of cookie management

sirepo/package_data/static/json/warpvnd-schema.json
```diff
@@ -1,7 +1,16 @@
...
-    "dynamicModules": [
-        "/static/js/ext/vtk.min.js"
-    ],
+    "dynamicFiles": {
+        "external": {
+            "js": [
+                "vtk.min.js"
+            ]
+        },
+        "local": {
```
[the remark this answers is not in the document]
mkeilman: OK.  I waffled on naming some of this stuff.  I like sirepo/external

sirepo/simulation_db.py
```diff
@@ -242,28 +245,31 @@ def get_schema(sim_type):
...
+
     if 'appModes' not in schema:
-        schema['appModes'] = pkcollections.Dict()
-    _merge_dicts(schema['commonAppModes'], schema['appModes'])
-    _merge_dicts(schema['commonModels'], schema['model'], depth=2)
-    _merge_dicts(schema['commonEnums'], schema['enum'])
-    common_views = schema['commonViews']
-    app_views = schema['view']
+        schema.appModes = pkcollections.Dict()
+    _merge_dicts(schema.common.localRoutes, schema.localRoutes, depth=2)
```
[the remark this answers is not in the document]
mkeilman: The loop is planned.  I'll work on making the merge depth sensible - it should probably default to "all the way down" right?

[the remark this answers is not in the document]
robnagler: I don't think you should generalize this right now. Just make it work for your case.

sirepo/simulation_db.py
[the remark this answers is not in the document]
robnagler: Perhaps this is an entirely different: externalModules, staticModules, dynamicModules. It's a very special case. That particular code looks very odd, especially since there are paths being generated.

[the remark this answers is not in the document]
robnagler: Here's another [DesignHint](https://github.com/radiasoft/pykern/wiki/DesignHints#learn-libraries-theres-an-api-for-that) which discusses the problem. I think static_file_path should be privatized and made _static_file_uri.

sirepo/simulation_db.py
```diff
@@ -64,6 +64,9 @@
 #: where users live under db_dir
 _LIB_DIR = 'lib'
 
+#: Properties of files loaded by html or javascript (see _init)
+LOADED_FILE_PROPS = []
```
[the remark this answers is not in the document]
mkeilman: I was preparing the putative ```schema.py``` and was referring to ```LOADED_FILE_PROPS``` from there.  Which raises a question: ```simulation_db``` knows about file paths etc., but eventually only the new file will use this structure, so where should it live?  And how will all that square with #1338?

PR #1388, opened 2018-10-17
Move toward single client cookie

mkeilman:
Notes:

- This is all on the client side right now, mainly to simplify interaction with the notification service
- Cookie (and notification) definitions live in ```schema-common```.  Notifications refer to cookie defs, which must exist, and cookie names/values must be valid; see ```validate_schema```
- Also introduces constants in the schema
- Currently we need various cookie timeouts (we bug users to log in every day e.g.)  So that is stored along with the cookie value
- Loading the simulations page removes all the old cookies; that's  a good candidate to move to the server side I think
- Does not address the 3rd-party ```sr_cookiesonsent``` cookie


sirepo/package_data/static/js/sirepo.js
```diff
@@ -622,86 +622,77 @@ SIREPO.app.factory('appDataService', function() {
...
-            return now.getTime() > v + (notification.delay || 0);
+        var cd = cookieDef(notification);
+        var c = cookieService.cleanExpiredCookie(cd);
+        var vcd = SIREPO.APP_SCHEMA.cookies.firstVisit;
+        var vc = cookieService.getCookie(vcd);
+        if(! c) {
+            var lstVisitDays = vc.t - cookieService.timeoutOrDefault(vcd);
+            // we need millisecond comparison here
+            var s = now.getTime() > (lstVisitDays + (notification.delay || 0)) * SIREPO.APP_SCHEMA.constants.oneDayMillis;
+            return s;
```
moellep: I'd suggest keeping the number of temporary variables to a minimum. Otherwise you don't know easily if variable "c" is used in many places, and it isn't easily searched.
```
if (! cookieService.cleanExpiredCookie(cookieDef(notification))) {
    var vcd = SIREPO.APP_SCHEMA.cookies.firstVisit;
    var vc = cookieService.getCookie(vcd);
    var lstVisitDays = vc.t - cookieService.timeoutOrDefault(vcd);
    // we need millisecond comparison here
    return now.getTime() > (lstVisitDays + (notification.delay || 0)) * SIREPO.APP_SCHEMA.constants.oneDayMillis;
}
```

Issue #1410, opened 2018-10-31
warpvnd: particle 3D trace full screen


moellep:
When full screen is selected on the 3D particle trace, the axis labels don't appear correct and the plot is cut off on a large screen.

Issue #1404, opened 2018-10-30
zoom plot with lattice overlay distorts flags


moellep:
When zooming on a plot which has a long lattice overly, the elements get stretched, which is OK for magnets and drifts, but the flag elements get distorted:

![screen shot 2018-10-30 at 3 15 35 pm](https://user-images.githubusercontent.com/6516910/47751157-aa76a300-dc56-11e8-940e-7a7c5d2fac99.png)

Issue #1419, opened 2018-11-05
Report notes enhancement


moellep:
If a user has entered a value for the report "notes" fields, it would be nice if:

- The notes text was visible on the modal pop-up, rather than requiring a click on the "click to enter notes" link. If no notes were entered, then the field could remain hidden.

- An info icon could appear on the report, possibly in the bottom left or right corner. The icon could behave like the field help text and show a popup if moused over, showing the notes info.

Issue #1411, opened 2018-10-31
warpvnd: particle trace 3d as PNG


moellep:
The 3D particle trace doesn't appear in the image when the Download as PNG is selected.

PR #1421, opened 2018-11-06
Lattice flag scaling (#1404)

mkeilman:
The flags now check the overall scale of the lattice plot and impose their own subscaling so they don't stretch too much as you zoom in - I presume it's OK to let them compress so they don't overlap, but that's simple to emend.  I just eyeballed an appropriate scale factor.

Also cleaned up the css a bit in researching #1410


mkeilman:
Notes for latest additions:

- Report notes now visible by default, as per #1419.  The user can still collapse them if they want
- vtk images can be downloaded as png (#1411).  This requires keeping a canvas object around to repaint, as the webgl draw buffer is updated asynchronously and can be empty by the time a user starts the download.  I borrowed the reportId added for keypress events to use as the key for storing the canvas object
- The pngs for the 3d particle trace do not include the axes at this time

PR #1458, opened 2018-11-29
Issue/1356 simulation href

### LINKED ISSUE #1356: Links of simulations in simulation list should be to simulations
When you hover (or copy link) of a simulation, it's just https://beta.sirepo.com/<app>, when it should be https://beta.sirepo.com/<app>#/source/<sid>.

e-carlin:
Resolves #1356 

This wasn't mentioned in the issue but the href's in the simulation dropdown menu are also broken ("open as new copy", "export as zip", etc.). I fixed the href for "open" because it shared code with the other work for this issue. Any preference on how to proceed with the other options in the dropdown? I can think of how to get a URL for all options except for "open as new copy". That opens in a modal which I believe I can't create a URL for. Let me know how to proceed.  

For testing, I ran test.sh. Is that all I need to run for tests? Should I write any tests for my work? It seemed to me that the relevant tests to my change would live in tests/karma is that correct?

### DIFF
--- a/sirepo/package_data/static/html/simulations.html
+++ b/sirepo/package_data/static/html/simulations.html
@@ -34,7 +34,7 @@
           </thead>
           <tbody>
             <tr data-ng-repeat="item in simulations.activeFolder.children | orderBy:['isFolder', simulations.sortField]">
-              <td><a data-ng-href="{{nav.simulationURL(item.simulationId)}}" data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
+              <td><a data-ng-href="{{ nav.simulationURL(item.simulationId) }}" data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
               <td>{{ item.lastModified }}</td>
             </tr>
             <tr data-ng-show="simulations.isWaitingForSim">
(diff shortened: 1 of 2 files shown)


mrakitin:
@e-carlin, GitHub supports automatic resolution of the issues with its keywords. If you edit your description to contain "Resolves #1356" (or "Closes #1356"), it will automatically close that issue once the PR is merged. See an example [here](https://github.com/NSLS-II/bluesky/pull/907).

sirepo/package_data/static/html/simulations.html
```diff
@@ -34,7 +34,7 @@
...
-              <td><a href data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
+              <td><a data-ng-href="{{nav.simulationURL(item.simulationId)}}" data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
```
robnagler: Spaces after `{{` and before `}}`. I call this "leave no traces coding". The indentation, spacing, brace placement, etc. should all match the existing code. We don't have a formatter for this type of code (please suggest one if one exists). 

```suggestion
              <td><a data-ng-href="{{ nav.simulationURL(item.simulationId) }}" data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
```

robnagler:
Karma test would be cool. It might not be worth it, because it probably will have to hardwire all kinds of context to validate something that's rather subtle.

We have an architectural issue (ie shortcut) that we don't name ever call. Please submit a new issue around the fact that everything should have a name. I would think there would be a clever way to do this. Our URL formatting should be automatic like the parsing. You should be able to pass any entry point name and its argument to a formatter that will format the URL so that it is parseable by the dispatcher.

sirepo/package_data/static/html/simulations.html
e-carlin: Thanks. I've used [prettier](https://github.com/prettier/prettier) on all other personal and professional JS projects. I personally really enjoy it but I also like opinionated tools. I've never used it with Angular templates. They seem to be [supported](https://prettier.io/blog/2018/11/07/1.15.0.html) but only since Nov 7, 2018. The React JSX support was very good.

e-carlin: It is commonly used in conjunction with eslint which I also like using. It looks like there is some Angular [support](https://github.com/Gillespie59/eslint-plugin-angular).

e-carlin:
@robnagler Can you elaborate on "we don't name call ever" and "everything should have a name"? I'm not sure I understand what you are saying.

sirepo/package_data/static/html/simulations.html
robnagler: We use jshint, but I don't think anybody is attached. If we could use something whose config works with PyCharm and Emacs, it would be great. This could be a project if you are interested. Not the highest priority, obviously, since there's only a few of us.

robnagler:
typo: "we don't name ever**y** call"

TL;DR just copy-and-paste to name the functions, we don't have support for generalized naming.

[There are only two hard things in Computer Science: cache invalidation and naming things](https://skeptics.stackexchange.com/q/19836) - Phil Karlton, Netscape. 

I actually heard the same thing a decade before from Dale Skeen in the context of distributed systems, which seems more relevant to this problem. I hardly thing we are doing Computer Science: this is just programming.

When things don't have a name, there's a lot of implicit coupling (read: assumptions aka bug fodder) and therefore copy-and-paste. I can only give [an example from BOP](https://github.com/biviosoftware/perl-Bivio/blob/d041266565b8e282a62bcbb1abd1f1fb846cc583/UI/XHTML/Widget/ForumTaskMenu.pm#L10):

```perl
my($_PARENT_MAP, $_PARENT_TASK) = _make_parent_map({
    FORUM_WIKI_VIEW => [qw(
        FORUM_WIKI_EDIT
        FORUM_WIKI_NOT_FOUND
        FORUM_WIKI_VERSIONS_LIST
        FORUM_WIKI_VERSIONS_DIFF
        SITE_WIKI_VIEW
    )],
    FORUM_MAIL_THREAD_ROOT_LIST => [qw(
        FORUM_MAIL_THREAD_ROOT_LIST
        FORUM_MAIL_THREAD_LIST
        FORUM_MAIL_FORM
        FORUM_MAIL_PART
[snip]
```

Every operation in the menu has a name. That name allows us to get access to meta data, e.g. don't display the entry if the user doesn't have permission. The name gives access to a URI in the appropriate context, and of course, the label to display to the user (with appropriate i18n). The fonts and classes, etc. are pulled from the appropriate file, too. The name is the explicit coupler between all of this.

In the code in the simulation list, we have none of this:

```javascript
 <td><a data-ng-href="{{ nav.simulationURL(item.simulationId) }}" data-ng-click="simulations.openItem(item)"><span class="glyphicon" data-ng-class="{'sr-user-item': ! simulations.fileManager.isItemExample(item), 'glyphicon-folder-close': item.isFolder, 'glyphicon-file': ! item.isFolder, 'sr-transparent-icon': ! item.isFolder && item.isExample, 'sr-transparent-icon-user': ! item.isFolder && ! item.isExample}"></span> <span data-ng-class="{ 'sr-user-item': ! simulations.fileManager.isItemExample(item) }">{{ item.name | simulationName }}</span></a></td>
```

There is no name for the concept of a menu, instead we have copy-and-paste `<td><a></a></td>`. We don't know that the menu item shouldn't be shown in certain cases. In this one line the word "example" is repeated four times. There's just no abstraction, everything is literal.

The flip side is that the simplicity of literal programming allows you (noob) to go into the code and make a simple and significant change (admittedly, caused by lack of said abstraction). However, as you noticed, there's a more complex problem: no name (URL) for every item. 

That's where the chicken-and-egg of complex systems arises: in a system as abstracted as BOP, it takes a long time to become an "expert". In a system as simple as Angular (which people complain about being complex), anybody can be an expert in a year. This is why so many programmers think it is easy to learn new languages: literal programming. And, why there were only three programmers at ViaWeb, where 30% of the code was Lisp macros.

We make an attempt at naming things on the [Python side of things](https://github.com/radiasoft/sirepo/blob/82e2ef7ad2fdd150b0af23557bcf3f7c57cd842b/sirepo/server.py#L99):

```python
@api_perm.require_user
def api_copySimulation():
```

Here we also supply some metadata. This is the underlying routine that "Open as new Copy" calls, but there's a lot of goop in between, which is a bummer. You know how to create the URI for this by supplying [the name to the URI formatter](https://github.com/radiasoft/sirepo/blob/82e2ef7ad2fdd150b0af23557bcf3f7c57cd842b/sirepo/uri_router.py#L110).

The solution to the problem is building a lot of abstractions that don't exist in most JS frameworks I've seen. I suppose you could use AngularJS/React components for this, but they are not the right answer imiho. You really want some dictionary based (JSON) structures that allow you to get them easily from the server, which is the ultimate arbiter of names, permissions, and user-visible text.

moellep:
Looks good to me @e-carlin - thanks for improving this.

e-carlin:
@robnagler Thanks for all of this! I hadn't considered being able to name things as one of the central constructs behind creating abstraction but after reading over this it now seems extremely obvious. 

I agree with your thoughts on highly abstracted systems being harder to understand to a noob. I do think some level of abstraction does actually make things easier though. Had you been using a component named menuRow (or something similar, naming is hard:)) for the `<td><a></a></td>` it may have actually helped me when reading over the code to be able to identify that as a row in the menu. But, that is a very simple abstraction.

I'm curious why you think components are not the right answer? I've used React components quite heavily in other projects always appreciated their abstraction.

robnagler:
We'll take this offline. :)

moellep:
I rolled back this fix for now - it caused an exception when a folder is present in the simulation list.

e-carlin:
Sorry about that. I was unaware of folders in the simulation list. I'll work on resolving now.

PR #1474, opened 2018-12-17
Fix bytes/unicode in sirepo.cookie

njsmith:
Without this, sirepo crashes immediately on py3, because we end up
passing a unicode string to cryptography's Fernet encryption, and it
wants a bytes.

After this change, sirepo seems to work fine on both py2 and py3.

### DIFF
--- a/sirepo/cookie.py
+++ b/sirepo/cookie.py
@@ -176,7 +176,8 @@ class _State(dict):
         return self._crypto().decrypt(base64.urlsafe_b64decode(value))
 
     def _deserialize(self, value):
-        v = value.split(_SERIALIZER_SEP)
+        v = value.decode("utf-8")
+        v = v.split(_SERIALIZER_SEP)
         v = dict(zip(v[::2], v[1::2]))
         assert v[_COOKIE_SENTINEL] == _COOKIE_SENTINEL_VALUE, \
             'cookie sentinel value is not correct'
@@ -229,7 +230,7 @@ class _State(dict):
             itertools.chain.from_iterable(
                 [(k, self[k]) for k in sorted(self.keys())],
             ),
-        )
+        ).encode("utf-8")
 
 
 @pkconfig.parse_none


sirepo/cookie.py
```diff
@@ -176,7 +176,8 @@ def _decrypt(self, value):
         return self._crypto().decrypt(base64.urlsafe_b64decode(value))
 
     def _deserialize(self, value):
-        v = value.split(_SERIALIZER_SEP)
+        v = value.decode("utf-8")
```
robnagler: We shouldn't hardcode utf-8. That's just like assuming ASCII. (or hardcoding ansi escape codes...) Of course, I'm tilting at windmills, but this implementation is incorrect.

HTTP *headers* are [required to be ASCII](https://tools.ietf.org/html/rfc7230#section-3):

> A recipient MUST parse an HTTP message as a sequence of octets in an encoding that is a superset of US-ASCII. [...]  
> Historically, HTTP has allowed field content with text in the ISO-8859-1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding.  In practice, most HTTP header field values use only a subset of the US-ASCII charset [USASCII]. Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

UTF-8 is not a superset of US-ASCII. Moreover, the cookie value is ours, and it has to be ASCII, because we determined the value characters. Therefore, we should encode the header octets as ASCII and if it doesn't work, throw a bad cookie exception.

Flask makes a mess of this, because [it requires UTF-8](http://flask.pocoo.org/docs/1.0/unicode/#automatic-conversion). Werkzeug repeats utf-8 about 100 times, and incorrectly uses sys.getdefaultencoding in werkzeug._compat. And, Python 2 & 3 screw up getdefaultencoding by returning 'ascii' for py2 and 'utf-8' for py3. If you are reading files, you should use locale.getpreferredencoding() afaik. At least that's something that we control in our deployment environment.

In this particular case, more work is required:

* The unit test should encode the values it passes as bytes.
* Enhance the unit test so that it tests invalid octets.
* All strings defined into cookie.py should be converted to bytes.
* set_value should assert key and value are both bytes or ascii. We get to control this so we should be rigid.
* set_log_user should convert uid with pykern.pkcompate.locale_str  I'm assuming here that logs are stored in the preferred locale.
* http_name should be decoded to bytes assuming ascii in _cfg_http_name.
* I think private_key may need to be converted to, or base64.urlsafe_b64decode might do the right thing.

We can talk about this, but I wanted to understand and to document the issues.

njsmith: Yeah, this definitely needs more work to figure out the details – I don't have a good picture of what's in the State object or the overall workflow.

But, I don't think the PR does put utf-8 into the header currently – I think the header value is `base64(encrypt(utf8(serialized state)))`, which is RFC-compliant (as long as your base64 encoder doesn't insert newlines).

robnagler:
I'm closing this, because we'll need to work on py3 as a distinct operation, once we can support py3 directly.

PR #1475, opened 2018-12-17
[WIP] Container runner

njsmith:
**(Messy prototype, posting for discussion only, do not merge.)**

Here's a first draft of container runner functionality. It can
successfully run 'myapp', while routing everything through docker,
including cancellation.

In terms of scope, I think it currently covers everything in the SOW1
milestone, except that it doesn't use TLS to connect to docker,
doesn't do multi-host slot management, and has no automated tests.
There's definitely enough to start discussing though :-).

In order to get something working quickly, this uses a hybrid model,
with threads and classic blocking libraries (redis-py, docker-py) on
one hand, plus Trio to manage the threads, and a bit of glue to
integrate the threads into Trio's cancellation system (so e.g.
control-C works cleanly). This involves a small amount of polling. The
core functionality is actually very straightforward; the only really
messy part is the kluges in the `ContainerJob` class.

To test it out:

* Create vagrant system; install redhat-dev, sirepo-dev, redhat-docker
* Check out this branch of sirepo
* Inside vagrant system:
  * ~~sudo dnf install redis python3-devel~~
  * ~~sudo systemctl enable redis docker~~
  * ~~sudo systemctl start redis docker~~
  * ~~virtualenv -p python3 py3env~~
  * ~~py3env/bin/pip install h5py~~
  * ~~py3env/bin/pip install -r src/radiasoft/sirepo/container-runner-requirements.txt~~
  * `sudo systemctl enable docker`
  * `sudo systemctl start docker`
  * `sudo dnf install libffi-devel`  # needed by pyenv install
  * `_bivio_pyenv_version 3.7.1 py37`  # maybe this should become the default for `bivio_pyenv_3`?
  * `pip install -e src/radiasoft/pykern`  # has to be installed before installing sirepo
  * `pip install -r src/radiasoft/sirepo/container-runner-requirements.txt`
  * `pip install h5py pytz==2015.7`  # workaround some dependency weirdness... not sure why h5py is missing, but pytz is related to celery stuff so should go away when celery does
  * In window 1: `sirepo service container-runner`
  * In window 2: `SIREPO_RUNNER_JOB_CLASS=container sirepo service http`

### DIFF
--- a/container-runner-requirements.txt
+++ b/container-runner-requirements.txt
@@ -1,6 +1,7 @@
 trio
-redis
-hiredis
 docker
+quart-trio
+hypercorn[trio]
+yarl
 -e src/radiasoft/pykern
 -e src/radiasoft/sirepo
--- a/container-runner-requirements.txt
+++ b/container-runner-requirements.txt
@@ -1,0 +1,6 @@
+trio
+redis
+hiredis
+docker
+-e src/radiasoft/pykern
+-e src/radiasoft/sirepo
--- a/container-runner-requirements.txt
+++ b/container-runner-requirements.txt
@@ -1,7 +1,0 @@
-trio
-docker
-quart-trio
-hypercorn[trio]
-yarl
--e src/radiasoft/pykern
--e src/radiasoft/sirepo
--- a/sirepo/pkcli/service.py
+++ b/sirepo/pkcli/service.py
@@ -74,6 +74,11 @@ def http():
         )
 
 
+def container_runner():
+    from sirepo.runner.container import serve
+    serve()
+
+
 def nginx_proxy():
     """Starts nginx in container.
 
--- a/sirepo/runner/__init__.py
+++ b/sirepo/runner/__init__.py
@@ -41,7 +41,7 @@ class State(aenum.UniqueEnum):
     RUN = 4
     STOP = 5
 
-_JOB_CLASSES = ('background', 'celery', 'docker')
+_JOB_CLASSES = ('background', 'celery', 'docker', 'container')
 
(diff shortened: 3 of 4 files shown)


robnagler:
Comments we discussed are attached.

[container-py.txt](https://github.com/radiasoft/sirepo/files/2688084/container-py.txt)

njsmith:
OK! This is switched to the new approach we talked about now, and seems to be working. I updated the try-it-out instructions in the first comment and I *think* they're accurate.

The one possibly-disruptive thing is that the HTTP framework we're using ("quart", like flask but async) currently can only be installed on python 3.7. I think this might only be for the sake of newer asyncio features which we don't care about b/c we're using trio, but switching to 3.7 might be easier than fighting with the packaging metadata? What I don't know is if switching to 3.7 will be disruptive somewhere else.

sirepo/cookie.py
```diff
@@ -176,7 +176,8 @@ def _decrypt(self, value):
         return self._crypto().decrypt(base64.urlsafe_b64decode(value))
 
     def _deserialize(self, value):
-        v = value.split(_SERIALIZER_SEP)
+        v = value.decode("utf-8")
```
robnagler: I'd like to merge this pull request, but we need to remove the cookie.py changes per earlier discussion. Once you've removed these changes, we can merge and close SOW1.

container-runner-requirements.txt
```diff
@@ -0,0 +1,7 @@
+trio
```
robnagler: Let's remove this file for now. It's a WIP, and we aren't going to deploy.

PR #1500, opened 2019-01-22
Axis refactor

mkeilman:
Most of the effort here was getting auxiliary code in place to support looping through the axes instead of treating them individually.  Some of the more prominent pieces are:

- Geometric objects and functions like points, line segments, etc. with a focus on plotting.  vtk has a lot of this but we should have our own for future non-vtk use.  Other libraries are probably overkill for our needs as well (could look into that though)
- 2D representations of 3D vtk objects (most notably a box but can be used for whatever). Keeps all of that out of the plotting directive

More needs to be done to extract this from ```particle3d```; stubs for that are included.  We should also have a generic vtk directive so we don't have to copy things like mouse handling etc.


moellep:
It looks like the karma plotVTK_test.js has one failure in boundaryIntersectionsWithPts(), line 146:
```
PhantomJS 2.1.1 (Linux 0.0.0) geometry should pass [rects]:  FAILED
        TypeError: undefined is not an object (evaluating 'bint1[2].equals') (line 146)
        error properties: Error
        tests/karma/plotVTK_test.js:146:24
        invoke@sirepo/package_data/static/js/ext/angular.js:5108:24
        workFn@tests/helpers/angular/angular-mocks.js:2404:26
        <Jasmine>
PhantomJS 2.1.1 (Linux 0.0.0): Executed 31 of 31 (1 FAILED) (0.221 secs / 0.187 secs)
```

Issue #1499, opened 2019-01-18
move db config and management to separate module


robnagler:
@njsmith notes:

> - sirepo.server.init normally gets the db dir from cfg.db_dir, but it
can also be overridden by an argument
> - then it stashes the result on app.sirepo_db_dir
> - then the app object gets manually passed through to other code like
sirepo.simulation_db, so it can read app.sirepo_db_dir

> AFAICT, there are no calls to sirepo.server.init(...) that pass the
> db_dir override argument. Maybe we should drop support for that
> option, and simply declare that the db dir is always available as
> "cfg.db_dir"? That seems like it would simplify things a lot.

We should move the relevant pieces from simulation_db as well.

PR #1504, opened 2019-01-24
SOW 2

njsmith:
So far this is very much an initial sketch, but it has the split
between the service and the API, the json-over-Unix-socket API, and a
very rudimentary attempt at redoing api_runCancel based on the
API (hidden behind a feature flag).

@robnagler: This is completely untested, because it depends on #1499 –
do you want me to take a stab at rearranging the db_dir config, or do
you want to do it? (I think I could get something working, but I feel
like without some hints it would probably be different than how you'd
do it :-).)

### DIFF
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -264,11 +264,11 @@ async def _handle_conn(job_tracker, lock_dict, stream):
             request = pkjson.load_any(request_bytes)
             if 'run_dir' in request:
                 request.run_dir = pkio.py_path(request.run_dir)
-            pkdp(request)
+            pkdp('runner request: {!r}', request)
             handler = _RPC_HANDLERS[request.action]
             async with lock_dict[request.run_dir]:
                 response = await handler(job_tracker, request)
-            pkdp(response)
+            pkdp('runner response: {!r}', response)
             response_bytes = pkjson.dump_bytes(response)
         except Exception as exc:
             await stream.send_all(
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -195,7 +195,7 @@ class _JobTracker:
                     if process.returncode:
                         _write_status(runner_client.JobStatus.ERROR, run_dir)
                     else:
-                        _write_status(runner_client.JobStatus.COMPLETE, run_dir)
+                        _write_status(runner_client.JobStatus.COMPLETED, run_dir)
             finally:
                 # _write_status is a no-op if there's already a status, so
                 # this really means "if we get here without having written a
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -142,6 +142,10 @@ class _JobTracker:
                 if disk_status == 'pending':
                     # We never write this, so it must be stale, in which case
                     # the job is no longer pending...
+                    pkdlog(
+                        'found "pending" status, treating as "error" ({})',
+                        disk_status_path,
+                    )
                     return runner_client.JobStatus.ERROR
                 return runner_client.JobStatus(disk_status)
         if run_dir in self.jobs and self.jobs[run_dir].jhash == jhash:
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -245,8 +245,9 @@ async def _cancel_job(job_tracker, request):
     job_info.process.terminate()
     with trio.move_on_after(_KILL_TIMEOUT_SECS):
         await job_info.finished.wait()
-    job_info.process.kill()
-    await job_info.finished.wait()
+    if job_info.returncode is None:
+        job_info.process.kill()
+        await job_info.finished.wait()
     return {}
 
 
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -181,7 +181,8 @@ class _JobTracker:
             try:
                 env = _subprocess_env()
                 run_log_path = run_dir.join(template_common.RUN_LOG)
-                # hack: switch back to py2 mode to run the actual command
+                # we're in py3 mode, and regular subprocesses will inherit our
+                # environment, so we have to manually switch back to py2 mode.
                 env['PYENV_VERSION'] = 'py2'
                 cmd = ['pyenv', 'exec'] + cmd
                 with open(run_log_path, 'a+b') as run_log:
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -45,7 +45,7 @@ def _catch_and_log_errors(exc_type, msg, *args, **kwargs):
     try:
         yield
     except trio.MultiError as multi_exc:
-        raise AssertionError("handle MultiErrors in _catch_and_log_errors")
+        raise AssertionError('handle MultiErrors in _catch_and_log_errors')
     except exc_type:
         pkdlog(msg, *args, **kwargs)
         pkdlog(pkdexc())
@@ -131,7 +131,7 @@ class _JobTracker:
                 return
             try:
                 env = _subprocess_env()
-                with open(run_dir / template_common.RUN_LOG, "a+b") as run_log:
+                with open(run_dir / template_common.RUN_LOG, 'a+b') as run_log:
                     process = trio.Process(
                         config['command'],
                         cwd=config['working_dir'],
@@ -175,9 +175,9 @@ async def _job_status(job_tracker, request):
     pkdc('job_status: {}', request)
     job_info = job_tracker.jobs.get(request['jid'])
     if job_info is None:
-        return {"status": _JobStatus.NOT_STARTED.value}
+        return {'status': _JobStatus.NOT_STARTED.value}
     else:
-        return {"status": job_info.status.value}
+        return {'status': job_info.status.value}
 
 
 async def _cancel_job(job_tracker, request):
@@ -195,9 +195,9 @@ async def _cancel_job(job_tracker, request):
 
 
 _HANDLERS = {
-    "start_job": _start_job,
-    "job_status": _job_status,
-    "cancel_job", _cancel_job,
+    'start_job': _start_job,
+    'job_status': _job_status,
+    'cancel_job', _cancel_job,
 }
 
 
@@ -211,7 +211,7 @@ async def _handle_conn(job_tracker, stream):
                     break
                 request_bytes += chunk
             request = pkjson.load_any(request_bytes)
-            handler = _HANDLERS[request["action"]]
+            handler = _HANDLERS[request['action']]
             response = await handler(job_tracker, request)
             response_bytes = pkjson.dump_bytes(response)
         except Exception as exc:
@@ -244,3 +244,8 @@ async def _main():
 def start():
     """Starts the container runner."""
     trio.run(_main)
+
+
+def start_dev():
+    """Temporary hack for testing: starts the dev server + container runner."""
+    import subprocess
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -307,7 +307,10 @@ def start():
 
 # Temporary (?) hack to make testing easier: starts up the http dev server
 # under py2 and the runner daemon under py3, and if either exits then kills
-# the other. Runner daemon output = green, flask output = blue.
+# the other.
+_RUNNER_DAEMON_OUTPUT_COLOR = 2  # green
+_FLASK_DEV_OUTPUT_COLOR = 4  # blue
+
 def _color(num):
     colors = curses.tigetstr('setaf')
     if colors is None:
@@ -323,6 +326,7 @@ async def _run_cmd(color, cmd, **kwargs):
                 return
             sys.stdout.buffer.raw.write(_color(color) + data + _color(0))
 
+    kwargs['stdin'] = subprocess.DEVNULL
     kwargs['stdout'] = subprocess.PIPE
     kwargs['stderr'] = subprocess.STDOUT
     async with trio.open_nursery() as nursery:
@@ -345,13 +349,15 @@ async def _dev_main():
             nursery.cancel_scope.cancel()
 
         nursery.start_soon(
-            _run_cmd_in_env_then_quit, 'py2', 4, ['sirepo', 'service', 'http'],
+            _run_cmd_in_env_then_quit,
+            'py2', _FLASK_DEV_OUTPUT_COLOR, ['sirepo', 'service', 'http'],
         )
         # We could just run _main here, but spawning a subprocess makes sure
         # that everyone has the same config, e.g. for
         # SIREPO_FEATURE_FLAG_RUNNER_DAEMON
         nursery.start_soon(
-            _run_cmd_in_env_then_quit, 'py3', 2, ['sirepo', 'runner', 'start'],
+            _run_cmd_in_env_then_quit,
+            'py3', _RUNNER_DAEMON_OUTPUT_COLOR, ['sirepo', 'runner', 'start'],
         )
 
 
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -8,6 +8,7 @@
 from __future__ import absolute_import, division, print_function
 
 import aenum
+import async_generator
 import collections
 import contextlib
 import functools
@@ -18,6 +19,7 @@ from pykern import pkjson
 import re
 import shlex
 from sirepo import mpi
+from sirepo import runner_client
 from sirepo import srdb
 from sirepo.template import template_common
 import sys
@@ -49,6 +51,36 @@ def _catch_and_log_errors(exc_type, msg, *args, **kwargs):
         pkdlog(pkdexc())
 
 
+# Used to make sure that if we get simultaneous RPCs for the same jid/run_dir,
+# then only one RPC handler runs at a time. Like defaultdict(trio.Lock), but
+# without the memory leak. (Maybe should be upstreamed to Trio?)
+class _LockDict:
+    def __init__(self):
+        # {key: ParkingLot}
+        # lock is held iff the key exists
+        self._waiters = {}
+
+    @async_generator.asynccontextmanager
+    async def __getitem__(self, key):
+        # acquire
+        if key not in self._waiters:
+            # lock is unheld; adding a key makes it held
+            self._waiters[key] = trio.hazmat.ParkingLot()
+        else:
+            # lock is held; wait for someone to pass it to us
+            await self._waiters[keys].park()
+        try:
+            yield
+        finally:
+            # release
+            if self._waiters[key]:
+                # someone is waiting, so pass them the lock
+                self._waiters[key].unpark()
+            else:
+                # no-one is waiting, so mark the lock unheld
+                del self._waiters[key]
+
+
 # Helper to call container.wait in a async/cancellation-friendly way
 async def _container_wait(container):
     while True:
@@ -67,32 +99,24 @@ async def _container_wait(container):
 def _write_status(status, run_dir):
     fn = run_dir.join('result.json')
     if not fn.exists():
-        pkjson.dump_pretty({'state': status}, filename=fn)
-        pkio.write_text(run_dir.join('status'), status)
+        pkjson.dump_pretty({'state': status.value}, filename=fn)
+        pkio.write_text(run_dir.join('status'), status.value)
 
 
 def _subprocess_env():
     env = dict(os.environ)
-    for k in list(env.items()):
+    for k in list(env):
         if _EXEC_ENV_REMOVE.search(k):
             del env[k]
-    env['SIREPO_MPI_CORES'] = mpi.cfg.cores
+    env['SIREPO_MPI_CORES'] = str(mpi.cfg.cores)
     return env
 
 
-class _JobStatus(aenum.Enum):
-    MISSING = 'missing'   # no data on disk, not currently running
-    PENDING = 'pending'   # data on disk is incomplete but it's running
-    ERROR = 'error'       # data on disk exists, but job failed somehow
-    CANCELED = 'canceled' # data on disk exists, but is incomplete
-    COMPLETE = 'complete' # data on disk exists, and is fully usable
-
-
 class _JobInfo:
-    def __init__(self, jid, run_dir, status, process):
-        self.jid = jid
-        self.status = status
+    def __init__(self, run_dir, jhash, status, process):
         self.run_dir = run_dir
+        self.jhash = jhash
+        self.status = status
         self.finished = trio.Event()
         self.process = process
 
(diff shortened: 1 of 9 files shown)


sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+async def _cancel_job(job_tracker, request):
+    job_info = job_tracker.jobs[request['jid']]
+    if job_info.status is not JobStatus.RUNNING:
+        return {}
+    simulation_db.write_result({'state': 'canceled'}, run_dir=job_info.run_dir)
+    # XX TODO: this is what api_runCancel used to do, but we can't really do
+    # it here. What should we do? (For a cancelled job, should we just delete
+    # the run-dir entirely, once we have the full hash in the jid?)
```
robnagler: canceled jobs can contain usable data. When we run parallel simulations, the user can cancel at any time, and still have the results already computed. Perhaps this might be a different state.

sirepo/runner_client.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    request_bytes = json.dumps(request).encode('ascii')
+    with socket.socket(socket.AF_UNIX) as sock:
+        sock.connect(_db_dir.join('runner.sock'))
+        sock.sendall(request_bytes)
+        sock.shutdown(socket.SHUT_WR)
+        response_bytes = bytearray()
+        while True:
+            chunk = sock.recv(4096)
+            if not chunk:
```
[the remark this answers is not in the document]
njsmith: It works correctly as is – in Python (unlike some other languages), strings are false iff they're empty; the actual characters don't come into it at all. But if you prefer `len(chunk)` to emphasize that, then that also works :-)

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,222 @@
...
+                    detach=True,
+                    init=True,
+                )
+            )
+            task_status.started()
+            # Returns a dict like: {'Error': None, 'StatusCode': 0}
+            # XX TODO but there may be a race condition where if the container
+            # finishes before our call to wait() starts, then it just errors
+            # out with docker.errors.NotFound
+            with _catch_and_log_errors(
```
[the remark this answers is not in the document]
njsmith: Unfortunately it's not – catch-all handlers in Trio programs need special error handling syntax, because when you have concurrency it's possible for multiple things to go wrong at the same time, and working with "multi-exceptions" is more complicated. This bit of Trio is actually [in flux](https://github.com/python-trio/trio/issues/611) right now and the new system will make it much easier to write this kind of decorator properly, plus currently we don't have any nurseries inside `_catch_and_log_errors`, so I figured we could fix it up later. But we will want to fix it up eventually, and then it'll be Trio-specific.

njsmith:
Status: the structure's there, it starts up (relatively conveniently with the `sirepo runner dev` I threw together), it has code for processes. The wiring to get service.py to actually use runner.py is the next big thing. Now I'm sorting through `simulation_db.py` and `service.py` to figure out how to get that wiring to work.

sirepo/pkcli/runner.py
```diff
@@ -131,7 +131,7 @@ def status(self, jid, run_dir):
                 return
             try:
                 env = _subprocess_env()
-                with open(run_dir / template_common.RUN_LOG, "a+b") as run_log:
+                with open(run_dir / template_common.RUN_LOG, 'a+b') as run_log:
```
robnagler: Overloaded `/` is not as readable as run_dir.join(template_common.RUN_LOG). Moreover, if the reader has to ask the question: what does `/` do, they can't search on [py path /](https://www.google.com/search?q=py+path+/) to get the answer whereas [py path join](https://www.google.com/search?q=py+path+join) produces the answer right on the google search page, because google can understand the question.  If google can't understand the question, that's a clue that the API has a bad name.

In general, avoid using or implementing overloaded operators like this. The `/` is a "cute" but cute is rarely maintainable. The use of `/` contradicts the meaning of `/` as divide: smaller result comes from a divide, is this removing or splitting something? This creates an unnecessary cognitive load. The word "join" is used in its proper English usage: connecting two things.

njsmith:
OK, I guess you'll want some tweaks but I think this has a working implementation of everything in SOW 2 now!

# How to run it

We're running flask under py2, and the runner daemon under py3. We already have pyenv set up to create both a py2 and a py3 environment; now we need sirepo installed in both. So something like:

```sh
$ pyenv global py2:py3
$ pip3 install -e src/radiasoft/pykern
$ pip3 install -r src/radiasoft/sirepo/requirements.txt
$ pip3 install -e src/radiasoft/sirepo
# Make sure it worked:
$ PYENV_VERSION=py3 pyenv exec sirepo --help
```

To spin up both the flask server and runner daemon to try it:

```sh
$ PYENV_VERSION=py3 pyenv exec sirepo runner dev
```

To run the test:

```sh
$ cd src/radiasoft/sirepo
$ pytest tests -k runner_test
```

**The test assumes you have the py3 env set up. We might want to change that before merging?**

Because of #1528, I haven't been able to run the full test suite.

# Other notes

I ended up switching to using `run_dir`+hash as the unique identifier for the job... so right now the runner doesn't actually use jids at all. I think eventually we'll want to switch to using `jid` as the unique identifier plus have ways to convert `jid` to-and-from `run_dir`, and make the runner responsible for managing run dirs in general, but (1) for now we can't change the actual `run_dir` path because server.py assumes that it can calculate it locally without talking to the runner, and (2) mapping jid to-and-from `run_dir` isn't trivial right now. So this seemed like the simplest approach for now.

njsmith:
Nice catch on the `complete` vs `completed` thing. It looks like there are several places in the existing code that are hard-coded to use `completed`, plus there might be existing files on disk using `completed`. So instead of switching the old code to use `complete` like in your patch, I switched the new code to using `completed`.

This also explains the test failure you saw: I copied the test's status-checking loop from `server_test.py`, so the test is looping until it sees `run.state == 'completed'` (i.e., what the old code uses). This didn't make any sense, because my new code was using `run.state = 'complete'`, so the loop check could never succeed. So now I have no idea how the test ever passed for me. But the output you pasted here shows exactly the failure mode we'd expect given this issue: the test gets a response with `run.state == 'complete'` (last line of the captured stdout), which is where it's supposed to stop looping. But, because it's checking for `run.state == 'completed'`, it keeps looping anyway, and then crashes trying to send the next request.

So my guess is that if you pull the latest version of my branch, the test should start passing for you.

This still doesn't explain why I'm having trouble running the *other* tests, but I guess it must be something unrelated...

robnagler:
Leaving it as completed is fine. Could also fix the code.

> Nice catch on the `complete` vs `completed` thing.

Fail fast. Let the computer do the thinking. :)

robnagler:
I realized that that we should be testing JobStatus. That was the lack of fail fast. When read_status runs, it should validate against JobStatus.

njsmith:
The new code does validate against `JobStatus`, both when reading from the run_dir in `runner.py`, and when parsing off the wire in `runner_client.py`. The problem here is that we were correctly validating against the wrong values...

njsmith:
I updated this so that the test automatically skips itself if you don't have the new py3 environment stuff set up. I guess eventually we'll want to make the py3 environment a standard thing, but I don't want to break your test suite in the mean time if this is merged :-).

robnagler:
pkdp and print are not allowed. This patch converts to pkdlog.

<details>

```patch
diff --git a/sirepo/pkcli/runner.py b/sirepo/pkcli/runner.py
index 82654a3d..c9a99adb 100644
--- a/sirepo/pkcli/runner.py
+++ b/sirepo/pkcli/runner.py
@@ -264,11 +264,11 @@ async def _handle_conn(job_tracker, lock_dict, stream):
             request = pkjson.load_any(request_bytes)
             if 'run_dir' in request:
                 request.run_dir = pkio.py_path(request.run_dir)
-            pkdp('runner request: {!r}', request)
+            pkdlog('runner request: {!r}', request)
             handler = _RPC_HANDLERS[request.action]
             async with lock_dict[request.run_dir]:
                 response = await handler(job_tracker, request)
-            pkdp('runner response: {!r}', response)
+            pkdlog('runner response: {!r}', response)
             response_bytes = pkjson.dump_bytes(response)
         except Exception as exc:
             await stream.send_all(
diff --git a/tests/runner_test.py b/tests/runner_test.py
index c5a39d94..7ee0de61 100644
--- a/tests/runner_test.py
+++ b/tests/runner_test.py
@@ -35,8 +35,9 @@ def test_runner_myapp():

     fc = srunit.flask_client()

+    from pykern.pkdebug import pkdlog
     from sirepo import srdb
-    print(srdb.runner_socket_path())
+    pkdlog(srdb.runner_socket_path())

     pkio.unchecked_remove(srdb.runner_socket_path())

@@ -56,9 +57,9 @@ def test_runner_myapp():
             {'simulationType': 'myapp',
              'search': {'simulationName': 'heightWeightReport'}},
         )
-        print(data)
+        pkdlog(data)
         data = data[0].simulation
-        print(data)
+        pkdlog(data)
         data = fc.sr_get(
             'simulationData',
             params=dict(
@@ -67,7 +68,7 @@ def test_runner_myapp():
                 simulation_type='myapp',
             ),
         )
-        print(data)
+        pkdlog(data)
         run = fc.sr_post(
             'runSimulation',
             dict(
@@ -78,13 +79,13 @@ def test_runner_myapp():
                 simulationType=data.simulationType,
             ),
         )
-        print(run)
+        pkdlog(run)
         for _ in range(10):
             run = fc.sr_post(
                 'runStatus',
                 run.nextRequest
             )
-            print(run)
+            pkdlog(run)
             if run.state == 'completed':
                 break
             time.sleep(1)
```

</details>

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+        return {}
+    job_info = job_tracker.jobs[request.run_dir]
+    if job_info.status is not runner_client.JobStatus.RUNNING:
+        return {}
+    job_info.status = runner_client.JobStatus.CANCELED
+    _write_status(runner_client.JobStatus.CANCELED, request.run_dir)
+    job_info.process.terminate()
+    with trio.move_on_after(_KILL_TIMEOUT_SECS):
+        await job_info.finished.wait()
+    job_info.process.kill()
```
robnagler: Should this only happen if the process is alive? I assume it is a nop if it is dead, but seems a bit confusing.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+            if 'run_dir' in request:
+                request.run_dir = pkio.py_path(request.run_dir)
+            pkdp('runner request: {!r}', request)
+            handler = _RPC_HANDLERS[request.action]
+            async with lock_dict[request.run_dir]:
+                response = await handler(job_tracker, request)
+            pkdp('runner response: {!r}', response)
+            response_bytes = pkjson.dump_bytes(response)
+        except Exception as exc:
+            await stream.send_all(
```
robnagler: If there is an error on the stream itself, you can't send to it. It's safer to protect against that rather than raise another exception. 

I think it shouldn't send back the error string, and instead, just send back "error". It breaks the abstraction to respond with raw errors. Rather, just log the error here. If there's a known error, translate it to a known error code for the sender.

Logging might look like: `pkdlog('Request={} error={} stack=', request, exc, pkdexc())`

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                pkjson.dump_bytes({'error_string': repr(exc)})
+            )
+            # Let's also log the full error here
+            raise
+        else:
+            await stream.send_all(response_bytes)
```
robnagler: If there's a BaseException (e.g. SystemExit) it will send a response anyway. I don't think that makes sense, because we don't know that response_bytes is valid here. I think the send should be inside the try, and any BaseExceptions are simply raised implicitly.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+class _JobTracker:
+    def __init__(self, nursery):
+        # XX TODO: eventually we'll need a way to stop this growing without
+        # bound, perhaps by clarifying the split in responsibilities between
+        # the on-disk simulation_db versus the in-memory status.
+        self.jobs = {}
+        self._nursery = nursery
+
+    def status(self, run_dir, jhash):
+        disk_in_path = run_dir.join('in.json')
```
robnagler: Reading from disk in this case needs to be protected by a try and report an error or missing status.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+        disk_status_path = run_dir.join('status')
+        if disk_in_path.exists() and disk_status_path.exists():
+            disk_in_text = pkio.read_text(disk_in_path)
+            disk_jhash = pkjson.load_any(disk_in_text).reportParametersHash
+            if disk_jhash == jhash:
+                disk_status = pkio.read_text(disk_status_path)
+                if disk_status == 'pending':
+                    # We never write this, so it must be stale, in which case
```
robnagler: log this

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+        await self._nursery.start(self._run_job, run_dir, jhash, cmd)
+
+    async def _run_job(
+            self, run_dir, jhash, cmd, *, task_status=trio.TASK_STATUS_IGNORED
+    ):
+        # XX TODO: there are all kinds of race conditions here, e.g. if the
+        # same jid gets started multiple times in parallel... we need to
+        # revisit this once jids are globally unique, and possibly add
+        # features like "if there is another job running with the same
+        # user/sim/job but a different hash, then auto-cancel that other job"
```
robnagler: and a [newer serial](https://github.com/radiasoft/sirepo/blob/3f7a4ba4333e8649775a4eac6e18d724442114bc/sirepo/simulation_db.py#L859).

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+        with _catch_and_log_errors(Exception, 'error in run_job'):
+            if run_dir in self.jobs:
+                # Right now, I don't know what happens if we reach here while
```
robnagler: log this

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                # the previous job is still running. The old job might be
+                # writing to the new job's freshly-initialized run_dir? This
+                # will be fixed once we move away from having server.py write
+                # directly into the run_dir.
+                assert self.jobs[jhash] == jhash
+                return
+            try:
+                env = _subprocess_env()
+                run_log_path = run_dir.join(template_common.RUN_LOG)
+                # hack: switch back to py2 mode to run the actual command
```
robnagler: Not a hack. It's going to be what we do for a while. It will eventually need to be encapsulated, because the docker container will need the same code.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                # Right now, I don't know what happens if we reach here while
+                # the previous job is still running. The old job might be
+                # writing to the new job's freshly-initialized run_dir? This
+                # will be fixed once we move away from having server.py write
+                # directly into the run_dir.
+                assert self.jobs[jhash] == jhash
+                return
+            try:
+                env = _subprocess_env()
+                run_log_path = run_dir.join(template_common.RUN_LOG)
```
robnagler: I've been using hungarian notation `run_log_f` for file and `run_d` for directory. Long local variable names also don't have a lot of value, especially when they are just cached values like this.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                # hack: switch back to py2 mode to run the actual command
+                env['PYENV_VERSION'] = 'py2'
+                cmd = ['pyenv', 'exec'] + cmd
+                with open(run_log_path, 'a+b') as run_log:
+                    process = trio.Process(
+                        cmd,
+                        cwd=run_dir,
+                        start_new_session=True,
+                        stdin=run_log,  # XX TODO: should be subprocess.DEVNULL?
```
robnagler: definitely don't want to pass the run_log to stdin. It should be devnull. Nobody should be reading stdin.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                    )
+                self.jobs[run_dir] = _JobInfo(
+                    run_dir, jhash, runner_client.JobStatus.RUNNING, process
+                )
+                async with process:
+                    task_status.started()
+                    # XX more race conditions here, in case we're writing to
+                    # the wrong version of the directory...
+                    await process.wait()
+                    if process.returncode:
```
robnagler: If `p` is used for `process`, you can write this with better encapsulation:

```py
s = getattr(runner_client.JobStatus, 'ERROR' if p.returncode else 'COMPLETED')
_write_status(s, run_dir)
```

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                        env=env,
+                    )
+                self.jobs[run_dir] = _JobInfo(
+                    run_dir, jhash, runner_client.JobStatus.RUNNING, process
+                )
+                async with process:
+                    task_status.started()
+                    # XX more race conditions here, in case we're writing to
+                    # the wrong version of the directory...
+                    await process.wait()
```
robnagler: we probably want to log the process status in an error

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+                    if process.returncode:
+                        _write_status(runner_client.JobStatus.ERROR, run_dir)
+                    else:
+                        _write_status(runner_client.JobStatus.COMPLETED, run_dir)
+            finally:
+                # _write_status is a no-op if there's already a status, so
+                # this really means "if we get here without having written a
+                # status, assume there was some error"
+                _write_status(runner_client.JobStatus.ERROR, run_dir)
```
robnagler: definitely want to log this path

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+        # features like "if there is another job running with the same
+        # user/sim/job but a different hash, then auto-cancel that other job"
+        with _catch_and_log_errors(Exception, 'error in run_job'):
+            if run_dir in self.jobs:
+                # Right now, I don't know what happens if we reach here while
+                # the previous job is still running. The old job might be
+                # writing to the new job's freshly-initialized run_dir? This
+                # will be fixed once we move away from having server.py write
+                # directly into the run_dir.
+                assert self.jobs[jhash] == jhash
```
robnagler: in other places self.jobs is indexed by run_dir.

sirepo/pkcli/runner.py
```diff
@@ -0,0 +1,366 @@
...
+            finally:
+                # _write_status is a no-op if there's already a status, so
+                # this really means "if we get here without having written a
+                # status, assume there was some error"
+                _write_status(runner_client.JobStatus.ERROR, run_dir)
+                # Make sure that we clear out the running job info and tell
+                # everyone the job is done, no matter what happened
+                job_info = self.jobs.pop(run_dir, None)
+                if job_info is not None:
+                    job_info.finished.set()
```
robnagler: need to validate the jhash, because could be a different job

njsmith:
Thanks the review! I'll take a look later today.

robnagler:
I see that lock_dict serialized rpc calls. I think _run_job is in a separate thread, which means that the comments about jhash in _run_job are valid. If an exception is thrown, there's no guarantee that the _run_job thread owns the process. I assume that cancel goes through finally, which is the code I'm concerned about.

sirepo/pkcli/runner.py
njsmith: Yeah, probably clearer if we make that explicit.

sirepo/pkcli/runner.py
njsmith: > If there is an error on the stream itself, you can't send to it. It's safer to protect against that rather than raise another exception.

In Python 3 we have implicit exception chaining, so if `stream.send_all` raises an exception, then both the original exception+traceback and the new exception+traceback will both be logged.

Also, we're already logging any leaking exceptions with `_catch_and_log_errors`, so that's not an issue.

Easy enough to just send back "error" though, I'll do that.

njsmith:
> pkdp and print are not allowed. This patch converts to pkdlog.

In a pytest test, the normal behavior of `print` (and any output for that matter) is that pytest captures it by default, and only shows it on the console if the test fails. This is really handy because it means you can leave in lots of chatty prints to help debug when things do go wrong, but without creating a bunch of noise during regular runs. (E.g. that's part of how I could debug your failures in `runner_test.py`.

It looks like `pkdlog` somehow circumvents pytest's capturing, so the chatty prints are dumped to the console regardless of whether there's a problem or not.

Would you rather we remove the chatty prints to keep the tests silent, keep them to make debugging easier, or figure out a way to get pytest's output capturing to work properly?

sirepo/pkcli/runner.py
njsmith: Actually, we might as well delete this bit and make the error indicator be "empty response" – adding fancier error handling code just produces more chances for things to go wrong :-)

njsmith: > In Python 3 we have implicit exception chaining, so if `stream.send_all` raises an exception, then both the original exception+traceback and the new exception+traceback will both be logged.

...actually, I'm wrong, because `pkdexc` is buggy on py3 :-( https://github.com/radiasoft/pykern/issues/29

sirepo/pkcli/runner.py
njsmith: the `else` in `try/else` only executes if there were no exceptions, so the old behavior on `BaseException` was to send nothing at all.

But since I just removed the `except` block in response to one of your previous comments, the `try` is now pointless, and I got rid of it entirely. So now the main code and the `send_all` are all straight-line code inside the `_catch_and_log_exceptions` block.

sirepo/pkcli/runner.py
njsmith: We check that the files exist before trying to read them. If they exist but trying to read them raises an error, then something weird is going on and it's probably best to crash...

sirepo/pkcli/runner.py
njsmith: rewrote the comment based on newer understanding :-)

sirepo/pkcli/runner.py
njsmith: the hacky bit is that py2 is normally the default, but we're in this special py3 environment so we have to take some special effort to prevent processes we spawn from inheriting that. docker won't have the problem, because containers won't inherit the runner's environment.

sirepo/pkcli/runner.py
njsmith: The `_path` here is specifically to distinguish the path string from the file object. Using `_f` to mean "not a file object" seems confusing :-).

`run_dir` seems to be used pretty consistently in `server.py` and `simulation_db.py`, so I used it everywhere here too, but we can switch if you want...

sirepo/pkcli/runner.py
njsmith: if there's an unexpected exception, then `_catch_and_log_errors` is already logging it :-)

njsmith:
OK, I think I addressed most of your comments. The big thing left is that we still have race conditions around `start_job`. (The other race conditions are fixed by the `LockDict` stuff.) I think we mostly get away with it for now, because the frontend is still trying hard to avoid tickling the backend's race conditions. But we should fix them. This needs a holistic approach. I think the two invariants we want are:

* only `runner.py`, and the jobs it runs, are allowed to write to the `run_dir`
* there are never two jobs running at the same time in the same `run_dir`

Together, these will fix the race conditions on disk, and the race conditions in memory. (In particular, note that if we enforce that there are never two jobs running at the same time in the same dir, then all the issues around one job's shutdown overlapping with another job's startup and causing stuff to be scribbled on the wrong job_info object just goes away.)

This will require some rearranging:

* Right now after `server.py` decides that it needs to start a new job, it directly goes blows away the run_dir and writes new stuff into it. If two `server.py` threads decide to do this at the same time, then bad stuff happens. We need to switch to `server.py` writing the job info into a temp dir, and then pass that to `server.py` to set up the real `run_dir`.
  * There may be some other cases where `server.py` writes directly to `run_dir`, but I guess we'll deal with the rest later when we're figuring out how to handle NERSC...
* `runner.py` needs to get some logic around `start_job` to check for existing jobs and handle them appropriately. This has some complications to make sure we get right: the policy for handling existing jobs should involve the serial, I guess (?) – not really sure what that is, but it certainly sounds relevant now that you've mentioned it :-) – and if we decide to cancel the old job then we need to make sure to wait for it to fully exit before starting a new one.

This is all doable, but it's complicated enough that I think we should bump it into SOW-3.

sirepo/pkcli/runner.py
robnagler: > > If there is an error on the stream itself, you can't send to it. It's safer to protect against that rather than raise another exception.
> 
> Easy enough to just send back "error" though, I'll do that.

To be clearer: `stream` is in error, you shouldn't send on it, because you'll get a cascade that could be avoided. I feel like in error code, we have to take pains to avoid that type of cascade, because it makes debugging harder, even if py3 gives you all the tracebacks.

robnagler:
> This is all doable, but it's complicated enough that I think we should bump it into SOW-3.

I agree about the race condition and SOW-3.  I'd like to discuss the solution, because we have to start decoupling based on the runner class (NERSC, subprocess, or Docker).

A key factor for NERSC (coming up soon now) is that a "job" is really anything that touches run_dir. This is also key for the runner in general, because we want to run anything to any "codes" inside a separate container to ensure encapsulation and to allow sirepo to migrate to py3. Therefore, I think we need to generalize the job to "do something with this particular run_dir, which might already exist."

robnagler:
> In a pytest test, the normal behavior of `print` (and any output for that matter) is that pytest captures it by default, and only shows it on the console if the test fails. 

We should have that as a goal without print, because pytest doesn't tell you what line number the print is on. The reason for pkdp, pkdlog, and pkdc is that I don't have to write anything to explain what's going on, because you can jump right to the line and see what the code is actually doing (as opposed to what the string says it is doing :).  

> It looks like `pkdlog` somehow circumvents pytest's capturing, so the chatty prints are dumped to the console regardless of whether there's a problem or not.

There are numerous bugs around how pytest and pykern interact, e.g. [pytest breaks capsys](https://github.com/radiasoft/pykern/issues/22) and a bunch of nulls problem that I have yet to write up.  Anyway, I agree with the need for better logging. 

Let's leave in the pkdlogs as is for now, because you don't see any test output when running python setup.py tests, which is the most common mode. Having it chatty for an individual test run is tolerable until https://github.com/radiasoft/pykern/issues/31 is fixed.

njsmith:
> A key factor for NERSC (coming up soon now) is that a "job" is really anything that touches run_dir. This is also key for the runner in general, because we want to run anything to any "codes" inside a separate container to ensure encapsulation and to allow sirepo to migrate to py3. Therefore, I think we need to generalize the job to "do something with this particular run_dir, which might already exist."

I think we'll want to distinguish between the current kind of job, and the data extraction jobs, because they have fundamentally different interactions with the run dir lifecycle. For example, the current "status check" operation is basically a check of what kind of run dir we have; checking on the status of a particular data extraction job will need very different inputs and outputs.

> Let's leave in the pkdlogs as is for now

Sounds good.

Issue #1489, opened 2019-01-04
DOE acknowledgment for Warp PBA is incorrect.


bruhwiler:
On this page
https://beta.sirepo.com/#/warppba

In the DOE acknowledgment at the bottom, the text 
"Basic Energy Sciences under Award No. DE-SC0013855"
should be changed to
"High Energy Physics under Award No. DE-SC0013855"

PR #1508, opened 2019-01-25
Simulation copy limit, plus Issue #1489

mkeilman:
Notes:

- The check for missing sim name in ```server.api_copySimulation()``` and its associated loop is gone.  The copy limit check is done at validation (```simulation_db.uniquify_name()```).  I left an assert in there in case I missed some copy mechanism that fails to supply a name
- ```server.api_copySimulation()``` also has a bit of refactoring to use dot notation for Dict fields
- This implementation also catches new sims and renamed sims that end with too high a number , regardless of the base name ("Sim 101" will fail even if nothing is already named "Sim").  One problem I'm still working on is that a bad rename fails silently (other than a javascript error) - I'll make an issue for that
- The same "template" is now used for both server and client (&lt;base name&gt; &lt;copy num&gt;)
- I put the limit in the schema so the client side can also use it - seems like we could invalidate the form beforehand

One meta-note: since Issue #1489 has an assigned priority, I added it to this PR as well.  I plan to do this from now on, using the highest priority of issues addressed


sirepo/simulation_db.py
```diff
@@ -1241,37 +1248,51 @@ def _validate_fields(data):
...
+    sim_id = s.simulationId
     n = s.name
+    f = s.folder
+    c = None
+    m = re.search(r'(.+) ([0-9]+)$', n)
+    if m:
+        n = m.group(1)
+        c = int(m.group(2))
+    if c is not None:
+        _assert_sim_copy_num(s.name, c)
```
robnagler: I don't think this is correct. The user should be able to name a sim anything they want. 

The goal was to eliminate a (potential) bug when the sim copy loop got to maxSimCopies, which it should never do.

sirepo/simulation_db.py
```diff
@@ -855,6 +855,8 @@ def user_dir_name(uid=None):
         return d
     return d.join(uid)
 
+def validate_name(data):
+    _validate_name(data)
```
robnagler: inline _validate_name(). I'm not sure why there's a need for a separate function.

sirepo/package_data/static/json/schema-common.json
```diff
@@ -105,7 +105,8 @@
         },
         "constants": {
             "clientCookie": "sirepo_cookie_js",
-            "oneDayMillis": 86400000
+            "oneDayMillis": 86400000,
```
robnagler: alphabetical order

sirepo/simulation_db.py
```diff
@@ -1241,37 +1248,51 @@ def _validate_fields(data):
...
+        n2 = '{} {}'.format(proposed_name, i)
         i += 1
-    data.models.simulation.name = n2
+    _assert_sim_copy_num(proposed_name, i - 1)
+    return n2
+
+
+def _validate_name_uniquify(data, starts_with):
+    """Uniquify data.models.simulation.name"""
+    data.models.simulation.name = _uniquify_name(data.models.simulation.name, starts_with)
```
robnagler: no need for a separate function. validate_name() should just do all the work.

sirepo/simulation_db.py
mkeilman: I had been calling this from server.py and forgot to delete once that went away

sirepo/simulation_db.py
robnagler: fewer interfaces are better. it reduces cognitive load.

mkeilman:
The latest checkin fixes #1509 - an error is expressed and the offending sim name reverted.  Since the autosave fails we don't need to re-save the simulation, just update the UI.

Issue #1194, opened 2018-05-16
Example folder for all apps


moellep:
All apps should have an Examples folder - existing examples should be moved into the folder via a db-upgrade. SRW examples should remain in their current locations. 

- [x] elegant
- [x] JSPEC
- [x] hellweg
- [x] shadow
- [x] Synergia
- [x] Warp PBA
- [x] Warp VND
- [x] zgoubi

messamer:
Can we please add the example folder for elegant?  I need it for work with Vermilion.

PR #1558, opened 2019-03-22
WIP #1194 - modified json and fixup for elegant

mkeilman:
If the solution is suitable then I will apply it to the other apps, but I wanted this available for @messamer first


sirepo/package_data/template/elegant/examples/bunchCompLSC.json
```diff
@@ -3932,7 +3932,7 @@
         "rpnVariables": [],
         "simulation": {
             "activeBeamlineId": 28,
-            "folder": "/",
+            "folder": "/Examples",
```
moellep: I'd recommend leaving the example files unchanged - the fixup will automatically correct them. That makes less change in the pull request as well.

Issue #1562, opened 2019-03-26
Avoid unqualified function imports


robnagler:
There should very few exceptions to the rule that you import modules, not module attributes. 

This task is to fix [this specific case](https://github.com/radiasoft/sirepo/blob/50d75e76ce03ca14681bc1458482bb79b334e64c/sirepo/simulation_db.py#L18), and look for other cases.

mkeilman:
To clarify: this is to replace all ```from x import y``` with ```import x``` and qualify all the uses in the code (```y``` -> ```x.y```)?

robnagler:
Not exactly. Replace all instances of `from sirepo.x import f1, f2` with `from sirepo import x` and qualify all uses to `x.f1()`. That is, you only need to qualify the import if it is a direct function execution.

Here's the reasoning: If I see `f1()` with this approach, I can know to look for the definition of `f1` inside the current module. If I see `x.f1()`, I know it is defined in module `x`. While one can use "etags" or other code traversal tools, it's still a cognitive load to know if a function is defined in one module (file) or not just by it's simple name. 

If you use always use qualified names, then you can change, say, `schema.validate_schema` to `schema.validate`, which is the same length as an unqualified `validate_schema`, but more informative and less cognitive load. I see `schema.validate`, and I know that there's a module that validates the schema (in general). That's valuable to me, at least, in thinking about how the application is put together.

In the WIP commit, there's this:

```py
import pykern as pk
import pykern.pkdebug
import sirepo as sr
import sirepo.template as tmplt
import sirepo.schema as sch
```

This is worse. First, always remember what [Ken Thompson said about `creat` in Unix](https://books.google.com/books?id=poFQAAAAMAAJ&q=%22spell+creat+with+an+e%22&dq=%22spell+creat+with+an+e%22&hl=en). Second, you've created a new name. There are [two problems in computer science](https://hilton.org.uk/blog/why-naming-things-is-hard), and naming is one of them. Aliasing creates problems both in program optimization and development time minimization. I now have to learn a new name. Third, this didn't pick up the gist of what this Issue is about: reducing cognitive load.

mkeilman:
OK, I jumped the gun there - I'll start over

mkeilman:
Latest commit is all I could find, except for:

- classes (like ```from flask_sqlalchemy import SQLAlchemy``` in oauth)
- lots of ```from pykern.pkdebug import pkdc, pkdexc, pkdlog, pkdp,...```

[...]

PR #1573, opened 2019-03-29
WIP #1562 - classes now qualified

mkeilman:
So things like ```from module.submodule import MyClass``` / ```c = MyClass()``` become ```from module import submodule``` / ```c = submodule.MyClass()```


sirepo/template/hellweg.py
```diff
@@ -98,17 +99,16 @@ def extract_beam_report(report, run_dir, frame):
 
 
 def extract_parameter_report(report, run_dir):
-    from rslinac.solver import BeamSolver
-    solver = BeamSolver(
+    beam_solver = solver.BeamSolver(
```
robnagler: I don't find that naming the instance `beam_solver` does much, especially with all the repetition. You could just call it `s`, and get as much clarity in the code.

sirepo/template/synergia.py
```diff
@@ -660,10 +660,10 @@ def _generate_parameters_file(data):
 
 
 def _import_bunch(lattice, data):
-    from synergia.foundation import pconstants, Reference_particle, Four_momentum
+    from synergia.foundation import pconstants
     if not lattice.has_reference_particle():
         # create a default reference particle, proton,energy=1.5
-        lattice.set_reference_particle(Reference_particle(pconstants.proton_charge, Four_momentum(pconstants.mp, 1.5)))
+        lattice.set_reference_particle(foundation.Reference_particle(pconstants.proton_charge, foundation.Four_momentum(pconstants.mp, 1.5)))
```
robnagler: This line is too long:

```py
lattice.set_reference_particle(
    foundation.Reference_particle(
        pconstants.proton_charge, 
        foundation.Four_momentum(pconstants.mp, 1.5),
    ),
)
```

sirepo/beaker_compat.py
```diff
@@ -25,7 +25,7 @@ def update_session_from_cookie_header(header):
     """
     maps = _init_maps()
     try:
-        cookie = SignedCookie(cfg.secret, input=header)
+        cookie = session.SignedCookie(cfg.secret, input=header)
```
robnagler: I think it would be clearer in this case to have `beaker.session.SignedCookie` since it is only used once, and we have a concept of `session` elsewhere in the code. In general, a fully qualified import in a single use is probably better than a partially qualified reference. It's obviously a judgement call, but I think it's worth changing in this case.

sirepo/beaker_compat.py
```diff
@@ -7,7 +7,7 @@
 """
 from __future__ import absolute_import, division, print_function
 
-from beaker import session
+from beaker.session import SignedCookie
```
robnagler: This is opposite of what I wrote. I'm confused.

mkeilman: Oops, I put it back the way it was instead of fully qualifying, one sec...

PR #1578, opened 2019-03-30
Python refactor import misc

mkeilman:
Notes:

- Looks like some zip tests are duplicated and can be removed from ```template_common_test.py```, but I'll defer that for the moment
- Not sure about ```scipy.ndimage.zoom``` - it seems like the try-except is specifically there for the zoom import...???


robnagler:
I'd ignore the image resizing code for now.

PR #1701, opened 2019-05-29
Webcon fft fixup

mkeilman:
Notes:

- Changed label to frequency from angular frequency
- The x units are now "1/s".  Note that normally we re-label the x-axis of plots as we zoom in to keep the tick values between 1-10 or so, but this is confusing with things like inverse units (that's why the units were "s-1" - the label would otherwise change to e.g. "m1/s").  The solution here is to preserve the scale of the axis regardless of zoom level
- y values normalized to the largest peak


bruhwiler:
If you used Hz instead of s^-1 then you could handle the units with the usual logic.

PR #1707, opened 2019-06-04
In dev env, use local python env even when running in docker

njsmith:
This is technically part of SOW-4, but it's self-contained and
ready-to-review. Up to you whether you want to treat it as its own PR
or wait for the rest of SOW-4.

### DIFF
--- a/sirepo/base_pkconfig.py
+++ b/sirepo/base_pkconfig.py
@@ -6,6 +6,9 @@ u"""Default config
 """
 from __future__ import absolute_import, division, print_function
 
+# Note: sirepo intentionally keeps this file empty. For discussion see:
+#   https://github.com/radiasoft/sirepo/pull/1707#discussion_r290323080
+
 def alpha():
     return {}
 
@@ -15,15 +18,7 @@ def beta():
 
 
 def dev():
-    return {
-        'sirepo': {
-            'runner_daemon': {
-                'docker_process': {
-                    'dev_env_in_container': True,
-                },
-            },
-        },
-    }
+    return {}
 
 def prod():
     return {}
--- a/sirepo/base_pkconfig.py
+++ b/sirepo/base_pkconfig.py
@@ -15,7 +15,15 @@ def beta():
 
 
 def dev():
-    return {}
+    return {
+        'sirepo': {
+            'runner_daemon': {
+                'docker_process': {
+                    'dev_env_in_container': True,
+                },
+            },
+        },
+    }
 
 def prod():
     return {}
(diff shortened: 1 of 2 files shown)


sirepo/base_pkconfig.py
```diff
@@ -15,7 +15,15 @@ def beta():
 
 
 def dev():
-    return {}
+    return {
```
robnagler: I like that you read up on pkconfig. Unfortunately, I am thinking of getting rid of base_config.py, because no one uses it, and we tend to use dynamic defaults ([example in pykern](https://github.com/radiasoft/pykern/blob/582f61bbc3eead69b01b26c17759f6c5f0efea13/pykern/pkcli/github.py#L244)). I would like your opinion on that.

In BOP, we used this committed config in a central file quite a bit, and it's fine, but there's just less config in Sirepo, and more hard coding -- for better or worse.

sirepo/runner_daemon/docker_process.py
```diff
@@ -155,3 +186,8 @@ def __init__(self, container, backend_info):
         )
     finally:
         await trio.run_sync_in_worker_thread(container.remove)
+
+
+cfg = pkconfig.init(
+    dev_env_in_container=(False, bool, 'mount host dev environment into container'),
```
robnagler: Maybe document this as "mount host ~/src and ~/.pyenv into container"

sirepo/base_pkconfig.py
njsmith: The dynamic defaults version seems fine to me. It reduces the number of files you have to cross-reference to figure out what's going on with a given option, which is nice. (OTOH I guess it increases the number of files you have to look at to figure out what's going on with a given channel.)

I'm happy to do it either way; I was just trying to make a plausible guess about what would fit in with your style :-)

robnagler: Let's go with "dynamic defaults". While you are in base_config.py, put a comment somewhere that says don't use this file. That will help others. Thanks.

njsmith:
Review addressed

sirepo/runner_daemon/docker_process.py
```diff
@@ -189,5 +189,9 @@ def __init__(self, container, backend_info):
 
 
 cfg = pkconfig.init(
-    dev_env_in_container=(False, bool, 'mount host dev environment into container'),
+    mount_dev_env_into_container=(
```
robnagler: Nit and don't fix now, just noting: when something returns a bool (like channel_in), you don't need to convert to a bool. If you are unsure, you can say `bool(expr)` if you care. If it is an inverse case, you can say `not bool(expr)`. Using an `if` introduces an unnecessary cognitive load.

Issue #1739, opened 2019-06-25
warpvnd: Allow different reports to use same simulation results


mkeilman:
With complicated conductors in play, highly granular cells mean that the field and field comparison reports can run for a long time, like hours.  They actually generate the same ```parameters.py``` file, so the exact same simulations get run twice.  I think we need a way to group reports so that they can use the data of their siblings.  elegant's bunch model does not work here because there is a single report, run once, that related reports sift, whereas this is two reports with different models.  It will be along the same lines, though, I expect.

robnagler:
We definitely need to distinguish between a simulation run and a report run. These are already being handled separately in the new runner. It would be great if we came up names for these things, including the directories in which they run.

moellep:
If a report takes longer than 5 minutes or so, it should be a simulation
instead (called animation because the first case of this was the warppba
animation frames). Reports will time-out otherwise and they also get
canceled if you leave the page. For warpvnd, we'll need to move some of the
3d reports into an animation. Animations also have the advantage that they
can show % completion and can have multiple plots all out of the same
directory - JSPEC is a simple example of this.

Paul

robnagler:
I think we need to distinguish between:

* start simulation
* real-time status of the simulation
* start extract of report
* status of report
* return of report

These are all different messages. A simulation may have multiple reports. A report may take a long time to extract. All of these operations need to be handled asynchronously by the client and the server.

robnagler:
To be clear, a "simulation" in Sirepo terms is the db for a collection of "simulation data and runs". For now, we have run a "job", but reports in the released runner are not jobs, and they need to be. The new runner has a concept of "extract job". There's asynchrony in extract job, which is [synchronized in the daemon](https://github.com/radiasoft/sirepo/blob/879429fd28b6c0b7137f63847b4ea10aa9c0c1bd/sirepo/runner_daemon/docker_process.py#L137) and shouldn't be.

PR #1744, opened 2019-07-03
Fete 3d animation

mkeilman:
Notes:

- This replaces the field and field comparison reports with "animations" so that the reports dip from the same simulations results (addressing #1739 ). This is for both 2d and 3d at the moment, as allowing two different report types depending on that was getting a little convoluted
- Consider this a starting point for moving from the historical report/animation duality to a unified approach (again see the discussion for #1739). I've gone with "calculation" or "calc" as a possible description distinct from "simulation"
- 3d conductor grid now has carats in the bottom plot
- Field comparison report allows selection of the y component of the field in 3d
- Heatmaps allow for optional min/max to set the range of values.  Useful for cross-sections that are uniform but vary along the other axis


mkeilman:
Setting the global min/max works for uniform cross-sections but washes out small variations.  Latest commit is an attempt to mitigate that.  I wonder if some kind of sensitivity dial would be useful...

PR #1775, opened 2019-07-25
add transmission Image to sample field

ahebnl:
1, add transmission Image to sample field
2, modify the name to Electron Beam with Arbitray Magnetic Field;
3, modify something about the "electron beam" and 'undulator'. still working on this.

### DIFF
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -1236,7 +1236,7 @@ def _fixup_beamline(data):
                 item['verticalCenterCoordinate'] = _SCHEMA['model']['sample']['verticalCenterCoordinate'][2]
             if 'cropArea' not in item:
                 for f in ['cropArea', 'areaXStart', 'areaXEnd', 'areaYStart', 'areaYEnd', 'rotateAngle', 'rotateReshape',
-                          'cutoffBackgroundNoise', 'backgroundColor', 'tileImage', 'tileRows', 'tileColumns',
+                          'cutoffBackgroundNoise', 'backgroundColor', 'tileImage', 'tileRows', 'tileColumns', 'transmissionImage',
                           'shiftX', 'shiftY', 'invert', 'outputImageFormat']:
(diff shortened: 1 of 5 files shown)


sirepo/template/srw.py
```diff
@@ -1772,6 +1772,7 @@ def _process_image(data):
         s = srwl_uti_smp.SRWLUtiSmp(
             file_path=path,
             area=None if not int(m['cropArea']) else (m['areaXStart'], m['areaXEnd'], m['areaYStart'], m['areaYEnd']),
+            extTr=int(m['transmissionImage']),
```
moellep: The extTr field doesn't exist on the srwl_uti_smp.SRWLUtiSmp module - this is used only when previewing the sample by selecting the "Download Processed Image" (the filter icon) next to the sample file name. I don't think the extTr value applies in that case.

Screen Shot 2019-07-26 at 11 03 13 AM

sirepo/template/srw.py
```diff
@@ -1236,7 +1236,7 @@ def _fixup_beamline(data):
                 item['verticalCenterCoordinate'] = _SCHEMA['model']['sample']['verticalCenterCoordinate'][2]
             if 'cropArea' not in item:
                 for f in ['cropArea', 'areaXStart', 'areaXEnd', 'areaYStart', 'areaYEnd', 'rotateAngle', 'rotateReshape',
-                          'cutoffBackgroundNoise', 'backgroundColor', 'tileImage', 'tileRows', 'tileColumns',
+                          'cutoffBackgroundNoise', 'backgroundColor', 'tileImage', 'tileRows', 'tileColumns', 'transmissionImage',
```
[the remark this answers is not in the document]
ahebnl: thanks for this revision!

sirepo/template/srw.py
ahebnl: yes! you are correct! thanks for your revision. :)

PR #1783, opened 2019-07-29
Landing page code links

mkeilman:
Note that those intermediate landing pages are no longer reachable (but not deleted)


bruhwiler:
The intermediate landing pages are still reachable.
You just click on the "Learn More" buttons, which are shown in this screenshot:

image

PR #1827, opened 2019-08-20
Fix #1802 purge user logic to handle guest users stored in db

### LINKED ISSUE #1802: Fix admin.purge_users to look for guest users
After the recent release and update of examples, all users got migrated into the db so the algorithm in [all_uids](https://github.com/radiasoft/sirepo/blob/3b446b020c9e696b77373b2d763c66c66dc6d291/sirepo/auth_db.py#L30) (used by [purge_users](https://github.com/radiasoft/sirepo/blob/3b446b020c9e696b77373b2d763c66c66dc6d291/sirepo/pkcli/admin.py#L42)) for detecting guest users to purge no longer works. We need a different algorithm that reads the db records found to see if the user is a guest.

e-carlin:


sirepo/auth_db.py
```diff
@@ -27,10 +27,10 @@
 thread_lock = threading.RLock()
 
 
-def all_uids():
+def guest_uids():
     with thread_lock:
         res = set()
-        for u in UserRegistration.query.all():
+        for u in UserRegistration.query.filter_by(display_name=None):
             if u.uid:
```
robnagler: don't need this test, uid is a not null constraint

sirepo/auth_db.py
```diff
@@ -27,10 +27,10 @@
 thread_lock = threading.RLock()
 
 
-def all_uids():
+def guest_uids():
     with thread_lock:
         res = set()
-        for u in UserRegistration.query.all():
+        for u in UserRegistration.query.filter_by(display_name=None):
```
robnagler: encapsulate this in `auth`, e.g. auth.guest_uids() and this function can go away. It would be a better implicit coupling with the concept that guests don't have display_names (currently only in `auth`).

sirepo/srtime.py
```diff
@@ -18,13 +18,11 @@
...
 _epoch = datetime.datetime.utcfromtimestamp(0)
 
-
-@api_perm.allow_visitor
-def api_adjustTime(days=None):
+def adjust_time(days):
     """Shift the system time by days
 
     Args:
-        days (str): must be integer. if None, no adjustment.
```
robnagler: why did you remove this comment?

sirepo/auth_db.py
```diff
@@ -67,6 +58,11 @@ def search_by(cls, **kwargs):
             with thread_lock:
                 return cls.query.filter_by(**kwargs).first()
 
+        @classmethod
+        def search_all(cls, **kwargs):
+            with thread_lock:
+                return cls.query.filter_by(**kwargs)
```
robnagler: Returning an iterator is not going to work with the thread_lock. The iteration needs to occur within the lock. There are two options: return an array or pass in an operator (code). 

I recommend redefining the function as follows:

```py
@classmethod
def search_all_for_column(cls, filter_by, column):
    with thread_lock
        return [getattr(r, column) for r in cls.query.filter_by(**filter_by)]
```

Maybe the name is wrong, but this is specifically what you need for guest_uids.

tests/pkcli/admin_test.py
```diff
@@ -18,27 +18,32 @@ def test_purge_users(monkeypatch):
...
-    pkok(dirs[0].check(dir=True), '{}: nothing deleted', res)
-    res = admin.purge_users(days=1, confirm=True)
-    pkeq(dirs, res, '{}: no users registered so one delete', res)
+    srtime.adjust_time(adjusted_time) 
+
+    monkeypatch.setattr(auth, 'guest_uids', lambda: [])
+    res = admin.purge_guest_users(days=days, confirm=False)
+    pkeq([], res, '{}: no guest users so no deletes')
+
+    monkeypatch.setattr(auth, 'guest_uids', lambda: [uid])
```
robnagler: I don't think you need to patch this any more. I think guest_uids will work, because auth_db is initialized. You might try it.

tests/pkcli/admin_test.py
```diff
@@ -18,27 +18,32 @@ def test_purge_users(monkeypatch):
...
-    pkeq([], res, '{}: all users registered so no deletes')
-    monkeypatch.setattr(auth_db, 'all_uids', lambda: [])
-    res = admin.purge_users(days=1, confirm=False)
-    pkeq(dirs, res, '{}: no users registered so one delete', res)
-    pkok(dirs[0].check(dir=True), '{}: nothing deleted', res)
-    res = admin.purge_users(days=1, confirm=True)
-    pkeq(dirs, res, '{}: no users registered so one delete', res)
+    srtime.adjust_time(adjusted_time) 
+
+    monkeypatch.setattr(auth, 'guest_uids', lambda: [])
```
robnagler: I think you need to split in two test cases so you can monkeypatch the no uid case to be different from the case below.

sirepo/auth/__init__.py
```diff
@@ -143,6 +143,14 @@ def complete_registration(name=None):
     cookie.set_value(_COOKIE_STATE, _STATE_LOGGED_IN)
 
 
+def guest_uids():
+    """All of the uids corresponding to guest users."""
+    res = set()
```
robnagler: This should not be a set. It should be an array.

tests/auth/github_test.py
```diff
@@ -34,6 +34,13 @@ def test_guest_merge(monkeypatch):
     s = fc.sr_auth_state(isLoggedIn=True, method='guest')
     fc.sr_get('authGithubAuthorized', query={'state': state})
     fc.sr_auth_state(method='github', uid=guest_uid)
+    fc.sr_post(
```
e-carlin: @robnagler This "fixed" the test now that complete_registration() is no longer in srunit. I'm still not sure I understand the order of operations when a user logs in so please double check that this makes sense.

robnagler: When the test was written, it was assumed that all UserRegistration records had display_name so needCompleteRegistration was set for guest users. When someone converted from guest to github, they already had a display_name. Now a guest will login with github and be asked to complete registration because needCompleteRegistration will be true and force an exception if they try to do anything else.

e-carlin: Makes sense. Thanks.

PR #1846, opened 2019-08-27
the image rotation for "plot3d" report.

ahebnl:
We added the "image rotation" function for all "plot3d" reporting and we have tested it. Now we request a pull. Thanks a lot!

### DIFF
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -302,7 +302,6 @@ def extract_report_data(filename, model_data):
         schema_values = [e for e in schema_enum if e[0] == str(subtitle_datum)]
         if len(schema_values) > 0:
             subtitle = subtitle_format.format(schema_values[0][1])
-
     info = pkcollections.Dict({
         'title': title,
         'subtitle': subtitle,
@@ -1868,6 +1867,7 @@ def _process_undulator_definition(model):
 
 #def _remap_3d(info, allrange, z_label, z_units, width_pixels, scale='linear'):
 def _remap_3d(info, allrange, z_label, z_units, width_pixels, rotate_angle, rotate_reshape, scale='linear'):
+    pkdlog('anhe info keys {}',info.keys())
     x_range = [allrange[3], allrange[4], allrange[5]]
     y_range = [allrange[6], allrange[7], allrange[8]]
     ar2d = info['points']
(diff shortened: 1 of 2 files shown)


sirepo/template/srw.py
```diff
@@ -1898,17 +1898,39 @@ def _remap_3d(info, allrange, z_label, z_units, width_pixels, scale='linear'):
...
             pass
-    z_label = z_label
-    if z_units:
-        z_label += ' [' + z_units + ']'
+    # rotate 3D image    
+    if rotate_angle:
+        rotate_reshape = (rotate_reshape == "1")
+        try:
+            from scipy import ndimage
+            pkdlog('Size before: {}  Dimensions: {}', ar2d.size, ar2d.shape)
```
moellep: Rather than using pkdlog() for debugging information like this, you can use pkdc() which only shows up if you turn on tracing for the file.

For example, if you change it to pkdc() and run the server like this:

```shell
$ export PYKERN_PKDEBUG_CONTROL=srw.py
$ sirepo service http
```

The debug statements will now appear, but they won't appear unless tracing is enabled. That way, on production, the items won't get printed in the logs.

sirepo/template/srw.py
```diff
@@ -1898,17 +1898,39 @@ def _remap_3d(info, allrange, z_label, z_units, width_pixels, scale='linear'):
...
+            x_range[1] = shape_rotate[0]/shape_before[0]*x_range[1]
+            y_range[0] = shape_rotate[1]/shape_before[1]*y_range[0]
+            y_range[1] = shape_rotate[1]/shape_before[1]*y_range[1]
+            pkdlog('x_range and y_range after rotate is [{},{}] and [{},{}]', x_range[0], x_range[1], y_range[0], y_range[1])
+
+            x_range[2] = ar2d.shape[1]
+            y_range[2] = ar2d.shape[0]
+            if info['title'] != 'Power Density': info['subtitle'] = info['subtitle'] + ' Image Rotate {}^0'.format(rotate_angle)
+        except Exception:
+            pkdlog('Cannot rotate the image - scipy.ndimage.rotate() cannot be imported.')
```
moellep: This is the correct place to use pkdlog() - it indicates that something is wrong, but the software is continuing anyway. On a production server, this message would appear in the logs and indicate there is a problem.

sirepo/template/srw.py
```diff
@@ -319,7 +318,10 @@ def extract_report_data(filename, model_data):
     if _DATA_FILE_FOR_MODEL[rep_name]['dimension'] == 3:
         width_pixels = int(model_data['models'][orig_rep_name]['intensityPlotsWidth'])
         scale = model_data['models'][orig_rep_name]['intensityPlotsScale']
-        info = _remap_3d(info, allrange, file_info[filename][0][3], file_info[filename][1][2], width_pixels, scale)
+        rotate_angle = model_data['models'][orig_rep_name]['rotateAngle']
+        rotate_reshape = model_data['models'][orig_rep_name]['rotateReshape']
```
moellep: For both rotate_angle and rotate_shape, defaulting the values to 0 and '0' would allow a report to be created without requiring these values. One example is the mirrorReport which appears on the beamline mirror editor and doesn't have many report fields.

```python
        rotate_angle = model_data['models'][orig_rep_name].get('rotateAngle', 0)
        rotate_reshape = model_data['models'][orig_rep_name].get('rotateReshape', '0')
```

sirepo/template/srw.py
ahebnl: Thanks for the "debugging" updates!

sirepo/template/srw.py
ahebnl: Get it! Thanks a lot!

sirepo/template/srw.py
ahebnl: Thanks for your updates!

PR #1876, opened 2019-09-09
Feat/async runner

e-carlin:
Initial pass at creating the async runner. This includes the supervisor client, supervisor, driver, and local process. This code can run simulations end to end in a local environment. 

It does not handle many error and certainly could use better abstractions. I'm opening to start a discussion.

To run:
- `SIREPO_FEATURE_CONFIG_JOB_SUPERVISOR=1 sirepo service http`
- `sirepo job_supervisor start`

@robnagler this branch has been rebased on the latest master.


sirepo/pkcli/job_supervisor.py
```diff
@@ -0,0 +1,126 @@
...
+            _http_send(r, write_to_driver)
+            self._srserver_reqs.task_done()
+
+    async def _process_driver_request(self, req, write_to_driver):
+        pkdc('Processing driver request. Request: {}', req)
+
+        action_dispatch = {
+            job.ACTION_DRIVER_READY_FOR_WORK: self._process_driver_ready_for_work,
+        }
+        process_fn = action_dispatch.get(req.action) or self._process_driver_results
```
e-carlin: @robnagler I tried to remove the if and use dispatch but I couldn't come up with anything better than this which feels like an if in disguise. The problem I'm encountering with dispatch is I don't want to list all of the possible actions and the function they should call because in all cases except for one (action_ready_for_work) they are results so they should be handled by _process_driver_results(). How should I handle this?

PR #1939, opened 2019-10-04
Feat/aync runner v2

e-carlin:


robnagler:
I have made some mods. I'm going to finish this up with @e-carlin on Monday. I want to move the slot allocations down to the specific driver and requests management to the driverbase.

robnagler:
A couple of thoughts:

* get_kind should roll up resource_class and other information that describes the kind of driver instance, not just the class. For now, we should have get_kind() return parallel-local, sequential-local, etc. This would simplify the queue management to (kind, uid). The supervisor doesn't need to know more (for now at least). This solves the NERSC problem, because there may only be one kind of driver instance for NERSC, for example.

* We need to unit test the supervisor independently. We can have a driver.srunit which does nothing except transition state (there's no agent, just queueing callbacks). This will be the default driver for a supervisor unit test. We also will need to mock the messages, that is, the supervisor should be able to be callable outside the web server context, that is, Base.do(msg) will be called with a mocked content and handler. Each of the objects should also be called that way, too.

* I think we then can get status working independently. That's the simplest case, but exercises the decision processes: which driver kind, create the instance, enqueueing, dequeueing, etc. This would also allow us to exercise error pathways. The  msg.content would describe the pathways to exercise.

robnagler:
I merged with master so we could get new structure (runner_api and job_api split) in master. This should also fix runner_test. Removed trio dependency.

PR #2015, opened 2019-10-30
Fix #1987 Confirmation page for emailed link

### LINKED ISSUE #1987: Login issue to Sirepo
I was setting up runs on my lab windows desktop without issue and signed in sending lund@frib.msu.edu to make an account.  Then I tried loggin in via other computers (laptops first running linux, then osx, then a 2nd lab desktop running windows) and could not get it to accept my email and send me a link to login and see my files using lund@frib.msu.edu.    

Have I been blocked some way?!?!   

This has me puzzled because the 2 students I am working with could connect with a 2nd computer logged making a connection on another computer.  This is the way I thought it should work.  

Please check and see if I triggered some issue.  Sorry for the problem.  The bad part is I was explaining to the students could connect from anywhere and continue working on their simulations and I could not connect in a chain of tries.  But they thought it was funny and I opened a 2nd account under another email without issue.  However, I want to go back to using lund@frib.msu.edu 

Thanks! Steve

Professor Steven M. Lund
Director, US Particle Accelerator School / FNAL
Michigan State University / FRIB
517-908-7291
lund@frib.msu.edu

### LINKED ISSUE #2010: appModes.default.localRoute is not really default route
The default local route should not require parameters. appModes.default.localRoute is currently `source`, which requires a simulationId. You don't have that when you want to redirect to a default. There's another default local route: the route in localRoutes which is set to isDefault. That's the one that should be used in http_reply.

### LINKED ISSUE #2011: GUI redirects
sirepo.js needs to complete reload when authState changes. #1987 is a POST, not a GET, so a redirect (302) doesn't work. I added code to detect [when javascript-redirect.html is rendered](https://github.com/radiasoft/sirepo/commit/f7624cb6ee89dfcce1042a1d021499a7615487dd#diff-4700cc17d0bc603c0c85d5b5dbbba738R1679), but this doesn't detect the 302 case. We can make the window.location redirect more robust, but it's good enough for now.

In order to make this happen, we probably shouldn't call flask.redirect directly. Rather, call redirect in http_reply.redirect (moves us toward #1275). I don't think there's a case where `http_reply.gen_redirect_for_anchor` would not work.

robnagler:
* Fix #2011 All redirects now go through javascript-redirect.html
* Fix #2010 Local redirect defaults to correct route
* Added sirepo.util.SRException and Redirect (base is Reply)
* http_reply.gen_sr_exception and gen_user_alert have been replaced by gen_exception
* uri_router.call_api catches these exceptions and converts to responses
* Factored out sirepo.uri from http_reply and uri_router
* srunit now supports redirects properly (redirect=True by default)

### DIFF
--- a/sirepo/api_auth.py
+++ b/sirepo/api_auth.py
@@ -11,7 +11,7 @@ from pykern import pkinspect
 from sirepo import api_perm
 from sirepo import auth
 from sirepo import cookie
-from sirepo import http_reply
+import sirepo.util
 
 
 def assert_api_def(func):
@@ -31,7 +31,7 @@ def check_api_call(func):
     a = api_perm.APIPerm
     if expect in (a.REQUIRE_COOKIE_SENTINEL, a.REQUIRE_USER):
         if not cookie.has_sentinel():
-            return http_reply.gen_sr_exception('missingCookies')
+            raise sirepo.util.SRException('missingCookies', None)
         if expect == a.REQUIRE_USER:
             return auth.require_user()
     elif expect == a.ALLOW_VISITOR:
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -106,6 +106,7 @@ def api_authState():
     if pkconfig.channel_in('dev'):
         # useful for testing/debugging
         v.uid = u
+    pkdc('state={}', v)
     return http_reply.render_static(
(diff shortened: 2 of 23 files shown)


robnagler:
Please squash merge with the comment at the end of this pull request so we get that in the history.

Issue #2030, opened 2019-11-05
Remove versions from animation args


robnagler:
also move animation args to schema

Issue #2029, opened 2019-11-05
webcon doesn't use get_simuation_frame?


robnagler:
There's correctorSettingAnimation in get_simulation_frame, but I don't think this exists as it is not referenced in webcon.js. There is a correctorSettingReport.

@moellep is there no need for get_simulation_frame?

moellep:
It should exist, but doesn't. Currently the reports reach into the epicsServerAnimation directory for data, see template/webcon.py get_settings_report(). Instead they should be one-frame animations.

robnagler:
I'm fixing get_simulation_frame so I'm going to remove it (or comment it out  if you like).

Issue #2027, opened 2019-11-05
api_downloadDataFile abuses frameIndex


robnagler:
If the frameIndex is -1, then the model is assumed to be a sequential model. This would be fixed by #2026 doing the right thing.  The code is confusing in api_downloadDataFile, because it checks f >= 0, and the code in javascript set -1. Rather, if something is a sentinel, test for it explicitly. f == -2 would be an error.

Issue #2024, opened 2019-11-04
avoid login loops


moellep:
There are a few odd cases where login loops occur:

* User comes into the app with a cookie uid which isn't valid. The app tries to log them out which puts it into a login loop.

* User's user dir is deleted and visits the simulations page. 

In these cases the auth state should get reset.

moellep:
This problem still occurs in this case:

User is logged in. Delete the user's user directory on the server. Navigate to the simulations page or to an simulation. This causes the client to get into a login loop where the client thinks the user is logged in, but the server sends back a login redirect which causes a request loop.

Changing the requestSender.handleSRException to reload the app would fix this problem.

robnagler:
call_api is not saving the cookies when a response an exception. That's part of the problem.

robnagler:
It seems that browsers don't follow [RFC 6265](https://tools.ietf.org/html/rfc6265) and do not save cookies when a 400 is returned. There's very clear text about this: 

> User agents MAY ignore Set-Cookie headers contained in responses with 100-level status codes but MUST process Set-Cookie headers contained in other responses (including responses with 400- and 500-level status codes). 

SRException was using 400, which is also a major cause of a login loop. When it returns 200, it works. I'll change the code.

robnagler:
I don't think there are any more loops

PR #2031, opened 2019-11-05
Improving protocols for simulationFrame and downloadDataFile and related version issues

robnagler:
This is not ready yet, just a placeholder for these issues:

- [x] #1557 simulationFrame will have computeJobHash and computeJobStart
- [x] #1894 cache control private
- [x] #1938 sim_in replaces model_data and frame_args replaces data for get_simulation_frame
- [x] #1952 unit testing animations
- [x] #1985 sim_data.compute_model replaces simulation_db.compute_job_model and template.simulation_dir_name
- [x] #2017 api_downloadDataFile caching issue
- [x] #2020 getFrameCount returns too many frames
- [x] #2023 GUI reload on saveSimulationData with different version
- [x] #2024 avoid login loops
- [x] #2027 api_downloadDataFile abuses frameIndex (compute_model is idempotent)
- [x] #2029 webcon doesn't use get_simuation_frame?
- [x] #2030 remove versions from animation args (move to schema)
- [x] #2032 check inputs in apis
- [x] #2038 custom errors need to be consistent
- [x] #2040 samesite
- [x] #2041 verify_user_dir


robnagler:
This is ready for @moellep to test.

moellep:
Founds several problems:

- Login with email, after confirm button, no user icon dropdown is show. Actually "New sim", "New Folder" are also missing. Refreshing the browser fixes the problem.

Screen Shot 2019-11-15 at 11 54 58 AM

- elegant twiss report is broken on the lattice page. Select the "Twiss Parameters" buttons to reproduce.

- warpvnd. All simulations show server error on the source page.

- SRW run partially coherent report with the "cacluate degree of coherence" toggle selected, but the extra reports cause server errors.

- synergia bunch report no longer sharing a single simulation output.

- synergia visualization - running a simulation on the visualization page causes server error. See "IOTA 6-6 with NLInsert" example.

robnagler:
Thanks @moellep!!! All bugs fixed, hopefully. Added tests for many of the problems.

moellep:
I confirmed the bugs were fixed. Here are a few more:

- warpvnd: running the Field Calc Status on the Source page produces no output. It looks like it is running the wrong simulation because the Visualization tab has data after running the source sim. Use the "Two Poles" as the example.

- zgoubi: twiss tab seems to be running > 1 simulation. Compare the "AGS Booster" example twiss tab on dev vs prod.

- webcon: running the Beam Steering on the Controls tab no longer produces any data. It looks like the code depends on the epicsAnimation run dir which is no longer present.

- webcon: changed asserts into UserAlert but the errors still show as "server error" rather than the custom error message. For example, run the Custer plot with only one column selected should show an error below:

Screen Shot 2019-11-18 at 10 08 34 AM

robnagler:
@moellep more fixes so when you have a chance to look at it. Thanks.

moellep:
Almost fixed. Below are some issues:

- In zgoubi and synergia, there is a twissReport model which is used on the lattice page and a twissReport2 model which is used on the Twiss tab (for zgoubi) or the Visualization tab (for synergia). The lattice page twissReport depends on the selected twissReport model values and the simulation.activeBeamlineId which is currently selected. Right now if you run the twiss report on the lattice page and then switch the selected beamline, the report does not recalculate.

For example, in zgoubi, open the Proton Storage Ring example, select the Twiss Parameters button and then change the selected beamline and select the button again.

Screen Shot 2019-11-19 at 9 15 52 AM

- In zgoubi, on the Twiss tab, the optics report should get its data from the twissReport2 directory. It shouldn't create an opticsReport directory and rerun the simulation.

moellep:
Also, the 3D warpvnd example is failing. Run the warpvnd "Lattice Gate 3D" example on the visualization tab (it takes awhile). It shows a server error page when the simulation completes.

moellep:
When running the warpvnd "Gridded TEC 2D Opt" example, running the simulation on the Optimization tab causes a frameId javascript failure when it tries to display the plot:

Screen Shot 2019-11-19 at 9 51 21 AM

PR #2050, opened 2019-11-13
WIP: Job download data file

e-carlin:


sirepo/job.py
```diff
@@ -32,7 +32,8 @@
 #: path supervisor registers to receive requests from server
 SERVER_URI = '/server'
 
-AGENT_URI = '/agent'
```
robnagler: I think  this line shouldn't be deleted

sirepo/runner_api.py
```diff
@@ -23,6 +24,26 @@
...
+        type=simulation_type,
+    )
+    f, c, t = sirepo.template.import_module(sim.type).get_data_file(
+        simulation_db.simulation_run_dir(sim.req_data),
+        sim.sim_data.compute_model(sim.model),
+        int(frame),
+        options=sim.req_data.copy().update(suffix=suffix),
+    )
+    return http_reply.headers_for_no_cache(
+        http_reply.as_attachment(flask.make_response(c), t, f),
```
robnagler: This should be app.response_class, not flask.make_response. Encapsulate this complete in http_reply.

sirepo/job_api.py
```diff
@@ -29,6 +31,21 @@
...
 
+@api_perm.require_user
+def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=None):
+#TODO(robnagler) validate suffix and frame
+    sim = sirepo.http_request.parse_params(
+        id=simulation_id,
+        model=model,
+        type=simulation_type,
+    )
+    f = _request(data=PKDict(sim.req_data, frame=frame, computeJobHash='x')).file
```
robnagler: Need to check the response to be "ok", and only then, read file. Sometimes data files aren't there.

sirepo/job_api.py
```diff
@@ -89,4 +108,5 @@ def _request_data(kwargs):
         libDir=lambda: str(sirepo.simulation_db.simulation_lib_dir(b.simulationType)),
         mpiCores=lambda: sirepo.mpi.cfg.cores if b.isParallel else 1,
         userDir=lambda: str(sirepo.simulation_db.user_dir_name(b.uid)),
+        tmpDir=lambda: str(sirepo.simulation_db.tmp_dir())
```
robnagler: Don't always add tmp_dir see above.

sirepo/job_api.py
robnagler:
```py
with simulation_db.tmp_dir() as d:
     r = _request(data=(), tmpDir=d)
     if r.state != 'ok':
         return r
     f = werkzeug.secure_filename(r.file)
     return sirepo.http_reply.gen_file(d.join(f))
```

sirepo/simulation_db.py
```diff
@@ -535,6 +535,25 @@ def process_simulation_list(res, path, data):
     ))
 
 
+def read_data_file(filepath):
```
robnagler: This doesn't belong here. and use `mimetypes.guess_type(f.basename) or application/octect`

sirepo/job.py
e-carlin: This was a duplicate definition. See line 30.

sirepo/simulation_db.py
e-carlin: Ok, where should I put it? In job_api?

robnagler: Yes, job_api. It's very specific.

I think we need to rename JOB_FILE (for sim.lib_file) and FILE to LIB_FILE_URI and DATA_FILE_URI. Everything associated with DATA_FILE should be data_file, DataFile, DATA_FILE, or dataFile, not just FIleReq or file. There are two different paths for different problems so they need to be distinguished clearly.

e-carlin: Agreed. Renamed to DATA_FILE

robnagler:
Job download data file #2050

PR #2058, opened 2019-11-14
Issue/2054 async docker cmd

### LINKED ISSUE #2054: job_driver.docker._cmd needs to be async
Use tornado.process.Subprocess, because otherwise the supervisor will hang while Docker commands start, some of which can take quite some time.

e-carlin:
Fix #2054

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -133,6 +133,12 @@ class DriverBase(PKDict):
     def websocket_on_close(self):
        self._websocket_free()
 
+    def _agent_on_exit(self, returncode):
+        k = self.pkdel('kill_timeout')
+        if k:
+            tornado.ioloop.IOLoop.current().remove_timeout(k)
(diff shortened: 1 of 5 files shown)


robnagler:
I think we might want to make terminate be an "add_callback" so that terminate is not asynchronous (too complicated). What do you think?

e-carlin:
I'm not sure that we can make it asynchronous. At some point we'll need to call ioloop.stop(). But, we can only call that once all of the callbacks have executed. To do that we will have to await for them to finish which will mean terminate must be async.

robnagler:
In the case of kill it should be asynchronous.

For termination, we call docker rm/stop synchronously using Python subprocess, not tornado.suprocess.

Perhaps we simply have to have an async termination. We'll need to manage that with signals from systemd and tornado. Control-c (SIGINT) should output something like "shutting down", and  SIGTERM should do something similar for journald.

Issue #2052, opened 2019-11-13
cascade environment better in job_agent/process


robnagler:
`PYKERN_PKCONFIG_CHANNEL` needs to be cascaded through job_supervisor, to job_agent, to job_process.

We will want to pass SIREPO_SRDB_ROOT to job_process based on where that is for job_agent. On NERSC, it'll be somewhere else. 

We should pass relative names for userDir, etc. and then join them in the appropriate places. This is becoming more important with the file transfers going both directions.

Issue #2051, opened 2019-11-13
Switch to random_base62 instead of uuid


robnagler:
We're using these unique_key more frequently (file transfers) so I think it be better to 1) use the bytes more efficiently, and 2) be longer. uuid's are 2^8x16 of randomness for 36 bytes (the dashes are taking up space, too). random_base62 are 2^6x36 bytes of (cyptographic) randomness for 36 bytes, or roughly twice the density, while still being ascii (which is convenient). We'll use 36 bytes, just because.

Issue #2046, opened 2019-11-12
uploadFile removes existing library file on errors


robnagler:
api_uploadFile [calls](https://github.com/radiasoft/sirepo/blob/8543ed6f37a66a1871f838aa36cc0f7255cc2138/sirepo/server.py#L576) validate_file, which [calls](https://github.com/radiasoft/sirepo/blob/8543ed6f37a66a1871f838aa36cc0f7255cc2138/sirepo/template/srw.py#L556) srw_is_valid_file_type, which today [throws an exception](https://github.com/radiasoft/sirepo/blob/8543ed6f37a66a1871f838aa36cc0f7255cc2138/sirepo/sim_data/srw.py#L269) when the file_type argument is not found (I set it to zip, when it should have been undulatorTable), instead of returning an error. However, fixing that caused the discovery that if the file is invalid (or file type) api_uploadFile will have already [removed the original file via overwriting](https://github.com/radiasoft/sirepo/blob/8543ed6f37a66a1871f838aa36cc0f7255cc2138/sirepo/server.py#L574).

@moellep why doesn't it validate the file first before overwriting the original?

robnagler:
This is another case where consistent errors would be useful. There is no `state` on the reply, just an `error`.

robnagler:
synergia has 

```py
assert file_type == 'bunch-particleFile'
```

It should return an error.

Issue #2028, opened 2019-11-05
srw background import needs work for remote agents


robnagler:
The workflow needs to be changed.

pkcli.srw._run_srw writes sirepo-data.json directly, which is not possible on remote execution (NERSC, departmental clusters):

```py
        #TODO(pjm): assumes the parent directory contains the simulation data,
        # can't call simulation_db.save_simulation_json() because user isn't set for pkcli commands
        simulation_db.write_json('../{}'.format(simulation_db.SIMULATION_DATA_FILE), parsed_data)
        simulation_db.write_result({
            'simulationId': sim_id,
        })
```

This needs to change to:

```py
        simulation_db.write_result(PKDict(data=parsed_data))
```

The code in srw.js should be moved into job/runner_api.py. It current gets the data back from importFile, and then:

```js
                   function(data) {
                        $scope.isUploading = false;
                        if (data.error) {
                            $scope.fileUploadError = data.error;
                        }
                        else if (data.models.backgroundImport) {
                            $scope.isUploading = true;
                            appState.loadModels(data.models.simulation.simulationId, function() {
                                simulationQueue.addTransientItem(
                                    'backgroundImport',
                                    appState.applicationState(),
                                    handleStatus);
                            });
                        }
                        else {
                            hideAndRedirect(data.models.simulation.simulationId);
                        }
                    });
```

PR #2060, opened 2019-11-17
Fix #2044 sim_data encapsulates local/remote lib file access

### LINKED ISSUE #2044: sim_data.lib_files* for job_process
When sim_data.lib_files* is called in job_process (via prepare_simulation), it needs to request the file list (lib_files) and the individual files (lib_files_copy) via a remote interface. 

I suggest we create a credential (job_key) in the job supervisor that is passed to the job agent that is passed to sim_data (via cfg) tell it to do make HTTP POST requests instead of getting the files locally. These would be synchronous requests in job_process and handled asynchronously in the supervisor. 

I think the fastest path would be to let nginx handle the file transfers. This could be accomplished by using the job_key as a dynamically linked directory name to the lib directory for the job. When the job exits, the link is removed by the supervisor. For development, tornado would handle the send file as a static file. We'd just have the proxy configured to go right to the directory root instead of hitting flask.

#1901 copy_related_files

robnagler:
sim_data will fully abstract lib_files. There are a lot of uses of lib_files, but most have been normalized. 

- [x] #2052 cascade environment including sim_data.lib_file_uri
- [x] #2051 longer unique key
- [x] #2046 removing library files on errors
- [x] #2044 lib file encapsulate

A basename is equivalent to py.path().basename. There was some misuse of the term so I fixed those places. lib_file_name is the basename without file_type so there are routines for manipulating lib_file_name_with_type, lib_file_name_with_model_field, etc.

We'll need a db upgrade to remove duplicates in the lib and resource directories. sim_data will transparently access resource or lib files. If you upload a file it goes in the lib. If you a delete file, you can only do that in the lib. It doesn't really delete resources, because you can't. We may want to distinguish between deletable (lib) and undeletable (resource) in the GUI. Not pressing, it'll work as is.

Writes of json will be atomic. The sim_data code has to write a file list so we can use static files to access lib files so it was another place where a non-atomic write would have caused problems. The json writes are moved to util, and simulation_db calls them with the old semantics (except they are all atomic).

The with/out file_type is quite confusing, because sometimes it is a tuple (model, field) and other times a string file_type. I'm not fixing all those places, just adding routines that clearly define what is being expected. "type" means file_type, a string. 

There's a concept of lib_file_write_path, which will be asserted to be a server-side only access. This is for upload and delete and the beamlist. lib_file_abspath will be either a write_path, resource_path, or a remote_path (copied locally).  There will be several other server-side only accesses, which will be asserted.

Adding lib_file_exists reduces some thrashing (getting the file remotely to see if it exists).

[SRW importer fdir](https://github.com/radiasoft/sirepo/blob/8543ed6f37a66a1871f838aa36cc0f7255cc2138/sirepo/template/srw_importer.py#L47) is not being changed from libdir which is unavailable to getcwd. In looking at srwl_bl, it seems fdir is used for reading mostly, and in this case, we would only want to use it for reading. Therefore, any reading files should be in that directory. fdir could be set to '.', because the code joins it with os.getcwd anyway. Noting here that we we'll have to watch this.

More test code has been fixed. You can wrap_in_request with a user (by default).


robnagler:
This is not ready to merge.

robnagler:
@moellep All the lib resource files are now in `template/<code>/lib` so we can list them more easily. I think I moved them all properly, but would be good to check.

robnagler:
@moellep please try this out

moellep:
Here are the problems I found:

- Selecting "Self-Extracting Simulation" on the Simulations list causes a server error.

- Selecting "Export as Zip" on the Simulations list causes a server error.

- Opening any SRW example causes a server error.

- Renaming a folder causes a server error.

- Importing an STL file into warpvnd causes a "not found" alert. Use the warpvnd lib file "stl-file.lattice_gate.stl" as an example import file.

- On the synergia visualization page, if you run a simulation and it tries to read the data as it is getting written, it redirects to a Server Error page. The old behavior was to show a "report not generated" message. To reproduce, open the "IOTA 6-6 With NL Element" example and run the Visualization a few times. The error occurs if the client asks for the data while the simulation is writing to output.

old message:
Screen Shot 2019-11-22 at 4 10 02 PM

robnagler:
The only bug not fixed is the sirepo.mpi_cores=4 with animation_test. The test takes longer to run with mpi, since the default timeout is so short. There does seem to be a weird cascade in this case, that is, other tests fail after test_srw fails due to the timeout, but those tests don't fail on their own (with cores=4). I don't think it's worth chasing down.

@moellep I merged the job branch into this branch, because we need everything. I think it's ready to go. I added tests for each of the cases (although partial_animation_test is not all that repeatable, but it helped diagnose the error so I left it there)

moellep:
Those problems are fixed, except for the warpvnd stl upload, which I think is related to the new filename bug below:

- elegant file upload creates incorrect filename. For example, if you
  download the raw datafile from the source tab and then select SDDS
  Input file and import it, the software creates the file
  "elegant.bun.bunchFile-sourceFile", but the simulations are looking
  for "bunchFile-sourceFile.elegant.bun".

- run-cancel failure caused client server error page. In SRW, was viewing the Coherent Results tab on the Beamline tab and switched to the 3D beamline tab while it was running the simulation. The run-cancel was sent, and failed and then the client was redirected to a Server Error page. The client shouldn't see server failures on canceled jobs.

```
10.10.10.1 - - [26/Nov/2019 16:30:59] "POST /run-simulation HTTP/1.1" 200 -
../sirepo/uri_router.py:91:call_api api=<function api_runCancel at 0x7f6217e45cf8> exception=[No such file or directory]: rename(u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/status.pkio-tmp', u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/status') stack=ENOENT: [No such file or directory]: rename(u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/status.pkio-tmp', u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/status')
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/threading.py", line 774, in __bootstrap
    self.__bootstrap_inner()
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/SocketServer.py", line 599, in process_request_thread
    self.finish_request(request, client_address)
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/werkzeug/serving.py", line 325
, in handle
    rv = BaseHTTPRequestHandler.handle(self)
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/werkzeug/serving.py", line 360
, in handle_one_request
    return self.run_wsgi()
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/werkzeug/serving.py", line 302
, in run_wsgi
    execute(self.server.app)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/werkzeug/serving.py", line 290
, in execute
    application_iter = app(environ, start_response)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/flask/app.py", line 2463, in _
_call__
    return self.wsgi_app(environ, start_response)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/flask/app.py", line 2446, in w
sgi_app
    response = self.full_dispatch_request()
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/flask/app.py", line 1949, in f
ull_dispatch_request
    rv = self.dispatch_request()
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/flask/app.py", line 1935, in d
ispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/uri_router.py", line 224, in _dispatch
    return call_api(route.func, kwargs)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/uri_router.py", line 85, in call_api
    r = flask.make_response(f(**kwargs) if kwargs else f())
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/runner_api.py", line 61, in api_runCancel
    simulation_db.write_result({'state': 'canceled'}, run_dir=run_dir)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/simulation_db.py", line 875, in write_result
    write_status(result['state'], run_dir)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/simulation_db.py", line 893, in write_status
    status.encode(),
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkio.py", line 34, in atomic_write
    n.rename(path)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/py/_path/local.py", line 456,
in rename
    return py.error.checked_call(os.rename, self.strpath, target)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/py/_error.py", line 86, in che
cked_call
    raise cls("%s%r" % (func.__name__, args))
../sirepo/http_reply.py:233:_gen_exception_error unsupported exception=<class 'py.error.ENOENT'> msg=[No such fi
le or directory]: rename(u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/s
tatus.pkio-tmp', u'/home/vagrant/src/radiasoft/sirepo/run/user/X6K7SQXI/srw/U0lXMj9D/watchpointReport4/status')
10.10.10.1 - - [26/Nov/2019 16:30:59] "POST /run-cancel HTTP/1.1" 200 -
10.10.10.1 - - [26/Nov/2019 16:30:59] "GET /srw HTTP/1.1" 200 -
10.10.10.1 - - [26/Nov/2019 16:30:59] "GET /static/html/server-error.html?20191126.162817 HTTP/1.1" 200 -

**Comment by robnagler** [comment]
Fixed with tests:
- elegant file upload creates incorrect filename 00e2ee54
- client shouldn't see server failures on canceled jobs 1aa007e2a59b8e579e876fdb39f99cf19e74069d

**Comment by moellep** [comment]
SRW file import fails. To reproduce, select download source for the CHX beamline and them import it:
```
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/srw_importer.py", line 25, in <module>
    _SIM_DATA, SIM_TYPE, _SCHEMA = sirepo.sim_data.template_globals()
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/sim_data/__init__.py", line 84, in template_globals
    c = get_class(sim_type or pkinspect.module_basename(pkinspect.caller_module()))
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/sim_data/__init__.py", line 62, in get_class
    ) else type_or_data
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/__init__.py", line 49, in assert_sim_type
    assert is_sim_type(sim_type), 'invalid simulation type={}'.format(sim_type)
AssertionError: invalid simulation type=srw_importer
```

robnagler:
The importer_test doesn't test this because of the fact that backgroundImport is a two step process (https://github.com/radiasoft/sirepo/issues/2028). Once the back in srw_importer.py was fixed (simple), the bigger issue is that the tmp dir goes away after the call so the client is sending a reference to a file which no longer exists. This is a security issue, because the client could specify any file. I will fix this by running the simulation in place.

robnagler:
rs4pi import is broken for a similar reason to srw. Temporary files do not exist across requests. They are temporary for one request only. I am going to fix this problem as well.

robnagler:
rs4pi is not broken, but it's pretty tricky to see that prepare_for_client is called in importFile.

PR #2086, opened 2019-11-27
WIP: Sbatch

e-carlin:
Here is the code to submit jobs to the "sbatch" backend. I've marked the PR as WIP because the code has a bad case of "works on my computer" syndrome. I still think it is worth discussing while we iron out the bits to make it run on everyone's machine.

[Here are the steps](https://docs.google.com/document/d/1Ca_cJXJk2nd7x8SDIr9qiyiRAs1sFV1iLGfruFyuPGE/edit?usp=sharing) I took to setup slurm on the radiasoft centos7 vm. You'll see many scary commands (ex chmod 777) and conflicting steps. I took the quickest route to getting Slurm working and spared little time for doing things "right". Again slurm "works on my machine" but we should build a vm with it preinstalled and configured in a safer manner.

For running the code on my machine I used the radiasoft centos7 vm and installed slurm on it following the instructions above. I then make a few modifications to the  sirepo code to get everything talking. Some of these steps will become pkconfig parameters. The others will be solved by the GUI sending requests with a field marking the user's desire for the job to run on NERSC.
Steps:

1. [Point the server](https://github.com/radiasoft/sirepo/blob/23d6bd40b54c6bc8c34d1c3fd78f15eb160d77cb/sirepo/job.py#L48) to where the supervisor is running (for me that is on v.radia.run).
2. [Set it so the sbatch](https://github.com/radiasoft/sirepo/blob/23d6bd40b54c6bc8c34d1c3fd78f15eb160d77cb/sirepo/job_driver/__init__.py#L202) driver is used
3. Make sure that whatever machine the supervisor is running on has the ssh key of the [machine the agent will run on](https://github.com/radiasoft/sirepo/blob/23d6bd40b54c6bc8c34d1c3fd78f15eb160d77cb/sirepo/job_driver/sbatch.py#L92)
4. start the server and the job_supervisor
5. Watch as parallel jobs are executed by slurm:metal:


robnagler:
Was sort of able to get the job_agent started on Cori to see what issues are. pyenv is finicky, but specifying absolute path for job_agent seems to work. Need to build shifter image with job_process, and add shifter support. runDir, etc. need work.

PR #2095, opened 2019-12-05
Sbatch auth

e-carlin:
Basic authentication failure handling for sbatch. This propagates and auth failed message back to the gui but the gui doesn't do anything special with it.

Open questions:
1. What should the GUI do?
2. Do we want to catch state AUTH_FAILED in the server and raise a 401 to the GUI?

### DIFF
--- a/sirepo/job.py
+++ b/sirepo/job.py
@@ -72,6 +72,7 @@ RUNNER_STATUS_FILE = 'status'
 
 UNIQUE_KEY_RE = re.compile(r'^\w$')
 
+AUTH_FAILED = 'auth_failed'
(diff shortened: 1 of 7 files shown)


e-carlin:
Thinking about this more, do we want to generalize this for all drivers to be an "agent start failed" message and then attach a reason (ex auth_failed)? In the local and docker case right now we just log an error if the start fails and then everything hangs indefinitely.

robnagler:
I definitely like the "hang indefinitely" approach. ;-)

I think we need to make the supervisor robust to such errors. Let's talk today.

PR #2139, opened 2019-12-20
Job testing

e-carlin:
Changes:
- Add the concept of computeJobSerial
- Get the tests to run with SIREPO_FEATURE_CONFIG_JOB = 1 || 0
- Add tests for sbatch. They will be skipped if machine that the tests are being run on doesn't have sbatch installed.
- Fix broken bits encountered along the way (ex getDataFile works again)


tests/conftest.py
```diff
@@ -246,6 +249,50 @@ def pytest_configure(config):
     )
 
 
+@pytest.fixture
+def sbatch_animation_fc(fc):
+    import functools
+    fc.sr_animation_run = functools.partial(
```
robnagler: Not sure I like this. My bad for starting this trend. We could just add sr_animation_run to srunit. conftest.py seems like the wrong place. sbatch_animation_fc doesn't need to exist, because it should be a mode on all runs in the test file, that is, job_run_mode would be configured with flask_client and be an attribute on fc.

sirepo/job_supervisor.py
```diff
@@ -288,10 +289,10 @@ def res(**kwargs):
         if self.db.computeJobHash != req.content.computeJobHash:
             return PKDict(state=job.MISSING, reason='computeJobHash-mismatch')
         if (
-            self.db.computeJobStart and req.content.computeJobStart
-            and self.db.computeJobStart != req.content.computeJobStart
-        ):
-            return PKDict(state=job.MISSING, reason='computeJobStart-mismatch')
+                req.content.computeJobSerial and
```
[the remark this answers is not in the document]
e-carlin: Does emacs do this for you automatically? Mine doesn't  and pylint complains (with indent back by 4). I don't care one way or the other but if emacs does the "right" thing for you I want to make it do the same for me.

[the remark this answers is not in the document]
robnagler: I fixed Emacs to do this "correctly". PEP8 may define this, but I didn't see it. 

My thinking is this: always indent by four. It's too complicated to have 8 spaces sometimes. PEP8 defines hanging indent as 4, which is what I think this is.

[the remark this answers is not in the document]
e-carlin: Makes sense.

e-carlin:
@robnagler I believe I've addressed all of your comments.

tests/conftest.py
```diff
@@ -274,6 +274,13 @@ def _job_supervisor_setup(request):
         PYKERN_PKDEBUG_WANT_PID_TIME='1',
         SIREPO_FEATURE_CONFIG_JOB='1',
     )
+    if 'status_test' in request.module.__name__:
```
e-carlin: @robnagler env must've not been cascaded properly in a previous iteration of my work because the status_test was passing until recently. The supervisor needs to be brought up even though it wasn't until this commit. I think adding the module.__name__ test is poor abstraction (same as sbatch) but I'm not sure of a better way to do it. Thoughts?

robnagler: Odd, it seems to have not made it into master. This turns up nothing:

```sh
git log -S 'status' -- tests/conftest.py
```

The module.name approach is similar to [Aspect Oriented Programming](https://en.wikipedia.org/wiki/Aspect-oriented_programming) which I don't think should be used liberally, but at the same time, it's useful. We could add a decorator to the test case (function), which might make it more readable.

One of the bugs in pytest's fixtures is that a test can't configure them. You have to do tricks to make them configurable: have a fixture that is a wrapper that must be called to configure. I don't like any of those, which is why went with name-based configuration. 

I don't particularly like the magic trick of the fixtures being the name of the arg. I think it would simply be clearer to introduce global names from conftest.py that are prefixed with conftest or whatever so you know where they are defined. 

I think we should push ahead with what we have, and replace pytest with something less magical (more explicitly coupled).

e-carlin: Thanks for the hint on AOP. 

I'll continue on as is for now. Let me know when you look into a pytest replacement. I'm curious how you evaluate the options.

robnagler: We are running into the makemake problem with pytest. About half of conftest.py is about signaling pytest to do something. conftest.py itself is not a module so at can't be tested, which was a problem. We should just expand pkunit and pykern.pkcli.test to support our testing policies. It would be work to replace the fixtures, but it would allow us to clean things up quite a bit. For example, we could import pkunit's methods (pkok, pkeq, etc.) into the test module. We wouldn't have to battle pytest's capturing of stdout and stderr. Things like work_dir would be initialized appropriately. We would enforce no imports at the global level so that we could insure appropriate initialization of flask client, pkconfig, etc.

My guess it would be a day or two of work to rewrite the parts of pytest we need, and we'd be able to remove a lot of code in pkunit, conftest, and pkcli.test.

e-carlin: That makes sense. Those would be big improvements (especially the output capturing, I still run up against it sometimes).

PR #2172, opened 2020-01-09
Issue/2162 stop agent before start

### LINKED ISSUE #2162: all drivers need to manage restarts when agent already running
For example, if a docker container loses its web connection, it will try to reconnect, but the supervisor may try to restart that container before it can reconnect, and the supervisor should stop the container before starting the new one.

e-carlin:
Fix #2162

### DIFF
--- a/sirepo/job_driver/docker.py
+++ b/sirepo/job_driver/docker.py
@@ -119,41 +119,30 @@ class DockerDriver(job_driver.DriverBase):
             self.host.slots[self.kind].in_use -= 1
             self.has_slot = False
 
-    async def _agent_start(self, msg):
-        self._agent_starting = True
-        try:
-            cmd, stdin, env = self._agent_cmd_stdin_env()
+    async def _do_agent_start(self, msg):
+        cmd, stdin, env = self._agent_cmd_stdin_env()
 #TODO(robnagler) remove PKDict after https://github.com/radiasoft/pykern/issues/50
-            c = PKDict(pykern.pkcollections.map_items(cfg[self.kind]))
-            p = (
-                'run',
-                # attach to stdin for writing
-                '--attach=stdin',
-                '--cpus={}'.format(c.get('cores', 1)),
-                '--init',
-                # keeps stdin open so we can write to it
-                '--interactive',
-                '--memory={}g'.format(c.gigabytes),
-                '--name={}'.format(self._cname),
-                '--network=host',
-                '--rm',
-                '--ulimit=core=0',
-                '--ulimit=nofile={}'.format(_MAX_OPEN_FILES),
-                # do not use a "name", but a uid, because /etc/password is image specific, but
-                # IDs are universal.
-                '--user={}'.format(os.getuid()),
-            ) + self._volumes() + (self._image,)
-            self._cid = await self._cmd(p + cmd, stdin=stdin, env=env)
-            pkdlog('cname={} cid={}', self._cname, self._cid)
-        except Exception as e:
-            self._agent_starting = False
-            pkdlog(
-                'agentId={} exception={}',
-                self._agentId,
-                e,
-                # TODO(e-carlin): read log
-            )
-            raise
+        c = PKDict(pykern.pkcollections.map_items(cfg[self.kind]))
+        p = (
+            'run',
+            # attach to stdin for writing
+            '--attach=stdin',
+            '--cpus={}'.format(c.get('cores', 1)),
+            '--init',
+            # keeps stdin open so we can write to it
+            '--interactive',
+            '--memory={}g'.format(c.gigabytes),
+            '--name={}'.format(self._cname),
+            '--network=host',
+            '--rm',
+            '--ulimit=core=0',
+            '--ulimit=nofile={}'.format(_MAX_OPEN_FILES),
+            # do not use a "name", but a uid, because /etc/password is image specific, but
+            # IDs are universal.
+            '--user={}'.format(os.getuid()),
+        ) + self._volumes() + (self._image,)
+        self._cid = await self._cmd(p + cmd, stdin=stdin, env=env)
+        pkdlog('cname={} cid={}', self._cname, self._cid)
 
     async def _cmd(self, cmd, stdin=subprocess.DEVNULL, env=None):
         c = DockerDriver.hosts[self.host.name].cmd_prefix + cmd
--- a/sirepo/job_driver/docker.py
+++ b/sirepo/job_driver/docker.py
@@ -121,8 +121,7 @@ class DockerDriver(job_driver.DriverBase):
 
     async def _do_agent_start(self, msg):
         cmd, stdin, env = self._agent_cmd_stdin_env()
-#TODO(robnagler) remove PKDict after https://github.com/radiasoft/pykern/issues/50
-        c = PKDict(pykern.pkcollections.map_items(cfg[self.kind]))
+        c = cfg[self.kind]
         p = (
             'run',
             # attach to stdin for writing
--- a/sirepo/job_driver/docker.py
+++ b/sirepo/job_driver/docker.py
@@ -106,9 +106,9 @@ class DockerDriver(job_driver.DriverBase):
                 o.send_ready.set()
 
     async def kill(self):
-        pkdlog('{}: stop cid={}', self.uid, self._cid)
         if '_cid' not in self:
             return
+        pkdlog('uid={} cid={}', self.get('uid'), self.get('_cid'))
         await self._cmd(
             ('stop', '--time={}'.format(job_driver.KILL_TIMEOUT_SECS), self._cid),
         )
--- a/sirepo/job_driver/sbatch.py
+++ b/sirepo/job_driver/sbatch.py
@@ -46,6 +46,10 @@ class SbatchDriver(job_driver.DriverBase):
         return cls
 
     async def kill(self):
+        # TODO(e-carlin): this won't work for the case where the ws drops, the
+        # agent is trying to reconnect and a new request comes in that starts
+        # an agent. In that case the script that starts job_agent should kill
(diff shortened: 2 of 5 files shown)


sirepo/job_driver/sbatch.py
```diff
@@ -112,10 +113,11 @@ def _agent_env(self):
                 script = f'''#!/bin/bash
 {self._agent_start_dev()}
 set -e
+pkill -u $USER -f 'sirepo job_agent start_sbatch' >& /dev/null || true
```
robnagler: I'm a bit tempted to be a bit more robust here. There's a problem if we are running alpha, beta, and prod on the same host (Cori). pkill was a quick hack. We should use a pid file.  Remove the pkill and after the `cd` do something like:

```sh
if [[ -r job_agent.pid ]]; then
     x=( $(<job_agent.pid) )
     if [[ ${x[0]} == $(hostname -f) ]]; then
          kill -KILL "${x[1]}" >& /dev/null || true
     else
          # you can ssh to a specific login node at NERSC s
          ssh "${x[0]}" kill -KILL "${x[1]}" >& /dev/null || true
     fi
     rm  job_agent.pid
fi
... start the process ...
echo "$(hostname -f) $!" > job_agent.pid
```

There are a number of questions here that might be better resolved by having the job_agent.py do all of the above. For example, how do we handle locking, or is that really the supervisor's responsibility? Should local agent do the same thing? Docker is different, of course.

sirepo/job_driver/docker.py
```diff
@@ -119,41 +119,30 @@ def slot_free(self):
...
-        except Exception as e:
-            self._agent_starting = False
-            pkdlog(
-                'agentId={} exception={}',
-                self._agentId,
-                e,
-                # TODO(e-carlin): read log
-            )
-            raise
+        c = PKDict(pykern.pkcollections.map_items(cfg[self.kind]))
```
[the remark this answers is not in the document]
e-carlin: Do I need to `.copy()`. The code doesn't modify `cfg[self.kind]` so copy seems unnecessary.

sirepo/job_driver/sbatch.py
e-carlin: Went with the job_agent.py route. Agent now writes pid file and kills existing agent on startup.

@robnagler on NERSC how is SSH auth handled from one login node to another? Keys or does the user have to supply their creds?

e-carlin:
@robnagler  ready for review.

sirepo/job_driver/sbatch.py
robnagler: It just works without a password. I'm guessing that the login nodes trust each other using identd or something.

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,49 @@ def default_command():
...
     i.start()
 
 
+def start_sbatch():
+    def kill_agent(pid_file):
+        if socket.getfqdn() == pid_file.fqdn:
+            os.kill(pid_file.pid, signal.SIGKILL)
+        else:
+            try:
+                subprocess.check_output(
```
robnagler: why check_output if it isn't checked? Use subprocess.run with `check_returncode=True`.

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,49 @@ def default_command():
...
+                    ('ssh', pid_file.fqdn, 'kill', '-KILL', str(pid_file.pid)),
+                    stderr=subprocess.STDOUT,
+                )
+            except subprocess.CalledProcessError as e:
+                pkdp(
+                    'cmd={cmd} returncode={returncode} output={output}',
```
robnagler: I think you don't need to merge stdout and stderr and it's better to keep them separate. There should be no output, and there might be stderr (only in an error).  See [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess).

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,49 @@ def default_command():
...
+                    **vars(e)
+                )
+    try:
+        f = pkio.py_path(_PID_FILE)
+        if f.exists():
+            kill_agent(pkjson.load_any(f))
+    except Exception as e:
+        pkdlog('error={} stack={}', e, pkdexc())
```
robnagler: Include `f` so we know what the error was about. It's arguable, but given there could be race conditions here, maybe better to just read the file, and catch the exception, and check if it was not_found ([EAFP](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/)).

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,49 @@ def default_command():
...
+    pkjson.dump_pretty(
+        PKDict(
+            fqdn=socket.getfqdn(),
+            pid=os.getpid(),
+        ),
+        _PID_FILE,
+        pretty=False,
```
robnagler: irrelevant here since the file is small so less code is better. Perhaps we could have `pkjson.dump_compact`, which would do all the right things.

sirepo/pkcli/job_agent.py
```diff
@@ -633,3 +670,8 @@ def __init__(self, stream):
         assert l < self._MAX, \
             'len(bytes)={} greater than _MAX={}'.format(l, _MAX)
         self.text.extend(t)
+
+
+def _terminate(dispatcher):
+    pkio.unchecked_remove(_PID_FILE)
```
robnagler: change the order of these two lines. If terminate gets in a loop (for whatever reason), then we'll want to pid to be killed.

sirepo/pkcli/job_agent.py
e-carlin: Thanks for the EAFP link. I went that route. Look right?

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,48 @@ def default_command():
...
+                pkdp(
+                    'cmd={cmd} returncode={returncode} stderr={stderr}',
+                    **vars(e)
+                )
+    try:
+        kill_agent(pkjson.load_any(pkio.py_path(_PID_FILE)))
+    except Exception as e:
+        if pkio.exception_is_not_found(e):
+            pass
+        pkdlog('error={} stack={}', e, pkdexc())
```
robnagler: Only log the stack and error if not found. You probably don't want to log not founds at all, since that's the normal case.

I think the read should happen before the kill_agent call. This is the "caveat" that not found could be thrown in kill_agent for another reason (process does not exist) so don't catch that exception in the same place.

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,48 @@ def default_command():
...
+        if socket.getfqdn() == pid_file.fqdn:
+            os.kill(pid_file.pid, signal.SIGKILL)
+        else:
+            try:
+                subprocess.run(
+                    ('ssh', pid_file.fqdn, 'kill', '-KILL', str(pid_file.pid)),
+                    capture_output=True,
+                ).check_returncode()
+            except subprocess.CalledProcessError as e:
+                pkdp(
```
robnagler: We don't commit pkdp, use pkdlog.

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,48 @@ def default_command():
...
+        return i.add_callback_from_signal(_terminate, d)
     signal.signal(signal.SIGTERM, s)
     signal.signal(signal.SIGINT, s)
     i.spawn_callback(d.loop)
     i.start()
 
 
+def start_sbatch():
+    def kill_agent(pid_file):
+        if socket.getfqdn() == pid_file.fqdn:
```
robnagler: pid_file.host would be clearer. I understand that getfqdn is being used here, but that's not the abstraction we really want to talk about. We use "host" everywhere else to mean "the name of a node".

socket.getfqdn() on my Mac nicely returns: `1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa`, which is useless. See https://stackoverflow.com/a/43330159 for discussion. It may return localhost, which is annoying. socket.gethostname() returns a fully qualified domain name on my Mac and on Linux/NERSC, it returns a name that is usable within the network context (possibly). 

Let's wrap this in a function that checks the result and gets the other. If gethostname does not return anything with a `.`, call getfqdn.

sirepo/pkcli/job_agent.py
```diff
@@ -55,14 +57,48 @@ def default_command():
...
+            os.kill(pid_file.pid, signal.SIGKILL)
+        else:
+            try:
+                subprocess.run(
+                    ('ssh', pid_file.fqdn, 'kill', '-KILL', str(pid_file.pid)),
+                    capture_output=True,
+                ).check_returncode()
+            except subprocess.CalledProcessError as e:
```
robnagler: I think there will be an error if the process isn't found, which is a case we should ignore. The process crashed, for example, before it could remove the pidfile

sirepo/pkcli/job_agent.py
e-carlin: Thanks for looking at this. Fix implemented.

sirepo/pkcli/job_agent.py
e-carlin: I've implemented just checking stderr for 'no such process'. Are you aware of anything more clever?

robnagler: For logging purposes, this is fine. If we get a system that starts outputting, say, "process does not exist", we can add a filter for that. Having extra logging isn't terrible, just makes it harder to see the real problems.

robnagler:
Fix #2195

PR #2182, opened 2020-01-13
Issue/2152 driver supervisor uri

### LINKED ISSUE #2152: Need driver dependent supervisor_uri
On an internal net, we need to be able to point to an internal port that doesn't go through the Internet gateway. On an external net (NERSC), it needs to use a public Internet uri.

e-carlin:
Fix #2152


etc/run-supervisor.sh
```diff
@@ -35,7 +34,7 @@ radia_run redhat-docker
         export SIREPO_JOB_DRIVER_SBATCH_SIREPO_CMD=/global/homes/${3::1}/$3/.pyenv/versions/py3/bin/sirepo
         export SIREPO_JOB_DRIVER_SBATCH_SRDB_ROOT='/global/cscratch1/sd/{sbatch_user}/sirepo'
         export SIREPO_JOB_SUPERVISOR_SBATCH_POLL_SECS=15
-        export SIREPO_JOB_SUPERVISOR_URI=http://$2:8001
+        export SIREPO_JOB_DRIVER_SBATCH_SUPERVISOR_HOST=$2
```
robnagler: This still needs to be URI, not HOST, because we have to allow for https and proxy host (alpha/beta/prod) or http and localhost/port (dev or curl).

sirepo/job.py
```diff
@@ -58,19 +51,13 @@
...
-DATA_FILE_ABS_URI = None
-
 #: how jobs request files (relative to `srdb.root`)
 SUPERVISOR_SRV_SUBDIR = 'supervisor-srv'
 
 #: how jobs request files (absolute)
 SUPERVISOR_SRV_ROOT = None
 
-DEFAULT_IP = '127.0.0.1'
+DEFAULT_HOST = socket.gethostname()
```
robnagler: we still need DEFAULT_IP, because there's no guarantee that we can lookup `gethostname` by DNS. Some routers block DNS lookups that return private IPs.

sirepo/job_driver/__init__.py
```diff
@@ -103,6 +101,21 @@ def get_ops_with_send_allocation(self):
...
+        if not self._has_remote_agent():
+            return
+        m = op.msg
+        d = pykern.pkio.py_path(m.simulation_lib_dir)
+        op.lib_dir_symlink = job.LIB_FILE_ROOT.join(
+            job.unique_key()
+        )
+        op.lib_dir_symlink.mksymlinkto(d, absolute=True)
+        m.pkupdate(
+            libFileUri=self.SUPERVISOR_URI + job.LIB_FILE_URI + '/' +
```
robnagler: encapsulate with dataFIleUri, because they both have to have slash trailing.

sirepo/job.py
```diff
@@ -177,27 +166,13 @@ def init():
             str,
             'shared secret between supervisor and server',
         ),
-        supervisor_uri=(
-            'http://{}:{}'.format(DEFAULT_IP, DEFAULT_PORT),
```
robnagler: This should be the default for dev. I think we use this value statically (DEFAULT_SUPERVISOR_URI) for the default case for all the specific cfg.supervisor_uri values.

sirepo/job_driver/sbatch.py
```diff
@@ -182,12 +186,22 @@ def init_class():
...
 
     cfg = pkconfig.init(
+        cores=(None, int, 'dev cores config'),
         host=pkconfig.Required(str, 'host name for slum controller'),
         host_key=pkconfig.Required(str, 'host key'),
-        cores=(None, int, 'dev cores config'),
         shifter_image=(None, str, 'needed if using Shifter'),
         sirepo_cmd=pkconfig.Required(str, 'how to run sirepo'),
         srdb_root=pkconfig.Required(_cfg_srdb_root, 'where to run job_agent, must include {sbatch_user}'),
+        supervisor_host=(
```
robnagler: This tuple could actually come from job.SUPERVISOR_URI_CFG_DECL or something like that so it can be used by all driver decls.

sirepo/pkcli/job_supervisor.py
```diff
@@ -48,10 +48,10 @@ def default_command():
         static_url_prefix=sirepo.job.LIB_FILE_URI + '/',
     )
     server = tornado.httpserver.HTTPServer(app)
-    server.listen(cfg.port, cfg.ip)
+    server.listen(cfg.port, cfg.host)
```
robnagler: this could be host or ip, but I prefer ip, since we know 127.0.0.1 works on all hosts. We don't know what host works on an arbitrary (curl) installation.

e-carlin:
@robnagler comments addressed. Ready for another review.

sirepo/job_driver/sbatch.py
e-carlin: Can you elaborate on this? I've changed it to be supervisor_uri in all drivers so each one of them can be overriden individually. They use job.DEFAULT_SUPERVISOR_URI as their default value.

robnagler: See 584358a8

e-carlin: Of course, thanks.

PR #2233, opened 2020-02-03
Test http async

e-carlin:


e-carlin:
@robnagler unfortunately there is no clear way to put a log message before `Exception after Future was cancelled` is [printed by Tornado](https://github.com/tornadoweb/tornado/blob/master/tornado/httpclient.py#L302). The print happens at a point where we don't have any control. We could put a log message after but that doesn't seem very useful.

To view the error output:
- Start the server
- **Don't** start the supervisor
- `sirepo http_test`

PR #2236, opened 2020-02-04
default to parallel if sbatch jobRunMode is no longer available

### LINKED ISSUE #2231: Can't get rid of `invalid jobRunMode=sbatch`
In https://github.com/radiasoft/sirepo/pull/2229 we added the ability to configure the jobRunMode and alert the user if the server changes the valid set of run modes (ex no longer supports sbatch). If the server has sbatch enabled, the user runs a simulation using it, the server then disables sbatch support, then the user tries to run the same simulation they can never get past the `error invalid jobRunMode=sbatch`

Steps to reproduce:
1. run the server and supervisor with sbatch enabled
2. Run srw partially coherent results with sbatch (let a few frames come in then cancel)
3. run the server and supervisor with only local enabled
4. Try to run the same srw simulation. You will get back the invalid jobRunMode error but then no amount of refreshing or clicking `start new simulation` will let you run the sim again without sbatch.

e-carlin:
Fix #2231

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -2973,7 +2973,7 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
             '<div class="clearfix"></div>',
             '<div data-ng-if="errorMessage()"><div class="text-danger"><strong>{{ ::appName }} Error:</strong></div><pre>{{ errorMessage() }}</pre></div>',
         ].join(''),
-        controller: function($scope) {
+        controller: function($scope, appState, authState) {
             $scope.appName = SIREPO.APP_SCHEMA.appInfo[SIREPO.APP_NAME].shortName;
 
             function callSimState(method) {
@@ -2998,6 +2998,14 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
                     || 'Simulation ' + $scope.simState.stateAsText() + ': ' + $scope.simState.getFrameCount() + ' animation frames';
             };
             $scope.start = function() {
+                // The available jobRunModes can change. Default to parallel if
+                // the current jobRunMode doesn't exist
+                if (
+                    appState.models[$scope.simState.model].jobRunMode
+                        in authState.jobRunModeMap === false
+                ) {
+                    appState.models[$scope.simState.model].jobRunMode = 'parallel';
+                }
                 appState.saveChanges($scope.simState.model);
                 $scope.simState.runSimulation();
             };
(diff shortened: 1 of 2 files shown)


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2998,6 +2998,14 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
                     || 'Simulation ' + $scope.simState.stateAsText() + ': ' + $scope.simState.getFrameCount() + ' animation frames';
             };
             $scope.start = function() {
+                // The available jobRunModes can change. Default to parallel if
+                // the current jobRunMode doesn't exist
+                if (
+                    appState.models[$scope.simState.model].jobRunMode
+                        in authState.jobRunModeMap === false
```
robnagler: We do not want to set jobRunMode on a model if it doesn't exist so the condition needs to be changed to:

```js
var j = appState.models[$scope.simState.model].jobRunMode;
if (j && j in authState.jobRunModeMap === false)
```

Just a note to JS novices like me: `undefined in {}` is a valid test, which I find interesting. Moreover, `! undefined in {} === false` and `undefined in {sbatch: 'sbatch'} === false` are both true. `!` has higher precedence than `in`.  I see why the `=== false` is necessary, but it's weird.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2998,6 +2998,14 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
...
+                ) {
+                    appState.models[$scope.simState.model].jobRunMode = 'parallel';
```
robnagler: There is a default value, which we could use. It's very hard to extract values from the schema. `SIREPO.INFO_INDEX_DEFAULT_VALUE = 2` is in sirepo-components, but you have to find the schema value, which is tricky. There should be "defaultValue(model, field)". Ugh. Just noting. Leave the copy-and-paste as it is.

sirepo/package_data/static/js/srw.js
```diff
@@ -1950,6 +1950,14 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
...
+                // The available jobRunModes can change. Default to parallel if
+                // the current jobRunMode doesn't exist
+                if (
+                    appState.models[$scope.simState.model].jobRunMode
+                        in authState.jobRunModeMap === false
+                ) {
+                    appState.models[$scope.simState.model].jobRunMode = 'parallel';
+                }
                 if ($scope.model == 'multiElectronAnimation') {
                     appState.saveChanges($scope.simState.model);
```
robnagler: I think there's a double save going on. May be able to delete this save.

e-carlin: I'm not sure but I'm leaning towards the double save being necessary.`getReportTitle()` is using appState which possibly implies the appState needs to be updated with these values for it to work. Also, the call to `saveChanges` within `saveAndRunSimulation` uses `simulation` as the name whereas the call to `saveChanges` here uses `multiElectronAnimation` as the name. I'm not familiar enough with the code to have a positive answer. Should I do a dive and investigate?

PR #2252, opened 2020-02-07
start on py2 to py3 conversion

moellep:
Here's a stab at py2 to py3 conversion. The main part is adding py2/py3 compatibility methods to sirepo.util and calling from cookie and md5 calls. elegant, srw and opal all seem to run OK. Other apps will need adjustments to calls like map(), xrange(), etc.

```python
# python 2/3 string and bytes compatibility
if sys.version_info < (3,):
    def b(x):
        return x
    def s(x):
        return x
else:
    def b(x):
        return bytes(x, 'utf-8')
    def s(x):
        return x.decode('utf-8')
```


sirepo/server.py
```diff
@@ -197,7 +197,12 @@ def api_findByName(simulation_type, application_mode, simulation_name):
 def api_findByNameWithAuth(simulation_type, application_mode, simulation_name):
     req = http_request.parse_params(type=simulation_type)
     #TODO(pjm): need to unquote when redirecting from saved cookie redirect?
-    simulation_name = urllib.unquote(simulation_name)
```
robnagler: This needs to be debugged.

sirepo/template/sdds_util.py
```diff
@@ -34,33 +35,47 @@
...
+_MAX_SDDS_INDEX = 19
+_sdds_lock = threading.RLock()
 
 
 def extract_sdds_column(filename, field, page_index):
     return process_sdds_page(filename, page_index, _sdds_column, field)
 
 
 def process_sdds_page(filename, page_index, callback, *args, **kwargs):
+    global _SDDS_INDEX
```
robnagler: This should be in runner_api (ensure only one request at at time)

sirepo/template/srw.py
```diff
@@ -701,7 +701,7 @@ def index_file_name(zf):
     # Compare index and zip contents
     # Does not include the index itself, nor any directories
     # also extract the filename since the index does not include path info
-    file_names_in_zip = map(lambda path: os.path.basename(path),  [f for f in zf.namelist() if not f.endswith('/') and f != index_file_name(zf)])
+    file_names_in_zip = list(map(lambda path: os.path.basename(path),  [f for f in zf.namelist() if not f.endswith('/') and f != index_file_name(zf)]))
```
robnagler: list comprehension

sirepo/util.py
```diff
@@ -130,6 +131,18 @@ def __init__(self, display_text, *args, **kwargs):
             **kwargs
         )
 
+# python 2/3 string and bytes compatibility
+if sys.version_info < (3,):
```
robnagler: feature test str.decode. Bad example for others to follow. Move this to pkcompat.

sirepo/util.py
```diff
@@ -130,6 +131,18 @@ def __init__(self, display_text, *args, **kwargs):
...
+    def b(x):
```
robnagler: Single letter names is not a good idea. to_bytes. from_bytes.

PR #2260, opened 2020-02-10
Issue/2245 agent start fail noisily

### LINKED ISSUE #2245: docker: detect startup failure of job_agent
If the container fails, there's no output and nothing in the log.

e-carlin:
Fix #2245

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -221,7 +221,6 @@ class DriverBase(PKDict):
         Returns:
             bool: True if the op was actually sent
         """
-        # TODO(e-carlin): s/self.ops/self._ops (across all drivers)
         self.ops[op.opId] = op
         if not self._websocket_ready.is_set():
             await self._agent_start(op)
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -44,10 +44,13 @@ class DriverBase(PKDict):
 
     __instances = PKDict()
 
+    _AGENT_STARTING_TIMEOUT_SECS = 5
+
     def __init__(self, req):
         super().__init__(
             _agentId=job.unique_key(),
             _agent_starting=False,
+            _agent_starting_timeout=None,
             _agent_start_lock=tornado.locks.Lock(),
             _cpu_slot_alloc_time=None,
             kind=req.kind,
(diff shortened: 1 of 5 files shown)


sirepo/job_driver/__init__.py
```diff
@@ -201,6 +221,7 @@ def pkdebug_str(self):
         Returns:
             bool: True if the op was actually sent
         """
+        self.ops[op.opId] = op
```
robnagler: I don't understand why this was added here.

sirepo/job_driver/__init__.py
```diff
@@ -163,27 +187,23 @@ def make_lib_dir_symlink(self, op):
...
-            q = self.op_q[n]
-            try:
-                op.op_slot = q.get_nowait()
-            except tornado.queues.QueueEmpty:
-                pkdlog('self={} op={} await op_q.get()', self, op)
-                op.op_slot = await q.get()
-                raise job_supervisor.Awaited()
-        finally:
-            # may be redundant, but need to track here
-            self.ops[op.opId] = op
```
robnagler: The op would be lost if it exited due to an assertion  (line 195).

sirepo/job_driver/__init__.py
```diff
@@ -163,27 +187,23 @@ def make_lib_dir_symlink(self, op):
...
+            op.op_slot = q.get_nowait()
+        except tornado.queues.QueueEmpty:
```
robnagler: The reason this was narrow is that it isolates the QueueEmpty to the function which could get that exception. Too wide a catch is a problem when catching a specific exception like this.

e-carlin: I believe it still does just that. I removed the try/finally but didn't change this try/except.

sirepo/job_driver/__init__.py
e-carlin: The insertion of the op in self.ops was moved into `prepare_send`. `prepare_send` is the only caller of this function (I should make it private in that case) so I don't think it will be lost.

sirepo/job_driver/__init__.py
e-carlin: It was moved from `op_ready` to here. I did this because the call to `await self._agent_start(op)` ([line 223](https://github.com/radiasoft/sirepo/blob/issue/2245-agent-start-fail-noisily/sirepo/job_driver/__init__.py#L223)) is where we will block when the agent fails to start. In the code to handle the failed start we destroy all ops in self.ops and in the old code self.ops wouldn't contain op.

PR #2271, opened 2020-02-17
Reproducible simulations

### LINKED ISSUE #2266: Permanent archive of simulation
* Add new GUI element to allow user to store simulation zip on server
* Implement API to store simulation zip on server in `<code>/archive/<sid>-<datetime>.zip`
* Store /rsmanifest.json in json file before zipping (in export archive, always)

e-carlin:
Below are the main pieces of the change:

GUI:
- button for archiving a simulation
- button for selecting an archived simulation to run  

Server:
- archive a simulation and store it in the db
- Unarchive a simulation in the db
- Run the simulation using the same image that it was run in previously

Fix #2266

### DIFF
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -112,6 +112,17 @@ def api_runCancel():
 @api_perm.require_user
 def api_runSimulation():
     r = _request_content(PKDict(fixup_old_data=True))
+    # TODO(e-carlin): I'm not sure where the best place to put this data on the request is
+    # Some options:
+    # 1. Put it on here
+    # 2. The GUI could put it on.
+    # 3. The job_supervisor could read sirepo-data.json and put it on if need be.
+    #    I like this option the best but the supervisor operates outside of the
+    #    flask context so it is not easy to get at a user's data
+    r.rsmanifest = sirepo.simulation_db.open_json_file(
+        r.simulationType,
+        sid=r.simulationId,
+    ).get('rsmanifest')
     # TODO(e-carlin): This should really be done in job_supervisor._lib_dir_symlink()
     # but that is outside of the Flask context so it won't work
     r.simulation_lib_dir = sirepo.simulation_db.simulation_lib_dir(r.simulationType)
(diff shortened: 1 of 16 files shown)


sirepo/job_api.py
```diff
@@ -111,6 +112,17 @@ def api_runCancel():
 @api_perm.require_user
 def api_runSimulation():
     r = _request_content(PKDict(fixup_old_data=True))
+    # TODO(e-carlin): I'm not sure where the best place to put this data on the request is
```
e-carlin: @robnagler any advice on this? I'm not sure where the best place to put the rsmanifest is on the reqeust (if rsmanifest is on the request it will be used to set the image for the simulation)

sirepo/job_driver/docker.py
```diff
@@ -104,6 +124,59 @@ def init_class(cls):
...
+        i = self._get_container_image(op.containerImage)
+        if i != self._image:
+            await self._change_image(op, i)
+        await super()._acquire_op_slot(op)
+
+    def _agent_starting_done(self):
+        self._image_change.set_complete()
+        super()._agent_starting_done()
+
+    async def _change_image(self, op, image):
```
e-carlin: @robnagler this is the bulk of the code for changing an image. self._image_change was added to control the whole flow of events. I think it could be simpler to just start a new agent with the different image and make sure that no user can have multiple agents allocated a cpu slot.

sirepo/package_data/static/json/schema-common.json
```diff
@@ -66,6 +66,7 @@
         "longName": "Sirepo"
     },
     "route": {
+        "archiveSimulation": "/archive-simulation",
```
e-carlin: @robnagler any thought on the names? These aren't my favorite.

sirepo/job_driver/docker.py
```diff
@@ -104,6 +124,59 @@ def init_class(cls):
...
+        )
+        r = PKDict()
+        try:
+            for k in self.op_q.keys():
+                r.setdefault(k, [])
+                while len(r[k]) < self.op_q[k].maxsize:
+                    r[k].append(await self.op_q[k].get())
+            await self.kill()
+            self._image = image
+            await self._agent_start(
```
e-carlin: @robnagler how should we go about cleaning up images? When we take a docker cotainers cpu slot should we kill the agent and rm the image (if it is not the up to date prod image)? Should we rm with each release? ...

robnagler:
Closing PR but leaving branch for posterity.

Issue #1435, opened 2018-11-15
Stop job after timeout


robnagler:
If the job takes too long to run, it should stop. Integrate with pkconfig so can pass environment vars.

njsmith:
By "stop" do you mean "we issue a `docker stop`", or something else?

Who sets the timeout? Is it a field in the job request, a piece of global configuration, ...?

How are environment variables connected? Do we want to pass an envvar into the job so it knows what its timeout is?

robnagler:
Partially implemented. Needs to send kill to remote process.

robnagler:
Timeouts need to be adjusted for OP_ANALYSIS vs OP_RUN (sequential). Configuration should probably be max_hours.parallel/sequential/analysis. We don't need to deprecate the old configuration, just rename it. @e-carlin please take care of this, and add a case to job_timeout test. You may need to hack myapp to have a long analysis.

BTW, sequential is what background import uses (srw python). Since we execute arbitrary code in other cases, it's probably not as big a deal now that arbitrary code is always contained.

PR #2290, opened 2020-02-20
check for idle docker agents every 30 min

### LINKED ISSUE #2105: Docker agent idle timeout
Need some sort of timeout on a docker agent that is idle. Driver side kill.

e-carlin:
Fix #2105

### DIFF
--- a/sirepo/job_driver/docker.py
+++ b/sirepo/job_driver/docker.py
@@ -241,7 +241,7 @@ class DockerDriver(job_driver.DriverBase):
 
         if not self._idle_timer:
             self._idle_timer = tornado.ioloop.IOLoop.current().call_later(
-                cfg.idle_check_mins * 60,
+                cfg.idle_check_secs,
                 _kill_if_idle,
             )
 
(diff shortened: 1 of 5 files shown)


sirepo/job_driver/docker.py
```diff
@@ -225,6 +230,21 @@ def _init_hosts(cls):
...
+            if not self.ops:
+                pkdlog('self={}', self)
+                self._idle_timer = None
+                await self.kill()
+            else:
+                self._start_idle_timeout()
+
+        if not self._idle_timer:
+            self._idle_timer = tornado.ioloop.IOLoop.current().call_later(
+                cfg.idle_check_mins * 60,
```
robnagler: let's use secs, since everything else in secs (except run time which is in hours)

sirepo/job_supervisor.py
```diff
@@ -493,6 +485,16 @@ def _create_op(self, opName, req, **kwargs):
         self.ops.append(o)
         return o
 
+    def _get_max_run_secs(self, op_name, kind, run_mode):
+        pkdp('{} {} {}', op_name, kind, run_mode)
```
robnagler: pkdlog

sirepo/template/myapp.py
```diff
@@ -25,6 +26,9 @@
 
 
 def get_data_file(run_dir, model, frame, options=None, **kwargs):
+    if 'suffix' in options and options.suffix == 'sr_long_analysis':
```
robnagler: options may be None so this will fail.

e-carlin: Thanks, this should be on my #1435 branch which is still WIP. Reverting.

sirepo/job_supervisor.py
e-carlin: This should've been on my #1435 branch. Reverted.

e-carlin:
@robnagler I addressed your comments. Two were related to accidental commits (I reverted them and cherry picked them to the correct branch) and the third (secs vs mins) I fixed. Ready for another review.

PR #2301, opened 2020-02-20
Issue/1435 send kill timeout

### LINKED ISSUE #1435: Stop job after timeout

e-carlin:
Fix #1435 

@robnagler I've left comments below on areas that I think need discussion.

### DIFF
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -301,7 +301,7 @@ class _ComputeJob(PKDict):
             dataFileKey=req.content.pop('dataFileKey')
         )
 
-    async def _receive_api_runCancel(self, req, op=None):
+    async def _receive_api_runCancel(self, req, timed_out_op=None):
         """Cancel a run and related ops
 
         Analysis ops that are for a parallel run (ex. sim frames) will not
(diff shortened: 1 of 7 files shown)


sirepo/job_supervisor.py
```diff
@@ -327,14 +321,24 @@ def _ops_to_cancel():
...
-            if op and op in self.ops:
+            if op and op in self.ops and op not in o:
                 o.append(op)
             return o
 
         r = PKDict(state=job.CANCELED)
         if (
-            not self._req_is_valid(req)
-            or self.db.status not in _RUNNING_PENDING
+# TODO(e-carlin): the 'and not op' check is not quite right. The problem is if
```
e-carlin: This could lead to strange UX. Read the code comment for context.

tests/job_timeout_test.py
```diff
@@ -53,3 +54,24 @@ def test_srw(fc):
...
+        'downloadDataFile',
+        PKDict(
+            simulation_type=d.simulationType,
+            simulation_id=d.models.simulation.simulationId,
+            model='heightWeightReport',
+            frame='-1',
+            suffix='sr_long_analysis',
+        ),
+    )
+    # TODO(e-carlin): This should be something more helpful than a 404
```
e-carlin: What's the best option here? 404 seems "fine" but not the most helpful.

e-carlin:
@robnagler the GUI interprets 'canceled' for sequential runs and analyses as bad data and re-runs the simulation/analyses up to 3x. [This is the code](https://github.com/radiasoft/sirepo/blob/32a8a83756692a1296cae613c876c843d2cc72c8/sirepo/package_data/static/js/sirepo-plotting.js#L193) that does the retry. Ultimately the user will see an error but it takes 3 retries before it appears. Maybe this is the UX we want, or maybe not, I'm not sure.

tests/job_timeout_test.py
robnagler: I think this is a very rare case so 404 is fine.

sirepo/job_supervisor.py
```diff
@@ -327,14 +321,24 @@ def _ops_to_cancel():
...
+# we are calling api_runCancel from run_timeout and the op that timed out was
+# an analysis op that is not normally checked for request validity. Ex
+# api_downloadDataFile does not check for a valid request because a data file
+# request is not valid (it has not computeJobHash for example). But we still
+# need to cancel downloadDataFile ops. We know we are called from run_timeout
+# if op is not None so that is how we are getting around the problem. This
+# could result in odd UX. Ex a running simulation being cancelled due to a
+# downloadDataFile request timeout in another browser window (only the
+# computJids must match between the two requests).
```
robnagler: Getting complicated. Let's talk about this in person. I want to understand completely.

sirepo/pkcli/job_agent.py
```diff
@@ -238,15 +238,17 @@ def terminate(self):
...
 
-    async def _cmd(self, msg):
+    async def _cmd(self, msg, **kwargs):
         if msg.opName == job.OP_ANALYSIS and msg.jobCmd != 'fastcgi':
             return await self._fastcgi_op(msg)
         c = _Cmd
         if msg.jobRunMode == job.SBATCH:
             c = _SbatchRun if msg.isParallel else _SbatchCmd
-        p = c(msg=msg, dispatcher=self)
         if msg.jobCmd == 'fastcgi':
```
robnagler: should this be elif?

sirepo/pkcli/job_agent.py
```diff
@@ -273,26 +275,25 @@ def _fastcgi_accept(self, connection, *args, **kwargs):
...
-        self._fastcgi_msg_q = None
-        self._fastcgi_cmd.destroy()
-        self._fastcgi_cmd = None
-        if msg:
-            await _reply_error(msg)
-        while q.qsize() > 0:
-            await _reply_error(q.get_nowait())
-            q.task_done()
+        # If self.fastcgi_cmd is None we initiated the kill so not an error
+        if self.fastcgi_cmd:
```
robnagler:
```py
if not self.fastcgi_cmd:
     return
```

I think we still want to log. 

Maybe put the logging back at the `except` and only call the handler if fastcgi_cmd?

sirepo/pkcli/job_agent.py
e-carlin: Yes.

sirepo/pkcli/job_agent.py
e-carlin: Ya that's clearer.

PR #2308, opened 2020-02-24
Issue/2135 more logging

### LINKED ISSUE #2135: improve job supervisor/agent logging
The supervisor should log more, e.g. which runDir is being used.

e-carlin:
Fix #2135


sirepo/job_driver/__init__.py
```diff
@@ -70,11 +70,7 @@ def __init__(self, req):
         )
         # Drivers persist for the life of the program so they are never removed
         self.__instances[self._agentId] = self
-        pkdlog(
-            'class={} agentId={_agentId} kind={kind} uid={uid}',
-            self.__class__.__name__,
-            **self
-        )
+        pkdlog('self={}', self)
```
robnagler: I think we can do without `self=`.

sirepo/job_driver/__init__.py
```diff
@@ -241,10 +237,9 @@ def receive(cls, msg):
 
     def send(self, op):
         pkdlog(
-            'op={} agentId={} opId={} runDir={}',
-            op.opName,
-            self._agentId,
-            op.opId,
+            'self={} op={} runDir={}',
```
robnagler: I think op= and self= are unnecessary. Op says it is an op. Self is a Driver.

sirepo/job_driver/__init__.py
```diff
@@ -200,12 +196,12 @@ def make_lib_dir_symlink(self, op):
 
     def pkdebug_str(self):
         return pkdformat(
-            '{}(agentId={}, kind={}, uid={}, ops={})',
+            '{}(agentId={:.6} kind={} uid={} ops={})',
```
robnagler: Since agentId is 6 (maybe make it 4 to be even more distinguishable) and uid is 8, we can output: 
```txt
JobDriver(wxyz parallel abcdefgh [op()...]
```

We know the line, function, and file for every entry so we have enough context. It's the IDs we need to be able to trace logic.

sirepo/job_driver/__init__.py
```diff
@@ -310,7 +304,7 @@ def _agent_env(self, env=None):
                 )
                 await self._do_agent_start(op)
             except Exception as e:
-                pkdlog('agentId={} exception={}', self._agentId, e)
+                pkdlog('self={} exception={}', self, e)
```
robnagler: drop self= everywhere.

sirepo/job_driver/__init__.py
```diff
@@ -336,23 +330,23 @@ def _receive(self, msg):
...
+                pkdlog('self={} no reply={}', self, c)
                 c.reply = PKDict(state='error', error='no reply')
             if i in self.ops:
                 #SECURITY: only ops known to this driver can be replied to
                 self.ops[i].reply_put(c.reply)
             else:
                 pkdlog(
-                    'agentId={} not pending opId={} opName={}',
-                    self._agentId,
+                    'self={} not pending opId={:.6} opName={}',
```
robnagler: This is when an opId might be good to be its own object (subclass of str) so we can implement pkdebug_str for it. It's a bit of a pain, because serialization fails (I think), but we could possibly implement a pkjson serializer function that allows us to serialize anything. Maybe pkcollections does that already with str.

sirepo/job_supervisor.py
```diff
@@ -193,15 +193,17 @@ def pkdebug_str(self):
         if not d:
             return '_ComputeJob()'
         return pkdformat(
-            '_ComputeJob({} {} ops={})',
+            '_ComputeJob(jid={} uid={} status={} ops={})',
```
robnagler: uid and jid are not necessary. Could just output jid, and it would be enough.

sirepo/pkcli/job_agent.py
```diff
@@ -201,15 +205,15 @@ def terminate(self):
...
             pkdc('m={}', m)
             return await getattr(self, '_op_' + m.opName)(m)
         except Exception as e:
             err = 'exception=' + str(e)
             stack = pkdexc()
             pkdlog(
-                'op={} exception={} stack={}',
-                m and m.get('opName'),
+                'opId={:.6} exception={} stack={}',
+                m and m.get('opId'),
```
robnagler: it would be good to know the opName

sirepo/job_supervisor.py
```diff
@@ -193,15 +193,17 @@ def pkdebug_str(self):
...
             d.get('computeJid'),
+            d.get('uid'),
             d.get('status'),
             self.ops,
         )
 
     @classmethod
     async def receive(cls, req):
-        pkdlog('{}', req)
+        if req.content.get('api') != 'api_runStatus':
```
robnagler: Is this only place that pkdlog of runStatus happens? If in error, it's fine to log. Just wondering if it is less noisy now.

sirepo/job_driver/__init__.py
e-carlin: I agree on truncating to 4 characters.

I can do away with ops= since they will say they are ops in the list. Also, eliminating kind= is fine since the value implies the type. 

How strongly do you feel about not having agentId= and uid=? AgentId and uid seem like important info that is not immediately obvious. The person debugging will have the line/function/file for the log message but not the pkdebug_str() function which is where they actually need to go to get the info they are after. 2 hops to find the info (and one hop requiring intricate knowledge of our system) doesn't seem worth it.

sirepo/job_driver/__init__.py
e-carlin: Hmm I'm 50/50. We would have to hand deserialize the opId in the message to an opId object which is work and I'm not sure is worth the upside. We also don't do that anywhere else and it seems like something we may want to build a structure for (deserializing message payloads, or parts of them, into objects). Thoughts?

e-carlin:
> Some suggestions to tighten up output. I'm debating about the opId:4 vs opId:6. I can't imagine that any given log would have collisions so seems like 4 is enough and would allow us to visually distinguish between opId and uid. agentId probably is the same. I don't think we'll see collisions.

I think 4 is plenty.

sirepo/job_supervisor.py
e-carlin: It also happens in [pkcli/job_supervisor.py](https://github.com/radiasoft/sirepo/blob/4e5514742b936dc9fb54b43cafecb3cd14771565/sirepo/pkcli/job_supervisor.py#L169). Eliminating the both of them makes less noise.

PR #2311, opened 2020-02-27
Adm job

e-carlin:
Screen Shot 2020-02-27 at 8 49 43 AM

This is building off of the admin page @mkeilman started on. It shows job information for all currently running jobs in the system. A user must have the 'adm' role to be able to see this.


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2634,6 +2643,48 @@ SIREPO.app.directive('bootstrapToggle', function() {
...
+            $scope.getAdmJobs = function (id) {
+                requestSender.sendRequest(
+                    'admJobs',
+                    dataLoaded,
+                    {
+                        id: id,
+                    });
+            };
+
+            appState.clearModels(appState.clone(SIREPO.appDefaultSimulationValues));
```
e-carlin: @moellep I've added this piece of code because I want all the header values (ex sectionTitle link, source tab, etc) to not be shown on the adm jobs list page (see image below). It looked like the [simulationsController was doing the same thing](https://github.com/radiasoft/sirepo/blob/master/sirepo/package_data/static/js/sirepo.js#L3295) when a user navigates to the simulations page. Does this look right? Could calling this / not calling something else break something I'm unaware of?

Screen Shot 2020-02-28 at 11 00 30 AM

moellep: Yes, this is fine for now - any models should get cleared out when someone views the jobs page, similar to the simulation list page. Really the SIREPO.appDefaultSimulationValues should access default-data.json for the simulation, but that can wait.

PR #2315, opened 2020-03-02
User job

e-carlin:
Users can see a list of their running jobs and navigate to them using the provided link.

Screen Shot 2020-03-02 at 10 34 57 AM

@robnagler this is based on #2311. Merge that one first.


sirepo/job_api.py
```diff
@@ -69,6 +71,14 @@ def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=No
         )
 
 
+@api_perm.require_user
+def api_jobs():
```
robnagler: This should be prefixed with something. There should be admJobs and maybe myJobs or ownJobs. The auth check should be first entry point for admJobs. 

Partitioning name spaces is important, because we may want to restrict access to particular APIs by IP in the proxy. We do this already [in rsconf for job supervisor](https://github.com/radiasoft/rsconf/blob/03b5d9d8cd905a917872721239bfd3461571436b/rsconf/package_data/sirepo_job_supervisor/nginx.conf.jinja#L7).

sirepo/job_api.py
```diff
@@ -69,6 +71,14 @@ def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=No
...
+    req = sirepo.http_request.parse_post().req_data
+    if req.adm:
```
robnagler: This is an unchecked value, and it is related to security so that's potentially dangerous. The value of `api_admJobs` was parsed by `uri_router` and validated against the schema in order to find the function. That's an important check. Next, if we decide to implement api-based roles, we would want api_admJobs to be distinct from api_ownJobs so that the role checks would be distinct.

Also,`req` is always the result of `parse_post`, not `req_data`. To refer to `req_data` directly, use `req.req_data`. Naming consistency is important.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -601,13 +601,23 @@ SIREPO.app.directive('logoutMenu', function(authState, authService, requestSende
               '<ul class="dropdown-menu">',
                 '<li class="dropdown-header"><strong>{{ ::authState.displayName }}</strong></li>',
                 '<li class="dropdown-header" data-ng-if="::authState.userName">{{ ::authState.userName }} via {{ ::authState.method }}</li>',
+                '<li data-ng-if="showAdmJobs()"><a data-ng-click="redirect(\'admJobs\')">Admin</a></li>',
```
robnagler: I would like these two entries to be links so users can bookmark them.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2634,6 +2644,82 @@ SIREPO.app.directive('bootstrapToggle', function() {
...
+            $scope.getRows = function() {
+                var d = $scope.data;
+                if (d) {
+                    var s = d.header.indexOf('Simulation id');
+                    var a = d.header.indexOf('App')
+                    if (s < 0 || a < 0) {
+                        throw new Error("'Simulation id' or 'App' not found on header=" + d.header);
+                    }
+                    var n = d.header.indexOf('Name')
+                    var m = is_adm();
```
robnagler: I think this should be passed in to this function or on $scope.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2634,6 +2644,82 @@ SIREPO.app.directive('bootstrapToggle', function() {
...
+            function getUrl(simulationId, app) {
+                return requestSender.formatUrlLocal(
+                    'source',
+                    {':simulationId': simulationId},
+                    app
+                );
+            };
+
+            function is_adm() {
+                return $location.path().includes('adm');
```
robnagler: This implicitly binds the string `adm` to the local route URL. Rather, there should be a feature of the route template or some other parameterization. It would be arguable to use the local route name as a binding token, but I'd prefer explicit parameterization.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2634,6 +2644,82 @@ SIREPO.app.directive('bootstrapToggle', function() {
...
+                        throw new Error("'Simulation id' or 'App' not found on header=" + d.header);
+                    }
+                    var n = d.header.indexOf('Name')
+                    var m = is_adm();
+                    var o = '';
+                    for (var i in d.rows) {
+                        var r = d.rows[i];
+                        o += '<tr>';
+                        for (var j =0; j < r.length; j++ ) {
+                            var v = r[j];
```
robnagler: I think we hit our single letter variable limit. It would be good to add a function or some type of denormalization.

sirepo/job_supervisor.py
```diff
@@ -295,6 +303,58 @@ def __db_write(self):
         sirepo.util.json_dump(self.db, path=self.__db_file(self.db.computeJid))
         return self
 
+    @classmethod
+    def _get_running_jobs(cls, want_adm=False):
+        def _get_jobs():
+            o = []
+            for i in filter(
+                    lambda x: x.db.status == job.RUNNING,
```
robnagler: This should be RUNNING or PENDING. If PENDING,  display queued time.

sirepo/job_supervisor.py
```diff
@@ -295,6 +303,58 @@ def __db_write(self):
...
+            ):
+                d = i.db
+                s = d.computeJobStart
+                l = d.lastUpdateTime
+                r = [
+                    d.simulationType,
+                    d.simulationId,
+                    _strftime(s),
+                    _strftime(l),
+                    round((l - s) / 60.0, 2),
```
robnagler: End users should not see driver details.

sirepo/job_supervisor.py
```diff
@@ -295,6 +303,58 @@ def __db_write(self):
...
+            o = []
+            for i in filter(
+                    lambda x: x.db.status == job.RUNNING,
+                    cls.instances.values(),
+            ):
+                d = i.db
+                s = d.computeJobStart
+                l = d.lastUpdateTime
+                r = [
+                    d.simulationType,
```
robnagler: I don't think simulationType and simulationId are useful for end users. The SimulationName should be enough. Users think in terms of what they see, and that's the name of the simulation. End users probably don't use more than one simulationType simultaneously so this will be a wasted column.

sirepo/job_supervisor.py
```diff
@@ -295,6 +303,58 @@ def __db_write(self):
...
+    @classmethod
+    def _get_running_jobs(cls, want_adm=False):
+        def _get_jobs():
+            o = []
+            for i in filter(
+                    lambda x: x.db.status == job.RUNNING,
+                    cls.instances.values(),
+            ):
+                d = i.db
+                s = d.computeJobStart
```
robnagler: These aliases do not add that much value, and they detract from readability.

sirepo/job_supervisor.py
robnagler: Why not use a format here, even an f string? `f'{d.lastUpdateTime - d.computeJobStart:.1f}'` An alternative might be `h:m:s` format so you wouldn't have to explain in the header. Jobs might run for hours.

sirepo/job_supervisor.py
```diff
@@ -295,6 +303,58 @@ def __db_write(self):
...
+
+        def _strftime(unix_time):
+            return datetime.datetime.utcfromtimestamp(
+                int(unix_time),
+            ).strftime('%Y-%m-%d %H:%M:%S')
+
+        c = [
+            'App',
+            'Simulation id',
+            'Start (UTC)',
```
robnagler: For adm, we probably want queued time so we can observe queueing. For end users, I don't think it's very useful (there isn't a lot of queueing in general).

sirepo/job_supervisor.py
```diff
@@ -414,10 +478,15 @@ def _ops_to_cancel():
...
-                    self.db.jobRunMode = req.content.jobRunMode
-                    self.db.computeJobSerial = int(time.time())
-                    self.db.pkupdate(status=job.PENDING)
+                    self.db.pkupdate(
+                        computeJobSerial=int(time.time()),
+                        computeModel=req.content.computeModel,
+                        driverDetails=o.driver.driver_details,
+                        # run mode can change between runs so we must update the db
+                        jobRunMode=req.content.jobRunMode,
+                        name=req.content.data.models.simulation.name,
```
robnagler: name => simName. It's easier to see what it is when it is in isolation. We probably want the simName in the history as it may change.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2634,6 +2644,81 @@ SIREPO.app.directive('bootstrapToggle', function() {
...
+                return requestSender.formatUrlLocal(
+                    'source',
+                    {':simulationId': simulationId},
+                    app
+                );
+            };
+
+            $scope.getJobs = function () {
+                requestSender.sendRequest(
+                    ($scope.wantAdm ? 'adm' : 'own') + 'Jobs',
```
robnagler: Repeating `Jobs` is going to make it easier to see where the API is used. 

```js
$scope.wantAdm ? 'admJobs' : 'ownJobs',
```

sirepo/job_supervisor.py
```diff
@@ -303,6 +363,10 @@ def __db_write(self):
             dataFileKey=req.content.pop('dataFileKey')
         )
 
+    @classmethod
+    async def _receive_api_ownJobs(cls, req):
+        return cls._get_running_jobs()
```
robnagler: How is this restricted just to the users's jobs?

Please add a test to validate this question.

e-carlin:

admJobs now looks like:
Screen Shot 2020-03-04 at 1 57 16 PM

ownJobs now looks like:
Screen Shot 2020-03-04 at 1 57 05 PM

PR #2318, opened 2020-03-03
Issue/2313 pending state on refresh

### LINKED ISSUE #2313: pending state not presented on refresh
Possibly already reported - when you start s background job which is waiting on a running background job, the UI shows "pending". If you refresh the browser, the browser shows no message, even though the job is still pending and will run later.

e-carlin:
Fix #2313


sirepo/job_supervisor.py
```diff
@@ -455,22 +455,22 @@ def _ops_to_cancel():
             jobCmd='compute',
             nextRequestSeconds=self.db.nextRequestSeconds,
         )
+        self.__db_init(req, prev_db=self.db)
+        self.db.pkupdate(
+            computeJobSerial=int(time.time()),
```
e-carlin: @robnagler is it a problem that the computeJobSerial is now written before the job is going to be run (before all of the awaits)? I don't think so but I want to confirm.

sirepo/job_supervisor.py
```diff
@@ -455,22 +455,22 @@ def _ops_to_cancel():
...
+            driverDetails=o.driver.driver_details,
```
robnagler: I think driver_details needs to get written every time, because the driver host might change.

sirepo/job_supervisor.py
robnagler: This is a bug. Definitely want computeJobSerial written right away, when the job is queued.

I thought we saved the time we queued. Technically that could be computeJobSerial, but maybe it should be a distinct value.

sirepo/job_supervisor.py
e-carlin: The instantiation of the op (which instantiates the driver) is outside of the loop so the driver details won't change.

e-carlin:
@robnagler  ready for another review.

PR #2352, opened 2020-03-11
Fix #2350 removed save_chdir and factored out exec()

### LINKED ISSUE #2350: code pkcli modules use pkrunpy and not save_chdir
run and run_background are always executed in run_dir so no need to save_chdir.
pkrunpy does the right thing for py2 and py3 so use that for all codes.

robnagler:
save_chdir is unnecessary for pkcli codes since job_cmd
executes in the run_dir.
exec(PARAMETERS_PYTHON_FILE) is now template_common.exec_parameters
and exec_parameters_with_mpi.
Also cleaned up pkcli.warppba step loops

### DIFF
--- a/sirepo/mpi.py
+++ b/sirepo/mpi.py
@@ -64,7 +64,7 @@ if MPI.COMM_WORLD.Get_rank():
     fn = 'mpi_run.py'
     pkio.write_text(fn, script)
     p = None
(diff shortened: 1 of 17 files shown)


robnagler:
tests pass. Tested warppba explicitly and compared results so confident those changes work.

Just would like you to try some stuff out (issue/2350 branch)

Issue #2351, opened 2020-03-11
JSPEC: Alert user when time step is out of range.


e-carlin:
#1158 brought up the issue of time step for jspec simulations. 

It mentioned two separate but related issues:
1. Letting the user define their own time step (and adjust total simulation time and number of steps accordingly). This is fixed in #2349
2. Alerting the user when their time step is out of range which can result in excessive runs or inaccurate simulations (see #1158 for more context on how to calculate this).

This issue is regarding point 2. 

There is some refactoring that needs to happen to handle this alert (and alerts generally in Sirepo). This warranted creating a new issue.

Definitions:
- **alert**: Either a warning (something happened but it didn't end the simulation) or an error (something happened and this thing ended the simulation).
- **post processing fn**: A function which is run after a simulation (parallel or sequential) that can do interpretation of the output (calculations, log parsing, etc). It will return the result to the supervisor.
- **template logs**: Logs which are written by the codes and can be parsed for information (ex causes of errors)
- **run.log**: stdout and stderr of the pkcli program. Sometimes codes write to stdout and to stderr, and other times they write an application specific log, e.g. elegant writes elegant.log.

Below are some areas that will need refactoring:
- Remove all try/excepts around pkcli run/run_background. Exceptions are going result in exit codes interpreted by the caller (job_cmd) and steps will be taken to handle them (described below).
- In the case of a non-zero exit code: The post processing fn will be called. It will parse the template logs or run.log. If there are no logs or they yield no information then parse the run.log for Python errors. TODO: need to think about cancel case.
- In the case of a zero exit code: parse the "template" logs. If there is information then pass that along as an alert. If there is no information then reply completed with no alerts.
- Elegant can [produce "errors" while it is executing](https://github.com/radiasoft/sirepo/blob/a54988544f8e42885a30a4de33f791b522c435cd/sirepo/template/elegant.py#L118) that do not end the simulation. These will now be called "alerts."
- In jspec, we are now going to compare the time step against the rate calculations (again see #1158 for more context). This calculation is going to happen in the new "post processing fn". The result of the calculation will then be returned to the supervisor as an alert.

These refactorings will break api's that the deprecated runner_api depends on. So, after the implementation the deprecated runner_api will be deleted.

sirepo.mpi currently has mpi_run.out, which needs to be removed. It should just be stderr and stdout of the parent process.

PR #2359, opened 2020-03-16
Issue/2351 refactor

e-carlin:
My apologies in advance for the size of this PR. I thought about breaking it up into separate PR's for each code but I think it is actually nice to see all of the changes together. It shows the pattern and someone with a fresher eye than mine may be able to see more repetition and refactor further. Happy to discuss via skype or break it up if need be.

This is just the refactor described in #2351. I'll open a separate PR later with the fix for JSPEC.

### DIFF
--- a/sirepo/pkcli/job_cmd.py
+++ b/sirepo/pkcli/job_cmd.py
@@ -203,7 +203,9 @@ def _on_do_compute_exit(success_exit, is_parallel, template, run_dir):
         # TODO(e-carlin): impl
         pass
 
-    def _post_processing(): return template.post_execution_processing(**kwargs)
+    def _post_processing():
+        if hasattr(template, 'post_execution_processing'):
+            return template.post_execution_processing(**kwargs)
 
     def _success_exit():
         # is_parallel is necessary because jspec we only want to calculate the rate calculation /time step error in parallel case
@@ -212,6 +214,7 @@ def _on_do_compute_exit(success_exit, is_parallel, template, run_dir):
             state=job.COMPLETED,
             alerts=_post_processing(),
         )
+
     if success_exit:
         return _success_exit()
     return _failure_exit()
--- a/sirepo/pkcli/job_cmd.py
+++ b/sirepo/pkcli/job_cmd.py
@@ -176,13 +176,11 @@ def _do_sbatch_status(msg, template):
 
 
 def _do_sequential_result(msg, template):
-    r = simulation_db.read_result(msg.runDir)
+    r = template_common.read_sequential_result(msg.runDir)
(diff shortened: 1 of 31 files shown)


sirepo/pkcli/job_cmd.py
```diff
@@ -178,16 +188,54 @@ def _do_sbatch_status(msg, template):
...
+            a = _parse_python_errors()
+        return PKDict(state=job.ERROR, error=a)
+
+    def _parse_python_errors():
+        f = run_dir.join(template_common.RUN_LOG)
+        if f.exists():
+            m = re.search(
+                # TODO(e-carlin): I don't actually understand why this works.
+                # Discuss with someone who might.
+                r'Traceback.*\n(\s+.*\n)*(.*?)\n*',
```
e-carlin: @robnagler I need some guidance with this. This regex "works" but I don't really understand why. I couldn't find any information on a good way to parse python tracebacks using a regular expression. [Here is an example](https://regex101.com/r/1c3D5z/1) of it in action. I think it may be more useful to have the last line (`AssertionError: a test error`) than the line I'm currently grabbing. Any suggestions for improvement?

robnagler:
I'm working on the hang first. I hope to get to this tomorrow.

moellep:
These changes look good and simplify the app code which is great. Error processing in the zgoubi case doesn't look correctly formatted - for example if you open the Storage Ring #2 example and change the number of turns to 100000, you will get an error:

Screen Shot 2020-03-17 at 9 41 23 AM

In this case, don't show the top "Error: (..." message and only show it in the zgoubi section, with proper newlines and full text.

e-carlin:
Thanks for the catch @moellep. The root cause of this is that exception parsing from a string is not something easily done in Python and I'm unfamiliar with how to got about rolling my own solution.

I've found [this library](https://boltons.readthedocs.io/en/latest/index.html) which can parse exceptions from logs.
```
from boltons import tbutils
f = run_dir.join(template_common.RUN_LOG)
pkdp(tbutils.ParsedException.from_string(pkio.read_text(f)).to_dict()['exc_msg'])
```

Is this worth introducing a dependency for? Any advice on parsing the logs if we don't want to go the tbutils route?

robnagler:
I don't think we need a dependency. You wrote that the regex works, but then said it wasn't working. I'm a little confused. 

I think the regex can parse what we need. We can run multiple regexes to get various bits that we're interested in. Don't need much reFu, just need to be clear what we want. 

The best way to do this is to select all the cases, and list them in a test. :)

moellep:
I cleaned it up a bit and added another test case with 3b4b2c6.

PR #2362, opened 2020-03-16
Issue/2351 alerts

### LINKED ISSUE #2351: JSPEC: Alert user when time step is out of range.

e-carlin:
**Heads up**: This branch is based on the issue/2351-refactor branch. Wait for that branch (pr #2359) to be merged before reviewing these changes and merging.

Fix #2351. This implements just the jspec / elegant changes.

### DIFF
--- a/sirepo/template/jspec.py
+++ b/sirepo/template/jspec.py
@@ -163,6 +163,25 @@ def get_data_file(run_dir, model, frame, options=None, **kwargs):
         return path.basename, f.read(), 'application/octet-stream'
 
 
+   # TODO(e-carlin): sort
+def get_rates(run_dir):
+    # TODO(e-carlin): single letter var name
+    f = pkio.py_path(run_dir).join(JSPEC_LOG_FILE)
+    assert f.exists(), 'non-existent log file {}'.format(f)
+    o = PKDict(
+        #TODO(pjm): x_range is needed for sirepo-plotting.js, need a better valid-data check
+        x_range=[],
+        rate=[],
+    )
+    for l in pkio.read_text(f).split("\n"):
+        m = re.match(r'^(.*? rate.*?)\:\s+(\S+)\s+(\S+)\s+(\S+)', l)
+        if m:
+            r = [m.group(1), [m.group(i) for i in range(2, 5)]]
+            r[0] = re.sub(r'\(', '[', r[0])
+            r[0] = re.sub(r'\)', ']', r[0])
+            o.rate.append(r)
+    return o
+
 def post_execution_processing(success_exit=True, is_parallel=False, **kwargs):
     if not success_exit or not is_parallel:
         return None
--- a/sirepo/template/jspec.py
+++ b/sirepo/template/jspec.py
@@ -182,10 +182,52 @@ def get_rates(run_dir):
             o.rate.append(r)
     return o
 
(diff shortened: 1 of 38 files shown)


sirepo/template/jspec.py
```diff
@@ -396,6 +425,41 @@ def _generate_parameters_file(data):
...
+            'The time step is too small. This can lead to long run times.'
+            ' Please consider decreasing the total time and/or increasing'
+            ' the number of steps.'
+        )
+    elif s > 4 * r:
+        w = (
+            'The time step is too large. This can lead to innacurate results.'
+            ' Please consider increasing the total time and/or decreasing'
+            ' the number of steps.'
+        )
```
moellep: I'd recommend formatting the error message with newlines, in the same spot as the text breaks because the UI shows the message in a PRE formatted box.

>            'The time step is too small. This can lead to long run times.\n'
>            'Please consider decreasing the total time and/or increasing\n'
>            'the number of steps.'

moellep:
Looks good.

robnagler:
@moellep if you want to merge, please do so. I'm still debugging.

PR #2363, opened 2020-03-17
Only skip job tests if SIREPO_FEATURE_CONFIG_JOB == 0 because SIREPO_FEATURE…

e-carlin:
…_CONFIG_JOB now defaults to 1

830595d6806ba07f021d97cccaba0964c166ccf2 changed the default value of job to True so checking != 1 meant the test was skipped when it shouldn't be.

On a side note, maybe we should have pykern test output what tests were skipped and why?

### DIFF
--- a/tests/adm_and_own_jobs_test.py
+++ b/tests/adm_and_own_jobs_test.py
@@ -11,8 +11,8 @@ from pykern.pkcollections import PKDict
 import time
 
 pytestmark = pytest.mark.skipif(
-    os.environ.get('SIREPO_FEATURE_CONFIG_JOB') != '1',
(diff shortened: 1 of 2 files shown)


robnagler:
> On a side note, maybe we should have pykern test output what tests were skipped and why?

This is a bug in pytest. If you skip all tests in a file, it doesn't exit [with "skipped" code](https://github.com/radiasoft/pykern/blob/7c37d751e6dca923679258af3a27c3ab9bfb92fe/pykern/pkcli/test.py#L59). 

The fix is to dump pytest.

e-carlin:
Ok, removed the test altogether.

PR #2373, opened 2020-03-18
Add premium user and alert users when simulation cancelled due to tim…

e-carlin:
- create role premium
- max run secs for parallel premium vs parllell
-  add premium to admJobs list
- text letting a user know that there simulation was cancelled due to a timeout

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -21,6 +21,7 @@ import sirepo.feature_config
 import sirepo.template
 import datetime
 import importlib
+import werkzeug.exceptions
 
 
 #: what routeName to return in the event user is logged out in require_user
@@ -166,12 +167,12 @@ def init_mock(uid=None, sim_type=None):
         login(sirepo.auth.guest, is_mock=True)
 
 
-def is_premium_user(uid):
-    with auth_db.thread_lock:
-        return auth_db.UserRole.search_by(
-            role=ROLE_PREMIUM,
-            uid=uid,
-        ) is not None
+def is_premium_user():
+    try:
+        check_user_has_role(ROLE_PREMIUM)
+        return True
+    except werkzeug.exceptions.Forbidden:
+        return False
 
 
 def logged_in_user():
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -30,6 +30,7 @@ LOGIN_ROUTE_NAME = 'login'
 METHOD_GUEST = 'guest'
 
 ROLE_ADM = 'adm'
+ROLE_PREMIUM = 'premium'
 
 #: key for auth method for login state
 _COOKIE_METHOD = 'sram'
@@ -122,6 +123,7 @@ def complete_registration(name=None):
 def get_all_roles():
     r = [_role_for_sim_type(t) for t in sirepo.feature_config.cfg().proprietary_sim_types]
     r.append(ROLE_ADM)
+    r.append(ROLE_PREMIUM)
     return r
 
 
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -166,6 +166,14 @@ def init_mock(uid=None, sim_type=None):
         login(sirepo.auth.guest, is_mock=True)
 
 
+def is_premium_user(uid):
+    with auth_db.thread_lock:
+        return auth_db.UserRole.search_by(
+            role=ROLE_PREMIUM,
+            uid=uid,
+        ) is not None
+
+
 def logged_in_user():
     """Get the logged in user
 
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -135,6 +135,8 @@ def api_runSimulation():
     # TODO(e-carlin): This should really be done in job_supervisor._lib_dir_symlink()
     # but that is outside of the Flask context so it won't work
     r.simulation_lib_dir = sirepo.simulation_db.simulation_lib_dir(r.simulationType)
+    if r.isParallel:
+        r.isPremiumUser = sirepo.auth.is_premium_user(r.uid)
     return _request(_request_content=r)
 
 
@@ -245,13 +247,6 @@ def _request_content(kwargs):
         computeJid=s.parse_jid(d, uid=b.uid),
         userDir=str(sirepo.simulation_db.user_dir_name(b.uid)),
     )
-    with sirepo.auth_db.thread_lock:
-        b.pkupdate(
-            isPremiumUser=sirepo.auth_db.UserRole.search_by(
-                role=sirepo.auth.ROLE_PREMIUM,
-                uid=b.uid,
-            ) is not None,
-        )
     return _run_mode(b)
 
 
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -136,7 +136,7 @@ def api_runSimulation():
     # but that is outside of the Flask context so it won't work
     r.simulation_lib_dir = sirepo.simulation_db.simulation_lib_dir(r.simulationType)
     if r.isParallel:
-        r.isPremiumUser = sirepo.auth.is_premium_user(r.uid)
+        r.isPremiumUser = sirepo.auth.is_premium_user()
     return _request(_request_content=r)
 
 
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -17,6 +17,7 @@ import pykern.pkio
 import re
 import requests
 import sirepo.auth
+import sirepo.auth_db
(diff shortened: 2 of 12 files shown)


sirepo/job_supervisor.py
```diff
@@ -280,6 +282,7 @@ def __db_init(self, req, prev_db=None):
             error=None,
             history=self.__db_init_history(prev_db),
             isParallel=c.isParallel,
+            isPremiumUser=c.isPremiumUser,
```
robnagler: add to history. I think isParallel may need to be in history, too.

sirepo/job_supervisor.py
```diff
@@ -271,6 +272,7 @@ def __db_init(self, req, prev_db=None):
         c = req.content
         self.db = PKDict(
             alert=None,
+            cancelledDueToTimeout=False,
```
robnagler: `cancelledAfterSecs=0`.  Then, when the timeout happens, fill in the number of seconds, so that the error message in the GUI can be more informative.

Add to history.

sirepo/job_supervisor.py
```diff
@@ -671,6 +684,8 @@ def _req_is_valid(self, req):
     def _status_reply(self, req):
         def res(**kwargs):
             r = PKDict(**kwargs)
+            if self.db.get('cancelledDueToTimeout'):
```
robnagler: Here's where we may want a lazy db upgrade. cancelledDueToTimeout should appear in the reply as False, always, so other code doesn't have to use get.

So we need some type of fixup that happens on ComputeJob init. We will want to add items to the history, too.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3125,6 +3125,10 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
                   '<div data-sbatch-cores-and-hours="simState"></div>',
                 '</div>',
               '</div>',
+            '<div data-ng-if="cancelledDueToTimeout()" class="alert alert-warning" role="alert">',
+                '<h4 class="alert-heading">Cancelled: Maximum runtime exceeded</h4>',
+                '<p>Your simulation ran longer than the maximum runtime. To increase your maximum runtime please upgrade to <a href="https://radiasoft.net/sirepo" target="_blank">Sirepo Premium</a>.</p>',
```
robnagler: If they are already Premium, we want them to upgrade to Enterprise. Therefore, authState needs isPremiumUser.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3144,6 +3148,10 @@ SIREPO.app.directive('simStatusPanel', function(appState) {
                 return callSimState('getAlert');
             };
 
+            $scope.cancelledDueToTimeout = function() {
```
robnagler: It's ok to have this as the name, but then the cancelledAfterSecs (probably formatted as a human readable time string, e.g. 5 minutes or 1 hour) has to be available to the template.

sirepo/package_data/static/js/srw.js
```diff
@@ -1835,6 +1835,10 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
         },
         template: [
             '<form name="form" class="form-horizontal" autocomplete="off" novalidate>',
+              '<div data-ng-if="simState.getCancelledDueToTimeout()" class="alert alert-warning" role="alert">',
+                  '<h4 class="alert-heading">Cancelled: Maximum runtime exceeded</h4>',
+                  '<p>Your simulation ran longer than the maximum runtime. To increase your maximum runtime please upgrade to <a href="https://radiasoft.net/sirepo" target="_blank">Sirepo Premium</a>.</p>',
```
robnagler: This should be a directive.

tests/job_timeout_test.py
```diff
@@ -12,7 +12,7 @@
 
 def setup_module(module):
     os.environ.update(
-        SIREPO_JOB_SUPERVISOR_MAX_HOURS_PARALLEL='0.002',
```
robnagler: I'm beginning to think we shouldn't have "MAX_HOURS" and instead be "MAX_SECS". I thought it was better to have the config be this way, but 0.002 is just weird. Thoughts?

sirepo/job_supervisor.py
e-carlin: I've added some fixup code. It is enough? Should we start versioning the db files?

https://github.com/radiasoft/sirepo/pull/2373/files#diff-940c7c8cb02a032b1e9c4befe02e6b81R312

sirepo/package_data/static/js/sirepo-components.js
e-carlin: authState has roles in it so I'm now checking those.

https://github.com/radiasoft/sirepo/pull/2373/files#diff-138a2b807a59f609d46eb468d799eaa1R326

sirepo/package_data/static/js/sirepo-components.js
e-carlin: I've mad cancelledAfterSecs available and I'm doing some conversion to format the time. Look ok?

https://github.com/radiasoft/sirepo/pull/2373/files#diff-138a2b807a59f609d46eb468d799eaa1R317

tests/job_timeout_test.py
e-carlin: I agree. I've opened another issue. I'll take care of the Sirepo code and then hand it off to you for rsconf

https://github.com/radiasoft/sirepo/issues/2382

e-carlin:

sirepo/job_api.py
```diff
@@ -134,6 +135,8 @@ def api_runSimulation():
     # TODO(e-carlin): This should really be done in job_supervisor._lib_dir_symlink()
     # but that is outside of the Flask context so it won't work
     r.simulation_lib_dir = sirepo.simulation_db.simulation_lib_dir(r.simulationType)
+    if r.isParallel:
+        r.isPremiumUser = sirepo.auth.is_premium_user(r.uid)
```
robnagler: I don't think it is correct to denormalize r.uid here. Rather, is_premium_user should always be on the auth_user. This avoids problems if there's a bug related to setting r.uid.

sirepo/job_supervisor.py
```diff
@@ -303,6 +308,20 @@ def __db_init_history(self, prev_db):
             PKDict(((k, v) for k, v in prev_db.items() if k in _HISTORY_FIELDS)),
         ]
 
+    @classmethod
+    def __db_load(cls, compute_jid):
+        def _fixup_fields(fields, db):
```
robnagler: I don't think this needs to be a function. It will be less code to inline the `if`.

sirepo/job_supervisor.py
```diff
@@ -594,11 +621,13 @@ def _create_op(self, opName, req, **kwargs):
...
-    def _get_max_run_secs(self, op_name, kind, run_mode):
+    def _get_max_run_secs(self, op_name, run_mode, req):
         if op_name in _UNTIMED_OPS or \
             (run_mode == sirepo.job.SBATCH and op_name == job.OP_RUN):
             return 0
-        t = cfg.max_hours[kind]
+        t = cfg.max_hours[req.kind]
+        if req.kind == job.PARALLEL and req.content.get('isPremiumUser'):
+            t = cfg.max_hours['parallel_premium']
         if op_name == sirepo.job.OP_ANALYSIS:
```
robnagler: I think if `op_name == sirepo.job.OP_ANALYSIS` should be before `if req.kind`, and that second if should be an `elif`. It made me pause to read this, because the code says OP_ANALYSIS takes precedence, but it's written in the opposite order.

sirepo/auth/__init__.py
```diff
@@ -164,6 +166,14 @@ def init_mock(uid=None, sim_type=None):
         login(sirepo.auth.guest, is_mock=True)
 
 
+def is_premium_user(uid):
+    with auth_db.thread_lock:
+        return auth_db.UserRole.search_by(
```
robnagler: `check_user_has_role` should share this code. It's subtle, but important. I note elsewhere that this should always be on auth_user.

sirepo/job_api.py
e-carlin: I think I understand the security implications but I'm not sure I understand "is_premium_user should always be on the auth_user". Should I not pass uid into is_premium_user and the function should get the current user (`logged_in_user()`)?

robnagler: Per our conversation: yes, you should rely on logged_in_user. auth_user is a BOP-concept. Sorry for the confusion.

Issue #2360, opened 2020-03-16
Create premium user role


robnagler:
- [x] create role "premium"
- [x] job_supervisor._get_max_run_secs will check role of user and set accordingly to cfg.max_hours.premium_parallel or just parallel.
- [x] #2280 improve error messages when running out of time
- [ ] configure values for prod/alpha/beta

@e-carlin @moellep anything else?

robnagler:
- [ ] Add premium to the admJobs list
- [ ] Add premium to simulation dump

e-carlin:
@robnagler a few questions:
1. I left a comment regarding #22800 (comment [here](https://github.com/radiasoft/sirepo/issues/2280#issuecomment-600316956))
2. You say "add premium to simulation dump". What is simulation dump?

robnagler:
I don't know that we need to work on #2280 just yet. That's not what I was thinking. When a user's job is killed for a timeout, then we have to put a different error message that indicates you can pay for premium.

The simulation dump is an admin script so it's my task.

robnagler:
Setting roles on actual users handled separately

Issue #1888, opened 2019-09-16
sirepo admin purge_guest_users fails if link to file doesn't exist


robnagler:
Not sure why symlinks to files don't exist, that's a separate issue:

```txt
sirepo admin purge_guest_users --days=30
Traceback (most recent call last):
  File "/home/vagrant/.pyenv/versions/py2/bin/sirepo", line 10, in <module>
    sys.exit(main())
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/sirepo/sirepo_console.py", line 18, in main
    return pkcli.main('sirepo')
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/pykern/pkcli/__init__.py", line 98, in main
    argh.dispatch(parser, argv=argv)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 174, in dispatch
    for line in lines:
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 277, in _execute_command
    for line in result:
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 260, in _call
    result = function(*positional, **keywords)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/sirepo/pkcli/admin.py", line 109, in purge_guest_users
    if (now - now.fromtimestamp(f.mtime())).days <= days:
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/py/_path/local.py", line 416, in mtime
    return self.stat().mtime
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/py/_path/local.py", line 545, in stat
    return Stat(self, py.error.checked_call(os.stat, self.strpath))
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/py/_error.py", line 86, in checked_call
    raise cls("%s%r" % (func.__name__, args))
py.error.ENOENT: [No such file or directory]: stat('/srv/sirepo/db/user/SNIP/srw/SNIP/intensityReport/magn_meas_chx.zip',)
```

[...]

PR #2394, opened 2020-03-24
Issue/2361 prune non premium

### LINKED ISSUE #2367: fastfelo: Missing report dir not being created
I've noticed this on dev - if you delete a report directory and visit the simulation, it raises an exception when it tries to save_chdir to the missing directoy.

To reproduce on dev, start the server and supervisor. Create a new SRW simulation and visit the source page and wait for all the reports to complete. Then delete the simulation's intensityReport directory and re-visit the simulation. Rather than creating the directory and starting a new simulation, the software assumes the simulations results are still available. The report is not fixed unless a field in the UI is changed which triggers it to recompute.

```
WOt8vWG2/intensityReport'', 'stack': 'OSError: [Errno 2] No such file or directory: '/home/vagrant/src/radiasoft/sirepo/run/user/GmRgpDHd/srw/WOt8vWG2/intensityReport'
  File "/home/vagrant/.pyenv/versions/py2/bin/sirepo", line 11, in <module>
    load_entry_point('sirepo', 'console_scripts', 'sirepo')()
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/sirepo_console.py", line 18, in main
    return pkcli.main('sirepo')
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkcli/__init__.py", line 108, in main
    res = argh.dispatch(parser, argv=argv)
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 174, in dispatch
    for line in lines:
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 277, in _execute_command
    for line in result:
  File "/home/vagrant/.pyenv/versions/2.7.16/envs/py2/lib/python2.7/site-packages/argh/dispatching.py", line 260, in _call
    result = function(*positional, **keywords)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/job_cmd.py", line 45, in default_command
    sirepo.template.import_module(msg.simulationType)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/job_cmd.py", line 119, in _do_fastcgi
    with pkio.save_chdir(m.runDir):
  File "/home/vagrant/.pyenv/versions/2.7.16/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkio.py", line 176, in save_chdir
    os.chdir(str(d))
', 'state': 'error'}}
```

e-carlin:
Fix #2367, #1888 

Part of #2360 
- revive pruning pkcli
- Differentiate between premium and non-premium users

### DIFF
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -284,15 +284,20 @@ class _ComputeJob(PKDict):
             history=self.__db_init_history(prev_db),
             isParallel=c.isParallel,
             isPremiumUser=c.get('isPremiumUser'),
-            jobRunMode=c.jobRunMode,
+            jobRunMode=c.get(
+                'jobRunMode',
+                self.db.jobRunMode if 'db' in self else None,
+            ),
             lastUpdateTime=0,
             simName=None,
-            nextRequestSeconds=_NEXT_REQUEST_SECONDS[c.jobRunMode],
             simulationId=c.simulationId,
             simulationType=c.simulationType,
 #TODO(robnagler) when would req come in with status?
             status=req.get('status', job.MISSING),
             uid=c.uid,
+        )
+        self.db.pkupdate(
+            nextRequestSeconds=_NEXT_REQUEST_SECONDS[self.db.jobRunMode],
         )
         if self.db.isParallel:
             self.db.parallelStatus = PKDict(
@@ -509,7 +514,11 @@ class _ComputeJob(PKDict):
         ):
             # Valid, completed, transient simulation
             # Read this first https://github.com/radiasoft/sirepo/issues/2007
-            return await self._receive_api_runStatus(req)
+            r = await self._receive_api_runStatus(req)
+            if r.state == job.MISSING:
+                # happens when the run dir is deleted w/o us knowing (ex admin cli deletion)
+                return await self._receive_api_runSimulation(req)
+            return r
         # Forced or canceled/errored/missing/invalid so run
         o = self._create_op(
             job.OP_RUN,
@@ -588,7 +597,7 @@ class _ComputeJob(PKDict):
             else job.SEQUENTIAL
         req.simulationType = self.db.simulationType
         # run mode can change between runs so use req.content.jobRunMode
-        # not self.db.jobRunmode
+        # not self.db.jobRunMode
         r = req.content.get('jobRunMode', self.db.jobRunMode)
         if r not in sirepo.simulation_db.JOB_RUN_MODE_MAP:
             # happens only when config changes, and only when sbatch is missing
@@ -691,7 +700,15 @@ class _ComputeJob(PKDict):
                         await self.run_dir_acquire(o)
                     await o.prepare_send()
                     o.send()
-                    return await o.reply_get()
+                    r =  await o.reply_get()
+                    # POSIT: All non runSimulation jobs that could run into
+                    # runDirNotFound call _send_with_single_reply()
+                    if r.get('runDirNotFound'):
+                        self.__db_init(req, prev_db=self.db)
+                        assert self.db.status == job.MISSING, \
+                            'expecting missing status={}'.format(self.db.status)
+                        return PKDict(state=job.MISSING)
+                    return r
                 except Awaited:
                     pass
             else:
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -284,10 +284,6 @@ class _ComputeJob(PKDict):
             history=self.__db_init_history(prev_db),
             isParallel=c.isParallel,
             isPremiumUser=c.get('isPremiumUser'),
-            jobRunMode=c.get(
-                'jobRunMode',
-                self.db.jobRunMode if 'db' in self else None,
-            ),
             lastUpdateTime=0,
             simName=None,
             simulationId=c.simulationId,
@@ -296,8 +292,20 @@ class _ComputeJob(PKDict):
             status=req.get('status', job.MISSING),
             uid=c.uid,
         )
+        r = c.get('jobRunMode')
+        if not r:
+            assert c.api != 'api_runSimulation', \
+                'api_runSimulation must have a jobRunMode'
+            # __db_init() will be called when runDirNotFound.
+            # The api_* that initiated the request may not have
+            # a jobRunMode (ex api_downloadDataFile). In that
+            # case use the existing jobRunMode because the
+            # request doesn't care about the jobRunMode
+            r = self.db.jobRunMode
+
         self.db.pkupdate(
-            nextRequestSeconds=_NEXT_REQUEST_SECONDS[self.db.jobRunMode],
+            jobRunMode=r,
+            nextRequestSeconds=_NEXT_REQUEST_SECONDS[r],
         )
         if self.db.isParallel:
             self.db.parallelStatus = PKDict(
@@ -703,14 +711,15 @@ class _ComputeJob(PKDict):
                     await o.prepare_send()
                     o.send()
                     r =  await o.reply_get()
-                    # POSIT: All non runSimulation jobs that could run into
-                    # runDirNotFound call _send_with_single_reply()
+                    # POSIT: any api_* that could run into runDirNotFound
+                    # will call _send_with_single_reply() and this will
+                    # properly format the reply
                     if not r.get('runDirNotFound'):
                         return r
                     self.__db_init(req, prev_db=self.db)
                     assert self.db.status == job.MISSING, \
                         'expecting missing status={}'.format(self.db.status)
-                    return PKDict(state=job.MISSING)
+                    return PKDict(state=self.db.status)
                 except Awaited:
                     pass
             else:
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -498,7 +498,7 @@ class _ComputeJob(PKDict):
             if c:
                 c.destroy(cancel=False)
 
-    async def _receive_api_runSimulation(self, req):
+    async def _receive_api_runSimulation(self, req, called_from_self=False):
         f = req.content.data.get('forceRun')
         if self._is_running_pending():
             if f or not self._req_is_valid(req):
@@ -516,8 +516,10 @@ class _ComputeJob(PKDict):
             # Read this first https://github.com/radiasoft/sirepo/issues/2007
             r = await self._receive_api_runStatus(req)
             if r.state == job.MISSING:
-                # happens when the run dir is deleted w/o us knowing (ex admin cli deletion)
-                return await self._receive_api_runSimulation(req)
+                # happens when the run dir is deleted w/o us knowing (ex pkcli.admin.purge_free_users)
+                assert not called_from_self, \
+                    'Infinite recursion detected. Already called from self.'
+                return await self._receive_api_runSimulation(req, called_from_self=True)
             return r
         # Forced or canceled/errored/missing/invalid so run
         o = self._create_op(
@@ -703,12 +705,12 @@ class _ComputeJob(PKDict):
                     r =  await o.reply_get()
                     # POSIT: All non runSimulation jobs that could run into
                     # runDirNotFound call _send_with_single_reply()
-                    if r.get('runDirNotFound'):
-                        self.__db_init(req, prev_db=self.db)
-                        assert self.db.status == job.MISSING, \
-                            'expecting missing status={}'.format(self.db.status)
-                        return PKDict(state=job.MISSING)
-                    return r
+                    if not r.get('runDirNotFound'):
+                        return r
+                    self.__db_init(req, prev_db=self.db)
+                    assert self.db.status == job.MISSING, \
+                        'expecting missing status={}'.format(self.db.status)
+                    return PKDict(state=job.MISSING)
                 except Awaited:
                     pass
             else:
--- a/sirepo/pkcli/admin.py
+++ b/sirepo/pkcli/admin.py
@@ -77,7 +77,7 @@ def move_user_sims(target_uid=''):
         shutil.move(lib_file, target)
 
 
-def purge_basic_users(days=180, confirm=False):
+def purge_free_users(days=180, confirm=False):
     """Remove basic users simulations
 
     Args:
@@ -87,6 +87,18 @@ def purge_basic_users(days=180, confirm=False):
     Returns:
         (list, list): dirs and uids of removed users (or to remove if confirm)
     """
+
+    def _check_expired_run_dir(path):
+        d = str(path).lower()
+        if 'report' in d or 'animation' in d:
+            for f in pkio.walk_tree(path):
+                # https://github.com/radiasoft/sirepo/issues/1888
+                if not f.exists():
+                    continue
+                if (n - n.fromtimestamp(f.mtime())).days <= days:
+                    return
+            else:
+                res.append(str(path))
 
     def _get_premium_uids():
         r = auth_db.UserRole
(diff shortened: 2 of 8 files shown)


sirepo/pkcli/admin.py
```diff
@@ -88,33 +88,42 @@ def purge_guest_users(days=180, confirm=False):
...
-        auth_db.UserRegistration.delete_all_for_column_by_values('uid', dirs_and_uids.values())
-
-    return dirs_and_uids
+    with auth_db.thread_lock:
+        p = _get_premium_uids()
+        n = srtime.utc_now()
+        r = {}
+        for d in pkio.sorted_glob(simulation_db.user_dir_name().join('*')):
+            u = simulation_db.uid_from_dir_name(d)
+            # TODO(e-carlin): When does a path end in /src?
```
e-carlin: @robnagler when does this happen?

sirepo/job_supervisor.py
```diff
@@ -509,7 +514,11 @@ def _ops_to_cancel():
         ):
             # Valid, completed, transient simulation
             # Read this first https://github.com/radiasoft/sirepo/issues/2007
-            return await self._receive_api_runStatus(req)
+            r = await self._receive_api_runStatus(req)
+            if r.state == job.MISSING:
+                # happens when the run dir is deleted w/o us knowing (ex admin cli deletion)
+                return await self._receive_api_runSimulation(req)
```
robnagler: This is infinitely recursive so we need some sort of counter.

sirepo/job_supervisor.py
```diff
@@ -509,7 +514,11 @@ def _ops_to_cancel():
         ):
             # Valid, completed, transient simulation
             # Read this first https://github.com/radiasoft/sirepo/issues/2007
-            return await self._receive_api_runStatus(req)
+            r = await self._receive_api_runStatus(req)
+            if r.state == job.MISSING:
+                # happens when the run dir is deleted w/o us knowing (ex admin cli deletion)
```
robnagler: `ex pkcli.admin.purge_basic_users` so a reader can find it. `admin cli deletion` is a bit too cryptic.

sirepo/job_supervisor.py
```diff
@@ -691,7 +700,15 @@ def _req_is_valid(self, req):
                         await self.run_dir_acquire(o)
                     await o.prepare_send()
                     o.send()
-                    return await o.reply_get()
+                    r =  await o.reply_get()
+                    # POSIT: All non runSimulation jobs that could run into
```
robnagler: runSimulation or here are the only places we need to check for runDirNotFound.

sirepo/job_supervisor.py
```diff
@@ -691,7 +700,15 @@ def _req_is_valid(self, req):
...
+                    # runDirNotFound call _send_with_single_reply()
+                    if r.get('runDirNotFound'):
```
robnagler: reverse the test so that the longer clause is less indented.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,53 @@ def move_user_sims(target_uid=''):
         shutil.move(lib_file, target)
 
 
-def purge_guest_users(days=180, confirm=False):
-    """Remove old users from db which have not registered.
+def purge_basic_users(days=180, confirm=False):
```
robnagler: basic => free?

sirepo/pkcli/admin.py
robnagler: @moellep I always forget where this code is. We have a "src" user.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,53 @@ def move_user_sims(target_uid=''):
...
+            for f in pkio.walk_tree(d):
+                # https://github.com/radiasoft/sirepo/issues/1888
+                if not f.exists():
+                    continue
+                if (n - n.fromtimestamp(f.mtime())).days <= days:
+                    break
+            else:
+                r[d] = u
+        if confirm:
+            pkio.unchecked_remove(*r.keys())
```
robnagler: we want to purge simulation output, but not the actual user.

sirepo/pkcli/job_agent.py
```diff
@@ -253,18 +253,25 @@ def terminate(self):
...
+        try:
+            if msg.opName == job.OP_ANALYSIS and msg.jobCmd != 'fastcgi':
+                return await self._fastcgi_op(msg)
+            c = _Cmd
+            if msg.jobRunMode == job.SBATCH:
+                c = _SbatchRun if msg.isParallel else _SbatchCmd
+            elif msg.jobCmd == 'fastcgi':
+                c = _FastCgiCmd
+            p = c(msg=msg, dispatcher=self, op_id=msg.opId, **kwargs)
+            if msg.jobCmd == 'fastcgi':
```
robnagler: for clarity, I think the `except` should be above this line. There's some global state being modified, and those operations should be atomic (or in error, i.e. exception raised). I would also move the `c` math into a function. It will clean up the code a little.

sirepo/pkcli/job_agent.py
```diff
@@ -321,6 +328,8 @@ def _fastcgi_accept(self, connection, *args, **kwargs):
             )
             # last thing, because of await: start fastcgi process
             await self._cmd(m, send_reply=False)
+        if not pkio.py_path(msg.runDir).exists():
```
robnagler: would this not want to be first, before starting fastcgi command?

sirepo/pkcli/job_agent.py
```diff
@@ -783,11 +793,20 @@ def start(self):
         self._subprocess.set_exit_callback(self._on_exit)
         return self
 
+    def _assert_run_dir_exists(self):
+        if self.cmd.msg.jobCmd not in ('prepare_simulation', 'compute') \
+           and not self.cmd.run_dir.exists():
```
robnagler: This part could be shared with the other `raise` above. Not critical, but would make the code simpler:

```
if self.cmd.msg.jobCmd not in ('prepare_simulation', 'compute'):
    _assert_run_dir(self.cmd.run_dir)
```

tests/pkcli/admin_test.py
```diff
@@ -10,66 +10,61 @@
...
-
-    res = admin.purge_guest_users(days=days, confirm=True)
-    pkeq(dirs_and_uids, res, '{}: one guest user so one dir and uid to delete', res)
-    pkok(not res.keys()[0].check(dir=True), '{}: directory deleted', res)
-    pkeq(
-        auth_db.UserRegistration.search_by(uid=res.values()[0]),
-        None,
-        '{}: expecting uid to deleted from db', res
+    u = fc.sr_login_as_guest()
+    sirepo.auth_db.UserRole.delete_roles(u, [sirepo.auth.ROLE_PREMIUM])
```
robnagler: how about login as email and then there's no need to reach inside the abstraction that guest always has a ROLE_PREMIUM.

sirepo/job_supervisor.py
e-carlin: Can you phrase what your saying a different way? I'm not sure what you're saying.

e-carlin:
@robnagler don't review yet. Still working on a few of your comments.

e-carlin:
@robnagler ready for another review. All comments addressed except [this one](https://github.com/radiasoft/sirepo/pull/2394#discussion_r398222015).

sirepo/job_supervisor.py
robnagler: The POSIT is that we know all the places that runDirNotFound needs to be caught. There are two: runSimulation and this location. "All non runSimulation" implies that runSimulation also needs to catch runDirNotFound. I thought it should be more explicit.

Probably being too pedantic.

sirepo/job_supervisor.py
```diff
@@ -284,16 +284,21 @@ def __db_init(self, req, prev_db=None):
             history=self.__db_init_history(prev_db),
             isParallel=c.isParallel,
             isPremiumUser=c.get('isPremiumUser'),
-            jobRunMode=c.jobRunMode,
+            jobRunMode=c.get(
+                'jobRunMode',
+                self.db.jobRunMode if 'db' in self else None,
```
robnagler: This is a little funky and deserves a comment, perhaps.

sirepo/job_supervisor.py
```diff
@@ -691,7 +700,15 @@ def _req_is_valid(self, req):
...
                     o.send()
-                    return await o.reply_get()
+                    r =  await o.reply_get()
+                    # POSIT: All non runSimulation jobs that could run into
+                    # runDirNotFound call _send_with_single_reply()
+                    if r.get('runDirNotFound'):
+                        self.__db_init(req, prev_db=self.db)
+                        assert self.db.status == job.MISSING, \
+                            'expecting missing status={}'.format(self.db.status)
+                        return PKDict(state=job.MISSING)
```
robnagler: self.db.status, not job.MISSING. Keeps the encapsulation.

sirepo/job_supervisor.py
robnagler: I don't see this change.

sirepo/job_supervisor.py
robnagler: I don't see this resolved.

sirepo/job_supervisor.py
e-carlin: Thanks. I rephrased it. Thoughts?

sirepo/job_supervisor.py
e-carlin: I added an assert and comment.

e-carlin: And changed the behavior slightly. After more thought we should never run into the None case.

e-carlin:
> Am I supposed to review? I did but maybe not ready?

Yes. I think it was ready. It looked like you possibly were commenting on outdated code? 
I believe 69ae657  and   93b96a3 addressed your comments up to that point. In any event all comments have been addressed now and ready for another pass.

sirepo/job_supervisor.py
```diff
@@ -284,16 +284,29 @@ def __db_init(self, req, prev_db=None):
...
             simulationId=c.simulationId,
             simulationType=c.simulationType,
 #TODO(robnagler) when would req come in with status?
             status=req.get('status', job.MISSING),
             uid=c.uid,
         )
+        r = c.get('jobRunMode')
+        if not r:
+            assert c.api != 'api_runSimulation', \
+                'api_runSimulation must have a jobRunMode'
```
robnagler: provide context eg. printing `c` in the error message.

sirepo/job_supervisor.py
```diff
@@ -509,7 +522,13 @@ def _ops_to_cancel():
         ):
             # Valid, completed, transient simulation
             # Read this first https://github.com/radiasoft/sirepo/issues/2007
-            return await self._receive_api_runStatus(req)
+            r = await self._receive_api_runStatus(req)
+            if r.state == job.MISSING:
+                # happens when the run dir is deleted w/o us knowing (ex pkcli.admin.purge_free_users)
+                assert not called_from_self, \
```
robnagler: `called_from_self` is a bit confusing. `self` is a meaningful concept, and it's a parameter on the call. Rather have it be a counter: recursion_depth, recursion_level, call_depth, etc. When the counter exceeds a threshold, it should throw an error. In this case, `1` is an appropriate threshold.

sirepo/job_supervisor.py
```diff
@@ -691,7 +710,16 @@ def _req_is_valid(self, req):
...
                     o.send()
-                    return await o.reply_get()
+                    r =  await o.reply_get()
+                    # POSIT: any api_* that could run into runDirNotFound
+                    # will call _send_with_single_reply() and this will
+                    # properly format the reply
+                    if not r.get('runDirNotFound'):
+                        return r
+                    self.__db_init(req, prev_db=self.db)
+                    assert self.db.status == job.MISSING, \
```
robnagler: This particular assertion is ok for now. I'm not sure we want it in the long run. At some point we have to trust the code to do what it is supposed to do. We aren't programming Eiffel. We have tests for this type of validation.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
         shutil.move(lib_file, target)
 
 
-def purge_guest_users(days=180, confirm=False):
-    """Remove old users from db which have not registered.
+def purge_free_users(days=180, confirm=False):
```
robnagler: Not exactly what's happening: expunge_free_simulations.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
+    """Remove basic users simulations
```
robnagler: Remove expired simulation results for non-premium users

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
-    if confirm:
-        pkio.unchecked_remove(*dirs_and_uids.keys())
-        auth_db.UserRegistration.delete_all_for_column_by_values('uid', dirs_and_uids.values())
-
-    return dirs_and_uids
+    with auth_db.thread_lock:
+        u = _get_premium_uids()
+        n = srtime.utc_now()
+        res = []
+        for p in pkio.sorted_glob(simulation_db.user_dir_name().join(
```
robnagler: Let's add this abstraction to simulation_db, which controls the structure of simulation run dirs.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
         confirm (bool): delete the directories if True (else don't delete) [False]
 
     Returns:
-        (list, list): dirs and uids of removed guest users (or to remove if confirm)
+        (list, list): dirs and uids of removed users (or to remove if confirm)
     """
 
+    def _check_expired_run_dir(path):
+        d = str(path).lower()
+        if 'report' in d or 'animation' in d:
```
robnagler: I don't see why this code needs to concern itself with the simulation run directory name.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
+            for f in pkio.walk_tree(path):
```
robnagler: I don't think this is necessary, since run directories are re-written every time we can just rely on the ctime for the run_dir.

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
+                # https://github.com/radiasoft/sirepo/issues/1888
+                if not f.exists():
+                    continue
+                if (n - n.fromtimestamp(f.mtime())).days <= days:
+                    return
+            else:
+                res.append(str(path))
+
+    def _get_premium_uids():
```
robnagler: Encapsulate this in auth_db.

I think, perhaps, auth_db should defined ROLES as PKDict(PREMIUM='premium', ...).

sirepo/pkcli/admin.py
```diff
@@ -77,44 +77,60 @@ def move_user_sims(target_uid=''):
...
-                break
-        else:
-
-            dirs_and_uids[d] = uid
-    if confirm:
-        pkio.unchecked_remove(*dirs_and_uids.keys())
-        auth_db.UserRegistration.delete_all_for_column_by_values('uid', dirs_and_uids.values())
-
-    return dirs_and_uids
+    with auth_db.thread_lock:
```
robnagler: This is insufficient and problematic. The problem is that Flask creates run directories without checking auth_db thread_lock, and in any event, flask is running in a different process than this pkcli so thread locks doesn't work.

We need to think about if locking is necessary when removing run directories. 

Since the supervisor controls run directories. We could send it a list of run directories to purge from this code.

The advantage of this is that the supervisor could actually store a different job state when a simulation is purged instead of missing. It would mean not hitting runDirNotFound on this case, that is, the supervisor would see the new state (e.g. job.PURGED_BASIC) and it would be able to return this to the GUI, which could then explain to the user: your simulations were purged, and you might want to upgrade to premium so this doesn't happen.

sirepo/pkcli/job_agent.py
```diff
@@ -208,6 +213,14 @@ def terminate(self):
         finally:
             tornado.ioloop.IOLoop.current().stop()
 
+    def _get_cmd_type(self, msg):
+        c = _Cmd
+        if msg.jobRunMode == job.SBATCH:
+            c = _SbatchRun if msg.isParallel else _SbatchCmd
```
robnagler: `return`, don't assign, then the code is cleaner:

```py
if msg.jobRunMode == job.SBATCH:
     return _SbatchRun if msg.isParallel else _SbatchCmd
elif msg.jobCmd == 'fastcgi':
     return _FastCgiCmd
return _Cmd
```

Easier to follow the logic, especially for the default case

e-carlin:
After discussion with @robnagler we decided that we're going to have a periodic callback in the supervisor that does the pruning at 2am every day. Closing while I work on that.

PR #2401, opened 2020-03-25
Issue/2079 supervisor in http

### LINKED ISSUE #2079: start job_supervisor in service.http
If feature_config.job_supervisor and job_supervisor is not running on a ping, start it. In dev, this will be the common case. Will have to catch control-C. We don't run sirepo.http except in dev or curl installer so it's not going to conflict with a "server install". 

If you are developing job_supervisor, you'll start it manually.

Will also need to support in the test fixture.

Add a ping api to job_supervisor? Not possible, because won't have a jid. Perhaps could create a special jid for this purpose.

e-carlin:
Fix #2079

### DIFF
--- a/sirepo/pkcli/service.py
+++ b/sirepo/pkcli/service.py
@@ -14,10 +14,15 @@ from pykern import pkio
 from pykern import pkjinja
 from pykern import pksubprocess
 from pykern.pkdebug import pkdc, pkdexc, pkdp
+import errno
+import os
 import py
+import threading
 import re
+import signal
 import sirepo.srdb
 import socket
+import subprocess
 
 
 def celery():
@@ -53,23 +58,61 @@ def flower():
 
 
 def http():
-    """Starts Flask server in http mode.
+    """Starts the Flask server and job_supervisor.
 
     Used for development only.
     """
-    from sirepo import server
 
-    with pkio.save_chdir(_run_dir()):
-        use_reloader = pkconfig.channel_in('dev')
-        app = server.init(use_reloader=use_reloader)
-        # avoid WARNING: Do not use the development server in a production environment.
-        app.env = 'development'
-        app.run(
-            host=cfg.ip,
-            port=cfg.port,
-            threaded=True,
-            use_reloader=use_reloader,
-        )
+    def _start_flask():
+        from sirepo import server
+
+        with pkio.save_chdir(_run_dir()):
+            use_reloader = pkconfig.channel_in('dev')
+            app = server.init(use_reloader=use_reloader)
+            # avoid WARNING: Do not use the development server in a production environment.
+            app.env = 'development'
+            app.run(
+                host=cfg.ip,
+                port=cfg.port,
+                threaded=True,
+                use_reloader=use_reloader,
+            )
+
+    def _start_job_supervisor():
+        def run():
+            e = os.environ.copy()
+            e.update(
+                PYENV_VERSION='py3',
+                SIREPO_JOB_DRIVER_MODULES='local',
+            )
+            p = subprocess.Popen(
+                ('pyenv', 'exec', 'sirepo', 'job_supervisor'),
+                preexec_fn=os.setsid, # TODO(e-carlin): only py2 use start_new_session=True, in py3
+                env=e,
+
+            )
+            if p.wait() != 0:
+                on_exit('supervisor exited with non-zero returncode')
+
+        if os.environ.get("WERKZEUG_RUN_MAIN"):
+            return
+
+        t = threading.Thread(target=run)
+        # t.daemon = True
+        t.start()
+        return t
+        # def _send_signal(signal, _):
+        #     os.killpg(p, signal)
+        #     raise KeyboardInterrupt
+
+        # signal.signal(signal.SIGTERM, _send_signal)
+        # signal.signal(signal.SIGINT, _send_signal)
+
+    def on_exit(error):
+        raise RuntimeError(error)
+
+    _start_job_supervisor()
+    _start_flask()
 
 
 def nginx_proxy():
--- a/sirepo/pkcli/service.py
+++ b/sirepo/pkcli/service.py
@@ -82,10 +82,8 @@ def http(driver='local', nersc_proxy=None, nersc_user=None, sbatch_host=None):
     Args:
         driver (string): The driver type to enable (one of local, docker, sbatch, nersc)
         nersc_proxy (string): A proxy nersc can use to reach the supervisor
-    # TODO(e-carlin): talk with rn
         nersc_user(string): A nersc user ??
         sbatch_host (string): A host to ssh into to start sbatch jobs
-    # TODO(e-carlin):
     """
     def _env(py_version):
         e = os.environ.copy()
@@ -148,9 +146,10 @@ def http(driver='local', nersc_proxy=None, nersc_user=None, sbatch_host=None):
             SIREPO_JOB_DRIVER_MODULES='local:sbatch',
             SIREPO_JOB_DRIVER_SBATCH_HOST='cori.nersc.gov',
             SIREPO_JOB_DRIVER_SBATCH_SHIFTER_IMAGE='radiasoft/sirepo:sbatch',
-            # TODO(e-carlin): /global/homes/??? NERSC is down so I can't ssh in and figure it out
-            # /global/homes/${3::1}/$3/.pyenv/versions/py3/bin/sirepo
-            SIREPO_JOB_DRIVER_SBATCH_SIREPO_CMD='/global/homes/{}/{}/.pyenv/versions/py3/bin/sirepo'.format(1, nersc_user),
+            SIREPO_JOB_DRIVER_SBATCH_SIREPO_CMD='/global/homes/{}/{}/.pyenv/versions/py3/bin/sirepo'.format(
+                nersc_user[0],
+                nersc_user,
+            ),
             SIREPO_JOB_DRIVER_SBATCH_SRDB_ROOT='/global/cscratch1/sd/{sbatch_user}/sirepo-dev',
             SIREPO_JOB_SUPERVISOR_SBATCH_POLL_SECS='15',
             SIREPO_JOB_DRIVER_SBATCH_SUPERVISOR_URI='http://{}:8001'.format(nersc_proxy),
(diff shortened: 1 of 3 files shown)


sirepo/pkcli/service.py
```diff
@@ -52,25 +59,159 @@ def flower():
...
+
+def http(driver='local', nersc_proxy=None, nersc_user=None, sbatch_host=None):
+    """Starts the Flask server and job_supervisor.
 
     Used for development only.
+
+    Args:
+        driver (string): The driver type to enable (one of local, docker, sbatch, nersc)
+        nersc_proxy (string): A proxy nersc can use to reach the supervisor
+        nersc_user(string): A nersc user ??
```
e-carlin: etc/run-supervisor.sh has the ability to pass in the nersc_user. The nersc_user is [only used to find the sirepo command](https://github.com/radiasoft/sirepo/blob/master/sirepo/job_driver/sbatch.py#L126). So, I think it is always nagler. Is that right? If so do we need to have it as a param?

sirepo/pkcli/service.py
```diff
@@ -52,25 +59,159 @@ def flower():
...
+                (
+                    'you need to get the host key in ~/.ssh/known_hosts'
+                    ' run: `ssh {} true`'.format(h)
+                )
+            x = m.group(0)
+            e.update(SIREPO_JOB_DRIVER_SBATCH_HOST_KEY=x)
+        return e
+
+    def _env_docker(env):
+        # TODO(e-carlin): py3 has shutil.which()
```
e-carlin: There are some differences between py2 and py3 for finding if commands exist or not. How do you want to solve this? Wrap it? Open an issue and link it to the py3 conversion work?

sirepo/pkcli/service.py
```diff
@@ -52,25 +59,159 @@ def flower():
...
+            SIREPO_SIMULATION_DB_SBATCH_DISPLAY='Vagrant Cluster',
+        )
+        if env['SIREPO_JOB_DRIVER_SBATCH_HOST'] == h:
+            # TODO(e-carlin): py3 has shutil.which()
+            assert distutils.spawn.find_executable('sbatch') is not None, \
+                'slurm not installed:  You need to run `radia_run slurm-dev`'
+        else:
+            env.update(SIREPO_JOB_DRIVER_SBATCH_CORES='4')
+
+    def _exit(*args):
```
e-carlin: @robnagler  Once again I'm stumped by process management. This works in all cases (to the best of my knowledge) except one. When the "parent" flask process dies unexpectedly (kill -9). Then "reloader" flask process process becomes a zombie. I'm unsure how to solve this. I'm using `preexc_fn=os.setsid` which I thought creates a process group so when I kill the parent then the child should die but that isn't the case. Before I dig into how flask might be monkeying around with the processes I figured I'd I ask you if you see a better way to solve this.

sirepo/pkcli/service.py
```diff
@@ -52,25 +59,159 @@ def flower():
...
+    app.env = 'development'
+    app.run(
+        host=cfg.ip,
+        port=cfg.port,
+        threaded=True,
+        use_reloader=use_reloader,
+    )
+
+
+def http(driver='local', nersc_proxy=None, nersc_user=None, sbatch_host=None):
```
robnagler: I guess I wasn't clear. The only drive we want is "local". It's the only one that can work in this simplistic mode. This is for the curl installer and for external developers. They won't be dealing with sbatch or docker configuration.

e-carlin: Argh sorry I missed that. Will do.

e-carlin:
@robnagler ok updated to be only local mode.

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+        e.update(
+            PYENV_VERSION=py_version,
+            SIREPO_JOB_DRIVER_MODULES='local',
+            SIREPO_MPI_CORES='2',
         )
+        return e
+
+    def _exit(*args):
+        for p in processes:
+            os.killpg(os.getpgid(p.pid), args[0])
```
robnagler: Just use kill and pid.

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+            p.wait()
```
robnagler: Need a timeout and then kill -KILL

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+        sys.exit()
```
robnagler: Only exit this way if we get a signal. Normal exit should pass through normal cases.

In the case of Control-C we should exit with a non-zero code (1 is fine).

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+
+    def _start(service, env):
+        c = ['pyenv', 'exec', 'sirepo']
+        c.extend(service)
+        processes.append(subprocess.Popen(
+            c,
+            cwd=str(_run_dir()),
+            env=env,
+            # TODO(e-carlin): py3 use start_new_session=True
```
robnagler: I don't think we need to set a session.

sirepo/pkcli/service.py
robnagler: I think there's too much work. Just use kill and don't set a session (noted elsewhere). 

The only time this doesn't work is sometimes with MPI, but let's not worry about that in this case. This code is either going to be used by a developer or inside a Docker container (curl installer) iwc the user will stop sirepo with docker rm -f.

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+
+        ))
+
+    processes = []
+    signal.signal(signal.SIGINT, _exit)
+    signal.signal(signal.SIGTERM, _exit)
+    _start(['job_supervisor'], _env('py3'))
+    _start(['service', 'flask'], _env('py2'))
+    p, _ = os.wait()
+    processes = filter(lambda x: x.pid != p, processes)
```
robnagler: Don't do so much work. Rather, send a kill on a process that doesn't exist (with try/except around kill).  Check the result of the kill. If it didn't get an error, then waitpid on the process for N seconds, then send a SIGKILL.

sirepo/pkcli/service.py
```diff
@@ -52,24 +59,67 @@ def flower():
...
+    app.env = 'development'
+    app.run(
+        host=cfg.ip,
+        port=cfg.port,
+        threaded=True,
+        use_reloader=use_reloader,
+    )
+
+
 def http():
```
robnagler: We need a separate entry point for simple "http" flask so that we don't start he supervisor. It can be called anything because it is called from run-server.sh anyway.

sirepo/pkcli/service.py
e-carlin: Ugh. Much simpler following what you described. Thanks for the help.

sirepo/pkcli/service.py
e-carlin: Are you looking for something different than the `def flask():` above?

sirepo/pkcli/service.py
```diff
@@ -13,63 +13,75 @@
...
-        app = server.init(use_reloader=use_reloader)
-        # avoid WARNING: Do not use the development server in a production environment.
-        app.env = 'development'
-        app.run(
-            host=cfg.ip,
-            port=cfg.port,
-            threaded=True,
-            use_reloader=use_reloader,
+    def _env():
+        e = os.environ
```
robnagler: This needs to be a copy, because you don't want to modify this process' environ. Don't need update for one value so:

```py
e = PKDict(os.environ)
e. SIREPO_JOB_DRIVER_MODULES = 'local'
return e
```

And this could be done once and used by _start. It doesn't need to be a routine, because it is invariant with respect to the two processes. Just need to define `e` before `_start` is defined.

sirepo/pkcli/service.py
```diff
@@ -13,63 +13,75 @@
...
+        e.update(
+            SIREPO_JOB_DRIVER_MODULES='local',
         )
+        return e
+
+    def _signal_exit(*args):
+        _exit(*args)
```
robnagler: Confusing names. _signal_exit should either be _signal_handler or _kill_and_exit. _exit should be renamed _kill.

sirepo/pkcli/service.py
```diff
@@ -13,63 +13,75 @@
...
+            env=_env(),
+        ))
+
+    processes = []
+    signal.signal(signal.SIGINT, _signal_exit)
+    signal.signal(signal.SIGTERM, _signal_exit)
+    _start(['job_supervisor'])
+    _start(['service', 'flask'])
+    p, _ = os.wait()
+    _exit(signal.SIGTERM)
```
robnagler: This is a little bit odd, but you should reset the signal handler after exiting `http` in a try/finally. This is global state, and if it is ever called directly, you'll want to reset it. `pkcli` functions should be directly callable.

The try/finally should also handle killing the processes. This is also global state. If there's some type of unexpected exception, then the try finally should stop the processes. In this case, it should be BaseException, not just Exception, because it is global state.

In thinking about it, the signal handler does not need to sys.exit. If the signal handler sends a kill, the os.wait in `http` will return, and then the normal sending of signals (to dead processes, but that doesn't matter) will proceed and a normal return from `http` can happen.

The os.wait may raise `ChildProcessError` so you'll need to catch that.

sirepo/pkcli/service.py
```diff
@@ -13,63 +13,75 @@
...
 import py
 import re
+import signal
 import sirepo.srdb
 import socket
+import subprocess
+import sys
 
 
-def celery():
```
robnagler: Remove rabbitmq as well.

sirepo/pkcli/service.py
e-carlin: Makes sense. I believe I resolved everything.

I'm using a try/finally to start/kill the subprocesses. Regarding the bit onBaseException, do I need to worry about it? If I'm killing in a finally then the type of exception doesn't matter. Do you want me to catch BaseException and do a pass to smother any errors? Seems like we'd want those exceptions to bubble up to the caller.

e-carlin:

sirepo/pkcli/service.py
robnagler: You are correct: finally/with is enough.

robnagler:
We will only merge when issue/1560 is merged.

PR #2405, opened 2020-03-26
Fix #2280

### LINKED ISSUE #2280: flashcap: show better message than Pending when waiting in user's queue
If a user if running a long running job, any requests to run another long job will show "Pending" until the first job completes or is canceled.

For example, if you are running a partially-coherent SRW simulation and then try to run an elegant simulation at the same time, the elegant simulation will show "Pending". Reloading the elegant page shows no status, but the server still has the simulation pending. Stopping the SRW simulation and reloading the elegant page shows the simulation completed.

Ideally, the Pending message would say something like "Waiting for SRW simulation to complete" until the simulation is ready to run.

e-carlin:
The hour-glass and pending text is now a link to the user's currently running simulations so they can understand what may be blocking their run.

There is a lot of blue already on the screen so I'm a little concerned that users won't see the link. If others agree I'm open to suggestions to make the link more obvious.

Screen Shot 2020-03-26 at 1 45 32 PM


e-carlin:
I didn't read comments (ex [this](https://github.com/radiasoft/sirepo/issues/2280#issuecomment-603847457)) about additional features. Closing until they're implemented.

PR #2414, opened 2020-03-31
Remove flask-sqlalchemy so db can be used by job_supervisor

### LINKED ISSUE #2409: eliminate flask-sqlalchemy
We are getting to the point where flask-sqlalchemy is in the way, because we want to access the db inside the supervisor.

e-carlin:
Fix #2409

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ pykern.pksetup.setup(
     install_requires=[
         # some "concrete" dependencies in requirements.txt
         'Flask>=1.1',
+        'SQLAlchemy',
         'aenum',
         'cryptography>=2.8',
         'numconv',
@@ -26,7 +27,6 @@ pykern.pksetup.setup(
 
         'Authlib>=0.13; python_version < "3"',
         'Flask-Mail; python_version < "3"',
-        'SQLAlchemy; python_version < "3"',
         'futures; python_version < "3"',
         'pyIsEmail; python_version < "3"',
         'uwsgi; python_version < "3"',
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@ pykern.pksetup.setup(
     install_requires=[
         # some "concrete" dependencies in requirements.txt
         'Flask>=1.1',
-        'Flask_SQLAlchemy',
         'aenum',
         'cryptography>=2.8',
         'numconv',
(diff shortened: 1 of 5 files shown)


sirepo/auth/email.py
```diff
@@ -67,10 +68,7 @@ def api_authEmailAuthorized(simulation_type, token):
         u = AuthEmailUser.search_by(token=token)
         if u and u.expires >= srtime.utc_now():
             n = _verify_confirm(req.type, token, auth.need_complete_registration(u))
-            u.query.filter(
-                (AuthEmailUser.user_name == u.unverified_email),
-                AuthEmailUser.unverified_email != u.unverified_email,
-            ).delete()
+            AuthEmailUser.delete_all_by_filter('changed_email', u)
```
e-carlin: @robnagler this is kind of funky but I'm not sure how to improve it. Since we don't want to expose the _session outside of auth_db this the only way I could come up with performing the operation.

One possibe other option is to make _session a class property of UserDBBase. Then we could create a classmethod on AuthEmailIUser called delete_all_changed_email (or something to that effect) and use the session within it.
```
# auth_db.py
class UserDbBase:
    _session = _session
...

#auth/email.py
class AuthEmailUser:
...
    @classmethod
    def delete_all_changed_email(cls, user):
        with auth_db.thread_lock:
            _session.query(cls).filter_by(cls.user_name == user.unverified_email and cls.unverified_email != user.unverified_eamil).delete()
           _session.commit()

```

robnagler: I think this is clearer to use _session in the subclass.

I don't like `@sqlalchemy.ext.hybrid.hybrid_method`. It is even funkier than the usual `==` hack in sqlalchemy. It's completely unclear what's being returned from this method.

setup.py
```diff
@@ -13,7 +13,6 @@
     install_requires=[
         # some "concrete" dependencies in requirements.txt
         'Flask>=1.1',
-        'Flask_SQLAlchemy',
```
robnagler: You'll need to add a dependency for sqlalchemy

e-carlin: Thanks. We had one below. I've removed the py3 constraint and moved it up.

sirepo/auth/email.py
e-carlin: Agreed, changed.

robnagler:
@moellep see any problems with a merge?

moellep:
Looks good.

PR #2428, opened 2020-04-07
Issue/2426 job cmd spin

### LINKED ISSUE #2426: fastfelo: job_cmd spins after cmd, sometimes
@moellep noticed this first, and I'm seeing it now. The request completed (in error), and job_cmd is running 100%:

```txt
24595 ?        Rsl    1:30 /home/vagrant/.pyenv/versions/3.7.2/envs/py3/bin/python /home/vagrant/.pyenv/versions/py3/bin/sirepo job_cmd /home/vagrant/src/radiasoft/sirepo/run/user/hXGJBnLw/agent-local/S6u3OInYqCMiUmq1ypdYTEdUK2kfI2JB/in-aTH2zMWof2hRg3WERRXRNyV13u5gY6pu.json
```

@e-carlin if you checkout issue/1560, and run warppba electron beam animation, then download raw data file for Plasma Animation, you'll get an error (I'm fixing), and the job_cmd may spin (the first time, refresh doesn't seem to do it).

e-carlin:
Fix #2426.

### DIFF
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -755,11 +755,13 @@ class _Process(PKDict):
         p = None
         try:
             p = self.pkdel('_subprocess').proc.pid
+            pkdlog('pid={}', p)
             os.killpg(p, signal.SIGKILL)
-        except ProcessLookupError:
+        except ProcessLookupError as e:
+            pkdlog('pid={} error={}', p, e)
             pass
         except Exception as e:
-            pkdlog('kill pid={} exception={}', p, e)
+            pkdlog('pid={} error={}', p, e)
 
     def start(self):
         # SECURITY: msg must not contain agentId
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -124,7 +124,11 @@ def start_sbatch():
 class _Dispatcher(PKDict):
 
     def __init__(self):
-        super().__init__(cmds=[], fastcgi_cmd=None)
+        super().__init__(
+            cmds=[],
+            fastcgi_cmd=None,
+            fastcgi_error_count=0,
+        )
 
     def format_op(self, msg, opName, **kwargs):
         if msg:
@@ -198,7 +202,7 @@ class _Dispatcher(PKDict):
     def terminate(self):
         try:
             x = self.cmds
-            self.cmds.clear()
+            self.cmds = []
             for c in x:
                 try:
                     c.destroy()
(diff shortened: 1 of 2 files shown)


sirepo/pkcli/job_cmd.py
```diff
@@ -117,6 +117,9 @@ def _do_fastcgi(msg, template):
             r = b''
             while True:
                 r += s.recv(int(1e8))
+                if not r:
+                    pkdlog('socket closed abruptly')
```
robnagler, line 121: log what agent, etc.

I can't add a comment to L134. Like we do in job_supervisor, we should have a max retries (or exceptions) in this loop. If there are too many retries, the job_cmd should exit with logging info. The job_agent should be prepared that the job_cmd exits prematurely -- it probably already handles this case.

If the job_agent finds itself looping on premature job_cmd exits, then it should exit or at least log the error back to the supervisor.

sirepo/pkcli/job_agent.py
```diff
@@ -755,11 +755,13 @@ def kill(self):
         p = None
         try:
             p = self.pkdel('_subprocess').proc.pid
+            pkdlog('pid={}', p)
             os.killpg(p, signal.SIGKILL)
-        except ProcessLookupError:
+        except ProcessLookupError as e:
+            pkdlog('pid={} error={}', p, e)
```
robnagler, line 761: probably should be logging self

sirepo/pkcli/job_cmd.py
e-carlin: The job_agent will [log information](https://github.com/radiasoft/sirepo/blob/8860e72cf5ea17fd82f67e1a8b2a2bd57a12e995/sirepo/pkcli/job_agent.py#L851) about the agent.

I've added the retries. All errors make it back to the supervisor.

sirepo/pkcli/job_agent.py
e-carlin: Yes, I've added a pkdebug_format()

sirepo/pkcli/job_cmd.py
```diff
@@ -109,33 +111,42 @@ def _do_compute(msg, template):
 def _do_fastcgi(msg, template):
     import socket
 
+    def _recv():
+        r = b''
+        while True:
```
robnagler, line 116: This loop needs a stop before it runs out of memory. I think if we set 1e8 as the max message size, we should enforce it. That is, it should look something like:

```py
b  = s.recv(int(_MAX_MSG))
if not b:
    pkdlog('socket closed abruptly msg={}', msg)
    # maybe a different exception that can be bubbled up easily below
    raise RuntimeError('socket closed abruptly') 
if len(b) + len(r) > _MAX_MSG:
    raise RuntimeError('message larger than {} bytes',  _MAX_MSG)
r += b
if r[-1:] == b'\n':
    return pkjson.load_any(r)
```

Something like that.

Issue #1890, opened 2019-09-16
User storage allocation management policy


robnagler:
We need a policy for managing user storage allocations. Here are some things we know:

* Users without simulations consume at least 35MB.
* Simulations can get very large, e.g. 30GB, but most are small (10MB)
* In prod, we get many users who never (>30 days) come back.
* Storage allocation management can happen daily at the moment.
* When we run out of disk space, we can purge a lot of users.
* Computations are caches, that is, they can be recomputed.
* Some simulations are special (referred to in articles), and must not be deleted.
* Some computations take a very long time to run (hours) so we would like to avoid re-running.
* Some computations take a short time to run and are very large.
* We probably can figure out how long it took for a simulation to run.
* The algorithm can know if a user is a guest or a registered user, but it cannot tell how "important" the user is. We may need some determination of importance related to when we purge computation results.

messamer:
Can I see the data for users who have never come back?   I would like a list with email, date last accessed, their storage usage.  Accounting/Sirepo is a good place for this data.

PR #2433, opened 2020-04-08
Purge non-premium users and re-create missing report dirs.

### LINKED ISSUE #2361: fastfelo: prune simulation results regularly
- [ ] revive pruning pkcli
- [ ] Differentiate between premium and non-premium users.
- [ ] add to rsconf (run nightly)

e-carlin:
Fix #2361 and #2367 

Also addresses #1890


sirepo/simulation_db.py
```diff
@@ -1132,13 +1132,17 @@ def _timestamp(time=None):
...
-def _user_dir():
+def _user_dir(uid=None):
     """User for the session
 
     Returns:
         str: unique id for user
     """
-    uid = sirepo.auth.logged_in_user()
+    if uid:
+        assert not util.in_flask_app_context(), \
```
robnagler, line 1142: I think it's not just flask_app_context, but also in job_cmd and job_agent?

sirepo/simulation_db.py
```diff
@@ -676,14 +676,14 @@ def sim_data_file(sim_type, sim_id):
     return simulation_dir(sim_type, sim_id).join(SIMULATION_DATA_FILE)
 
 
-def simulation_dir(simulation_type, sid=None):
+def simulation_dir(simulation_type, sid=None, uid=None):
```
robnagler, line 679: Could we not add a "acting on behalf of user" feature so that we don't have to pass uid everywhere?  I think this would be more secure, and maybe build it into sirepo.auth.

Let's discuss.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+        f = pkio.sorted_glob(pkunit.work_dir().join(
+            'db',
+            'user',
+            fc.sr_auth_state().uid,
+            'myapp',
+            '*',
+            '*',
+            ),
+        )
+        f = filter(_filter, f)
```
robnagler, line 44: Not relevant to this problem, but I wrote it, and it relates to py3.

py3: `list(filter())`  filter returns a generator. You could also just do `next(filter())` to find out if there's an element.

I would argue a list comprehension might be clearer in this case:

```py
f = [x in for x in _dirs() if 'heightWeightReport' in str(x)]
```py

Where `_dirs` would be a function that would return the sorted_glob.

File: sirepo/srtime.py [hunk 4]
```diff
@@ -44,15 +55,24 @@ def api_adjustTime(days=None):
...
+        'You can only adjust time in internal test'
     adjust_time(days)
     return http_reply.gen_json_ok({
         'adjustedNow': utc_now().isoformat(),
         'systemNow': datetime.datetime.utcnow().isoformat(),
     })
 
 
+def init():
+    if not pkconfig.channel_in_internal_test():
```
robnagler, line 68: does this have to protect itself from re-entrancy.

Conditional should be inverted:

```py
if pkconfig.channel_in_internal_test():
    return
<do_work>
```

sirepo/srtime.py
```diff
@@ -44,15 +55,24 @@ def api_adjustTime(days=None):
     """
     from sirepo import http_reply
 
+    assert pkconfig.channel_in_internal_test(), \
+        'You can only adjust time in internal test'
```
robnagler, line 59: 'API forbidden' is clearer, because that's what it is. I don't like idiomatic assert messages.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+:copyright: Copyright (c) 2019 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+import pytest
+from pykern.pkcollections import PKDict
+import os
+import time
+
+SECS = 3
```
robnagler, line 13: What type of SECS/DAYS and no need to make public.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+    def _make_user_premium(uid):
+        sirepo.auth_db.UserRole.add_roles(uid, [sirepo.auth.ROLE_PREMIUM])
+        r = sirepo.auth_db.UserRole.search_all_for_column('uid')
+        pkunit.pkeq(1, len(r), 'One user with role premium r={}', r)
+        pkunit.pkeq(r[0], uid, 'Expected same uid as user')
+
+    def _run_sim(data):
+        from pykern import pkunit
+        c = None
+        try:
```
robnagler, line 67: this should be modularized. It's copied from elsewhere.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+        r = fc.sr_post(
+            'authEmailLogin',
+            PKDict(email=user, simulationType=fc.sr_sim_type),
+        )
+        fc.sr_email_confirm(fc, r)
+
+    def _make_user_premium(uid):
+        sirepo.auth_db.UserRole.add_roles(uid, [sirepo.auth.ROLE_PREMIUM])
+        r = sirepo.auth_db.UserRole.search_all_for_column('uid')
+        pkunit.pkeq(1, len(r), 'One user with role premium r={}', r)
```
robnagler, line 61: Replace two lines with: `pkeq(r, [uid])`

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+        def _filter(elem):
+            return 'heightWeightReport' in str(elem)
+
+        f = pkio.sorted_glob(pkunit.work_dir().join(
+            'db',
+            'user',
+            fc.sr_auth_state().uid,
+            'myapp',
+            '*',
+            '*',
```
robnagler, line 41: Use `*heightWeightReport*` and remove filter

Or, better, use `walk_tree(_user_dir, file_re('heightWeightReport'))`

You don't need to be so specific, because this is a test, just user_dir.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+    from pykern import pkunit
+    from pykern.pkdebug import pkdp
+    import sirepo.auth
+
+    def _check_run_dir(should_exist=False):
+        def _filter(elem):
+            return 'heightWeightReport' in str(elem)
+
+        f = pkio.sorted_glob(pkunit.work_dir().join(
+            'db',
```
robnagler, line 36: user_dir is not available?

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+            'user',
+            fc.sr_auth_state().uid,
+            'myapp',
+            '*',
+            '*',
+            ),
+        )
+        f = filter(_filter, f)
+        if should_exist:
```
robnagler, line 45: `pkeq(int(should_exit), len(f))`

This is a test, and we can design it to be simpler, e.g. should_exist could be expect_runs which could be an int. The caller would be simpler and just as clear.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+            pkunit.pkeq(1, len(f))
+            return
+        pkunit.pkeq(0, len(f))
+
+    def _login_as_user(user):
```
robnagler, line 50: modularize this in srunit

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+        _make_user_premium(fc.sr_auth_state().uid)
+
+    def _status_eq(next_req, status):
+        pkunit.pkeq(
+            status,
+            fc.sr_post('runStatus', next_req).state
+        )
+
+    fc = auth_fc
+    m = 'heightWeightReport'
```
robnagler, line 112: Nice use of functions to make the meat of the test clear.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -0,0 +1,128 @@
...
+                r = fc.sr_post('runStatus', r.nextRequest)
+                time.sleep(1)
+            else:
+                pkunit.pkfail('Never entered running state')
+        except Exception:
+            if c:
+                fc.sr_post('runCancel', c)
+            raise
+
+    def _register_both_users():
```
robnagler, line 90: register_both_users is unnecessary. It obscures the flow a bit, too. _register is good.

sirepo/srtime.py
e-carlin: I don't believe that it has to protect itself. None of the state that is set could be mucked up by re-entrancy. Agree?

e-carlin:

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+        return db_file.mtime() < (
+            sirepo.srtime.utc_now_as_float() -
+            (cfg.purge_free_after_days * 24 * 60 * 60)
+        )
+
+    def _purge_sim(db_file):
+        # OPTIMIZATION: We assume the uids_of_paid_users doesn't change very
+        # frequently so we don't need to check again. A user could run a sim
+        # at anytime so we need to check that they haven't
+        if not _is_older_than_days(db_file):
```
robnagler, line 904: At this point, we want to check the value inside the db_file, because the db file might be updated for any reason, e.g. db upgrade, but the true test is to validate the computeJobStart. The use of mtime is an efficiency hack to trim a large list quickly to likely candidates.

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+            return
+        d = pkcollections.json_load_any(db_file)
+        if d.status == job.FREE_USER_PURGED:
+            # TODO(e-carlin): should we touch the file so we don't look at again for a bit?
```
robnagler, line 908: We should touch the file, and indeed, the file is touched below when it is purged.

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+            # TODO(e-carlin): should we delete the file after some time?
+            return
+        p = _get_run_dir_path(d)
+        pkio.unchecked_remove(p)
+        d.status = job.FREE_USER_PURGED
+        _ComputeJob.db_write(d)
```
robnagler, line 914: seems like db_update would bea  better call. Indeed, why not just instantiate a ComputeJob instance?

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+        sims_purged.append((db_file, p))
```
robnagler, line 915: too much info. just output the computeJid.

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+
+    try:
+        sims_purged = []
+        for u, v in (await _get_uids_and_files()).items():
+            with sirepo.auth.set_user(u):
+                for f in v:
+                    _purge_sim(f)
+            await tornado.gen.sleep(0)
+        pkdlog('sims purged: {}', sims_purged)
```
robnagler, line 924: 'jids={}'

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+            # TODO(e-carlin): should we delete the file after some time?
+            return
+        p = _get_run_dir_path(d)
+        pkio.unchecked_remove(p)
+        d.status = job.FREE_USER_PURGED
+        _ComputeJob.db_write(d)
+        sims_purged.append((db_file, p))
+
+    try:
+        sims_purged = []
```
robnagler, line 918: I think in py2 this needed to be defined above. It looks a bit odd, but it's ok, I guess. :)

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+        for u, v in (await _get_uids_and_files()).items():
```
robnagler, line 919: maybe make a generator? even with await it might work?

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+            with sirepo.auth.set_user(u):
+                for f in v:
+                    _purge_sim(f)
+            await tornado.gen.sleep(0)
+        pkdlog('sims purged: {}', sims_purged)
+    except Exception as e:
+        pkdlog('error={} stack={}', e, pkdexc())
```
robnagler, line 926: where were we? which user and which sim.

sirepo/srtime.py
```diff
@@ -24,14 +26,23 @@ def adjust_time(days):
     Args:
         days (str): must be integer. If None or 0, no adjustment.
     """
+    def _adjust_supervisor_srtime():
+        import sirepo.job_api
+
+        if not sirepo.util.in_flask_app_context():
```
robnagler, line 32: invert the condition.

sirepo/util.py
```diff
@@ -179,6 +179,11 @@ def json_dump(obj, path=None, pretty=False, **kwargs):
     return res
 
 
+def in_flask_app_context():
+    import flask
+    return flask.current_app
```
robnagler, line 184: `return bool(flask.current_app)`

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
...
+        _purge_free_simulations,
+    )
+
+
+async def _purge_free_simulations():
+    async def _get_uids_and_files():
+        r = PKDict()
+        p = sirepo.auth_db.UserRole.uids_of_paid_users()
+        for f in pkio.sorted_glob(_DB_DIR.join('*')):
+            u = str(f.basename).split('-')[0]
```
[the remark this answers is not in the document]
e-carlin: Thanks for str() catch.

A module level fn in sim_data so we don't need the simType?

sirepo/job_supervisor.py
```diff
@@ -809,3 +854,75 @@ def send(self):
 
     def __hash__(self):
         return hash((self.opId,))
+
+
+def _purge_free_simulations_set():
+    def tomorrow_2am_mst():
```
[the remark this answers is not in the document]
e-carlin: To clarify, the start will be something like next 2am and the period will be 24hours?

I hope none of the physicists see my frequency vs period blunder:)

bruhwiler:
The "frequency vs period blunder" has been observed and noted! 👍

sirepo/job_supervisor.py
```diff
@@ -322,7 +349,7 @@ def __db_update(self, **kwargs):
         return self.__db_write()
 
     def __db_write(self):
-        sirepo.util.json_dump(self.db, path=self.__db_file(self.db.computeJid))
+        self.db_write(self.db)
```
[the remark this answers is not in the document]
e-carlin: In that case _purge_free_sims() [will call](https://github.com/radiasoft/sirepo/blob/2bfa64c538c1558630902d9076b1adf2c0ad3826/sirepo/job_supervisor.py#L914)  __db_write_file()? So we'll be calling a dunder private method from outside the class?

sirepo/job_supervisor.py
e-carlin: What about using lastUpdatetime? It would be closer to mtime and I think closer to "last time of interaction."

sirepo/job_supervisor.py
e-carlin: Yes but what about ten days after we've purged it? After that it will be checked every time which is expensive. Should we do some sort of archiving after a long time (ex 30 days after purging).

robnagler: The problem is that we can't query the db efficiently. For now, this is not a big issue. It happens once a day, and it will finish in minutes, even with thousands of simulations. Let's get this out the door, and worry about the efficiency problem later. 

We will eventually want to denormalize some state in SQL so we can write a query against it. However, this will probably mean replacing sqlite with postgres for production, which I think we can wait to do for some time.

sirepo/job_supervisor.py
robnagler: Is lastUpdateTime always set? If so, yes.

sirepo/job_supervisor.py
e-carlin: I thought about that but chose this because __db_update() calls many other __db_* fn's and those would all require the same classmethod vs instance method pull apart weirdness that db_write required. This would all be resolved if we created an instance. I didn't go that route because we'd need a special get instance pathway for this code (no request, no chaching in cls.instances, etc). That seemed no better but looking at it now I think that was misguided. What do you think?

robnagler: ah, just leave it alone :)

sirepo/job_supervisor.py
e-carlin: I also find it strange. In the same vein I struggle routinely with when to use variables from the outer scope vs passing them into nested functions. I've settled on mostly passing in but open to better ideas...

robnagler: No answers from me. Nested functions are powerful and nice, but they also create this complexity, much like storing on the object or passing to a method...

robnagler:
@bruhwiler i tried to help @e-carlin understand period vs frequency by showing him [this code](https://github.com/radiasoft/zgoubi/blob/190cd5397703951a54a0e6dca53200d9854657aa/zgoubi/ff.f#L99).

sirepo/srtime.py
robnagler: I think it's a practical to have it protect itself. There are some inits (init_apis) that are protected by the caller, but otherwise, the function should protect itself even if the values are idempotent. If someone adds something to the init, they might think it is protect.

sirepo/job_supervisor.py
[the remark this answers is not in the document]
robnagler: We'll enter a "start" and a period. At startup, we'll run it immediately, because we don't know how long it is down, and then we'll schedule the next one at today + start if > now or today + start + period * N, where N is large enough to be greater than now. For example, if period is 1:0:0 and start is 10:0:0 and the supervisor starts at 18:0:0 UTC, then we would have the next purging event at 19:0:0 UTC or in the above equation N would be 9.

sirepo/job_supervisor.py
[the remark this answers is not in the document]
robnagler: Ah, didn't notice. _purge_free_sims should be a class method of ComputeJob since it operates on ComputeJob.

e-carlin:
@robnagler  ready for another review.

PR #2436, opened 2020-04-10
Fix #2232 mpmd implementation

### LINKED ISSUE #2232: mpmd: initial sirepo app
Get this thing going

mkeilman:
This is far enough along to try a merge I think...


robnagler:
Why is this called rsradia?

mkeilman:
> Why is this called rsradia?

Mirroring "jupyter-rs-radia"

robnagler:
The reason for the "rs" in jupyter-rs-radia is that it is a global name space. The name space for sirepo is owned by us so we don't need the rs prefix. It's important, because the only codes that have rs were the ones we were developing (rslinac and rs4pi). The other codes are known by their public names: elegant, srw, etc. Please remove the rs.

mkeilman:
Sorry - one day I'll get naming right.  One problem: the library "radia" is already associated with "Radia" (big R) so I can't also use ```radia.py``` for the template file - ```import radia``` finds the wrong module.  But everything is based on the app name, so how do I reconcile?

robnagler:
You must have this at the top of your file for py2:

```py
```

For py3, this is a non-issue.

PR #2437, opened 2020-04-10
Fix #2280 inform user why simulation pending

### LINKED ISSUE #2280: flashcap: show better message than Pending when waiting in user's queue

e-carlin:
Fix #2280 .

This branch is base off of issue/2416-agent-starting-timeout. Please review and merge that first.


sirepo/job_supervisor.py
```diff
@@ -173,6 +175,24 @@ def cache_timeout_set(self):
...
+    def clear_status(self, op, exception=None):
+        if not exception:
+            self.__db_update(jobStatusMessage=None)
+            return
+        if not isinstance(exception, asyncio.CancelledError):
+            # All non CancelledError exceptions should be preserved in the db
+            # for future debugging
+            return
+        # If we initiated the cancel then clear the jobStatusMessage
+        # TODO(e-carlin): Race condition - Another run may have started so we
```
e-carlin, line 187: @robnagler any thoughts on how you want to handle this? I'm inclined to just get rid of the exception handling code altogether. In the case of an exception (cancelledError or not) we just leave the jobStatusMessage in the db.

e-carlin: There is another race condition. If the runSim deletes the self._cancelledSerial before we get a chance to use it then we also won't remove the jobStatusMessage. This pushes me to be even more in favor of just leaving it on any exception.

robnagler: I think we shouldn't clear_status. It's only used for auditing. We can know, for example, that something was waiting for a cpu_slot when it was canceled.

PR #2444, opened 2020-04-15
fix #2443 cpu slot allocation for same driver

### LINKED ISSUE #2443: flashcap: job_driver cpu slot allocation does not wake all waiters

e-carlin:


sirepo/job_driver/__init__.py
```diff
@@ -414,3 +420,31 @@ def init(job_supervisor_module):
...
+
+    async def wait(self):
+        assert self.locked(), 'Do not wait. Lock is not locked.'
+        await self._event.wait()
+
+    def _acquire(self):
+        assert not self.locked(), 'The lock is locked. It cannot be acquired'
+        self._event.clear()
+
+    def _relese(self):
```
robnagler, line 441: typo, but why add this method? Just use __enter__ and __exit__.

sirepo/job_driver/__init__.py
```diff
@@ -101,13 +102,17 @@ def cpu_slot_free_one(self):
     async def cpu_slot_ready(self):
         if self.cpu_slot:
             return
+        if self._cpu_slot_alloc_lock.locked():
+            await self._cpu_slot_alloc_lock.wait()
+            raise job_supervisor.Awaited()
         try:
             self.cpu_slot = self.cpu_slot_q.get_nowait()
```
robnagler, line 109: The cpu_slot_q should be the new abstraction, not adding another lock.

sirepo/job_driver/__init__.py
```diff
@@ -101,13 +102,17 @@ def cpu_slot_free_one(self):
     async def cpu_slot_ready(self):
         if self.cpu_slot:
             return
+        if self._cpu_slot_alloc_lock.locked():
```
robnagler, line 105: All this can be encapsulated in the new abstraction. The only reason we have it literally is that we don't have good abstractions from tornado so we have to fix them up individually.

I think all Queues probably could use the same object that would manage the resource was allocated. 

I need to think about what the proper abstraction shoudl be.

robnagler:
I think there's a larger question here: global driver state vs op state. Agent startup is global state to the driver and so is cpu slots. When a driver is not ready to process an op, there probably should be a lock on modifying global state that all ops wait for. Cancel is of course a special case.

Let's talk today about this.

e-carlin:
@robnalger code has been fixed. I rebased this branch on issue/1560 because I couldn't get the tests to pass  otherwise. I committed the cardinal sin of rebase - don't do it on a public branch. I know you've checked this branch out. I think if you run `git checkout issue/2433-cput-slot-wait && git rebase --hard origin issue/2433-cput-slot-wait` then all will be well. I don't think you have any local changes but this will blow them away if you do.

sirepo/job_driver/__init__.py
```diff
@@ -198,11 +203,12 @@ async def op_ready(self, op):
 
     def pkdebug_str(self):
         return pkdformat(
-            '{}(a={:.4} k={} u={:.4} {})',
+            '{}(a={:.4} k={} u={:.4} c={} {})',
             self.__class__.__name__,
             self._agentId,
             self.kind,
             self.uid,
+            bool(self.get('cpu_slot')),
```
robnagler, line 211: why not print the slot number?

sirepo/job_supervisor.py
```diff
@@ -768,7 +768,7 @@ def make_lib_dir_symlink(self):
         self.driver.make_lib_dir_symlink(self)
 
     def pkdebug_str(self):
-        return pkdformat('_Op({}, {:.4})', self.opName, self.opId)
+        return pkdformat('_Op({}, {:.4} s={})', self.opName, self.opId, bool(self.get('op_slot')))
```
robnagler, line 771: ditto

sirepo/job_driver/__init__.py
```diff
@@ -101,15 +102,19 @@ def cpu_slot_free_one(self):
...
-        try:
-            self.cpu_slot = self.cpu_slot_q.get_nowait()
-        except tornado.queues.QueueEmpty:
-            self.cpu_slot_free_one()
-            pkdlog('{} await cpu_slot_q.get()', self)
-            self.cpu_slot = await self.cpu_slot_q.get()
-            raise job_supervisor.Awaited()
-        finally:
-            self._cpu_slot_alloc_time = time.time()
+            try:
```
robnagler, line 105: indentation is wrong

sirepo/job_driver/__init__.py
```diff
@@ -101,15 +102,19 @@ def cpu_slot_free_one(self):
...
+                self.cpu_slot = self.cpu_slot_q.get_nowait()
```
robnagler, line 106: I think the lock should be here. Consider this:

* X has a slot
* Y has no slot and ends up waiting below
* X frees a slot
* Z has no slot and gets to run before Y
* Z gets slot before Y

e-carlin: Ah ya I hadn't considered that. That needs to be fixed. Thanks.

But another problem is created by putting the lock there:

- A has a slot
- B has a slot
- C has no slot grabs the lock and waits below
- D waits on the lock
- Agent dies unexpectedly 
- A and B returns their slots
- C gets slot, releases the lock, and __raises await__
- D gets the slot and __does *not* raise awaited__

As I wrote out this example I realized it could never actually happen. All ops are destroyed when the websocket dies unexpectedly and cpu_slot_ready is the first `await` call after waiting on the websocket. With these two guarantees we can be sure that when we acquire the lock the state that was assumed before waiting on the lock is still true. But: 
- It doesn't seem very robust to work on those two assumptions which could change in the future
- This creates an odd case where we have an await and then we don't raise awaited. It is different than all of the other prepare send code.  For example, we raised awaited after waiting on the cpu_slot_q just a couple lines below even though with the guarantees we don't technically have to.

I think a lock with more introspection could solve our problem. We could know if we actually awaited and raise awaited in that case.

Thoughts?

robnagler: If the agent dies, then [aren't all ops destroyed](https://github.com/radiasoft/sirepo/blob/e5ea65387c02110c27548e11f1d3dc9ff3038fb8/sirepo/job_driver/__init__.py#L152)?

e-carlin: Yes they are. I still think the code needs more work than just moving the lock.

> As I wrote out this example I realized it could never actually happen. All ops are destroyed when the websocket dies unexpectedly and cpu_slot_ready is the first await call after waiting on the websocket. With these two guarantees we can be sure that when we acquire the lock the state that was assumed before waiting on the lock is still true. But:
>
>  - It doesn't seem very robust to work on those two assumptions which could change in the future
> - This creates an odd case where we have an await and then we don't raise awaited. It is different than all of the other prepare send code. For example, we raise awaited after waiting on the cpu_slot_q just a couple lines below even though with the guarantees we don't technically have to.

robnagler: The await behavior is the same for [`with self._agent_start_lock`](https://github.com/radiasoft/sirepo/blob/f991659143dd5d029c4c17798c017117462d75ee/sirepo/job_driver/__init__.py#L306). This is probably a bug, too.

You are right: there needs to be an Awaited raised if the lock is waited on. We can't ask this question of any of the tornado.locks so we have to use a queue to do the locking. I think we should write a lock with a queue so that raises Awaited itself. Or we could adopt the run_dir acquire/release approach, but encapsulate it so that we can manage the owner. Owner isn't needed for the agent_start/cpu_slot locks, because we can use a contextmanager to release the lock on cancel.

Let's chat tomorrow.

e-carlin: @robnagler and I have had a couple calls today to discuss this. I'm just putting down some notes from our discussions.

We are in agreement that we need to wrap a tornado lock/event/queue/?? so we can ask if we will await and raise awaited when we do. How exactly to implement this is still unclear.

[run_dir_mutex](https://github.com/radiasoft/sirepo/blob/f991659143dd5d029c4c17798c017117462d75ee/sirepo/job_supervisor.py#L235), [agent_start_lock](https://github.com/radiasoft/sirepo/blob/f991659143dd5d029c4c17798c017117462d75ee/sirepo/job_driver/__init__.py#L306), and [cpu_slot_q_lock](https://github.com/radiasoft/sirepo/blob/f991659143dd5d029c4c17798c017117462d75ee/sirepo/job_driver/__init__.py#L125) are all doing the same thing. They are guarding a resource that only one thing should be accessing at a time. #2433 arose because we had multiple ops for the same driver all trying to acquire a cpu slot with no coordination.

How we go about solving this is still unclear. 

One question we don't have an answer for yet:
After an op waits on a mutex should it acquire the resource it is after and then raise awaited? Or, should it raise awaited immediately and try to get the resource the next time through the flow? Pros for getting the resource immediately are it follows other patterns (ex waiting on an item from a queue) and means the first thing woken up from the mutex will be the first to get the resource (keeps ordering). The downside to this is that we may grab resources "out of order" which could lead to deadlock. For example, a cpu_slot could be acquired but the websocket is no longer alive (can't really happen because all ops are destroy in this case but it illustrates the idea). In addition we may cycle indefinitely (or until max awaited is met) trying to get a resource.

robnagler: See #2467

robnagler:
Replaced by #2473

PR #2450, opened 2020-04-17
fix #2204 finish async tester

### LINKED ISSUE #2204: asynchronous test driver
Make parallel http requests in multiple processes to hammer the alpha server.

### LINKED ISSUE #2443: flashcap: job_driver cpu slot allocation does not wake all waiters

e-carlin:
This branch is based on the issue/1560 branch. Please merge that first. It also requires the cpu slot allocation fix from #2444 to ever successfully finish. #2444 is not currently in this branch once #2444 is stable I'll merge it in.

### DIFF
--- a/sirepo/pkcli/test_http.py
+++ b/sirepo/pkcli/test_http.py
@@ -114,10 +114,8 @@ def default_command():
 
 
 class _App(PKDict):
-    def __init__(self, sim_type, client, **kwargs):
+    def __init__(self, **kwargs):
         super().__init__(
-            sim_type=sim_type,
-            client=client,
             **kwargs
         )
         self.client.app = self
@@ -158,7 +156,7 @@ async def _cancel_all_tasks(tasks):
     for t in tasks:
         t.cancel()
     # We need a gather() after cancel() because there are awaits in the
-    # finally blocks (ex awiat post('run-cancel)). We need return_exceptions
+    # finally blocks (ex await post('run-cancel)). We need return_exceptions
     # so the CancelledErrors aren't raised which would cancel the gather.
     await asyncio.gather(*tasks, return_exceptions=True)
 
@@ -183,7 +181,7 @@ class _Client(PKDict):
 
     async def get(self, uri, caller):
         uri = self._uri(uri)
-        with self._timer('{} {}'.format(uri, caller.pkdebug_str())):
+        with self._timer(uri, caller):
             return self.parse_response(
                 await self._client.fetch(
                     uri,
@@ -247,7 +245,7 @@ class _Client(PKDict):
         data.simulationType = self.sim_type
         uri = self._uri(uri)
 
-        with self._timer('{} {}'.format(uri, caller.pkdebug_str())):
+        with self._timer(uri, caller):
             return self.parse_response(
                 await self._client.fetch(
                     uri,
@@ -262,11 +260,16 @@ class _Client(PKDict):
             )
 
     @contextlib.contextmanager
-    def _timer(self, description):
+    def _timer(self, uri, caller):
         s = time.time()
         yield
-        if 'run-status' not in description:
-            pkdlog('{} elapsed_time={}', description, time.time() - s)
+        if 'run-status' not in uri:
+            pkdlog(
+                '{} {} elapsed_time={}',
+                uri,
+                caller.pkdebug_str(),
+                time.time() - s,
+            )
 
 
     @property
@@ -340,7 +343,7 @@ class _Sim(PKDict):
                         raise
                     finally:
                         if c:
-                            await self._cancel(e)
+                            await self._cancel(error=e)
             except asyncio.CancelledError:
                 # Don't log on cancel error, we initiate cancels so not interesting
                 raise
@@ -365,7 +368,7 @@ class _Sim(PKDict):
             self.get('_waiting_on', '<unknown>'),
         )
 
-    async def _cancel(self, error):
+    async def _cancel(self, error=False):
         c = self._app.client.post(
             '/run-cancel',
             PKDict(
@@ -456,6 +459,7 @@ def _init():
     if cfg:
         return
     cfg = pkconfig.init(
+        emails=(['one@radia.run', 'two@radia.run', 'three@radia.run'], list, 'emails to test'),
         server_uri=('http://127.0.0.1:8000', str, 'where to send requests'),
         run_min_secs=(90, int, 'minimum amount of time to let a simulation run'),
         run_max_secs=(120, int, 'maximum amount of time to let a simulation run'),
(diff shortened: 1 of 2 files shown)


sirepo/pkcli/test_http.py
```diff
@@ -322,60 +288,230 @@ def _uri(self, uri):
...
+    )
+    try:
+        t = asyncio.gather(*s)
+        _register_signal_handlers(t)
+        await t
+    except Exception as e:
+        await _cancel_all_tasks(s)
+        if isinstance(e, asyncio.CancelledError):
+            # Will only be cancelled by a signal handler
+            sys.exit(1)
```
robnagler, line 515: I don't like sys.exit, but I think I understand why it is here. Is there not a way to simply exit tornado?

sirepo/pkcli/test_http.py
```diff
@@ -95,44 +106,84 @@
...
+            '{}(sim_type={} email={})',
+            self.__class__.__name__,
+            self.sim_type,
+            self.client.email,
+        )
+
+
+async def _cancel_all_tasks(tasks):
+    for t in tasks:
+        t.cancel()
```
robnagler, line 159: Can this throw an exception?

sirepo/pkcli/test_http.py
```diff
@@ -95,44 +106,84 @@
...
+    # We need a gather() after cancel() because there are awaits in the
+    # finally blocks (ex awiat post('run-cancel)). We need return_exceptions
```
robnagler, line 161: typo await

sirepo/pkcli/test_http.py
```diff
@@ -144,68 +195,59 @@ async def get(self, uri):
...
             self._headers.Cookie = resp.headers['Set-Cookie']
         if 'json' in resp.headers['content-type']:
-            self.json = pkjson.load_any(resp.body)
-            return self.json
+            return pkjson.load_any(resp.body)
         try:
-            b = resp.body.decode()
+            b = pkcompat.from_bytes(resp.body)
         except UnicodeDecodeError:
             # Binary data files can't be decoded
```
robnagler, line 229: This is subtle. I am not sure we want to detect this case. Rather, we should know what we are expecting and assert that. If we are expecting binary data, pass that in to parse_response.

sirepo/pkcli/test_http.py
```diff
@@ -322,60 +288,230 @@ def _uri(self, uri):
...
+                            '{} expecting error instead of frame={}'.format(
+                                self.pkdebug_str(),
+                                f,
+                            )
+                    except Exception:
+                        e = True
+                        raise
+                    finally:
+                        if c:
+                            await self._cancel(e)
```
robnagler, line 343: pass a keyword so the code is more readable

sirepo/pkcli/test_http.py
```diff
@@ -322,60 +288,230 @@ def _uri(self, uri):
...
-    yield
-    if 'run-status' not in description:
-        pkdlog('{} elapsed_time={}', description, time.time() - s)
+    s = _get_sims(
+        await _get_apps(
+            await _get_clients('one@b.c', 'two@b.c', 'three@b.c'),
+        ),
+    )
+    try:
+        t = asyncio.gather(*s)
```
robnagler, line 508: I would think we would want the signal handlers to be set before the gather, but maybe not. If a login fails, there's no server side state to manage.

_get_sims is the only code that needs to know about the gather, and therefore the code should be:

try:
    t = await _get_sims()
    _register_signal_handlers(t)
    await t

sirepo/pkcli/test_http.py
```diff
@@ -322,60 +288,230 @@ def _uri(self, uri):
...
+        finally:
+            if c:
+                await self._cancel(e)
+
 def _init():
     global cfg
     if cfg:
         return
     cfg = pkconfig.init(
         server_uri=('http://127.0.0.1:8000', str, 'where to send requests'),
```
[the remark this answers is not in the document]
e-carlin: Can I just import pkcli.service and [read its config](https://github.com/radiasoft/sirepo/blob/c71be82eff79dd2f961b00f73801a52b2e315e55/sirepo/pkcli/service.py#L180)?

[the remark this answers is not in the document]
robnagler: A couple things:

* the function needs to be public so pkcli._commands needs to ignore functions that have a particular prefix, e.g. internal_.
* the cfg should be exported via a function, e.g. internal_cfg(), which is called internally to handle initialization.
* Move `import sirepo.srdb` to `_run_dir`, it will limit imports.

Then I think it would be ok to import pkcli.service into test_http.

sirepo/pkcli/test_http.py
e-carlin: I used `sys.exit(1)` so we would have a non-zero exit code after a C-c.  If you prefer it could just be `return`. AFAIK no ioloop cleanup has to happen.

sirepo/pkcli/test_http.py
e-carlin: [I don't think so.](https://github.com/python/cpython/blob/master/Lib/asyncio/tasks.py#L233)

sirepo/pkcli/test_http.py
robnagler: I don't know that we care about the return code from a sigint or term. I do care about sys.exits in the code as they make the code non-modular. Rather use a return

sirepo/pkcli/test_http.py
e-carlin: Yes, there is no state to cleanup with login requests. Registering the signal handler after the login simplifies the handler code. If we registered it before the login we'd have to convert the login coroutines into tasks before we could cancel them. There is no upside to cancelling them so this seems like unnecessary confusion.

Yes and no, _get_sims() is the only code that needs to know about the gather but it isn't the only code that needs to know about the list of tasks that are going to be gathered. The exception handling code needs a reference to the list of tasks so it can cancel them. So I went with this:
```python
def _sims():
  # ...
  return (sims, asyncio.gather(*sims))

try:
  s, t = await _sims()
  _register_signal_handlers(t)
  await t
except Exception as e:
  await cancel_all_tasks(s)
  # ...
```

Having an await return another awaitable seems odd but maybe it is only odd to me because it is new...

robnagler: I don't think it is a problem that `_sims()` returns the tasks to wait on. I think the `await` on `t` is problematic since the `t` is passed to the signal handlers, too. Seems like that's the global state that needs to be managed, not sims.

I don't think `s` needs to be returned. Nobody uses it. There's a reference to `_sims` in the `except`, but I don't think that's correct or useful if it is `s`. There is a `cancel_all_tasks(s)`, but I think that should be `t`, not `s`.

sirepo/pkcli/test_http.py
```diff
@@ -5,21 +5,32 @@
...
 import random
+import signal
+import sys
 
-CODES = PKDict(
+cfg = None
+
+_sims = []
+
+_CODES = PKDict(
```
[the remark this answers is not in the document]
e-carlin: A tuple can't be copy()'ed or shuffle()'ed which I need to do on these lists. I could use deepcopy() instead of copy() and cast to a list then back to a tuple but that doesn't seem worth it. Agree?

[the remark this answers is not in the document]
robnagler: The tuple should be deepcopied. When you need to shuffle, you can use `sample(x, k=len(x))`

sirepo/pkcli/test_http.py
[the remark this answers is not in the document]
e-carlin: Can I skip the first two bullet points and just put the test_http.cfg globally on the module? AFAIK, putting it on the module would do everything `_init()` currently does.

For my own knowledge, what is the advantage of moving the `import sirepo.srdb` into `_run_dir()`? I understand it's advantage in tests where we don't want global imports confusing aspects of the state but I'm unsure of the use case here.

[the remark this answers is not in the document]
robnagler: I don't think I was clear, and the underscores turned into italics.

It is a non-explicit coupling to import the module to init cfg. Some modules need to init later. Rather, there should be function in pkcli.service that gets called to return the cfg. We do this explicitly in feature_config. I think we should do it in other places when we share config this way.

The pkcli.service function is not a command. It technically could be, because it is not doing anything so we could skip the first bullet. If we exported service.cfg() to get the cfg, and also used that same function in the module to refer to the cfg, it would work fine.

Adopting this same model for test_http would work, too. Instead of `cfg` being a value, it is a function, and is `_cfg` is the value, just like `feature_cfg.cfg`. 

There is no "use case" for moving import srdb into `_run_dir`. It's about limiting imports and ensuring initialization order. srdb is only needed in _run_dir so we are reducing the import surface and ensuring initialization order isn't mucked with.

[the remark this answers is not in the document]
e-carlin: Ah, make sense now. Thanks for the detailed explanation.

e-carlin:

PR #2470, opened 2020-04-30
Issue/2425 load flash dynamically

e-carlin:
#2461 was closed inadvertently (maybe because the base branch was pruned). Reopening here.


e-carlin:

sirepo/pkcli/admin.py
```diff
@@ -77,9 +125,21 @@ def move_user_sims(target_uid=''):
         shutil.move(lib_file, target)
 
 
+def proprietary_sim_type_dir(sim_type):
+    return srdb.root().join(cfg.proprietary_code_dir, sim_type)
```
robnagler, line 129: This doesn't need to be a cfg. It's "our" namespace so we can do what we want with it, and it isn't something we can change by just by changing the cfg. 

proprietary_sim_type_dir is proprietary_code_dir. No need for a different name.

sirepo/template/flash.py
```diff
@@ -410,6 +408,17 @@ def _cell_size(f, refine_max):
...
 
+def _extract_zip(data, run_dir):
+    import os
+    import sirepo.pkcli.flash
+    import stat
+    import zipfile
+    zipfile.ZipFile(run_dir.join(_SIM_DATA.proprietary_lib_file_basename(data))).extractall()
+    # extractall() doesn't maintain file permissions
+    # https://bugs.python.org/issue15795
+    os.chmod(run_dir.join(sirepo.pkcli.flash.EXE_NAME), stat.S_IXUSR)
```
robnagler, line 419: I think this import should be the other way around. We don't want to import pkcli.flash here. pkcli.flash should get definitions from the template, or better yet, sim_data.

sirepo/template/flash.py
```diff
@@ -424,10 +433,11 @@ def _cell_size(f, refine_max):
     ],
 }
 
+
 def _generate_parameters_file(data):
     res = ''
     names = {}
-    for line in pkio.read_text(_FLASH_UNITS_PATH[data.models.simulation.flashType]).split('\n'):
+    for line in pkio.read_text(_SETUP_UNITS_FILE).split('\n'):
```
robnagler, line 440: I think this should be in sim_data, too.

tests/pkcli/roles_test.py
```diff
@@ -0,0 +1,44 @@
...
+            sirepo.auth.role_for_sim_type(fc.sr_sim_type),
+        )
+
+    def _check_file(exists=True):
+        pkunit.pkeq(
+            [_PROPRIETARY_FILE] if exists else [],
+            [x.basename for x in pkio.walk_tree(fc.sr_user_dir(), _PROPRIETARY_FILE)],
+        )
+
+    _PROPRIETARY_FILE = 'CapLaser.zip'
```
robnagler, line 32: this is not a global so it shouldn't be all caps.

e-carlin:

Issue #2472, opened 2020-05-01
conftest sometimes does not stop on max_case_run_secs


robnagler:
In debugging #2443 something is hung. Need code to send a kill.

Issue #2467, opened 2020-04-28
asyncio, resources, errors, locks, etc.


robnagler:
Quoted from https://github.com/radiasoft/sirepo/pull/2444#discussion_r416960700:

> @robnagler and I have had a couple calls today to discuss this. I'm just putting down some notes from our discussions.
> 
> We are in agreement that we need to wrap a tornado lock/event/queue/?? so we can ask if we will await and raise awaited when we do. How exactly to implement this is still unclear.
> 
> run_dir_mutex, agent_start_lock, and cpu_slot_q_lock are all doing the same thing. They are guarding a resource that only one thing should be accessing at a time. #2443 arose because we had multiple ops for the same driver all trying to acquire a cpu slot with no coordination.
>
> How we go about solving this is still unclear.
>
> One question we don't have an answer for yet:
> After an op waits on a mutex should it acquire the resource it is after and then raise awaited? Or, should it raise awaited immediately and try to get the resource the next time through the flow? Pros for getting the resource immediately are it follows other patterns (ex waiting on an item from a queue) and means the first thing woken up from the mutex will be the first to get the resource (keeps ordering). The downside to this is that we may grab resources "out of order" which could lead to deadlock. For example, a cpu_slot could be acquired but the websocket is no longer alive (can't really happen because all ops are destroy in this case but it illustrates the idea). In addition we may cycle indefinitely (or until max awaited is met) trying to get a resource.

robnagler:
I wrote [AsyncPython](https://github.com/radiasoft/pykern/wiki/AsynchPython), and it was unfinished

We have had bugs, like #2346 and #2416 and in the past #1082. Deadocks and bugs related to resource management are not new.

We have tried to figure out how to program asynchronously with cooperative multitasking. It's really hard, which is why people write trio, curio, etc. I don't think they fix the problem. I haven't seen a big program that manages resources over long periods of time in Python.

When I start reading about the problem of resources and concurrency, I always end up at shared nothing architectures such as [Erlang](http://www.diva-portal.org/smash/get/diva2:9492/FULLTEXT01.pdf) and [Tandem](https://www.hpl.hp.com/techreports/tandem/TR-85.7.pdf). We have create a shared everything architecture, and we run into bug after bug after bug.

Fortunately, the problem is relatively simple. We can keep it all in our heads (mostly), and we can fix bug after bug while writing tests.

[...]

PR #2473, opened 2020-05-01
fix #2443 job_driver cpu slot allocation does not wake all waiters 

### LINKED ISSUE #2443: flashcap: job_driver cpu slot allocation does not wake all waiters

robnagler:
Also fixes:
* asyncio, resources, errors, locks, etc. #2467
* configure SIREPO_CONFTEST_MAX_CASE_RUN_SECS #2471
* conftest sometimes does not stop on max_case_run_secs #2472

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -152,9 +152,7 @@ class DriverBase(PKDict):
         # must be before cpu_slot_q, because reserves queue position
         # of op relative to other ops, and this also limits ops to
         # one per opName.
-        await self._op_slot_ready(op)
-        # once local resources are acquired, ask for global
-        await op.cpu_slot.alloc('Waiting for CPU resources')
+        await self._slots_ready(op)
 
     @classmethod
     def receive(cls, msg):
@@ -262,19 +260,6 @@ class DriverBase(PKDict):
     def _has_remote_agent(self):
         return False
 
-    async def _op_slot_ready(self, op):
-        """Only one op of each type allowed"""
-        n = op.opName
-        if n in (job.OP_CANCEL, job.OP_KILL):
-            return
-        if n == job.OP_SBATCH_LOGIN:
-            l = [o for o in self.ops.values() if o.opId != op.opId]
-            assert not l, \
-                'received {} but have other ops={}'.format(op, l)
-            return
-        await op.op_slot.alloc('Waiting for another simulation to complete')
-        await op.run_dir_slot.alloc('Waiting for access to simulation state')
-
     def _receive(self, msg):
         c = msg.content
         i = c.get('opId')
@@ -326,6 +311,21 @@ class DriverBase(PKDict):
     def _receive_error(self, msg):
 #TODO(robnagler) what does this mean? Just a way of logging? Document this.
         pkdlog('{} msg={}', self, msg)
+
+    async def _slots_ready(self, op):
+        """Only one op of each type allowed"""
+        n = op.opName
+        if n in (job.OP_CANCEL, job.OP_KILL):
+            return
+        if n == job.OP_SBATCH_LOGIN:
+            l = [o for o in self.ops.values() if o.opId != op.opId]
+            assert not l, \
+                'received {} but have other ops={}'.format(op, l)
+            return
+        await op.op_slot.alloc('Waiting for another simulation to complete')
+        await op.run_dir_slot.alloc('Waiting for access to simulation state')
+        # once local resources are acquired, ask for global
+        await op.cpu_slot.alloc('Waiting for CPU resources')
 
     def _websocket_free(self):
         pass
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -53,7 +53,9 @@ def assign_instance_op(req, jobRunMode, op):
         )
     op.driver = res
     op.driver.ops[op.opId] = op
-    op.op_slot = None
+    op.cpu_slot = op.driver.cpu_slot_q.sr_slot_proxy(op)
+    q = op.driver.op_slot_q.get(op.opName)
+    op.op_slot = q and q.sr_slot_proxy(op)
 
 
 class DriverBase(PKDict):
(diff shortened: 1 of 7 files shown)


sirepo/job_supervisor.py
```diff
@@ -79,6 +79,60 @@ class Awaited(Exception):
...
+            'no secret in message: {}'.format(self.content)
+        assert s == sirepo.job.cfg.server_secret, \
+            'server_secret did not match'.format(self.content)
+        self.handler.write(await _ComputeJob.receive(self))
+
+
+class SlotProxy(PKDict):
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs, _value=None)
```
e-carlin, line 106: **kwargs must be the last thing with no trailing comma in py2. I guess it is valid in py3. Still looks kind of odd.

robnagler:
Still testing. OPAL/Docker is hanging

sirepo/job_driver/__init__.py
```diff
@@ -224,16 +148,11 @@ async def prepare_send(self, op):
...
-            pkdlog('{} {} await _websocket_ready', self, op)
-            await self._websocket_ready.wait()
-            raise job_supervisor.Awaited()
-        await self.cpu_slot_ready()
-        # must be last, because reserves queue position of op
-        # relative to other ops even it throws Awaited when the
-        # op_slot is assigned.
-        await self.op_ready(op)
+        await self._agent_ready(op)
+        # must be before cpu_slot_q, because reserves queue position
```
e-carlin, line 152: This comment should reworded to match new architecture and be moved in _slots_ready()

sirepo/job_supervisor.py
```diff
@@ -757,6 +747,7 @@ def _create_op(self, opName, req, **kwargs):
             req_content=req.copy_content(),
             task=asyncio.current_task(),
         )
+#TODO(robnagler) move into _Op()
```
e-carlin, line 750: Yes, this should be moved. There would be some attributeError issues if the op was destroyed before this was called.

sirepo/job_driver/__init__.py
```diff
@@ -88,54 +85,12 @@ def __init__(self, req):
...
-        finally:
-            self._cpu_slot_alloc_time = time.time()
-
     def destroy_op(self, op):
         """Clear our op and (possibly) free cpu slot"""
         self.ops.pkdel(op.opId)
-        if not self.ops:
-            # might free our cpu slot if no other ops
-            self.cpu_slot_free_one()
+        op.cpu_slot.free()
```
e-carlin, line 91: This should be guarded with an if too. It may be better to move the assignment of the slot proxies to the op in op initialization so we can remove the guard if's.

sirepo/job_driver/__init__.py
```diff
@@ -53,7 +53,9 @@ def assign_instance_op(req, jobRunMode, op):
         )
     op.driver = res
     op.driver.ops[op.opId] = op
-    op.op_slot = None
+    op.cpu_slot = op.driver.cpu_slot_q.sr_slot_proxy(op)
```
e-carlin, line 56: I don't think this will work for docker. cpu_slot_q is a dict in the docker case.

sirepo/job_driver/sbatch.py
```diff
@@ -37,17 +37,12 @@ def __init__(self, req):
             # before it is overwritten by prepare_send
             _local_user_dir=pkio.py_path(req.content.userDir),
             _srdb_root=None,
+            # we allow one of each op type in. This is essentially a no-op (ha ha)
+            # but makes it easier to code the other cases.
+            cpu_slot_q=job_supervisor.SlotQueue(len(job_driver.OPS_WHICH_NEED_SLOTS)),
```
e-carlin, line 42: Grammar nit - I'm not totally positive but I think "that" is better than "which" in this case since the phrase depends on need_slots (you use "that" for restrictive clauses).

sirepo/job_supervisor.py
```diff
@@ -304,44 +328,18 @@ async def receive(cls, req):
...
-                jobStatusMessage=m,
-            )
+        s = self.db.jobStatusMessage
+        p = 'Exception: '
+        if status is not None:
+            # POSIT: no other status begins with exception
+            assert not s or not s.startswith(p), \
+                f'Trying to overwrite existing jobStatusMessage="{s}" with status="{status}"'
+        if exception is not None:
+            status = f'{p}{e}, while {s}'
```
e-carlin, line 341: i think it should be {exception} not {e}. I don't see e defined anywhere.

sirepo/job_supervisor.py
robnagler: See https://github.com/radiasoft/sirepo/issues/2475

sirepo/job_driver/__init__.py
robnagler: `DockerDriver.cpu_slot_q` is `SlotQueue`, which [is initialized from `host.cpu_slot_q`](https://github.com/radiasoft/sirepo/blob/01352458a3d7fd84e9f552d7b9ac45e864e2d586/sirepo/job_driver/docker.py#L53).

sirepo/job_driver/sbatch.py
robnagler: ha ha. The grammarian caught a bug: OPS_WHICH_NEED_SLOTS was not defined.

sirepo/job_supervisor.py
robnagler: Good catch. BTW, did you happen to see how status is being set now differently than before? It allows the exception to be cascaded to the jobStatusMessage.

sirepo/job_supervisor.py
robnagler: 374a8cb5

sirepo/job_driver/__init__.py
robnagler: 374a8cb5

sirepo/job_driver/__init__.py
robnagler: There's always a cpu_slot. I changed sbatch to create cpu_slots (for every driver) the same number of ops. This will work properly because ops is the restrictive case.

sirepo/job_driver/sbatch.py
robnagler: 374a8cb5

sirepo/job_supervisor.py
robnagler: 374a8cb5

e-carlin: No I missed that. Thanks for pointing it out.

robnagler: Do you think it is a good idea, and does the algorithm work correctly?

e-carlin: I think you want `{type(exception)}` or something to that effect. Just doing `{e}` prints out the message of the exception. In the case of CancelledError this is empty so we get jobStatusMessage="Exception: , while Waiting for another simulation to complete"

I think it's a good idea. But, I do think we (maybe just me if you have a grasp) should take time and think about what we want to be in the fields  alert, error, statusMessage, and internalError. It may be worth documenting them. They all may contain a message about what went wrong and it isn't immediately clear what should go where and when they should/shouldn't overlap.

For example, if a user cancels a simulation that is waiting on a resource then the jobStatusMessage='exception: cancelledErrror while waiting for ...'., the status=cancelled, and the error=None. The message has 'exception' but error=None. Is that confusing? Not to me because I know how this all works but may be to someone else. In addition currently we don't display the jobStatusMessage to user's for non running/pending sims but if we change that in the future then they will see some info about 'exception ...' even though from their perspective there was no "error" (exceptions are just how we cancel things internally).

e-carlin: Thinking about this lead me to uncover that the python folks broke CancelledError in py 3.8. It now  inherits from BaseException instead of Exception. This is going to break us in more than a few places. I sent an email to sw to discuss.

https://bugs.python.org/issue32528

PR #2485, opened 2020-05-05
Fix #2481. Eliminate purge_free_start.

### LINKED ISSUE #2481: supervisor_purge_free_sims_test.py:75:_status_eq expect=free_user_purged != actual=completed
I'm rerunning the build, because it failed with, which I assume is spurious:

```txt
_______________________ test_myapp_free_user_sim_purged ________________________
  File "/rscontainers/sirepo/tests/supervisor_purge_free_sims_test.py", line 93, in test_myapp_free_user_sim_purged
    _status_eq(next_req_free, 'free_user_purged')
  File "/rscontainers/sirepo/tests/supervisor_purge_free_sims_test.py", line 75, in _status_eq
    fc.sr_post('runStatus', next_req).state
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/pykern/pkunit.py", line 217, in pkeq
    pkfail('expect={} != actual={}', expect, actual)
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/pykern/pkunit.py", line 230, in pkfail
    raise PKFail('{} {}'.format(call, msg))
pykern.pkunit.PKFail: tests/supervisor_purge_free_sims_test.py:75:_status_eq expect=free_user_purged != actual=completed
```

e-carlin:
We now purge sims every purge_free_period. We keep a cache of
purged jids so the purging process should be fast enough. Startup
may be slow since we will populate the cache.


sirepo/job_supervisor.py
```diff
@@ -257,16 +256,26 @@ def pkdebug_str(self):
         )
 
     @classmethod
-    async def purge_free_simulations(cls, init=False):
+    async def purge_free_simulations(cls):
+        def _next_call_secs():
+            t = datetime.datetime.strptime(cfg.purge_free_period, "%H:%M:%S")
```
robnagler, line 261: see above

sirepo/job_supervisor.py
```diff
@@ -153,8 +153,7 @@ def init():
             sequential=(.1, float, 'maximum run-time for sequential job'),
         ),
         purge_free_after_days=(1000, int, 'how many days to wait before purging a free users simulation'),
-        purge_free_start=('02:00:00', str, 'time to first start purging free users simulations (%H:%M:%S)'),
-        purge_free_period=('01:00:00', str, 'how often to purge free users simulations after start time (%H:%M:%S)'),
+        purge_free_period=('12:00:00', str, 'how often to purge free users simulations (%H:%M:%S)'),
```
robnagler, line 156: I don't like the name purge_free_period.  `purge_non_premium_task_secs` and `purge_non_premium_after_days`. 

It's ridiculous there isn't a better way to parse.

BTW, there should be a _cfg_function that does this conversion. The default value should be None, btw, and we shouldn't purge simulations if it is None.

```py
purge_non_premium_task_secs=(None, _cfg_secs, 'when to clean up simulation runs of free users (%H:%M:%S)'),
```

with this here for now, but could maybe get moved into pkconfig:

```py
def _cfg_secs(value):
    try:
        return int(secs)
    except Exception:
        pass
    t = datetime.datetime.strptime(value, '%H:%M:%S') # single quotes, not double (as copied :)
    return int(...total_seconds())
```

note the `int` on `total_seconds` which returns a `float`.

sirepo/job_supervisor.py
```diff
@@ -293,6 +302,7 @@ def _purge_sim(db_file):
             d.status = job.FREE_USER_PURGED
```
robnagler, line 302: `JOB_RUN_PURGED`.  We don't care why, just that it was purged.

sirepo/job_supervisor.py
```diff
@@ -293,6 +302,7 @@ def _purge_sim(db_file):
...
             cls.__db_write_file(d)
             jids_purged.append(db_file.purebasename)
+            cls._cache_purged_jid(d.computeJid)
```
robnagler, line 305: `jids_purged` is unnecessary, and use purebasename since that's what's used on [278](https://github.com/radiasoft/sirepo/pull/2485/files#diff-940c7c8cb02a032b1e9c4befe02e6b81R278).

I don't think the abstraction in this case helps. Rather:

```
cls._cache_purged_jid.add(db_file.purebasename)
```

sirepo/job_supervisor.py
```diff
@@ -312,7 +322,10 @@ def _purge_sim(db_file):
         except Exception as e:
             pkdlog('u={} f={} error={} stack={}', u, f, e, pkdexc())
         finally:
-            cls._purge_free_simulations_set(s, init)
+            tornado.ioloop.IOLoop.current().call_later(
+                _next_call_secs(),
```
robnagler, line 326: simplified to `cfg.purge_non_premium_task_period`

sirepo/job_supervisor.py
```diff
@@ -666,6 +679,7 @@ def _set_error(op, compute_job_serial):
             simName=req.content.data.models.simulation.name,
             status=job.PENDING,
         )
+        self._cache_purged_jid(self.db.computeJid, invalidate=True)
```
robnagler, line 682:
```py
self._purged_jids_cache._discard(self.__db_file(self.db.computeJid).purebasename))
```

perhaps `self.__db_file()` should be an attribute of self so it is calculated once.

sirepo/job_supervisor.py
e-carlin: Done. But why don't we care? In the future we may purge for other reasons (purge premium after n days or move to an archive after y days). I guess this could be a YAGNI thing but it seems like upside with no downside (the abstraction is exactly the same, just a phrase). More information in the db would be better than less. My general rule of thumb for writing code that is in line with your style is that when we don't really know how something is going to be used more info is better than less. Is that misguided and/or not true in this case?

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2485/commits/8cb87e0aaa7ead347b40f12da4f6895376ca1e92

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2485/commits/8cb87e0aaa7ead347b40f12da4f6895376ca1e92

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2485/commits/8cb87e0aaa7ead347b40f12da4f6895376ca1e92

sirepo/job_supervisor.py
e-carlin: Can you give some context on why jids_purged is unnecessary? I have the same general question as [this comment](https://github.com/radiasoft/sirepo/pull/2485#discussion_r420877358) on when info is useful and when not. If I was debugging a bug possibly related to purging of sims I think this is info I'd want (when a purge for a particular jid happened in relation to everything else going on for that jid).

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2485/commits/3cae5f2b5b9cf09fc482886f2745bbeecd1421e6

sirepo/job_supervisor.py
robnagler: I understand that it may seem strange, but offer too specific codes require too specific explanations with no extra benefit. In this case, we know the simulation was purged along with isPremiumUser so we don't have more information. What we are logging is the act of purging, not the act of purging, because they are a non-premium user.

sirepo/job_supervisor.py
robnagler: We might want to log the time of the purge (or rather, the update to the db). I think that's useful. However, it's unlikely that we'll be able to pinpoint the time of the purge relative to the complaint that a wrong jid was purged. It's more like we'll get a complaint a week or month later, and we will look at the db first to see if it is purged, because that's easier to do than pinpointing when the purge happened in the logs.

sirepo/job_supervisor.py
e-carlin: Ok, thanks.

sirepo/job_supervisor.py
e-carlin: Make sense, thanks.

e-carlin:

PR #2491, opened 2020-05-07
Fix #2483 sbatch creds and agent_timeout_secs use

### LINKED ISSUE #2483: Permission denied: '/global/cscratch1/sd/nersc'
I'm logged in correctly, but it is using the wrong value for the login name to create the scratch directory path.

robnagler:
* Fix report_test not running zgoubi since py3 change 4c2f4a00
* Fix srunit.sr_run_sim to iterate timeout loop correctly 7ef9c71a
* Fix run-supervisor.sh ssh instructions to include nersc_user a174eb5b

### DIFF
--- a/sirepo/job_driver/sbatch.py
+++ b/sirepo/job_driver/sbatch.py
@@ -67,7 +67,7 @@ class SbatchDriver(job_driver.DriverBase):
                 'how long to wait before reading the agent log on start',
             ),
             agent_starting_secs=(
-                cls._AGENT_STARTING_SECS * 3,
+                cls._AGENT_STARTING_SECS_DEFAULT * 3,
                 int,
                 'how long to wait for agent start',
             ),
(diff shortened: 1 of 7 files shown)


sirepo/job_driver/sbatch.py
```diff
@@ -87,33 +87,32 @@ def init_class(cls, job_supervisor_module):
...
-            m.userDir = '/'.join(
-                (
-                    str(self._srdb_root),
-                    sirepo.simulation_db.USER_ROOT_DIR,
-                    m.uid,
-                )
+        c = m.pkdel('sbatchCredentials')
+        if self._srdb_root is None or c:
+            if c:
+                self._creds = c
```
e-carlin, line 93: If someone were to hack our api and send a message with sbatchCredentials on it when an agent is already started then they will never be deleted from the driver (self._creds). We don't log self._creds and we deleted them from the message so I don't think it is a big deal but worth mentioning.

sirepo/job_driver/sbatch.py
```diff
@@ -87,33 +87,32 @@ def init_class(cls, job_supervisor_module):
...
+            if not self.get('_creds') or 'username' not in self._creds:
+                self._raise_sbatch_login_srexception('no-creds', m)
+            self._srdb_root = self.cfg.srdb_root.format(
```
e-carlin, line 96: Again this would only happen if someone hacked our api but if they put sbatchCredentials on a message with a different username than the agent is currently running under. Then the self._srdb_root will change (and thus the userDir and runDir will change) but the agent is still running as the previous user. There will probably be permission denied problems when trying to read/write files for the new user. If self._srdb_root is True we probably shouldn't change it.

sirepo/job_driver/sbatch.py
robnagler: Correct.

sirepo/job_driver/sbatch.py
robnagler: Correct. Hacking APIs is not important as long as it doesn't lead to a breach. It will lead to random server errors, which is fine.

Issue #2400, opened 2020-03-25
Exceeding cori maximum queue time should result in error


robnagler:
We saw a timeout after 27 minutes so perhaps the 30 minutes on the debug queue is fuzzy.

nickgoldring:
I'll just note that whenever I have submitted jobs to the debug queue in the past, the runtime limit has been respected by ±15 seconds.

robnagler:
@nickgoldring did you set it to 30 minutes?

shifter may be different, because it transfers the image to the compute nodes.

nickgoldring:
I usually set the runtime limit to 30 minutes in the batch script although sometimes I decrease it via

`#SBATCH -t HH:MM:SS`

robnagler:
Did you set the hours explicitly when you ran the job? It defaults to .4 (24 minutes) so maybe that's what happened.

nickgoldring:
I did not!  Thanks for pointing this out.  I will assign more hours and rerun the simulations in Sirepo.  I guess that the only time limit should be 48 hrs from the regular queue?

robnagler:
The software toggles between the debug and batch queues based on the number of cores and hours. We don't have a maximum, but we should. I'll change this Issue.

robnagler:
It does result in TIMEOUT so that's good enough

PR #2511, opened 2020-05-13
Fix #2500. Allow user to manually select sbatch queue.

### LINKED ISSUE #2500: Manually select NERSC queue
Currently we choose the queue (qos) for the user based on the number of hours and cores they select. Users should be able to override this and explicitly select a queue.

e-carlin:
If they don't select a queue one will be chosen for them based on
the hours and cores. If they do select a queue but they enter hours
or cores more than the queue allows then the hours and cores will be
reduced to their maximum allowable value.


sirepo/pkcli/job_agent.py
```diff
@@ -651,13 +651,11 @@ async def _prepare_simulation(self):
...
         i = self.msg.shifterImage
         s = o = ''
-        q = "debug" if self.msg.sbatchHours < 0.5 \
-            and self.msg.sbatchCores < 62 * 32 else "regular"
         if i:
 #TODO(robnagler) provide via sbatch driver
             o = f'''#SBATCH --image={i}
 #SBATCH --constraint=haswell
-#SBATCH --qos={q}
+#SBATCH --qos={self.msg.sbatchQueue}
```
robnagler, line 658: This needs to be validated. It's a user input. We can't just insert it. It should match one of the known queues. This can be denormalized in the job_agent.

sbatchHours and sbatchCores are semi-validated (float conversion and comparisons), but it's wonky. They need to be actually validated, too. This should happen in job_api. sbatchCores should be a positive int, and sbatchHours should be a positive float.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3137,24 +3137,73 @@ SIREPO.app.directive('sbatchLoginModal', function() {
...
+            function setQueue() {
+                if (model.sbatchHours <= max.Hours.debug && model.sbatchCores <= max.Cores.debug) {
+                    model.sbatchQueue = 'debug';
+                    return;
+                }
+                if (model.sbatchHours <= max.Hours.regular && model.sbatchCores <=  max.Cores.regular) {
+                    model.sbatchQueue = 'regular';
+                    return;
+                }
+                model.sbatchQueue = 'realtime';
```
robnagler, line 3190: This is not necessarily true. We should not "trim" to realtime, because it is a special queue. If the user requests too many cores or hours, we should simply limit them to the max.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3137,24 +3137,73 @@ SIREPO.app.directive('sbatchLoginModal', function() {
...
                 '<div data-model-field="\'sbatchHours\'" data-model-name="simState.model" data-label-size="3" data-field-size="3"></div>',
                 '<div data-model-field="\'sbatchCores\'" data-model-name="simState.model" data-label-size="3" data-field-size="3"></div>',
+                '<div data-model-field="\'sbatchQueue\'" data-model-name="simState.model" data-label-size="3" data-field-size="3"  data-ng-click="sbatchQueueFieldIsDirty = true"></div>',
             '</div>',
         ].join(''),
-        controller: function($scope) {
+        controller: function($scope, $element) {
+            $scope.sbatchQueueFieldIsDirty = false;
+            var model = appState.models[$scope.simState.model];
+            var coresPerNode = 32;
```
robnagler, line 3157: This is an unnecessary complication for the code. Just list the number of cores.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3137,24 +3137,73 @@ SIREPO.app.directive('sbatchLoginModal', function() {
...
+            var max = {
+                Hours: {
+                    debug: 0.4,
```
robnagler, line 3160: .5 seems to be valid so use that.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3137,24 +3137,73 @@ SIREPO.app.directive('sbatchLoginModal', function() {
...
+                    regular: 47
```
robnagler, line 3161: Is it 47 or 48?

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3137,24 +3137,73 @@ SIREPO.app.directive('sbatchLoginModal', function() {
...
                 '<div data-model-field="\'sbatchCores\'" data-model-name="simState.model" data-label-size="3" data-field-size="3"></div>',
+                '<div data-model-field="\'sbatchQueue\'" data-model-name="simState.model" data-label-size="3" data-field-size="3"  data-ng-click="sbatchQueueFieldIsDirty = true"></div>',
             '</div>',
         ].join(''),
-        controller: function($scope) {
+        controller: function($scope, $element) {
+            $scope.sbatchQueueFieldIsDirty = false;
+            var model = appState.models[$scope.simState.model];
+            var coresPerNode = 32;
+            var max = {
```
robnagler, line 3158: This is nersc, not general. For a departmental cluster, sbatch may have only one queue, for example. I think we are going to want to send this configuration from the server. 

We are hacking this into auth_state now so we might as well pile it in there. It's an abstraction (job_run_mode_map). This could contain all the configuration. We will have to get the data from sbatch.cfg, which is in the supervisor. Therefore, we may have to move it into job.py. 

Thoughts?

sirepo/package_data/static/js/sirepo-components.js
e-carlin: I'm confused about the rules. Seems that they are fuzzy. We have cores for debug [set to < 62](https://github.com/radiasoft/sirepo/pull/2511/files#diff-1773844367e9b768846ca4b000047165L654) even though NERSC supports 64 cores for the debug queue.

https://github.com/radiasoft/sirepo/issues/2400 (timeout issue where Nick said that the rules are fuzzy)

https://docs.nersc.gov/jobs/policy/#haswell

What should we do?

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/fb55c2011646a1e6d9593bfb0c069e4c4223b959

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/fb55c2011646a1e6d9593bfb0c069e4c4223b959

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/fb55c2011646a1e6d9593bfb0c069e4c4223b959

sirepo/package_data/static/js/sirepo-components.js
robnagler: I think #2400 was a red herring. 64 cores and half hour for debug queue.

Also, we'll want the premium queue to be selectable, too.

sirepo/job_api.py
```diff
@@ -257,6 +257,10 @@ def _request_content(kwargs):
 
 
 def _run_mode(request_content):
+    def _validate_value(value, *obj_types):
+        assert type(value) in obj_types and value > 0, \
+            f'value={value} expecting a {obj_type} > 0'
```
robnagler, line 262: obj_types?

sirepo/job_api.py
```diff
@@ -274,8 +278,11 @@ def _run_mode(request_content):
                 request_content.computeJid,
             )
         )
+    _validate_value(m.sbatchCores, int)
+    _validate_value(m.sbatchHours, float, int)
     return request_content.pkupdate(
         jobRunMode=j,
         sbatchCores=m.sbatchCores,
         sbatchHours=m.sbatchHours,
+        sbatchQueue=m.sbatchQueue,
```
robnagler, line 287: queue name needs to be validated.

e-carlin: That's done in the agent. I assumed that when you said the queue [types could be denormalized in the job_agent](https://github.com/radiasoft/sirepo/pull/2511#discussion_r424773587) you wanted the check there too? Seems a little strange to have the job_api depend on the job_agent.

sirepo/job_api.py
e-carlin: Sorry. Was type which I thought might yield funny results with isinstance() because type is a type... so changed it to obj_type (object type). Then realized hours could be a float or an int so changed to obj_types. Now just types:)
https://github.com/radiasoft/sirepo/pull/2511/commits/255438815245fa038582a6c8722f3ec1f8ee52bf

sirepo/package_data/static/js/sirepo-components.js
e-carlin: 48.

https://github.com/radiasoft/sirepo/pull/2511/commits/255438815245fa038582a6c8722f3ec1f8ee52bf

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/255438815245fa038582a6c8722f3ec1f8ee52bf

sirepo/job_api.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/579c3fbbfc59f452081b73a6b75a90de77873122

The queue values are hardcoded right now. Should they come from pkconfig?

e-carlin: ^ That comment was for another one of your comments. Disregard.

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/579c3fbbfc59f452081b73a6b75a90de77873122

The queue values are hardcoded. Should they come from pkconfig?

robnagler: That's fine. They aren't changing. As we add clusters, we'll figure out the model for configuration.

sirepo/job_api.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/29dd6577c388758d8ee08269639ad663149ae0af

e-carlin:
@robnagler all comments addressed. If it looks good lmk and I'll merge it.

sirepo/job.py
```diff
@@ -51,6 +51,9 @@
 #: where user lib file directories are linked for static download (job_supervisor)
 LIB_FILE_ROOT = None
 
+# POSIT: These are the same queues as in schema-common.common.enum.SbatchQueue
+SBATCH_QUEUES = ('debug', 'premium', 'realtime', 'regular')
```
robnagler, line 55: I think we should call this NERSC, not SBATCH.  Probably should be a frozenset.

sirepo/job.py
```diff
@@ -102,6 +105,21 @@
 #: categories of jobs
 KINDS = frozenset((SEQUENTIAL, PARALLEL))
 
+#: Max values for each sbatch queue
+SBATCH_QUEUE_MAXES = PKDict(
```
robnagler, line 109: SBATCH_QUEUE_MAXES => NERSC_QUEUE_MAX. I don't like "MAXES".  LIMITS might be ok, but MAX is sufficient. Plural is unnecessary.

sirepo/job_api.py
```diff
@@ -271,8 +271,30 @@ def _run_mode(request_content):
...
+        _validate_and_add(field, _v)
+
+    def _number(field, *types):
+        def _v(value):
+            assert type(value) in types and value > 0, \
+                f'value={value} expecting a {types} > 0'
+        _validate_and_add(field, _v)
+
+    def _validate_and_add(field, validate):
+        v = getattr(compute_model, field)
```
robnagler, line 292: `getattr` is not right. It should be compute_model.get(field).

sirepo/job_api.py
```diff
@@ -271,8 +271,30 @@ def _run_mode(request_content):
...
+        validate(v)
+        setattr(request_content, field, v)
```
robnagler, line 294: `setattr` is not right. It should be request_content[field] = v

sirepo/job_api.py
```diff
@@ -271,8 +271,30 @@ def _run_mode(request_content):
...
+
+    _number('sbatchCores', int)
```
robnagler, line 296: I think this is too complicated with all these names for one or two lines of code that are self-describing. The error messages are being made generic, when they could be literal and more complete. Here's a rewrite:

```py
m = compute_model
c = request_content
 assert m.sbatchQueue in sirepo.job.NERSC_QUEUES, \
    f'sbatchQueue={m.sbatchQueue} not in NERSC_QUEUES={sirepo.job.NERSC_QUEUES}'
c.sbatchQueue = m.sbatchQueue
for f in 'sbatchCores', 'sbatchHours':
    assert m[f] > 0, f'{f}={m[f]} must be greater than 0'
    c[f] = m[f]
```

The error messages then are self-explanatory. 

sbatchHours can be a float and still tested against 0 in Python. If sbatchHours is not a number, this will fail in a way, but it's an unlikely case, because the GUI is supposed to be sending valid values. We are just protecting against a hacker sending a string where a number should have been sent. It's a balance between producing useful error messages in the important case (number is negative -- which the GUI might protect against).

One Python "bug" is that assertions are not functions. This is annoying. We'd like to able to pass them around, not having to pass them in a wrapped function. In this case, the improved quality of the self-documenting assertions and reduced number of abstractions outweighs the duplication.

At least imiho. :)

sirepo/package_data/static/json/radia-schema.json
```diff
@@ -161,7 +161,8 @@
             "precision": ["Precision (T)", "Float", 0.00010, "", ""],
             "result": ["Result", "String", ""],
             "sbatchCores": ["Cores", "Integer", 1],
-            "sbatchHours": ["Hours", "Float", 0.4]
+            "sbatchHours": ["Hours", "Float", 0.4],
+            "sbatchQueue": ["Queue", "SbatchQueue", "debug"]
```
robnagler, line 165: SbatchQueue => NERSCQueue, because that's what it is for now. Leave sbatchQueue as the field, since that may be used if we add another HPC with multiple queues. Then, we'll have to solve the 2-case. The word NERSCQueue here flags it as a 1-case.

sirepo/pkcli/job_agent.py
```diff
@@ -651,13 +651,11 @@ async def _prepare_simulation(self):
     def _sbatch_script(self):
         i = self.msg.shifterImage
         s = o = ''
-        q = "debug" if self.msg.sbatchHours < 0.5 \
-            and self.msg.sbatchCores < 62 * 32 else "regular"
         if i:
 #TODO(robnagler) provide via sbatch driver
```
robnagler, line 655: Irrelevant comment?

sirepo/pkcli/job_agent.py
```diff
@@ -651,13 +651,11 @@ async def _prepare_simulation(self):
...
             o = f'''#SBATCH --image={i}
```
robnagler, line 656: Add a comment:

```py
#POSIT: job_api has validated values
```

sirepo/job_api.py
e-carlin: I wanted an AttributeError to be raised when field isn't found. Since we rarely use the `compute_model[field]` syntax I'm slightly embarrassed to say I didn't think of using it and came up with this instead...:)

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

sirepo/package_data/static/json/radia-schema.json
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

sirepo/job_api.py
e-carlin: Much better. Amazing how good I am at creating gobs of code and how good your are at deleting them:)

https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

sirepo/job.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

sirepo/job.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2511/commits/779ddedc7bc17f3c385612be4fee7c6b6a503204

e-carlin:

e-carlin:
> Looks good. I think I'm not supposed to merge this.

Yes. I'll merge shortly.

PR #2515, opened 2020-05-14
Fix #2410. All role lookups by email or uid.

### LINKED ISSUE #2410: pkcli.roles should allow lookup by email
If not found by uid, lookup by email if configured.

e-carlin:


sirepo/pkcli/roles.py
```diff
@@ -5,50 +5,59 @@
...
     sirepo.server.init()
-    assert sirepo.auth_db.UserRegistration.search_by(uid=uid), \
-        'uid {} not found'.format(uid)
+
+    # POSIT: Uid's are from the base62 charset so an '@' implies an email.
+    if '@' in uid_or_email:
+        u = sirepo.auth.email.AuthEmailUser.search_by(user_name=uid_or_email)
+    else:
+        u = sirepo.auth_db.UserRegistration.search_by(uid=uid_or_email)
+    assert u, \
```
robnagler, line 57: This should be a pkcli.command_error, not an assert.

sirepo/pkcli/roles.py
```diff
@@ -5,50 +5,59 @@
...
+        'uid_or_email={} not found'.format(uid_or_email)
     if roles:
         a = sirepo.auth.get_all_roles()
         assert set(roles).issubset(a), \
             'roles={} not a subset valid roles={}'.format(roles, a)
```
robnagler, line 62: We've started this naming mechanism of x=y, which is fine, but the tags should be unique: all_roles={}.

sirepo/pkcli/roles.py
```diff
@@ -5,50 +5,59 @@
...
     Args:
-        uid (str): Uid of the user
+        uid_or_email (str): Uid or email of the user
         *args: The roles to assign to the user
     """
-    _check_uid_and_roles(uid, args)
-    sirepo.auth_db.UserRole.add_roles(uid, args)
-    admin.audit_proprietary_lib_files(uid)
+    u = _check_uid_or_email_and_roles(uid_or_email, args)
+    sirepo.auth_db.UserRole.add_roles(u, args)
```
robnagler, line 24: While this is probably ok, there is a problem that there's no communication with the supervisor. We add the role to sqlite, but the files aren't there. Really, sqlite should be updated after, but of course, then the auditor won't work.

We need some type of locking, but it's not important to solve right now.

sirepo/pkcli/roles.py
```diff
@@ -5,50 +5,59 @@
...
 
-def _check_uid_and_roles(uid, roles):
+def _check_uid_or_email_and_roles(uid_or_email, roles):
     sirepo.server.init()
-    assert sirepo.auth_db.UserRegistration.search_by(uid=uid), \
-        'uid {} not found'.format(uid)
+
+    # POSIT: Uid's are from the base62 charset so an '@' implies an email.
+    if '@' in uid_or_email:
+        u = sirepo.auth.email.AuthEmailUser.search_by(user_name=uid_or_email)
```
robnagler, line 54: This works, and is fine, but it doesn't work for other auth types (github). Probably should note that.

We don't have to worry about the '@'. Probably could search for a list of valid auth types. Don't fix Just noting.

sirepo/pkcli/roles.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2515/commits/af8a2bb2a236d342c64f98613d2da0402e268f95

sirepo/pkcli/roles.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2515/commits/af8a2bb2a236d342c64f98613d2da0402e268f95

sirepo/pkcli/roles.py
e-carlin: I added a comment regarding GitHub etc. not working

https://github.com/radiasoft/sirepo/pull/2515/commits/af8a2bb2a236d342c64f98613d2da0402e268f95

e-carlin:
@robnagler all comments addressed

PR #2521, opened 2020-05-15
Fix #2518. Deleted run dirs do not result in an error.

### LINKED ISSUE #2518: missing animation directory breaks simulation
Similar to the earlier report bug which has been fixed - if you run a long-running job and then delete the animation directory, the report shows an error and the simulation can never be successful run again.

To reproduce: in SRW run the Partially Coherent Results simulation. Then delete the simulation's multiElectronAnimation directory and refresh the browser page. The "Report not generated" error is shown in the report panel, and selecting the "Start New Simulation" button results in "server error".

e-carlin:
There are a few fixes:
1. In __db_init we use prev_db to get the jobRunMode.
2. When a run_dir is purged we initiliaze a new supervisor db.
Previously we didn't initialize a new db so the gui thought
there were frames available and requested them resulting in
"Report not generated." Initializing the db solves this by
re-initializing parallelStatus so the frameCount is 0 and no frames
are requested.
3. If a job status is in job_run_purged we need to add it to the
purged_jids_cache. Without this we read the db file for purged jobs
each time purge_free_simulations() runs.

### DIFF
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -298,16 +298,12 @@ class _ComputeJob(PKDict):
             # at anytime so we need to check that they haven't
             if d.lastUpdateTime > _too_old:
                 return
+            cls._purged_jids_cache.add(db_file.purebasename)
             if d.status == job.JOB_RUN_PURGED:
-                cls._purged_jids_cache.add(db_file.purebasename)
                 return
             p = sirepo.simulation_db.simulation_run_dir(d)
             pkio.unchecked_remove(p)
-            _update_db(d)
-            cls._purged_jids_cache.add(db_file.purebasename)
-
-        def _update_db(db):
-            n = cls.__db_init_new(db, db)
+            n = cls.__db_init_new(d, d)
             n.status = job.JOB_RUN_PURGED
             cls.__db_write_file(n)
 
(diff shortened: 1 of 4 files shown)


sirepo/job_supervisor.py
```diff
@@ -299,13 +299,18 @@ def _purge_sim(db_file):
             if d.lastUpdateTime > _too_old:
                 return
             if d.status == job.JOB_RUN_PURGED:
+                cls._purged_jids_cache.add(db_file.purebasename)
```
robnagler, line 302: Move this one line up, and delete line 307.

sirepo/job_supervisor.py
```diff
@@ -299,13 +299,18 @@ def _purge_sim(db_file):
...
                 return
             p = sirepo.simulation_db.simulation_run_dir(d)
             pkio.unchecked_remove(p)
-            d.status = job.JOB_RUN_PURGED
-            cls.__db_write_file(d)
+            _update_db(d)
```
robnagler, line 306: Why a function here? It seems to only be used once.

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2521/commits/bf9c680f975b12ee38bfa0b0e494fb3e3abe90f7

sirepo/job_supervisor.py
e-carlin: I just liked the abstraction, not that it hid much... removed

https://github.com/radiasoft/sirepo/pull/2521/commits/bf9c680f975b12ee38bfa0b0e494fb3e3abe90f7

e-carlin:

PR #2522, opened 2020-05-15
Issue/2509 srw icons

mkeilman:
Upgraded beamline icons.  Notes:

1. Now reading in svg files instead of joining strings
2. I had to make the classes and reference urls unique since they all end up on the same page
3. The viewboxes needed tweaking to make the icons fit.  Still some centering to do in a few cases


moellep:
I added some image adjustments by type to get things on the same scale and centered.

PR #2528, opened 2020-05-20
Fix #2501. Allow NERSC project/repo selection.

### LINKED ISSUE #2501: Allow user to select project/repo
#2278 is affected by this, because if you haven't logged in, you don't know which repo you are to use, and you need to present that list to the user. Even if you have logged, the list could have changed. 

The sbatch agent on login should parse the output of `hpssquota` to get a list of repos:

```txt
cori08:~> hpssquota
usage: hpssquota [-h] [-u USER] [-r REPO]

hpss quota info

optional arguments:
  -h, --help            show this help message and exit
  -u USER, --user USER  user name for HPSS quota info
  -r REPO, --repo REPO  repo name for HPSS quota info
HPSS Usage for User nagler
REPO                     STORED [GB]      USER QUOTA [GB]     PERCENT USED [%]
------------------------------------------------------------------------------
m1253                            0.0             133120.0                  0.0
------------------------------------------------------------------------------
TOTAL USAGE [GB]                 0.0
```

This has to be cached, because it is slow. This can be done before the agent is started, actually. Kind of a simple test that the ssh connection is good.

e-carlin:
Users can now enter a NERSC repo and it will be used for running
the job. The field is optional. If no repo is supplied then one
won't be supplied to sbatch and the default repo for the user
will be used.

I'd like to test this before merging. Cori is down right now but when it is back up and we sort out networking issues with the VM on our server I will.

### DIFF
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -283,7 +283,7 @@ def _validate_and_add_sbatch_fields(request_content, compute_model):
         assert m.sbatchQueue in sirepo.job.NERSC_QUEUES, \
             f'sbatchQueue={m.sbatchQueue} not in NERSC_QUEUES={sirepo.job.NERSC_QUEUES}'
         c.sbatchQueue = m.sbatchQueue
-        c.sbatchRepo = m.sbatchRepo
+        c.sbatchProject = m.sbatchProject
     for f in 'sbatchCores', 'sbatchHours':
         assert m[f] > 0, f'{f}={m[f]} must be greater than 0'
         c[f] = m[f]
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -283,6 +283,7 @@ def _validate_and_add_sbatch_fields(request_content, compute_model):
         assert m.sbatchQueue in sirepo.job.NERSC_QUEUES, \
             f'sbatchQueue={m.sbatchQueue} not in NERSC_QUEUES={sirepo.job.NERSC_QUEUES}'
         c.sbatchQueue = m.sbatchQueue
+        c.sbatchRepo = m.sbatchRepo
     for f in 'sbatchCores', 'sbatchHours':
         assert m[f] > 0, f'{f}={m[f]} must be greater than 0'
         c[f] = m[f]
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -16,6 +16,7 @@ import datetime
 import json
 import os
 import re
+import shutil
 import signal
 import sirepo.auth
 import sirepo.tornado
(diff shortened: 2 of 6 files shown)


sirepo/pkcli/job_agent.py
```diff
@@ -644,19 +645,39 @@ async def _prepare_simulation(self):
                 opName=job.OP_ANALYSIS,
             ),
             send_reply=False,
+            op_id=self.msg.opId,
         )
         await c.start()
         await c._await_exit()
 
     def _sbatch_script(self):
+        def _assert_account():
```
robnagler, line 654: account => repo

sirepo/job_api.py
```diff
@@ -283,6 +283,7 @@ def _validate_and_add_sbatch_fields(request_content, compute_model):
         assert m.sbatchQueue in sirepo.job.NERSC_QUEUES, \
             f'sbatchQueue={m.sbatchQueue} not in NERSC_QUEUES={sirepo.job.NERSC_QUEUES}'
         c.sbatchQueue = m.sbatchQueue
+        c.sbatchRepo = m.sbatchRepo
```
robnagler, line 286: NERSC calls these things projects (canonically) and repos (parenthetically) so let's call them projects. It'll be a bit less confusing in general, because repos mean something specific to us.

sirepo/pkcli/job_agent.py
```diff
@@ -644,19 +645,39 @@ async def _prepare_simulation(self):
                 opName=job.OP_ANALYSIS,
             ),
             send_reply=False,
+            op_id=self.msg.opId,
```
robnagler, line 648: why was this added?

sirepo/pkcli/job_agent.py
```diff
@@ -644,19 +645,39 @@ async def _prepare_simulation(self):
...
         await c._await_exit()
 
     def _sbatch_script(self):
+        def _assert_account():
+            if not self.msg.sbatchRepo:
+                return ''
+            r = subprocess.run(
+                # Info on sed https://stackoverflow.com/a/49430740/5518313
+                # -n == --silent. --silent is not available on BSD
+                "hpssquota | sed -n '/^-\{76\}/,//{//!p;}' | cut -d' ' -f1",
```
robnagler, line 660: Don't use shell pipelines in python for parsing:

* There are no guarantees that the pipeline works so you would need to check that. The less complex a shell command the better. 
* Avoids the use of `shell=True` (in this case)
* Testing is harder. You can mock the output of hpssquota and then write a test for the parser.
* It's not Python so it's harder to understand for Pythonistas viz your comment about how to use sed

Also:

* Goal is to validate the repo, not to get a list of repos and check against it so a regex is a better solution
* Before inserting it into a regex, it needs to be validated for specials. I choose a simple regex, which may not be right, but probably is. NERSC doesn't publish syntax.

```py
o = subprocess.check_output(['hpssquota'], text=True)
p = self.msg.sbatchProject
assert re.search(r'^[-\w]+$', p), \
    'invalid NERSC project={r}'
assert re.search(r'{}\s+\d\.'.format(p), o), \
    'sbatchProject={p} is invalid. hpssquota={}'
```py

sirepo/pkcli/job_agent.py
e-carlin: used project https://github.com/radiasoft/sirepo/pull/2528/commits/ca5b244468a0e5f89041b56fb49a4eac225a8a87

sirepo/job_api.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2528/commits/ca5b244468a0e5f89041b56fb49a4eac225a8a87

sirepo/pkcli/job_agent.py
e-carlin: [pkdebug_str() expects its](https://github.com/radiasoft/sirepo/blob/5295ae9b2762c2537a5d24e8b613f2a9c6e4982c/sirepo/pkcli/job_agent.py#L466)

sirepo/pkcli/job_agent.py
e-carlin: Thanks for the detailed information. For the final regex I think it should be `r'{}\s+\d+\.'` (added a + after \d). This will match `project 123.0`. 
https://github.com/radiasoft/sirepo/pull/2528/commits/ca5b244468a0e5f89041b56fb49a4eac225a8a87

robnagler: Correct

PR #2534, opened 2020-05-21
Fix ops 508. computeModel not in data

e-carlin:


sirepo/job_supervisor.py
```diff
@@ -300,6 +300,11 @@ def _purge_sim(db_file):
             cls._purged_jids_cache.add(db_file.purebasename)
             if d.status == job.JOB_RUN_PURGED:
                 return
+            # compute_model isn't guaranteed to be in data but simulation_db.simulation_run_dir()
+            # expects it
+            d.pksetdefault(
+                computeModel=sirepo.sim_data.split_jid(db_file.purebasename).compute_model
```
e-carlin, line 306: @robnagler maybe computeModel should be a required field and we should create a fixup in __db_load() for it? Also, we should probably use __db_load here so we get the fixups it applies.

robnagler: When would it not be in data?

sirepo/sim_data/__init__.py
```diff
@@ -655,8 +655,13 @@ def _organize_example(cls, data):
             dm.simulation.folder = '/Examples'
 
 
-def uid_from_jid(jid):
-    return jid.split(_JOB_ID_SEP)[0]
+def split_jid(jid):
+    """Reverse of parse_jid. Take a jid and split it into its parts"""
+    s = jid.split(_JOB_ID_SEP)
```
robnagler, line 660: you want to use `zip`:

```py
return PKDict(zip(
    ('uid', 'simulation_id', 'compute_model'), 
    jid.split(_JOB_ID_SEP),
))
```

e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/c462bafae19f5d6ad2d76998d1460fbfccf747b5

sirepo/job_supervisor.py
e-carlin: We don't require it in [__db_init_new()](https://github.com/radiasoft/sirepo/blob/c1f31e8b286fb373dfcf1779c12316a931084ae2/sirepo/job_supervisor.py#L390). It is only put on in [runSimulaiton()](https://github.com/radiasoft/sirepo/blob/c1f31e8b286fb373dfcf1779c12316a931084ae2/sirepo/job_supervisor.py#L689). In addition our upgrade code [doesn't put it on](https://github.com/radiasoft/sirepo/blob/c1f31e8b286fb373dfcf1779c12316a931084ae2/sirepo/pkcli/db.py#L84). I think it should be a required field.

I've fixed this in  https://github.com/radiasoft/sirepo/pull/2534/commits/788682ba9f66212284a3762464e80326e208fc36. Our db upgrade code now includes it. In addition, __db_load() fixes up the data and puts computeModel on. Our purge code now uses __db_load() so it will get all fixups.

sirepo/job_supervisor.py
```diff
@@ -290,14 +290,14 @@ def _get_uids_and_files():
             if r:
                 yield u, r
 
-        def _purge_sim(db_file):
-            d = pkcollections.json_load_any(db_file)
+        def _purge_sim(compute_jid):
+            d = cls.__db_load(compute_jid)
```
robnagler, line 294: Why is this called cmopute_jid, when elsewhere it's just jid.

sirepo/job_supervisor.py
```diff
@@ -275,7 +275,7 @@ def _get_uids_and_files():
             for f in pkio.sorted_glob(_DB_DIR.join('*{}'.format(
                     sirepo.simulation_db.JSON_SUFFIX,
             ))):
-                n = sirepo.sim_data.uid_from_jid(f.purebasename)
+                n = sirepo.sim_data.split_jid(f.purebasename).uid
```
robnagler, line 278: This is an implicit coupling used for optimization. I think it might be better to document: jid=f.purebasename.

sirepo/job_supervisor.py
```diff
@@ -319,7 +319,7 @@ def _purge_sim(db_file):
             for u, v in _get_uids_and_files():
                 with sirepo.auth.set_user(u):
                     for f in v:
-                        _purge_sim(f)
+                        _purge_sim(f.purebasename)
```
robnagler, line 322: jid=f.purebasename

sirepo/sim_data/__init__.py
```diff
@@ -655,8 +655,12 @@ def _organize_example(cls, data):
             dm.simulation.folder = '/Examples'
 
 
-def uid_from_jid(jid):
-    return jid.split(_JOB_ID_SEP)[0]
+def split_jid(jid):
+    """Reverse of parse_jid. Take a jid and split it into its parts"""
```
robnagler, line 659: I struggle with comments for things like this, because it's obvious. It doesn't actually tell me what it returns. "parts" is too generic (tuple, list, dict?).  Also I think "simulation_id" is too long, and we used "sid" elsewhere.

```py
"""Split jid into named parts

Args:
    jid (str): properly formed job identifier
Returns:
    PKDict: parts named uid, sid, compute_model.
"""
```

e-carlin: Makes sense. Thanks.

While we're on this road, why "properly formed"? Isn't that by definition? Is it to imply we don't do any checking and if you put malformed in you will get malformed out?

sirepo/job_supervisor.py
```diff
@@ -388,6 +388,7 @@ def __db_init_new(cls, data, prev_db=None):
             cancelledAfterSecs=None,
             computeJid=data.computeJid,
             computeJobHash=data.computeJobHash,
+            computeModel=data.computeModel,
```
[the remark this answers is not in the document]
e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/9ce89be2fa5fb91607e6a20d55f8bf0cb9a69f45

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/9ce89be2fa5fb91607e6a20d55f8bf0cb9a69f45

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/9ce89be2fa5fb91607e6a20d55f8bf0cb9a69f45

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/9ce89be2fa5fb91607e6a20d55f8bf0cb9a69f45

sirepo/sim_data/__init__.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2534/commits/9ce89be2fa5fb91607e6a20d55f8bf0cb9a69f45

e-carlin:
@robnagler  all comments addressed.

PR #2538, opened 2020-05-23
Fix #2537 protect zip extract in mpi4py

### LINKED ISSUE #2537: Only run setup_magnetic_measurement_files for mpi rank 0
[setup_magnetic_measurement_files](https://github.com/radiasoft/sirepo/blob/6dc7ad602f1d64b10192f2fcbfcdefb9db681bd3/sirepo/package_data/template/srw/parameters.py.jinja#L261) extracts a zip, but this should only be done in rank 0 when MPI is running, e.g.

```py
def setup_magnetic_measurement_files(filename, v):
    import os
    import re

    import zipfile
    z = zipfile.ZipFile(filename)

    #z.extractall()
    rank = 0
    comMPI = None
    try:
        from mpi4py import MPI
        comMPI = MPI.COMM_WORLD
        rank = comMPI.Get_rank()
    except: pass
    if rank == 0: z.extractall()
    if comMPI is not None: comMPI.Barrier()

    for f in z.namelist():
        if re.search(r'\.txt', f):
            v.und_mfs = os.path.basename(f)
            v.und_mdir = os.path.dirname(f) or './'
            return
    raise RuntimeError('missing magnetic measurement index *.txt file')
```

robnagler:
Abort mpi if extract fails
communicate zip file via bcast
If too many files, also raise exception

### DIFF
--- a/sirepo/package_data/template/srw/parameters.py.jinja
+++ b/sirepo/package_data/template/srw/parameters.py.jinja
@@ -260,7 +260,7 @@ def setup_magnetic_measurement_files(filename, v):
     f = None
     r = 0
     try:
-        import mpi4py
+        import mpi4py.MPI
         if mpi4py.MPI.COMM_WORLD.Get_size() > 1:
             c = mpi4py.MPI.COMM_WORLD
             r = c.Get_rank()
--- a/sirepo/package_data/template/srw/parameters.py.jinja
+++ b/sirepo/package_data/template/srw/parameters.py.jinja
@@ -256,16 +256,38 @@ varParam = srwl_bl.srwl_uti_ext_options([
 {% if setupMagneticMeasurementFiles %}
 def setup_magnetic_measurement_files(filename, v):
     import os
-    import re
-    import zipfile
-    z = zipfile.ZipFile(filename)
-    z.extractall()
-    for f in z.namelist():
-        if re.search(r'\.txt', f):
-            v.und_mfs = os.path.basename(f)
-            v.und_mdir = os.path.dirname(f) or './'
-            return
-    raise RuntimeError('missing magnetic measurement index *.txt file')
+    c = None
+    f = None
+    r = 0
+    try:
+        import mpi4py
+        if mpi4py.MPI.COMM_WORLD.Get_size() > 1:
+            c = mpi4py.MPI.COMM_WORLD
+            r = c.Get_rank()
+    except Exception:
+        pass
+    if r == 0:
+        try:
+            import zipfile
+            z = zipfile.ZipFile(filename)
+            f = [x for x in z.namelist() if x.endswith('.txt')]
+            if len(f) != 1:
+                raise RuntimeError(
+                    '{} magnetic measurement index (*.txt) file={}'.format(
+                        'too many' if len(f) > 0 else 'missing',
+                        filename,
+                    )
+                )
+            f = f[0]
+            z.extractall()
+        except Exception:
+            if c:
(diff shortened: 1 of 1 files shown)


moellep:
The generated source only works sometimes, other times I see  "ValueError: could not convert string to float" or "IndexError: list index out of range".

To reproduce, export the Flux Animation source from the NSLS-II ESM beamline (tabulated undulator) example. Running the file sometimes works, but often it fails with the errors above.

```
mpiexec -n 4 python nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py
  File "nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py", line 258, in <module>
    main()
  File "nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py", line 256, in main
    srwl_bl.SRWLBeamline(_name=v.name, _mag_approx=mag).calc_all(v, op)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwl_bl.py", line 2551, in calc_all
    _yp = _v.ebm_yp - _v.und_yp)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwl_bl.py", line 609, in set_und_tab
    curFldCnt = srwl_uti_read_mag_fld_3d(curFilePath, '#')
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwlib.py", line 6481, in srwl_uti_read_mag_fld_3d
    locArBz[i] = float(curLineParts[2])
ValueError: could not convert string to float:
```

robnagler:
I ran lots of times with 4 cores. Didn't fail.  Here's the script:

```sh
set -ex
for i in $(seq 10); do
    rm -rf x
    mkdir x
    cd x
    unzip /vagrant/NSLS-II_ESM_beamline_tabulated_undulator.zip
    mpiexec -n 4 python run.py
    cd ..
done
```

Any other way I can try to reproduce?

moellep:
Try with this file - the exported source from the Flux report. You don't need to unzip the zip file first, just have it in the same directory. 

[nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py.gz](https://github.com/radiasoft/sirepo/files/4684386/nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py.gz)

moellep:
Sorry - that was the wrong example - I'll add the correct one.

moellep:
Attached is the example which reproduces it, exported from the flux report. If the zip files has already been unzipped and you comment out the z.extractall() command, it will run ok.

[nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py.gz](https://github.com/radiasoft/sirepo/files/4684438/nsls-ii-esm-beamline-tabulated-undulator-spectral-flux-27-5m.py.gz)

moellep:
I think the problem is you have to have this: 
```python
    from mpi4py import MPI
    if MPI.COMM_WORLD.Get_rank():
```
instead of this:
```python
    import mpi4py
```

The second case raises an exception which is quietly caught. exception: module 'mpi4py' has no attribute 'MPI'

robnagler:
Thanks. I added that fix. Also extractall [silently overwrites](https://docs.python.org/3/library/zipfile.html#default-behaviors-of-extraction) so we don't have to protect that (no need to comment out).

Rerunning the flux report doesn't cause an error.

PR #2552, opened 2020-06-03
Fix #2544 proprietary_sim_types added to sim_types

### LINKED ISSUE #2544: add values from proprietary_sim_types to sim_types
proprietary_sim_types should be unioned with sim_types so that any proprietary_sim_types is guaranteed in sim_types.  This makes configuration management easier.

default proprietary sim types can include flash for dev, but not alpha, etc.

robnagler:
sim_types is now a frozenset

### DIFF
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -101,9 +101,10 @@ def _init():
     )
     a = ALL_CODES if pkconfig.channel_in_internal_test() else NON_ALPHA_CODES
     s = set(_cfg.sim_types) if _cfg.sim_types else set(a)
-    if _cfg.proprietary_sim_types:
-        s.update(_cfg.proprietary_sim_types)
-    # need more cases like this
+    s.update(_cfg.proprietary_sim_types)
+    # jspec imports elegant, but elegant won't work if it is not a valid
+    # sim_type so need to include here. Need a better model of
+    # dependencies between codes.
     if 'jspec' in s and 'elegant' not in s:
         s.add('elegant')
     x = s.difference(a)
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -77,36 +77,19 @@ def _init():
     from pykern import pkconfig
     global _cfg
 
-    @pkconfig.parse_none
-    def _cfg_sim_types(value):
-        res = pkconfig.parse_set(value)
-        if not res:
-            return frozenset(_codes())
-        for c in res:
-            assert c in _codes(), \
-                'invalid sim_type={}, expected one of={}'.format(c, _codes())
-        if 'jspec' in res:
-            res = set(res)
-            res.add('elegant')
-        return frozenset(res)
-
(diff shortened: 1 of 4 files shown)


robnagler:
Actually hold off. Need to validate types.

robnagler:
@e-carlin ready

sirepo/feature_config.py
```diff
@@ -111,4 +99,15 @@ def _codes():
             display_test_boxes=(pkconfig.channel_in_internal_test(), bool, 'Display test boxes to visualize 3D -> 2D projections'),
         ),
     )
+    a = ALL_CODES if pkconfig.channel_in_internal_test() else NON_ALPHA_CODES
+    s = set(_cfg.sim_types) if _cfg.sim_types else set(a)
+    if _cfg.proprietary_sim_types:
+        s.update(_cfg.proprietary_sim_types)
+    # need more cases like this
+    if 'jspec' in s and 'elegant' not in s:
```
e-carlin, line 107: Why check explicitly for elegant?

sirepo/pkcli/admin.py
```diff
@@ -99,13 +99,12 @@ def setup_dev_proprietary_code(sim_type, rpm_url):
       sim_type (str): simulation type
       rpm_url (str): Url of the rpm (file:// or http://)
     """
-    import sirepo.pkcli.admin
     import urllib.request
 
     assert pkconfig.channel_in('dev'), \
         'Only to be used in dev. channel={}'.format(pkconfig.cfg.channel)
 
-    d = sirepo.pkcli.admin.proprietary_code_dir(sim_type)
```
e-carlin, line 108: Geez, who wrote that code!

sirepo/feature_config.py
```diff
@@ -111,4 +99,15 @@ def _codes():
             display_test_boxes=(pkconfig.channel_in_internal_test(), bool, 'Display test boxes to visualize 3D -> 2D projections'),
         ),
     )
+    a = ALL_CODES if pkconfig.channel_in_internal_test() else NON_ALPHA_CODES
+    s = set(_cfg.sim_types) if _cfg.sim_types else set(a)
+    if _cfg.proprietary_sim_types:
```
e-carlin, line 104: Why check?

sirepo/feature_config.py
robnagler: New comment:

```py
    # jspec imports elegant, but elegant won't work if it is not a valid
    # sim_type so need to include here. Need a better model of
    # dependencies between codes.
```

sirepo/feature_config.py
robnagler: removed. old code was frozenset so was complicated to udpate.

sirepo/pkcli/admin.py
robnagler: As Buffett said "I've made so many mistakes, if I tried to kick myself, my legs would be exhausted."

Issue #2542, opened 2020-05-26
Radia symmetries not applied


mkeilman:
No symmetries on objects unless they are in a container

PR #2559, opened 2020-06-05
Mpmd updates

mkeilman:
Various and sundry:

- Radia solution stored with model
- Corrected field map sdds columns
- User can select beam axis, which is drawn in svg
- Workaround for symmetry issue #2542


sirepo/template/radia.py
```diff
@@ -394,13 +420,18 @@ def _get_sdds():
...
 
 
-def _read_data(sim_id, view_type, field_type):
+def _read_path(sim_id, h5path):
     try:
         with h5py.File(_geom_file(sim_id), 'r') as hf:
-            g = template_common.h5_to_dict(hf, path=_geom_h5_path(view_type, field_type))
-        return g
+            return template_common.h5_to_dict(hf, path=h5path)
     except IOError:
```
robnagler, line 427: This is too coarse of a test. If the file is "not found", then we should know that. There's a call for this in pkio, which you should use so it is compatible with py.path and future not founds.

sirepo/template/radia.py
```diff
@@ -394,13 +420,18 @@ def _get_sdds():
...
-        return {}
+        return PKDict()
```
robnagler, line 428: I think it shouldn't have a solution if there is no solution. Return None, not an empty dict, but see below.

sirepo/template/radia.py
```diff
@@ -394,13 +420,18 @@ def _get_sdds():
...
+
+
+def _read_data(sim_id, view_type, field_type):
+    res = _read_path(sim_id, _geom_h5_path(view_type, field_type))
+    res.solution = _read_solution(sim_id)
```
robnagler, line 433: If read path fails, you shouldn't try to read the solution, and there shouldn't be a solution attribute. This is where errors need to be documented by if statements.

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
         return get_application_data(data)
 
 
-def _save_fm_sdds(name, f_data, file_path):
+def _read_solution(sim_id):
+    try:
+        return _read_path(sim_id, 'solution')
+    except KeyError:
```
robnagler, line 455: _read_path should tell you the result. KeyError could be raised for other reasons (ie. other defects) and therefore it's too coarse. Better to return an explicit "no solution" (None) result and test for that here.

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+        # not solved yet
+        return []
```
robnagler, line 457: There is no solution so return None.

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+
+
+def _rot_flat_list(l, r):
```
robnagler, line 460: `rot` doesn't mean anything to me. `l` and `r` don't mean anything. function args should be long enough to be readable. Internal variables to a function can be single letter.

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+    return r.apply(numpy.reshape(l, (-1, 3)))
```
robnagler, line 461: -1 and 3 don't mean anything to me. Is -1 mean "inverse"? Does 3 mean three dimensions. Why not get those from `l`?

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+
+
+def _save_field_csv(f_type, vectors, rot, file_path):
+    #pkdp('SAVE TYPE {} V {} PATH {}', f_type, vectors, file_path)
+    # rotate so the beam axis is aligned with z
+    data = ['x,y,z,' + f_type + 'x,' + f_type + 'y,' + f_type + 'z']
+    # mm -> m
+    pts = 0.001 * _rot_flat_list(vectors.vertices, rot).flatten()
```
robnagler, line 469: f_type is field type? file_path could be shortened to path. `rot` is not clear to me.  If _rot_flat_list is supposed to flatten why is `flatten` called?

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+        # not solved yet
+        return []
+
+
+def _rot_flat_list(l, r):
+    return r.apply(numpy.reshape(l, (-1, 3)))
+
+
+def _save_field_csv(f_type, vectors, rot, file_path):
+    #pkdp('SAVE TYPE {} V {} PATH {}', f_type, vectors, file_path)
```
robnagler, line 465: `pkdc` can be used for this purpose.

sirepo/template/radia.py
```diff
@@ -418,15 +449,45 @@ def _read_or_generate(geom_id, data):
...
+    # rotate so the beam axis is aligned with z
+    data = ['x,y,z,' + f_type + 'x,' + f_type + 'y,' + f_type + 'z']
```
robnagler, line 467: seems odd to me. Is there a way to document what this is? Header?

sirepo/template/radia.py
mkeilman: -1 means "numpy will figure this dimension out automatically".  "3" does mean 3 dimensions since these are vectors.  I will rename to ```_rotate_flat_vector_list(vectors, rotation)```  which should provide context

robnagler: "rotation" is a matrix? If you call it "vectors", then "rotation" might better be "matrix", because we already know that we are rotating by the name of the function. You could also call it rotation_matrix.

mkeilman: rotation is a scipy Rotation object, more general than a matrix

robnagler: then "scipy_rotation" or "numpy_rotation". Unfortunately, numpy's API doesn't define it as an object or even give it a useful name. We have to help the reader to know what the thing is.

sirepo/template/radia.py
mkeilman: ```_rotate_flat_vector_list()``` (proposed new name) does not flatten, it accepts a flat list.  We need the resultant list un-flattened to craft the sdds columns, but for csv it needs to be flat again.

I'll change ```f_type``` to ```field_type``` where it is a passed parameter

Issue #1275, opened 2018-07-19
Encapsulate Flask better


robnagler:
#1245 is an example of this. This Issue should collect others and perhaps as "sub" issues.  As I noted in #1274 we probably want to drop Flask. It's just not the right basis for an application server in Python.

Things like testing are tricky, of course, but I think part of the problem is that we don't encapsulate enough of Flask so that the tests don't have to go through Flask. If we pulled all of Flask out of server.py, it would be a good thing...

robnagler:
This is still important. I particularly don't like how flask assumes global variables, which are really "thread locals". I think we need to use instances for our apis whose instances are then used to handle requests. This would allow passing in variables via `self`.

robnagler:
Don't need this specific issue. We are refactoring as we go, and eventually will replace flask with tornado.

robnagler:
Easier to have this issue open to collect flash encapsulations.

PR #2586, opened 2020-06-11
flash and roles

### LINKED ISSUE #2513: Replace sirepo.server._app.sirepo_db_dir with srdb.root
I think we should use srdb.root everywhere and remove some unnecessary coupling.

### LINKED ISSUE #2561: propietary codes can't be in all_codes
If you don't set proprietary_codes and a proprietary codes is in ALL_CODES, then the proprietary code system is bypassed. Deny by default. Change ALL_CODES to VALID_CODES which are used to validate names, but not used to initialize sim_types (when no value). Was getting to flash on alpha as an unprivileged user, but start simulation was failing with `Error: Simulation library file "flash.rpm" does not exist`, which was implicitly deny by default.

Fix b69ddbdc138da265f67e80002668ef23370e4a8e

### LINKED ISSUE #2576: dev: unknown uid check at start
I had removed the run dir, and was trying to access flash, but it was failing for forbidden. The problem was that the uid was stale. I tried to add the role, and it failed because no such user. It was not until I accessed another FOSS code that it got `user_dir_not_found`, and blew away the user.

When we are checking a role, we should do a validation of the user if we get forbidden.

### LINKED ISSUE #2577: dev: pull flash.rpm when creating run dir
When the run dir is created, we need to populate it with the flash.rpm. Otherwise, the user will not get the symlink to the flash.rpm, and `sirepo admin audit-proprietary-lib-files` must be run manually on the uid. There needs to be a config param where to pull the flash.rpm (could be local).

### LINKED ISSUE #2579: service.http: sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) table user_role_t already exists
There's a race condition creating auth db.

### LINKED ISSUE #2583: auth._create_roles_for_user needs to audit
It calls auth_db directly, which breaks the abstraction.

robnagler:
- moved role auditor into auth_db so better encapsulated fix #2583
- removed `app` from init_apis and added `util.flask_app` to get app #1275
- refactored srdb root and added proprietary_code_dir there so can be used in auth_db #2583 
- added uid to simulation_dir and simlation_lib_dir to allow auth_db to get directory for any user #2583 
- service.flask calls pkcli.setup_dev which pulls proprietary RPMs from rsconf (default) or any url fix #2577
- added sleep between supervisor and flask startup to fix db create race condition fix #2579
- VALID_CODES includes all possible codes. Removed "all codes" which is now sim_types fix #2561
- Removed sirepo_db_dir and replaced with srdb.root fix #2513
- check_user_has_role uses logged_in_user and fixed a redirect issue in user_not_found fix #2576


sirepo/auth_db.py
```diff
@@ -131,27 +130,77 @@ def delete_roles(cls, uid, roles):
...
+        r = UserRole.has_role(
+            uid,
+            sirepo.auth.role_for_sim_type(t),
+        )
+        for f in pykern.pkio.sorted_glob(
+            sirepo.srdb.proprietary_code_dir(t).join('*'),
+        ):
+#TODO(robnagler) ensure no collision on names with uploaded files
+# (restrict suffixes in user uploads)
+            p = sirepo.simulation_db.simulation_lib_dir(t, uid=uid).join(f.basename)
```
e-carlin, line 183: Why not use sirepo.auth.set_user() instead of passing in the uid?

sirepo/pkcli/roles.py
```diff
@@ -15,41 +15,45 @@
 import sirepo.server
 
 
-def add_roles(uid_or_email, *args):
+def add_roles(uid_or_email, *roles):
     """Assign roles to a user
     Args:
         uid_or_email (str): Uid or email of the user
         *args: The roles to assign to the user
```
e-carlin, line 22: *roles

sirepo/pkcli/roles.py
```diff
@@ -15,41 +15,45 @@
...
+    return sirepo.auth_db.UserRole.search_all_for_column(
+        'role',
+        uid=_args(uid_or_email, []),
+    )
 
 
 # TODO(e-carlin): This only works for email auth or using a uid
 # doesn't work for other auth methods (ex GitHub)
-def _check_uid_or_email_and_roles(uid_or_email, roles):
+def _args(uid_or_email, roles):
```
e-carlin, line 56: check_args? I see that saying uid_or_email_and_roles is repetitive because those are the arg names. But _args seems too far to the terse side.

sirepo/pkcli/service.py
```diff
@@ -45,10 +46,13 @@ def cfg():
...
     from sirepo import server
+    import sirepo.pkcli.setup_dev
 
     with pkio.save_chdir(_run_dir()):
-        use_reloader = pkconfig.channel_in('dev')
-        app = server.init(use_reloader=use_reloader)
+        sirepo.pkcli.setup_dev.default_command()
+        # above will throw better assertion, but just in case
+        assert pkconfig.channel_in('dev')
+        app = server.init(use_reloader=True)
```
e-carlin, line 55: `u=True` and then use u here and on line 64.

sirepo/auth_db.py
robnagler: #2588 discusses this in detail. set_user is global state, which is bad, because this is local context: simulation_lib_dir for uid. Also set_user requires that flask context be configured, because it sets the cookie (or set_cookie_for_utils) which is too much coupling here. We aren't logging in the user here; we are performing a database operation in an unspecified context.

sirepo/pkcli/roles.py
robnagler: I changed it to _parse_args. It's a private function that's used in all the other functions in the module. Therefore, it's critical to understand how it works, and the name is less use, since it does some critical functions: initializing, parsing, and validating. We could call it _preamble or _setup, but those are as unclear.

sirepo/pkcli/service.py
robnagler: In this case, I don't think the abstraction adds value. It's a symbolic substitution for True, but the word True speaks for itself, and there are only two lines copied in the same module. I think this is where Go's "a little copying" applies. If it were in two functions, the explicit coupling would be valuable, but in this case, I think it's better to copy.

Issue #2590, opened 2020-06-15
FLASH: add species selection panel


ncook882:
Within the Source tab of the CapLaser simulation (and future iterations on this configuration), users should be able to select the fill species and wall species from a list of available species. 

The initial fill species options should include Hydrogen gas (H2), Helium gas, and Argon gas. The initial wall species options should be limited to Alumina (Al2O3).  Opacity and EOS files will need to be templated (`op_wallFileName`, `op_fillFileName`, `eos_wallTableFile`, etc.). The fillA and fillZ values should be automatically updated based upon the selection. I will provide tabulated EOS files and corresponding A/Z values for each.

Users should also be permitted to view and change the relevant initial conditions for the fill and wall species. In particular, temperature, and density should be adjustable. For now, we should fix the electron, ion, and radiation temperatures be equal (e.g. `sim_teleWall` = `sim_tionWall`, etc.)

PR #2593, opened 2020-06-15
Fix 2581. Fix CapLaser by using new IONMIX table files.

e-carlin:


e-carlin:
This is obsolete. Work for #2590 will subsume this.

PR #2594, opened 2020-06-15
Fix #2592 auth_db init at module level

### LINKED ISSUE #2592: Remove call to auth_db.init()
auth_db can be initialized on import instead.

robnagler:
Removed direct calls to auth_db.init
made init() private and call
reduced race condition time in _migrate_db_file()
We can mitigate race condition by starting sirepo
without starting job_supervisor. This would be normal
for older installations.


robnagler:
@e-carlin did you see this?

e-carlin:
> @e-carlin did you see this?

I missed it. Thx for the bump. LGTM.

PR #2609, opened 2020-06-19
Fix #2604. Check for an error response from the supervisor.

### LINKED ISSUE #2604: max message size limit should not cause Not Found error
If a requested file exceeds the max message size, it should cause a server error or a 413 too large, not a "Not Found".

To reproduce an example, run on dev:

SIREPO_JOB_MAX_MESSAGE_SIZE=1048576 sirepo service http

Then open the Compact Storage Ring example in elegant, set the particles to 17500 and then download "Raw Data File" on the first phase space report.

e-carlin:


sirepo/job_api.py
```diff
@@ -63,14 +63,15 @@ def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=No
...
-            _request(
+            r = _request(
                 computeJobHash='unused',
                 dataFileKey=t.basename,
                 frame=int(frame),
                 isParallel=False,
                 req_data=req.req_data,
                 suffix=s,
             )
+            assert not r.state == 'error', f'r={r}'
```
robnagler, line 74: change 'r='  to 'error state in request='. It's the only thing on the assertion line so it's too terse.

e-carlin: https://github.com/radiasoft/sirepo/pull/2609/commits/412a07788212fb1968337d210073456d1ef9152f

PR #2611, opened 2020-06-23
Issue/2598 radia encapsulation

### LINKED ISSUE #2598: Radia geometry directory breaks encapsulation
Similar to #2065 but more problematic. There seems to be a directory "geometry" that holds results that is independent of run_dir. This won't work on NERSC, and Radia is supposed to work on NERSC.

We need to keep to the concept of run_dirs and not invent new directories. We shouldn't see code like this:

```py
def _get_res_file(sim_id, filename):
    return simulation_db.simulation_dir(SIM_TYPE, sim_id) \
        .join(_GEOM_DIR).join(filename)
```

_GEOM_DIR is not a computeJob run_dir. It's problematic to refer to sim_id within code templates as well. The run_dir is an abstraction that has nothing to do with simulation_dir on the server. They are in fact completely different roots on NERSC and the server.

The exception in #2597 was caused by this improper use of simulation_dir and simulation_ids.

What we can have is a single computeJob that stores the state. Then, if some reports need the state, they will run off the same computeJob run_dir. This is what elegant does, for example.

mkeilman:
Fix #2598

To preface, this solution uses```get_application_data()``` to fetch or compute results from a stored Radia dump file.  However, that function is unaware of the run_dir, so I provide it in the input data once it is known.  sim_id has been expunged.

1) All compute models fold into 'geometry'
2) A new model ```radiaReq``` persists the input to ```get_application_data()```
3) In ```get_application_data()```, if run_dir is not known or the desired file does not exist, the client calls ```panelState.requestData()``` and ```pkcli.run()``` executes generated python.
4) ```template.extract_report_data()``` knows about run_dir, and passes it to ```get_application_data()``` to return the requested data, plus radiaReq, to the client
5) The client saves the radiaReq model so the run_dir is provided in future requests

After all that, some jiggery-pokery was still required on the client side to re-render properly after a solve: 'computeJobHash-mismatch' is happening so there is something incorrect even though it "works"...

### DIFF
--- a/sirepo/package_data/static/html/radia-visualization.html
+++ b/sirepo/package_data/static/html/radia-visualization.html
@@ -1,7 +1,7 @@
 <div class="container-fluid">
   <div class="row">
       <div class="sr-plot">
-        <div data-radia-viewer="" data-model-name="magnetDisplay"></div>
+        <div data-radia-viewer="" data-model-name="magnetDisplay" data-viz="viz"></div>
       </div>
       <div data-radia-field-paths="" data-model-name="fieldPaths"></div>
       <div data-field-integral-table="" data-model-name="fieldPaths"></div>
--- a/sirepo/package_data/static/js/radia.js
+++ b/sirepo/package_data/static/js/radia.js
@@ -128,6 +128,14 @@ SIREPO.app.factory('radiaService', function(appState, fileUpload, panelState, re
         return (appState.models.fieldTypes || {}).path;
     };
 
+    self.getRadiaData = function(inData, handler, regenOnFail) {
+        // store inData on the model, so that we can refer to it if
+        // getApplicationData fails
+        appState.models.radiaReq = inData;
+        appState.saveQuietly('radiaReq');
+        requestSender.getApplicationData(inData, handler);
+    };
+
     self.getSelectedObj = function() {
         return self.selectedObj;
     };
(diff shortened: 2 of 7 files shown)


moellep:
Not necessarily related to this pull request, but here are a few things I notice when using the app on this branch:

- Open the Dipole example and select "Discard Changes to Example"
- Select "New Path" and type in a name "test". For some reason the Magnet Viewer disappears. 
- Enter a New Path of x: -10 to 10, y: 15 to 15, z: 0 to 0 and select OK
- The server does some computation and the app then shows a javascript error "property 'length' of undefined"
- Reloading the webpage and the new Field Path is shown.
- The Field Integrals shows the line, but the field arrays are empty. Selecting the Download icon on the Field Integrals causes a javascript error, "Cannot read property of B".
- Selecting the Solve button on the Solver panel runs something on the server, but then nothing appears to update in the UI.

Is the Field Integrals and Solver panels dependent on the Field Paths? If so, it might make sense to hide the Field Integrals and Solver panels until a Field Path is present. Also, for the Dipole example, it would be good for the example to include a valid field path for the problem.

The Dipole and Undulator and Wiggler examples don't show any color on the 3D view, only black facets.

robnagler:
I'm not going to merge this due to the rundir issue. Something like #2624 needs to happen first.

PR #2613, opened 2020-06-23
Fix radiasoft/sirepoflash#4. Discharge configuration in source tab.

e-carlin:
**Do not** merge this PR. Lmk when it is approved and I will merge it. It has changes that are going to conflict with the other open flash PRs and I want to make sure they all get sorted out properly. Ideally we would merge the other flash PRs before this one.

This PR covers more than just the changes requested in the issue.
While working on the issue I ran into numerous problems with the ":"
in the model names (ex jquery selector didn't like it,
werkzeug.secure_filename stripped them). After discussion with @robnagler 
I tried using ^ instead of : but the problems persisted. Ultimately,
it was best just to use no special characters. This make some of the
model names more challenging to read but simplifies everything else.


moellep:
Looks good. Removing the ":" from the model names makes sense.

PR #2634, opened 2020-07-01
Fix #2619. Add a madx matching example.

### LINKED ISSUE #2619: Madx matching
Create a madx example that handles matching. Use [this matching notebook](https://github.com/radiasoft/rscon/blob/master/webcon/madx_examples/madx_matching_example01.ipynb) as an example

e-carlin:
Refactored the visualization tab to create a report for each
output file that is produced (similar to elegant and opal).

@mkeilman can you take a look at the source tab and make sure it still looks ok to you? It looked fine to me but I refactored/deleted a fair bit of code and I want to make sure I didn't mess it up.


mkeilman:
Looks fine to me!

PR #2635, opened 2020-07-02
Fix #2612. Validate model names contain no special characters.

### LINKED ISSUE #2612: validate model names
Model names must conform to [^A-Za-z0-9_-]. We need a centralized location where we can validate that the schema for a code complies with this. We have validate_models() but not all codes call this.

e-carlin:


sirepo/srschema.py
```diff
@@ -200,6 +202,16 @@ def _validate_enum(val, sch_field_info, sch_enums):
...
         raise AssertionError(util.err(sch_enums, 'enum {} value {} not in schema', type, val))
 
+def _validate_model_name(model_name):
+    """Ensure model name contain no special characters
+
+    Args:
+        model_name (str): name to validate
+    """
+
+    m = re.findall(r'[^\w]', model_name)
```
robnagler, line 212: `\W` can be used in place of `[^\w]`. 

I think there is little value in printed the invalid characters if you are printing the model name. I think instead the error string should say "model name must contain word characters only".  Therefore, you don't the findall, just a search.

Question: do we want to limit model names to `^[a-z]\W*$`. If so, you could say "must be a Python identifier" in error.

e-carlin: Ah I didn't know about the uppercase inverse in regex. Thanks.

Yes I think a python identifier is better. That way PKDict lookup using dot notation will always work and we still solve the other problems.

I'm not sure about `^[a-z]\W*$`. For example we have [_APERTURE_ELEMENT](https://github.com/radiasoft/sirepo/blob/71f3056a6acaf6519761aef9fdae7de7bc238f7b/sirepo/package_data/static/json/madx-schema.json#L515) which doesn't pass that regex. How about `^[^\d\W]\w*$`?

robnagler: I think positive is better `^[a-z_]\w*$` with ignore case on the regex.

e-carlin: Done https://github.com/radiasoft/sirepo/commit/a846d89d5ca063e9944da7578e4adcba94194a7a

PR #2640, opened 2020-07-02
Fix #2508. Use max_secs instead of max_hours.

### LINKED ISSUE #2508: Switch max_hours to max_secs
max_hours is inconvenient. For example, 10 minutes is repeating decimal. Rather say 10:0 with pkconfig.parse_secs.

e-carlin:


sirepo/job_supervisor.py
```diff
@@ -145,14 +145,14 @@ def init():
     _DB_DIR = sirepo.srdb.root().join(_DB_SUBDIR)
     cfg = pkconfig.init(
         job_cache_secs=(300, int, 'when to re-read job state from disk'),
-        max_hours=dict(
-            analysis=(.04, float, 'maximum run-time for analysis job',),
```
robnagler, line 149: does the change from float affect anything? I don't think so, just asking.

e-carlin: Not really but it is a breaking change if anyone was using floats in their config.

robnagler: The question is: does changing from float to int (which was committed) break anything? I assume you hit all cases so probably not.

PR #2656, opened 2020-07-10
Fix radiasoft/sirepoflash#5. Create CapLaser3D example.

e-carlin:
This requires at least 2 cores (SIREPO_MPI_CORES=2) to run.


sirepo/sim_data/flash.py
```diff
@@ -80,14 +86,14 @@ def _lib_file_basenames(cls, data):
         t = data.models.simulation.flashType
         if t == 'RTFlame':
             return ['helm_table.dat']
-        if t == 'CapLaserBELLA':
+        if 'CapLaser' in t:
```
robnagler, line 89: I don't like this idiom. It should be a function, which can use this idiom once. If we rename "CapLaser" to "FooBar", we have to change all references that look like this without changing any other references that look like, say, SimulationCapLaserBella, which might not be changed.

e-carlin: Thanks. Resolved https://github.com/radiasoft/sirepo/commit/e9ae6dd7e3e310cfb6be1fbb10246a33ffeb9410

PR #2661, opened 2020-07-14
Fix #2650 only cancel ops already in queue

### LINKED ISSUE #2650: Server Error- Sirepo Beamline report
I'm seeing a server error on beamline tab in alpha: https://alpha.sirepo.com/srw#/beamline/j0lipE6K

Screen Shot 2020-07-08 at 11 01 11 AM

robnagler:
runCancel takes snapshot of queue before doing anything,
and then intersects that with the list of current ops
at the time it starts to cancel. This avoids canceling
ops "in the future".
In myapp, setting favoriteTreat to agent_start_delay=N where
N is a number of seconds will cause agent to sleep for N seconds
before sending OP_LIVE.

### DIFF
--- a/sirepo/tornado.py
+++ b/sirepo/tornado.py
@@ -5,7 +5,7 @@ u"""Wrappers for Tornado
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from pykern.pkdebug import pkdlog, pkdexc
-import asyncio
+import sirepo.util
 import tornado.queues
 
 
@@ -24,7 +24,7 @@ class Queue(tornado.queues.Queue):
             # on the queue and before this task finishes the await.
             x = super().get()
             return await x
-        except (asyncio.CancelledError, concurrent.futures._base.CancelledError):
+        except sirepo.util.ASYNC_CANCELLED_ERROR:
             if x:
                 try:
                     r = x.result()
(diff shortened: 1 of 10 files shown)


sirepo/tornado.py
```diff
@@ -24,7 +24,7 @@ async def get(self):
             # on the queue and before this task finishes the await.
             x = super().get()
             return await x
-        except asyncio.CancelledError:
+        except (asyncio.CancelledError, concurrent.futures._base.CancelledError):
```
e-carlin, line 27: It is unclear to me when we need to handle both of these types of cancelledErrors. It looks like CPython [wraps all concurrent.futures._base.CancelledError's in an asyncio.CancelledError](https://github.com/python/cpython/blob/568fb0ff4aa641329261cdde20795b0aa9278175/Lib/asyncio/tasks.py#L245) for tasks. Tornado doesn't seem to provide the same guarantees. So maybe anytime we are interacting with a Tornado primitive that could be cancelled we need to catch both? Since we are using tornado throughout I think that means everywhere. Agree?

e-carlin:
@robnagler added handling of futures.CancelledError https://github.com/radiasoft/sirepo/pull/2661/commits/856df36d921eab3c90799ca6388177be8fddab92.

PR #2663, opened 2020-07-15
Fix radiasoft/sirepflash#10. Remove unsupported geometries for CapLaserBELLA

e-carlin:


sirepo/package_data/static/js/flash.js
```diff
@@ -9,8 +9,24 @@ SIREPO.app.config(function() {
...
+    self.capLaserBELLAGridgeometyEnum = [];
+
+    function initGridgeometry() {
+        if (self.originalGridgeometryEnum.length) {
+            return;
+        }
+        var e = SIREPO.APP_SCHEMA.enum.Gridgeometry;
+        self.originalGridgeometryEnum = e;
+        e.forEach(function(v) {
+            if (self.isFlashType('CapLaserBELLA') && ['polar', 'spherical'].includes(v[0])) {
```
moellep, line 24: We've been using Array.indexOf() rather than Array.includes() because MSIE doesn't support that method.

sirepo/package_data/static/js/flash.js
```diff
@@ -9,8 +9,24 @@ SIREPO.app.config(function() {
...
+    self.originalGridgeometryEnum = [];
+    self.capLaserBELLAGridgeometyEnum = [];
+
+    function initGridgeometry() {
+        if (self.originalGridgeometryEnum.length) {
+            return;
+        }
+        var e = SIREPO.APP_SCHEMA.enum.Gridgeometry;
+        self.originalGridgeometryEnum = e;
+        e.forEach(function(v) {
```
moellep, line 23: Rather than alter the schema, it is cleaner to use PanelState.showEnum()

moellep: I'll make this change and merge with the pull request along with a few other changes

PR #2666, opened 2020-07-16
Fix radiasoft/sirepoflash#12. Update existing CapLaserBELLA simulations.

e-carlin:
In f4c83274144d1de2b8868f3a33d2e4bf952acb46 CapLaser was renamed to
CapLaserBELLA. A fixup was not created to change the flashType.
This creates the fixup.

### DIFF
--- a/sirepo/sim_data/flash.py
+++ b/sirepo/sim_data/flash.py
@@ -7,6 +7,7 @@ u"""simulation data operations
 from __future__ import absolute_import, division, print_function
 from pykern import pkio
 from pykern.pkdebug import pkdp
+import re
 import sirepo.sim_data
 import sirepo.util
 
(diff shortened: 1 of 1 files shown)


sirepo/sim_data/flash.py
```diff
@@ -22,10 +23,13 @@ def fixup_old_data(cls, data):
         dm = data.models
         for m in list(dm.keys()):
             n = m.replace(':', '').replace('magnetoHD', '')
+            n = re.sub('CapLaser$', 'CapLaserBELLA', n)
```
robnagler, line 26: Something is wrong here. The m != n test should be before this line and/or needs to be done for this line as well.

e-carlin: Can you give an example of what problem you see? I'm not seeing it.

robnagler: OK, I see the new case, which is confusing so perhaps this would be clearer:

```txt
n = re.sub(
    'CapLaser$', 
    'CapLaserBELLA', 
    m.replace(':', '').replace('magnetoHD', ''),
)
```

Seeing the difference between `m` and `n` on line 26 was confusing to me.  Perhaps you could iterate the transforms (they could all be regexes):

```py
n = m
for x in (':', ''), ('magnetoHD', ...)...:
    n = re.sub(x[0], x[1], n)
```

e-carlin: Done https://github.com/radiasoft/sirepo/commit/7abf69f208caf462152e0803f71d0d6c7996fb35

PR #2675, opened 2020-07-20
Fix #2673. Make nginx root /usr/share/nginx/html

### LINKED ISSUE #2673: Nginx server root in /usr/share/nginx/html/
See https://github.com/radiasoft/sirepo/commit/1c0f50d78f6780b04865392a6e5a7247041fa952#r40753198

e-carlin:


sirepo/package_data/nginx_proxy.conf.jinja
```diff
@@ -29,8 +29,7 @@ http {
 
     server {
         listen {{ ip }}:{{ nginx_proxy_port }};
-        # We know this is empty in the nginx container
```
robnagler, line 32: This was a good catch btw.

PR #2676, opened 2020-07-20
Fix radiasoft/sirepoflash#14. Fix radiasoft/sirepoflash#13. Grid evolution plot columns

e-carlin:
Allow the user to select what columns are plotted on the Grid Evolution
plot. Dynamically update the column options based on the output.


sirepo/package_data/static/js/flash.js
```diff
@@ -131,6 +132,17 @@ SIREPO.app.controller('VisualizationController', function (appState, flashServic
...
         }
+        if (! self.gridEvolutionColumnsSet && data.gridEvolutionColumns) {
+            self.gridEvolutionColumnsSet = true;
+            SIREPO.APP_SCHEMA.enum.GridEvolutionColumn = [];
+            for (var i = 0; i < data.gridEvolutionColumns.length; i++) {
+                var e = data.gridEvolutionColumns[i];
+                if (e[0] === '#') {
+                    continue;
+                }
+                SIREPO.APP_SCHEMA.enum.GridEvolutionColumn.push([e, e]);
```
robnagler, line 143: I think we had problems when we did this for job stuff. I think this probably needs to be on the server.

sirepo/package_data/static/js/flash.js
```diff
@@ -122,6 +122,7 @@ SIREPO.app.controller('VisualizationController', function (appState, flashServic
     var self = this;
     self.flashService = flashService;
     self.plotClass = 'col-md-6 col-xl-4';
+    self.gridEvolutionColumnsSet = false;
```
robnagler, line 125: this isn't in the same context as SIREPO.APP_SCHEMA.enum.GridEvolutionColumn. I think they should either both be on APP_SCHEMA or not.

sirepo/template/flash.py
```diff
@@ -683,6 +666,14 @@ def _generate_parameters_file(data):
     return res
 
 
+def _grid_evolution_columns(run_dir):
+    try:
+        with open(run_dir.join(_GRID_EVOLUTION_FILE)) as f:
```
robnagler, line 671: we shouldn't have naked opens. i think this has to be pkcompat

sirepo/package_data/static/js/madx.js
```diff
@@ -5,6 +5,7 @@ var srdbg = SIREPO.srdbg;
 
 SIREPO.app.config(function() {
     SIREPO.PLOTTING_COLOR_MAP = 'afmhot';
+    SIREPO.SINGLE_FRAME_ANIMATION = ['twissAnimation'];
```
robnagler, line 8: should this be in the schema or on the server?

sirepo/package_data/static/js/flash.js
e-carlin: I spoke with @moellep about this implementation. Maybe we misunderstood eachother. We do this sort of enum manipulation elsewhere (ex [in srw](https://github.com/radiasoft/sirepo/blob/9d7d346753a90486f6a77dd0e1c7c5c44b17c3a0/sirepo/package_data/static/js/srw.js#L85))

Can you elaborate a bit on what the on the server implementation would entail?

sirepo/template/flash.py
e-carlin: We have naked opens all over the template code. Is there something different about this one or are they all wrong?

In pkcompat what would the open function do? Require that we always specify an encoding so Python doesn't default?

sirepo/package_data/static/js/madx.js
e-carlin: The schema seems fine. It isn't dynamic. @moellep thoughts?

sirepo/package_data/static/js/flash.js
robnagler: Here's a commit where we had problems with jobRunMode changing: https://github.com/radiasoft/sirepo/commit/3484285758dc924b2781f721c78d38bca1a95f17

We eventually moved the dynamicity out of the schema for this reason.

Maybe this can't happen with FLASH or SRW.

sirepo/template/flash.py
robnagler: Actually, we added pkio.open_text for this reason. The only places where there are naked "text" opens is where I haven't gotten to fixing them yet, e.g. ml.py. Most of the opens are binary. The proper form is e.g. from srw.py:

```txt
with pkio.open_text(filename) as f:
```

I would like avoid proliferating decoding of utf-8 everywhere.

sirepo/package_data/static/js/flash.js
e-carlin: All of that code is still in the code base. What exactly did we move out (I can probably answer that question and will look more into it tomorrow)? My recollection of jobRunMode was that we had it hardcoded in the schema. That didn't work because the config could change and add/remove run modes. So we had to have the server dynamically send the available jobRunModes (on the authState). I think flash is doing the exact same thing (albeit at a different point in the lifecycle, status instead of authState, maybe that is the problem?).

Here is how I see it:
- jobRunMode: gets the list of [runModes from the server](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/auth/__init__.py#L542) and populates them [on the enum](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/package_data/static/js/sirepo.js#L144). 
- Flash: get the list of [column values from the server](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/template/flash.py#L38) and populates them [on the enum](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/package_data/static/js/flash.js#L143)

One problem I do see is in Flash is that jobRunMode has code for handling when the jobRunMode set by the user isn't known by the [gui or the server](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/package_data/static/js/sirepo-components.js#L3325) or [just the server](https://github.com/radiasoft/sirepo/blob/09e28d2286e7cf0e7f608bd287eaedd8ebf0d589/sirepo/job_supervisor.py#L755). Flash doesn't have any of this code. So if the user has a column selected that is no longer available then the report generation code will fail and they will get back "report not generated." In that case should we choose a default value for them, send back a more helpful message, or keep report not generated?

robnagler: I guess I didn't remember how jobRunModeMap works. There's a difference in the way this works, for sure. 

JobRunMode does not change for a particular server. If it does, it's rarely done, and likely would be coupled with some other release so the GUI would refresh. 

GridEvolutionColumn is being changed as the result of a simulation. This is modifying global state in the app (the schema) based on user data that is generated dynamically for a particular simulation. I don't know what the values are, but could they possibly be added dynamically in a list, e.g. like lib files are?

Definitely need to handle "not there" cases. We want to know if it happens, at a minimum.

sirepo/template/flash.py
e-carlin: These file are [ASCII encoded](http://flash.uchicago.edu/~jbgallag/2012/flash4_ug/node15.html#SECTION05224000000000000000). Should I still use pkio.open_text? Will we ever run on a machine local.getprefferedencoding() isn't ascii compatible?

sirepo/package_data/static/js/madx.js
moellep: SINGLE_FRAME_ANIMATION is a UI constant. I don't think the server needs it.

sirepo/package_data/static/js/flash.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2676/commits/6867d3d5b0cd825d636f681fbe341e7967026ceb

sirepo/package_data/static/js/flash.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2676/commits/6867d3d5b0cd825d636f681fbe341e7967026ceb

sirepo/package_data/static/js/madx.js
e-carlin: https://github.com/radiasoft/sirepo/issues/2687

sirepo/template/flash.py
e-carlin: @robnagler ping

robnagler: I'm getting concerned about getprefferedencoding. I think we might want to use utf-8 instead, because fundamentally, pkio.open_text is not how we would handle i18n. It would be explicit based on the logged in user (not command line user).

Therefore, I think the change is to open_text always be utf8. We have to leave pkcompat.locale_str the way it is, because locale means something. We might want to change those references anyway to something else eg. to_bytes and from_bytes.

e-carlin: https://github.com/radiasoft/sirepo/pull/2676/commits/4dfa1f80bac0b73eb40a7d8189d81a158474af16

PR #2677, opened 2020-07-21
Fix #2584. Remove flask_mail

### LINKED ISSUE #2584: replace flask.mail
flask.mail makes for an annoying coupling to flask.

e-carlin:
This is still a work in progress. I need to get postfix setup on my machine so I can test it.


sirepo/comsol_register.py
```diff
@@ -6,51 +6,38 @@
...
-            MAIL_PORT=587,
-            MAIL_SERVER=cfg.mail_server,
-            MAIL_USERNAME=cfg.mail_username,
-            MAIL_PASSWORD=cfg.mail_password,
-        )
-        _mail = flask_mail.Mail(a)
-    req = http_request.parse_json()
-    msg = flask_mail.Message(
+    smtp.SMTP(
+        smtp_server=cfg.mail_server,
```
robnagler, line 27: This should be moved into smtp. we don't need two smtp logins and moreover, we probably only want to send from one address/name (support) so add those with appropriate defaults to smtp cfg. recipient config needs to stay in comsol. 

I don't think we need to deprecate the config in auth.email. Just remove it. I can add it to the rsconf db in advance of pushing. Nobody else is using smtp (for comsol or auth).

sirepo/smtp.py
```diff
@@ -0,0 +1,39 @@
...
+        s = None
+        try:
+            # TODO(e-carlin): Understand line endings issue. EmailMessage
+            # defaults to \n RFC expects \r\n
+            # See https://docs.python.org/3/library/email.policy.html#email.policy.default
+            m = email.message.EmailMessage()
+            m['From'] = sender
+            m['To'] = recipient
+            m['Subject'] = subject
+            m.set_content = body
```
robnagler, line 31: In [the documentation](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.set_content), it says this is function:

```py
set_content(*args, content_manager=None, **kw)
```

sirepo/smtp.py
```diff
@@ -0,0 +1,39 @@
...
+            with smtplib.SMTP(self.smtp_server, self.smtp_port) as s:
+                s.starttls()
+                s.ehlo()
+                s.login(self.smtp_user, self.smtp_password)
+                s.send_message(m)
+        finally:
```
robnagler, line 37: why would this be necessary with the `with`?

sirepo/smtp.py
```diff
@@ -0,0 +1,39 @@
...
+        try:
+            # TODO(e-carlin): Understand line endings issue. EmailMessage
+            # defaults to \n RFC expects \r\n
+            # See https://docs.python.org/3/library/email.policy.html#email.policy.default
+            m = email.message.EmailMessage()
+            m['From'] = sender
+            m['To'] = recipient
+            m['Subject'] = subject
+            m.set_content = body
+            with smtplib.SMTP(self.smtp_server, self.smtp_port) as s:
```
robnagler, line 32: I don't think you need the `try`. [`SMTP.__exit__`](https://github.com/python/cpython/blob/3.8/Lib/smtplib.py#L275) seems to do the right thing. [`quit`](https://github.com/python/cpython/blob/3.8/Lib/smtplib.py#L986) is designed for connection reuse, which we aren't doing.

sirepo/smtp.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(*args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.smtp_port = 587
+
+    def send(self, subject, sender, recipient, body):
+        if isinstance(sender, tuple):
+            sender = f'{sender[0]} <{sender[1]}>'
+        s = None
+        try:
+            # TODO(e-carlin): Understand line endings issue. EmailMessage
```
robnagler, line 24: I think [send always inserts `\r\n`](https://github.com/python/cpython/blob/3.8/Lib/smtplib.py#L968). `flatten` (in the referenced code) has to parse everything anyway.

sirepo/smtp.py
```diff
@@ -0,0 +1,39 @@
...
+
+class SMTP(PKDict):
+    def __init__(*args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.smtp_port = 587
+
+    def send(self, subject, sender, recipient, body):
+        if isinstance(sender, tuple):
+            sender = f'{sender[0]} <{sender[1]}>'
+        s = None
```
robnagler, line 22: remove

sirepo/comsol_register.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
e-carlin: Thanks https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
e-carlin: Doh, I had the try initially, move to with, forgot to delete the try https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
e-carlin: Nice, thx. https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2677/commits/5562b8e45c4790e870646d055a019adb1ceb5dc1

sirepo/smtp.py
```diff
@@ -0,0 +1,36 @@
...
+    m.set_content(body)
+    with smtplib.SMTP(cfg.server, cfg.port) as s:
+        s.starttls()
+        s.ehlo()
+        s.login(cfg.user, cfg.password)
+        s.send_message(m)
+
+
+cfg = pkconfig.init(
+    from_email=('support@sirepo.com', str, 'Email addres of sender'),
```
robnagler, line 30: addres => address

e-carlin: https://github.com/radiasoft/sirepo/pull/2677/commits/ddb33cd41cecd4a85c547fa764aa47e1f7fb0d03

Issue #2652, opened 2020-07-09
Simple top-level URLs for marketing materials


robnagler:
@messamer requests: Top level URLs for marketing materials, e.g. `/plans` should redirect to `/en/plans.html`.

robnagler:
I think these should be redirects, because it will add complexity. We need to add them to the schema or some lookup table to ensure we don't have collisions.

e-carlin:
@messamer we now have top-level marketing urls. So sirepo.com/plans goes to sirepo.com/en/plans (well it will be on alpha once #2790 is closed). Others can be added if you wish.

PR #2690, opened 2020-07-25
Fix #2520. Better message when user sim is purged.

### LINKED ISSUE #2520: Inform non-premium users when run_dir is purged after x days.
Currently the status is just set to Job_run_purged. We should tell users why it was purged and give them info on upgrading so it doesn't happen again.
Screen Shot 2020-05-15 at 10 08 42 AM

e-carlin:


sirepo/job_supervisor.py
```diff
@@ -431,6 +432,7 @@ def __db_load(cls, compute_jid):
                 'isPremiumUser',
                 'jobStatusMessage',
                 'internalError',
+                # TODO(e-carlin): dbUpdateTime?
```
e-carlin, line 435: @robnagler what's the proper way to do a fixup for this. Just use srtime.utc_now_as_float()? Try and look back in the history and use another time (ex lastUpdateTime)?

robnagler: `__db_write` should set `dbUpdateTime` to now before writing. This should cover all cases except [this one](https://github.com/radiasoft/sirepo/blob/6d285a3f4e33e9f4871aa63417f5451f8947d4a0/sirepo/job_supervisor.py#L839), which seems to be a bug. There's a db_init when the run_dir is not found, but there's no db_update/write.

sirepo/job_supervisor.py
```diff
@@ -883,7 +883,10 @@ def res(**kwargs):
         ):
             return PKDict(state=job.MISSING, reason='computeJobSerial-mismatch')
         if self.db.isParallel or self.db.status != job.COMPLETED:
-            return res(state=self.db.status)
+            return res(
+                state=self.db.status,
+                dbUpdateTime=_strf_unix_time(self.db.dbUpdateTime),
```
robnagler, line 888: Seems to me this is presentation logic. Why do we need to send back a formatted time stamp?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -158,6 +160,14 @@ SIREPO.app.factory('authState', function(appDataService, appState, errorService,
         controller.showWarning = true;
         controller.warningText = 'Server reported an error, please contact support@radiasoft.net.';
     };
+
+    self.premiumOrEnterprise = function() {
+        if (self.roles.indexOf('premium') >= 0) {
```
robnagler, line 165: This string should come from the server.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -312,10 +312,12 @@ SIREPO.app.directive('cancelledDueToTimeoutAlert', function(authState) {
         template: [
             '<div data-ng-if="simState.getCancelledAfterSecs()" class="alert alert-warning" role="alert">',
               '<h4 class="alert-heading">Cancelled: Maximum runtime exceeded</h4>',
-              '<p>Your simulation ran for {{getTime()}}. To increase your maximum runtime please upgrade to <a href="https://radiasoft.net/sirepo" target="_blank">Sirepo {{ premiumOrEnterprise() }}</a>.</p>',
+              '<p>Your simulation ran for {{getTime()}}. To increase your maximum runtime please upgrade to <a href="https://radiasoft.net/sirepo" target="_blank">Sirepo {{ authState.premiumOrEnterprise() }}</a>.</p>',
```
robnagler, line 315: The plan is "Sirepo Premium". This string should be in the schema, and looked up based on the servicePlan or something like that in authState. 

`https://radiasoft.net/sirepo` should not be hardwired here. It should be a looked up in the schema. #2652 needs to be addressed, and it's basically the same idea. We need to name this redirect so that we know 1) that the redirect happens here, and 2) we can control it for on premise installs.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -1445,6 +1440,40 @@ SIREPO.app.directive('simplePanel', function(appState, panelState) {
...
+        },
+        template: [
+            '<div class="col-sm-12" ng-bind-html="message()"><br><br></div>',
+        ].join(''),
+        controller: function($scope, $sce) {
+            $scope.message = function() {
+                if ($scope.simState.isStatePurged()) {
+                    return $sce.trustAsHtml([
+                        '<div>Simulation data purged on ' + $scope.simState.getDbUpdateTime() + ' UTC.</div>',
+                        '<div>Upgrade to <a href="https://radiasoft.net/sirepo" target="_blank">Sirepo ' + authState.premiumOrEnterprise() + '</a> for persistent data storage.</div>',
```
robnagler, line 1457: See previous comment about redirects to pay.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -1445,6 +1440,40 @@ SIREPO.app.directive('simplePanel', function(appState, panelState) {
...
+            simState: '=simulationStoppedStatus',
+        },
+        template: [
+            '<div class="col-sm-12" ng-bind-html="message()"><br><br></div>',
+        ].join(''),
+        controller: function($scope, $sce) {
+            $scope.message = function() {
+                if ($scope.simState.isStatePurged()) {
+                    return $sce.trustAsHtml([
+                        '<div>Simulation data purged on ' + $scope.simState.getDbUpdateTime() + ' UTC.</div>',
```
robnagler, line 1456: `getDbUpdateTime` should return a time which is rendered in the local time zone. There's no need to use UTC for end-users.

sirepo/package_data/static/js/srw.js
```diff
@@ -2194,6 +2189,10 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 srwService.computeModel($scope.model),
                 handleStatus
             );
+
+            $scope.simState.notRunningMessage = function() {
```
robnagler, line 2193: This seems strange as there's no indication here that the simulation is or is not running. This is just displaying stateAsText with "Simulation " in front of it.

sirepo/job_supervisor.py
```diff
@@ -483,19 +488,14 @@ def _get_header():
...
-
         def _strf_seconds(seconds):
             # formats to [D day[s], ][H]H:MM:SS[.UUUUUU]
             return str(datetime.timedelta(seconds=seconds))
 
         def _get_rows():
             def _get_queued_time(db):
                 m = i.db.computeJobStart if i.db.status == job.RUNNING \
-                    else int(time.time())
+                    else int(sirepo.srtime.utc_now_as_float())
```
robnagler, line 498: thanks for fixing these.

sirepo/job_supervisor.py
e-carlin: From my experience, JS has [weak capabilities for formatting timestamps](https://css-tricks.com/everything-you-need-to-know-about-date-in-javascript/). We are already doing it on the server so I just continued down that path. Thinking about it more it looks like [angularjs provides what we need](https://docs.angularjs.org/api/ng/filter/date). Should I use that instead?

sirepo/package_data/static/js/sirepo.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2690/commits/6ae62e9e3521eeb67104be4c4eda58da7c49f4d4

sirepo/job_supervisor.py
robnagler: Using a date formatter is fine. I would ideally like to encapsulate this so we don't become more dependent on angular. We would want all dates formatted the same way so make a routine (e.g. appState.formatDate) and put all the code there.

sirepo/package_data/static/js/sirepo-components.js
e-carlin: Is this right? https://github.com/radiasoft/sirepo/pull/2690/commits/68d5a2d040d51f8ae5ba41517c9b0571c45a1f72

sirepo/job_supervisor.py
e-carlin: I'll make those changes. But, we still have a problem. If a runStatus is the first request a __db_write() is never called so dbUpdateTime won't be on the db object.

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/blob/6d285a3f4e33e9f4871aa63417f5451f8947d4a0/sirepo/job_supervisor.py#L234

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/pull/2690/commits/68d5a2d040d51f8ae5ba41517c9b0571c45a1f72

sirepo/package_data/static/js/srw.js
e-carlin: Can you say that a different way? I don't understand the issue. 

If the simulation isn't running it will say something like `Simulation completed`. This code is [only called](https://github.com/radiasoft/sirepo/blob/6d285a3f4e33e9f4871aa63417f5451f8947d4a0/sirepo/package_data/static/js/sirepo-components.js#L1462) when a simulation isn't running

sirepo/job_supervisor.py
robnagler: I was making the point that [this case](https://github.com/radiasoft/sirepo/blob/6d285a3f4e33e9f4871aa63417f5451f8947d4a0/sirepo/job_supervisor.py#L839) needs to call db_write, because it is creating a db object, returning it, and not writing it.

sirepo/package_data/static/js/sirepo-components.js
robnagler: Seems fine to me. The chaos of it bothers me, but at this stage, I don't see a way out of it.

sirepo/package_data/static/js/srw.js
robnagler:
```js
   self.simState.notRunningMessage = function() {
        return 'Simulation ' + self.simState.stateAsText();
    };
```

is called here:

```js
             var m = 'notRunningMessage';
                var res = $scope.simState[m] && $scope.simState[m]();
                if (res) {
                    return res;
                }
                res = 'Simulation ' + $scope.simState.stateAsText();
```

The concept 'Simulation ' + stateAsText can be reused in all cases. They are all the same format.

The problem is `simState.getFrameCount()` needs to be called in the case that it is running or has run. That's a separate concept but related to simState. Therefore, I propose a single function on simState that does it all.

sirepo/package_data/static/js/sirepo-components.js
bruhwiler: I should be pointing at sirepo.com and not at radiasoft.net

robnagler: Payments go through radiasoft.net/sirepo

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2690/commits/baaf84a3bdfc7d0bba4ec6793c0eb670645b610f

Should I fix the [jobs page](https://github.com/radiasoft/sirepo/blob/baaf84a3bdfc7d0bba4ec6793c0eb670645b610f/sirepo/job_supervisor.py#L511) too?

sirepo/package_data/static/js/sirepo-components.js
e-carlin: https://github.com/radiasoft/sirepo/commit/baaf84a3bdfc7d0bba4ec6793c0eb670645b610f

sirepo/job_supervisor.py
robnagler: Yes, please

sirepo/package_data/static/js/sirepo-components.js
robnagler: @e-carlin Please change upgradePlanLink  to /en/plans.html

e-carlin: sirepo.com/en/plans.html ? I get a 404 . On radiasoft.net I get redirected to the home page.

sirepo/package_data/static/js/srw.js
e-carlin: I've abstracted `'Simulation' + stateAsText()`. https://github.com/radiasoft/sirepo/pull/2690/commits/52c29708d04620911b0275d8c8fef7b493ebc23e . I have a feeling you want something more but I'm not sure.

`simState.getFrameCount()` needs to be called when we **care** about frames. Ex in srw we care about particles not frames so we have a different notRunningMessage that doesn't display info about frames.

sirepo/job_supervisor.py
e-carlin: Ok, I'll fix that. 

Can you answer my original question:

> robnagler what's the proper way to do a fixup for this. Just use srtime.utc_now_as_float()? Try and look back in the history and use another time (ex lastUpdateTime)?

We need to add dbUpdateTime for all of the jobs in the db that don't have it so we can display it in the GUI.

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2690/commits/4d21167b3a5bfc67863e22bdd19621a3288bbb87

sirepo/job_supervisor.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2690/commits/c8a7613fdb8c6f1d53c4cc8e1569628943fcd152 fix all time formats.

https://github.com/radiasoft/sirepo/pull/2690/commits/c8a7613fdb8c6f1d53c4cc8e1569628943fcd152 fix missing db_write

sirepo/package_data/static/js/srw.js
e-carlin: @robnagler https://github.com/radiasoft/sirepo/pull/2690/commits/71a952b86573f08ba59907d977720a7a4aeb0a7e

mkeilman:
> Or in between the solve and reset button as Rob says?

I think the latter: like this?

```
                        '<div data-sim-status-panel="viz.simState"></div>',
                        '<div data-ng-show="(viz.solution || []).length">',
                                '<div><strong>Step Count:</strong> {{ viz.solution[3] }}</div>',
                                '<div><strong>Max |M|: </strong> {{ utilities.roundToPlaces(viz.solution[1], 4) }} A/m</div>',
                                '<div><strong>Max |H|: </strong> {{ utilities.roundToPlaces(viz.solution[2], 4) }} A/m</div>',
                        '</div>',
                        '<div data-ng-hide="(viz.solution || []).length">No solution found</div>',

```

I was leveraging the notRunningMessage thing at the time...

robnagler:
Would it be possible to avoid code like: `(viz.solution || []).length` and `utilities.roundToPlaces(viz.solution[1], 4)`. Instead we should have a formatter that shares the concept of 4 decimal place rounding which is probably used elsewhere. And, we'd return a dict instead of an array with the solution so the fields are labeled appropriately, possibly being a model?

mkeilman:
Just added a stab at the solution thing.  I left the roundToPlaces calls to be addressed in its own issue, is that good?

robnagler:
Good enough.

robnagler:
@e-carlin good to go?

e-carlin:
Good on my end.

Issue #2691, opened 2020-07-25
Generate HTML instead of copy-and-paste (specific case)


robnagler:
[This code](https://github.com/radiasoft/sirepo/pull/2690/files#diff-2cd9b0ed6866eb2c75933e0e8026d414R25) should be encapsulated:

```html
<div data-simulation-stopped-status="visualization.simState"></div>
<div class="col-sm-12" data-ng-show="visualization.simState.hasTimeData()" data-simulation-status-timer="visualization.simState"></div>
```

However, I think it should be encapsulated in javascript. It needs to be generated once. This code should be changed at the same time:

```js
        "visualization": {
            "route": "/visualization/:simulationId",
            "config": {
              "controller": "VisualizationController as visualization",
              "templateUrl": "/static/html/madx-visualization.html"
            }
        }
```

It might look like:

```js
        "visualization": {
            "route": "/visualization/:simulationId",
            "config": {
              "controller": "VisualizationController as visualization",
              "template": "<div data-visualization-tab=''></div>"
            }
        }
```

VisualizationTab would be a widget would know how to find `madx-visualization.html`, because it knows the code name, and it knows the name VisualizationController. There's no need to have all this repetition. Ultimately, these localRoutes should be generated upon reading the schema. They are quite redundant and out of control, but that's not the point at the moment.

The file `madx-visualization.html` would be preprocessed on the server (and cached) with a simple algorithm:

```py
return re.sub(
    '^(\w+)\(([^\)]*)\)$', 
    lambda m: _dispatch(m.group(1), m.group(2).split('\s*,\s*')),
    template, 
    flags=re.MULTILINE,
)
```

`_dispatch` will call the function named by `m.group(1)` with the `split arguments`, which are required to be identifiers, e.g.

```txt
sim_state_text(visualization)
```

Initially, the function `sim_state_text` might look like:

```py
def html_sim_state_text(scope):
    return f'''<div data-simulation-stopped-status="{scope}.simState"></div>
<div class="col-sm-12" data-ng-show="{scope}.simState.hasTimeData()" data-simulation-status-timer="{scope}.simState"></div>'''
```

This function could be used in other places to generate the html for inline templates by preprocessing the javascript before sending to the server or, better, having a lookup table for inline templates that are preprocessed.

I don't particularly care about the syntax now. The above syntax does not appear in any templates in `html/*.html`.

BTW, you can use single quotes for attribute values (as I did above). This avoids `"x=\"\""`, which is hard to read.

[...]

Issue #1561, opened 2019-03-26
AngularJS migration


robnagler:
This task is for brainstorming how/when we should migrate from AngularJS.

End of Life for AngularJS 1.7 (our and last version) is 6/30/2021. The conversion is not simple unlike #1560 (py3 migration), which should happen first anyway.

Browser frameworks are evolving very rapidly. The longer we wait, the better off we will be in terms of which framework is best for us. And, the longer we wait, the more difficult the conversion.

robnagler:
Whatever we choose, it needs to be dispatch-driven, not state-variable-driven. I'm working on #1495, and the control flow is very hard to follow, especially between the server and the client. 

There are 251 `if` statements in sirepo.js right now. That's way too many.  Defects are more common with state variables, because they have to agree with other state variables and their uses. Code which uses dispatch is going to be less buggy.

PR #2695, opened 2020-07-25
WIP #2691 html.render: non-jinja templating

robnagler:
The template syntax looks like:

```txt
  <div class="row">
    sim_settings_and_status(scope=visualization);
    <div class="col-md-6 col-xl-4" data-ng-if="visualization.simState.hasFrames()">
```

We could make it recursive and nested. Not important right now. I want to explore how we could simplify the html templates, and then the inline templates in the js.

### DIFF
--- a/sirepo/html.py
+++ b/sirepo/html.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+u"""html template
+
+:copyright: Copyright (c) 2020 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdc, pkdlog, pkdp
+import pykern.pkio
+import re
+import sys
+
+
+_CALL_RE = re.compile(r'(\w+)\(([^\)]*)\);', flags=re.MULTILINE)
+
+_ARGS_RE = re.compile(r'\s*,\s*')
+
+_ARG_RE = re.compile(r'^(\w+)=(.*)$', flags=re.DOTALL)
+
+_FUNC_MAP = PKDict()
+
+_FUNC_PREFIX = '_widget_'
+
+def render(path):
+    def _dispatch(match):
+        f = match.group(1)
+        assert f in _FUNC_MAP, \
+            f'function={f} not in map path={path}'
+        k = PKDict(ctx=PKDict(path=path))
+        for a in  _ARGS_RE.split(match.group(2).strip()):
+            m = _ARG_RE.search(a)
+            assert m, \
+                f'invalid keyword argument={a} function={f} path={path}'
+            assert m.group(1) not in k, \
+                f'duplicate keyword argument={a} function={f} path={path}'
+            k[m.group(1)] = m.group(2)
+        return _FUNC_MAP[f](**k)
+
+    return _CALL_RE.sub(_dispatch, pykern.pkio.read_text(path))
+
+
+def _init():
+    import sys, inspect
+    for n, o in inspect.getmembers(sys.modules[__name__]):
+        if n.startswith(_FUNC_PREFIX) and inspect.isfunction(o):
+            _FUNC_MAP[n[len(_FUNC_PREFIX):]] = o
+
+
+def _widget_sim_settings_and_status(ctx, scope):
+    return f'''<div class="col-md-6 col-xl-4">
+      <div data-basic-editor-panel="" data-view-name="simulationSettings"></div>
+    </div>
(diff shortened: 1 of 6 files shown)


sirepo/html.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
```
e-carlin, line 1: Should this file be named srhtml? When do you add the sr prefix (ex srtime?).

sirepo/html.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def _init():
+    import sys, inspect
+    for n, o in inspect.getmembers(sys.modules[__name__]):
+        if n.startswith(_FUNC_PREFIX) and inspect.isfunction(o):
+            _FUNC_MAP[n[len(_FUNC_PREFIX):]] = o
+
+
+def _widget_sim_settings_and_status(ctx, scope):
```
moellep, line 50: I'd rather see the html formatting on the client in javascript. I think we could have a similar html.js module for this. Adding it too the server makes it difficult to switch between python/javascript modes of thinking and make it harder to track down where things are being referenced in the UI.

sirepo/html.py
robnagler: I thought about srhtml, but realized we want it in all cases except for the few jinja files that are loaded before boot. I will address this more in response to https://github.com/radiasoft/sirepo/pull/2695#pullrequestreview-455890322

The sr prefix is a mistake as is the pk prefix except in rare instances like pkdlog and PKDict, which we want to use "everywhere". If we just always used fully qualified names, sirepo.time and sirepo.db, we wouldn't have any problems. It was a bit a problem in py2, but we had absolute_imports so it wasn't a problem.

sirepo/html.py
```diff
@@ -0,0 +1,62 @@
...
+                f'duplicate keyword argument={a} function={f} path={path}'
+            k[m.group(1)] = m.group(2)
+        return _FUNC_MAP[f](**k)
+
+    return _CALL_RE.sub(_dispatch, pykern.pkio.read_text(path))
+
+
+def _init():
+    import sys, inspect
+    for n, o in inspect.getmembers(sys.modules[__name__]):
```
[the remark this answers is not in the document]
robnagler: I somehow didn't see that when I looked in pkinspect, and yes, I wrote it. :)  One of the problems with emergent APIs is that you never know when they are "regular" and can be assumed.

sirepo/html.py
robnagler: #1561 looms in my mind. How are we ever going to decouple the code from AngularJS so we can migrate to something else? When this has worked in the past, it was through declarative programming: _what not how_.

I realize that we have the problem of switching between python/javascript, and I agree with you. Yet, we have many more modes, and some modes are being ignored such as test mode. When I'm thinking in test mode, I think in code not DOM. 

Javascript in the browser is deficient for test mode in a couple of ways. There's no file system to store intermediate inputs/outputs. This is critical when debugging code generation, because you want to see the output of the code generator. One of the problems with Lisp macros is that all the code generation happens in the interpreter, and you have to do testing at the repl, which seems to be a favorite of lispers, but is insufficient for building large systems. Automation is the only solution, and tracing intermediate results is key to understanding when automation goes wrong.

Another problem with Javascript in the browser is that we lose control of the language processor version (see https://github.com/radiasoft/sirepo/pull/2663#pullrequestreview-455896648). It was very hard with py2/py3, but we could actively choose to be in one or the other. It becomes a least common denominator thing, which to me devolves to assembly language, which is how I think about Javascript. This is why people program in Typescript, in a way, so they can have a consistent execution environment (assuming no bugs in the language processor, which is the C++ problem that they replicated in typescript).

Ideally, we'd build a declarative language that could speak about all the ways we need to interact in Javascript. The schema would drive all of it. Then, we would only need to test the widgets in browsers, independent of the data.

The other thing about generating HTML and Javascript in Javascript is that it is confusing to me to ready. I don't like the templateUrl, because they are not abstracted.

While I understand when we first started writing Sirepo, we didn't know what we are doing. Today, we have lots of examples of what needs to be encapsulated.

For this particular problem, the files are requested by angular directly. Rather than figuring out how angular caches the templates, I decided the server could generate what needs to be cached. This is similar to the google tag solution, and could replace it. The goal would be to start slowly as we modified code, and refine the language over time. I would also like to encaspulate bootstrap as a part of this process, because it will be replaced one day. 

Naturally, this will not make Sirepo easy for the novice to understand. I don't think in its present form it is easy to understand (see [schema constants](https://github.com/radiasoft/sirepo/commit/68d5a2d040d51f8ae5ba41517c9b0571c45a1f72#diff-e804deb9cc0f9aa7d43101b42eb28f55R145)).

PR #2705, opened 2020-07-29
Fix #2703. Elapsed time should be wallclock - start

### LINKED ISSUE #2703: Elapsed time doesn't update as time increases
As a simulation is running, the "Elapsed time" clock does not reflect the actual run time.  
IMG_6702
IMG_6701
IMG_6703

e-carlin:
@robnagler is this right? Was there a reason elapsed time was ever calculated using lastUpdateTime?


sirepo/job_supervisor.py
```diff
@@ -860,7 +860,7 @@ def res(**kwargs):
                 r.update(self.db.parallelStatus)
                 r.computeJobHash = self.db.computeJobHash
                 r.computeJobSerial = self.db.computeJobSerial
-                r.elapsedTime = self.db.lastUpdateTime - self.db.computeJobStart
+                r.elapsedTime = int(sirepo.srtime.utc_now_as_float()) - self.db.computeJobStart
```
robnagler, line 863: I don't think this is right. If the simulation is canceled, the time should not change.

robnagler:
We want a clock to run in javascript. It should stop when status changes to canceled, completed, etc. 

The server will still have to return elapsedTime, but it should be based on dbUpdateTime, not lastUpdateTime. lastUpdateTime is insufficient, because some templates don't update it. We will update dbUpdateTime when the simulation is canceled, completed, etc.

e-carlin:
I decided to calculate elapsedTime on the server in all cases. I think it covers everything and is simpler. Lmk

https://github.com/radiasoft/sirepo/pull/2705/commits/1cf8fc1cf1c7c64948feadd10393b252e4b443ba

PR #2721, opened 2020-07-31
fix #2707 move common code variable logic into code_variable module

### LINKED ISSUE #2707: refactor lattice code variables
Lattice applications which use dynamic user-defined variables (historically named RPNValues) need to share more code. For example, the template's get_application_data() has a lot of code which can be refactored across the elegant, opal and madx applications.

moellep:


sirepo/pkcli/madx.py
```diff
@@ -63,7 +63,6 @@ def _generate_ptc_particles_file(run_dir, data, twiss):
     if 'bunchReport' in data.report:
         v.summaryData = twiss
         simulation_db.write_json(run_dir.join(template.BUNCH_PARTICLES_FILE), v)
-        return
```
robnagler, line 66: was this intentional?

sirepo/template/madx.py
```diff
@@ -135,38 +131,22 @@ def background_percent_complete(report, run_dir, is_running):
...
-    if data.method == 'validate_rpn_delete':
-        model_data = simulation_db.read_json(
-            simulation_db.sim_data_file(data.simulationType, data.simulationId))
-        data.error = cv.validate_var_delete(
-            data.name,
-            model_data,
-            _SCHEMA
-        )
+    if _code_var(data.variables).get_application_data(data, _SCHEMA, ignore_array_values=True):
         return data
```
robnagler, line 137: Missing a return statement when get_application_data returns false (which I don't think it should).

sirepo/template/code_variable.py
```diff
@@ -63,6 +66,29 @@ def eval_var_with_assert(self, expr):
...
+            return True
+        if args.method == 'recompute_rpn_cache_values':
+            self.recompute_cache(args.cache)
+            return True
+        if args.method == 'validate_rpn_delete':
+            model_data = simulation_db.read_json(
+                simulation_db.sim_data_file(args.simulationType, args.simulationId))
+            args.error = self.validate_var_delete(args.name, model_data, schema)
+            return True
+        return False
```
robnagler, line 90: Shouldn't this throw an exception? see false case below which is not handled properly.

sirepo/template/code_variable.py
```diff
@@ -63,6 +66,29 @@ def eval_var_with_assert(self, expr):
...
+            if err:
+                args.error = err
+            else:
+                args.result = v
+            return True
+        if args.method == 'recompute_rpn_cache_values':
+            self.recompute_cache(args.cache)
+            return True
+        if args.method == 'validate_rpn_delete':
+            model_data = simulation_db.read_json(
```
robnagler, line 86: This is problematic. It bypasses the concept of validating the simulationId and simulationType. It also decouples the values that are in the GUI at the time get_application_data is called from the values on disk on the server, which could have been saved after those values were first retrieved. This is a known problem elsewhere, but it seems like we are continually avoiding the problem that what the GUI things of the model is different from what is in the db. If we passed the serial, it would guarantee that the two are in agreement.

sirepo/template/code_variable.py
```diff
@@ -63,6 +66,29 @@ def eval_var_with_assert(self, expr):
         assert not err, err
         return float(v)
 
+    def get_application_data(self, args, schema, ignore_array_values=False):
```
robnagler, line 69: I don't like the switch from `data` to `args`. It should validate the arguments before using them.

sirepo/server.py
```diff
@@ -254,7 +254,9 @@ def api_getApplicationData(filename=None):
     """
     req = http_request.parse_post(template=True, filename=filename or None)
     with simulation_db.tmp_dir() as d:
+        assert 'method' in req.req_data
```
robnagler, line 257: We should wrap get_application_data in something that does the checking of the inputs and outputs.

sirepo/pkcli/madx.py
moellep: Yes, both data files are needed, one for json, one for data download.

sirepo/template/madx.py
moellep: It used to have assertion failure in all the templates, so I moved it to server.py - it will now raise an assertion if the get_application_data() did not produce output.

robnagler: Fast fail would suggest to fail in the common code as well as if no output was produced. If you throw the exception in side code_variables.py, you'll get the stack of what happened. If you throw it outside, you have no idea why it wasn't produced. Is there a case where checking the return value should not result in an exception?

sirepo/template/code_variable.py
moellep: The True/False is indicating if the data.method was handled by the call. If it threw an exception, it would always have to be the very last call in get_application_data().

robnagler: Or, https://github.com/radiasoft/sirepo/pull/2721#discussion_r463821235 would dispatch based on method. Each method would map to a function on a class (ApplicationData). code_variables.get_application_data would be a base class.

PR #2729, opened 2020-08-03
Fix #2702. s/job_status/job_situation/.

### LINKED ISSUE #2702: Change job_status to job_situation
[Running was set](https://github.com/radiasoft/sirepo/blob/467924ccfb4ce3f59afead568569b45824d1cf30/sirepo/job_supervisor.py#L789) when it should be running. set_job_status should assert that it is in the valid set of statuses.

e-carlin:


sirepo/job_supervisor.py
```diff
@@ -784,7 +784,7 @@ async def _run(self, op):
         op.task = asyncio.current_task()
         op.pkdel('run_callback')
         try:
-            with op.set_job_status('Running'):
+            with op.set_job_situation('Running'):
```
robnagler, line 787: `Running` => `Entered __create._run`

e-carlin: https://github.com/radiasoft/sirepo/pull/2729/commits/1dda0d9dee04e8591c8d22291da4194e77f3583f

PR #2730, opened 2020-08-03
Fix #2688. Never output null or epoch to user

### LINKED ISSUE #2688: Never output "null" or "epoch" to user
If there is no data, the output should be n/a or blank. I guess @remy278 may have a comment about this. I saw this on the jobs page just now for a simulation that I hadn't run before:

Screen Shot 2020-07-24 at 2 05 35 PM

e-carlin:


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2832,16 +2832,21 @@ SIREPO.app.directive('jobsList', function(requestSender, appState, $location, $s
...
                 for (var i = getStartIndex(); i < row.length; i++) {
-                    var v = row[i];
-                    var t = $scope.data.header[i][1];
-                    if (t === 'DateTime') {
-                        v = appState.formatDate(v);
-                    }
-                    else if (t === 'Time') {
-                        v = appState.formatTime(v);
-                    }
+                    var v = typeDispatch[$scope.data.header[i][1]](row[i]);
```
robnagler, line 2849: Is there a bug here if the string could contain html?

Would be nice to encapsulate the n/a, since it is used in two different files. The only odd case is formatTime '0'. If we tested the result of the formatter for falsey, it might work. Then, we'd just have to return null in the case of falsey in formatDate. formatTime could stay the same, and formatString could either return null for falsey or just falsey.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2832,16 +2832,21 @@ SIREPO.app.directive('jobsList', function(requestSender, appState, $location, $s
...
 
             function getRow(row, nameIndex, simulationIdIndex, appIndex) {
+                function formatString(str) {
+                    if (! str) {
+                        return 'n/a';
+                    }
+                    return str;
+                }
+
+                var typeDispatch = {
```
robnagler, line 2842: Nice encapsulation.

sirepo/package_data/static/js/sirepo-components.js
e-carlin: > Is there a bug here if the string could contain html?
What kind of bug are you thinking of? An injection attack? The server could send back html and we would render it (it doesn't do this currently).

https://github.com/radiasoft/sirepo/pull/2730/commits/0c277ac46a3f83391c6f896c3a5524cd57098632

Issue #2706, opened 2020-07-30
refactor editor business logic


moellep:
The UI has too many incomplete ways to handle editor business logic:
- Through the controller with handleModalShown() or responding to sr-tabSelected events
- Through modal editor directive's like zgoubi's srCaviteEditor which get added when a modal is dynamically created.

This doesn't allow for popover editors used in the SRW and Shadow beamlines. Instead the controller handles all the interaction which leads to problems like the SRW BeamlineController being 500 lines of code with handlers for all the fields.

Instead of dynamically creating the field handler on appState.showModalEditor(), it should be done on the advancedEditorPane directly based on the viewName provided. For example, an angular directive could be dynamically loaded named srGratingEditor whenever the grating editor is created.

This change would greatly simplify the code and keep the business logic related to an editor in one place, rather than grouped into the controller. Common business logic could be kept in the app's service.

This issue overrides #1690

PR #2735, opened 2020-08-04
Changes for #2706 refactor editor business logic

moellep:
- create dynamic View directives when an UI editor is created for app business logic
- allow angular expressions in field label tooltips
- added PanelState enableFields() and showFields() to allow mass field updates
- update SRW app to use new view and beamline logic
- no longer need to create modal dialogs linked to the controller in html templates

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -49,9 +49,9 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState, $compi
             var viewInfo = appState.viewInfo($scope.viewName);
             var i;
 
-            function camelToSnakeCase(v) {
+            function camelToKebabCase(v) {
                 return v.replace(/[\w]([A-Z])/g, function(m) {
-                    return m[0] + "_" + m[1];
+                    return m[0] + "-" + m[1];
                 }).toLowerCase();
             }
 
@@ -73,19 +73,16 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState, $compi
             }
             // create a View component for app business logic
             $element.append($compile(
-                '<div data-' + camelToSnakeCase($scope.viewName)
+                '<div data-' + camelToKebabCase($scope.viewName)
                     + '-view="{{ fieldDef }}"'
                     + ' data-model-name="modelName" data-model-data="modelData">'
                     + '</div>')($scope));
 
             $scope.isColumnField = function(f) {
-                return typeof(f) == 'string' ? false : true;
+                return typeof(f) != 'string';
             };
             $scope.isField = function(f) {
-                if ($scope.isColumnField(f) || $scope.isLabel(f)) {
-                    return false;
-                }
-                return true;
+                return !($scope.isColumnField(f) || $scope.isLabel(f));
             };
             $scope.isLabel = function(f) {
                 if ($scope.isColumnField(f)) {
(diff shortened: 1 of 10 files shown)


moellep:
This is a start on isolating editor logic. More work could be done to make it more declarative. It reduced the logic in the controllers and removed the coupling between directives and controllers.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -49,6 +49,12 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState) {
             var viewInfo = appState.viewInfo($scope.viewName);
             var i;
 
+            function camelToSnakeCase(v) {
```
robnagler, line 52: yay!

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -49,6 +49,12 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState) {
...
+                return v.replace(/[\w]([A-Z])/g, function(m) {
+                    return m[0] + "_" + m[1];
```
robnagler, line 54: Does angular accept underscore as well as dashes?

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -49,6 +49,12 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState) {
             var viewInfo = appState.viewInfo($scope.viewName);
```
robnagler, line 49: Is there a way to avoid using $scope here? It seems like all these values could be on an object, and then refer to them via that object.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -57,13 +63,21 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState) {
...
             }
+            // create a View component for app business logic
+            $element.append($compile(
+                '<div data-' + camelToSnakeCase($scope.viewName)
+                    + '-view="{{ fieldDef }}"'
+                    + ' data-model-name="modelName" data-model-data="modelData">'
+                    + '</div>')($scope));
+
             $scope.isColumnField = function(f) {
                 return typeof(f) == 'string' ? false : true;
```
robnagler, line 82: `return typeof(f) != 'string'`

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -57,13 +63,21 @@ SIREPO.app.directive('advancedEditorPane', function(appState, panelState) {
...
             };
```
robnagler, line 83: `return !($scope.isColumnField(f) || $scope.isLabel(f))`

sirepo/package_data/static/js/sirepo-components.js
moellep: Yes, I guess it does. This was unintentional. I think I wanted camelToKebobCase() but I copied the wrong function.

sirepo/package_data/static/js/sirepo-components.js
moellep: You can think of $scope as self here. The args come in from html attributes, so you could grab "data-view-name" attribute from the html element. Attributes like modelData are somehow replaced with an actual object by angularjs, so that would be tricky to replace.

robnagler: The question is: do they have to come from html elements. Would it not be better to parameterize another way? I need to think about this.

PR #2745, opened 2020-08-05
Fix #2740. Use chrome instead of phantomjs for testing.

### LINKED ISSUE #2740: jshint es6
I started using let and const because we are now using es6. jshint didn't like this (#2739). When I changed to es6 in etc/jshint.conf there were many errors. We should use es6 in jshint.conf and resolve those errors.

e-carlin:
Phantomjs doesn't support ES6 and it is deprecated.
Use Chrome Headless for testing instead.

This depends on radiasoft/download#115.

### DIFF
--- a/etc/jshint.conf
+++ b/etc/jshint.conf
@@ -4,7 +4,7 @@
     "globalstrict": true,
     "predef": ["angular", "$", "console", "d3", "SIREPO", "saveAs", "Colorbar", "Detector", "Modernizr", "Split", "domtoimage", "vtk", "katex"],
     "curly": true,
-    "esversion": 5,
+    "esversion": 6,
     "freeze": true,
     "futurehostile": true,
     "noarg": true,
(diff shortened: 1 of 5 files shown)


e-carlin:
Yes https://github.com/radiasoft/sirepo/pull/2745/commits/51774c2ded4dccf7b7faf294a97e94f55bfe94fb

I think adding it to beamsim-codes make sense.

robnagler:
It's about 100MB, because it brings in wayland... This is how we get to 10GB images. Go ahead and fix it including test.sh

e-carlin:
I added chrome to common https://github.com/radiasoft/download/pull/115/commits/264145987744952a5172b3927748c89f1236e24d

I also removed es6 checks from test.sh https://github.com/radiasoft/sirepo/pull/2745/commits/0f03d29a5664b789d5f1a7c6bb79c740f4d78fe7

Issue #2667, opened 2020-07-16
Async tester


e-carlin:
We made the async tester. Need to test it again and get it running on a schedule.

PR #2752, opened 2020-08-05
WIP #2667. Async tester fixups.

e-carlin:


sirepo/pkcli/test_http.py
```diff
@@ -328,10 +306,18 @@ async def post(self, uri, data, caller):
                         'Content-type',  'application/json'
                     ),
                     method='POST',
-                    **self._FETCH_DEFAULT_ARGS
+                    **self._fetch_default_args
                 ),
             )
 
+    @property
```
e-carlin, line 313: @robnagler this is wonky but I wasn't sure how else to implement this. If we define _fetch_default_args as a constant on the class like:
```python
class _Client:
  _FETCH_DEFAULT_ARGS = PKDict(..., validate_cert=cfg.validate_cert)
```

I get an error `nonetype has no attribute validate_cert`. This is because cfg is None when this code is evaluated. Do you have a better way of implementing this?

sirepo/pkcli/test_http.py
```diff
@@ -524,3 +515,23 @@ async def _run_sim_until_completion(self):
...
+        return
+    c = sirepo.pkcli.service.cfg()
+    cfg = pkconfig.init(
+        emails=(['one@radia.run', 'two@radia.run', 'three@radia.run'], list, 'emails to test'),
+        server_uri=(
+            'http://{}:{}'.format(c.ip, c.port),
+            str,
+            'where to send requests',
+        ),
+        run_min_secs=(90, int, 'minimum amount of time to let a simulation run'),
```
robnagler, line 532: pkconfig.parse_seconds

sirepo/pkcli/test_http.py
robnagler: I wouldn't bother with making it a property. That's an optimization. :)  I would just call the function fetch_default_args. It doesn't need to be a constant.

e-carlin: :)

https://github.com/radiasoft/sirepo/pull/2752/commits/355420a7e87bf9b04bf43496ba6731e81ed70d2d

sirepo/pkcli/test_http.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2752/commits/355420a7e87bf9b04bf43496ba6731e81ed70d2d

PR #2759, opened 2020-08-07
Website lines

cwmcallister:
This should eliminate the thin line at the bottom of each public plan.
And all instances where that could have been possible.
@messamer @robnagler


sirepo/package_data/static/en/css/landing.css.map
```diff
@@ -1,7 +1,7 @@
...
+"sources": ["file:///Users/connor/Sites/cm-sirepo/sirepo/package_data/static/en/css/main.scss","file:///Users/connor/Sites/cm-sirepo/sirepo/package_data/static/en/css/variables.scss","file:///Users/connor/Sites/cm-sirepo/sirepo/package_data/static/en/css/responsive.scss"],
```
robnagler, line 4: It would be good to not have your home directory hardwired into this file. Is there a way to generate it with relative paths?

cwmcallister: I certainly agree! The task runner wasn't in the repo from the original build. I must have copied a bit too much when I was setting it up. I cleaned up the sass files and grunt process, so those paths should be relative now.

PR #2773, opened 2020-08-10
Fix #2766. Put fastcgi sock in configured directory (/tmp default)

### LINKED ISSUE #2766: Do not bind a unix domain socket on a mounted volume in docker
The job_agent tries to create a socket in the current directory inside a container. It should use /var/tmp, which should be a local file system. The directory has to be shared with job_cmd from job_agent so it should be an absolute path, probably configurable to workaround other issues that might arise when deploying.

e-carlin:


sirepo/pkcli/job_agent.py
```diff
@@ -332,11 +333,9 @@ async def _fastcgi_op(self, msg):
         if not self.fastcgi_cmd:
             m = msg.copy()
             m.jobCmd = 'fastcgi'
-            self._fastcgi_file = 'job_cmd_fastcgi.sock'
+            self._fastcgi_file = str(pkio.py_path(cfg.fastcgi_sock_dir).join('job_cmd_fastcgi.sock'))
```
robnagler, line 336: Since this is now in a public directory, it needs to be unique. This won't matter when running inside docker, but it will matter when running locally and running the tests and on sbatch. Something like this:

`cfg.fastcgi_sock_dir.join(f'sirepo_job_cmd-{cfg.agent_id:8}.sock')`

The only place `str()` is needed is on the bind_unix_socket. I shortened the name since we have to give more context. 8 chars of the agent id are enough in this context. Could use something else, but this would help with debugging in local environment.

sirepo/pkcli/job_agent.py
```diff
@@ -54,6 +54,7 @@ def start():
 
     cfg = pkconfig.init(
         agent_id=pkconfig.Required(str, 'id of this agent'),
+        fastcgi_sock_dir=('/var/tmp', str, 'directory where fastcgi socket will be placed'),
```
robnagler, line 57: `pkio.py_path('/var/tmp'), pkio.py_path,`

I think we might actually want to use `/tmp`. We don't want the sockets lying around on reboots of virtual machines. Also, it's shorter, which matters since agent_id is long.

On NERSC, we have a problem because the agent is running natively and job_cmd is in shifter. Please login to Cori and see if /tmp is visible inside and outside the container.

sirepo/pkcli/job_agent.py
```diff
@@ -332,11 +333,9 @@ async def _fastcgi_op(self, msg):
         if not self.fastcgi_cmd:
             m = msg.copy()
             m.jobCmd = 'fastcgi'
-            self._fastcgi_file = 'job_cmd_fastcgi.sock'
+            self._fastcgi_file = str(pkio.py_path(cfg.fastcgi_sock_dir).join('job_cmd_fastcgi.sock'))
             self._fastcgi_msg_q = sirepo.tornado.Queue(1)
             pkio.unchecked_remove(self._fastcgi_file)
-            # Avoid OSError: AF_UNIX path too long (max=100)
```
robnagler, line 338: I would document this actually in the definitely of fastcfg_sock_dir, e.g. `directory of fastcgi socket, must be less than 50 chars` -- just to pick a random number that will be less than base file name.

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2773/commits/56c0225deec58e83d9eb23d034a73a052299b95a

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2773/commits/56c0225deec58e83d9eb23d034a73a052299b95a

I confirmed /tmp is visible inside and outside shifter.

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2773/commits/56c0225deec58e83d9eb23d034a73a052299b95a

PR #2775, opened 2020-08-10
WIP: #2652 marketing urls (ex plans => en/plans)

e-carlin:
@robnagler I took a stab at this. The code needs to be cleaned up but does the general idea seem ok?

### DIFF
--- a/sirepo/uri_router.py
+++ b/sirepo/uri_router.py
@@ -120,9 +120,9 @@ def init(app, simulation_db):
         register_api_module(importlib.import_module('sirepo.' + n))
     for f in pkio.sorted_glob(simulation_db.STATIC_FOLDER.join('en','*.html')):
         @sirepo.api_perm.allow_visitor
-        def x():
-            return sirepo.http_reply.gen_redirect(f'/en/plans.html')
-        _add_api_func(f'api_{f.purebasename}', x)
+        def x(filename=f):
+            return sirepo.http_reply.gen_redirect(f'/en/{filename.basename}')
+        _add_api_func(f'{_FUNC_PREFIX}{f.purebasename}', x)
     _init_uris(app, simulation_db)
 
     sirepo.http_request.init(
--- a/sirepo/uri_router.py
+++ b/sirepo/uri_router.py
@@ -7,12 +7,14 @@ u"""Handles dispatching of uris to server.api_* functions
 from __future__ import absolute_import, division, print_function
 from pykern import pkcollections
 from pykern import pkinspect
+from pykern import pkio
 from pykern.pkdebug import pkdc, pkdexc, pkdlog, pkdp
 import flask
 import importlib
 import inspect
 import re
 import sirepo.api_auth
+import sirepo.api_perm
 import sirepo.auth
 import sirepo.cookie
 import sirepo.http_reply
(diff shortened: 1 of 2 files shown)


sirepo/uri_router.py
```diff
@@ -116,6 +118,11 @@ def init(app, simulation_db):
 
     for n in _REQUIRED_MODULES + tuple(sorted(feature_config.cfg().api_modules)):
         register_api_module(importlib.import_module('sirepo.' + n))
+    for f in pkio.sorted_glob(simulation_db.STATIC_FOLDER.join('en','*.html')):
```
robnagler, line 121: We don't necessarily want these all at the top level. 

This is in an awkward place in the code. It's very detailed code when the rest is fairly highly-level code. I would embed in a method or put in init_uris.

sirepo/uri_router.py
```diff
@@ -184,6 +188,15 @@ def uri_for_api(api_name, params=None, external=True):
     return res
 
 
+def _add_api_func(name, func, module=None):
+    if not name.startswith(_FUNC_PREFIX) or not inspect.isfunction(func):
+        return
+    s = f'function is duplicate: func={name}'
+    assert name not in _api_funcs, \
+        s + ('' if not m else f' module={m}')
```
robnagler, line 196: too complicated. module=None is fine to output.

sirepo/uri_router.py
```diff
@@ -116,6 +118,11 @@ def init(app, simulation_db):
 
     for n in _REQUIRED_MODULES + tuple(sorted(feature_config.cfg().api_modules)):
         register_api_module(importlib.import_module('sirepo.' + n))
+    for f in pkio.sorted_glob(simulation_db.STATIC_FOLDER.join('en','*.html')):
+        @sirepo.api_perm.allow_visitor
```
robnagler, line 122: I don't like this style, because it obscures the names. Right now we can find all apis with grep `^def api_`. It's very useful to be able to do this.

sirepo/uri_router.py
```diff
@@ -116,6 +118,11 @@ def init(app, simulation_db):
...
+        def x(filename=f):
+            return sirepo.http_reply.gen_redirect(f'/en/{filename.basename}')
```
robnagler, line 124: filename should not be a parameter. It should be the lexically scoped value. Otherwise, I don't think this code will do the right thing, because Flask will pass filename. `f'/en/{filename.basename}'` should be evaluated in the enclosing scope.

robnagler:
One other comment. When we know the list of redirects, we can use them to canonicalize URIs. When the redirect target is the result of formatUri, the next step is "does this match a redirect?" If so, then use the redirect source. This way you never render URIs incorrectly. 

In BOP we had SEO optimization prefix support similar to Amazon's `/anything-can-go-here/db/XYZZY`, which was table driven by the customer. This lookaside was handled seamlessly by format_uri and the router. Amazon's use is a bit more chaotic, because they have a bazaar-style uri rendering.

e-carlin:
Thanks for the review. I don't know what you mean by "redirect map" so lets talk in person when you can.

e-carlin:
Discussed with @robnagler about using a different approach. Closing and will open another PR with the discussed pattern.

PR #2784, opened 2020-08-11
Fix #2762 add use_reloader, hide_guest_warning in dev

### LINKED ISSUE #2762: radia-run.sh use dev channel
~~Allow override on the command line (docker -e). Also, maybe use dev by default when running sirepo.~~

Need to use `dev` channel, because `sirepo service http` asserts dev among other things. Original problem was to use alpha to avoid use of the reloader so just allow the reloader to be configured. A few other pieces of code need to be configured too.

robnagler:
generate random server secret
set use_reloader to false in radia-run.sh
take out unnecessary vars
reuse code in feature_config to make more readable

### DIFF
--- a/container-conf/radia-run.sh
+++ b/container-conf/radia-run.sh
@@ -1,17 +1,22 @@
 #!/bin/bash
 #
-# Start sirepo
+# Docker CMD to start Sirepo in development mode.
 #
-. ~/.bashrc
+# DO NOT USE on a public server or in an environment with multiple users.
+#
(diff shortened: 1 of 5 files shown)


robnagler:
not ready

robnagler:
@moellep Now it's ready. I think this should be merged, but I did more than absolutely necessary, of course. :) I don't think there's a high risk, but you review and decide when to merge (before or after 8/19/20)

PR #2786, opened 2020-08-12
Fix #2652. Top-level marketing urls.

### LINKED ISSUE #2652: Simple top-level URLs for marketing materials

e-carlin:
@robnagler one possible issue: the URLS are formatted like v.radia.run:8000/*elegant in the address bar. Is this what you want or should the * be removed?

### DIFF
--- a/sirepo/server.py
+++ b/sirepo/server.py
@@ -390,7 +390,7 @@ def api_robotsTxt():
         # We include dev so we can test
         if pkconfig.channel_in('prod', 'dev'):
             u = [
-                sirepo.uri.api('root', params={'simulation_type': x})
+                sirepo.uri.api('root', params={'path_info': x})
                 for x in sorted(feature_config.cfg().sim_types)
             ]
         else:
@@ -402,15 +402,14 @@ def api_robotsTxt():
 
 
 @api_perm.allow_visitor
-def api_root(simulation_type):
+def api_root(path_info):
     try:
-        req = http_request.parse_params(type=simulation_type)
+        req = http_request.parse_params(type=path_info)
     except AssertionError:
-        if simulation_type == 'warp':
-            return http_reply.gen_redirect(sirepo.uri.app_root('warppba'))
-        if simulation_type == 'fete':
-            return http_reply.gen_redirect(sirepo.uri.app_root('warpvnd'))
-        sirepo.util.raise_not_found('Invalid simulation_type={}', simulation_type)
+        u = sirepo.uri.root_redirect(path_info)
+        if u:
+            return http_reply.gen_redirect(u)
+        sirepo.util.raise_not_found(f'uknown path={path_info}')
     return _render_root_page('index', PKDict(app_name=req.type))
 
 
--- a/sirepo/server.py
+++ b/sirepo/server.py
@@ -403,14 +403,12 @@ def api_robotsTxt():
 
 @api_perm.allow_visitor
 def api_root(path_info):
-    try:
-        req = http_request.parse_params(type=path_info)
-    except AssertionError:
-        u = sirepo.uri.root_redirect(path_info)
-        if u:
-            return http_reply.gen_redirect(u)
-        sirepo.util.raise_not_found(f'uknown path={path_info}')
-    return _render_root_page('index', PKDict(app_name=req.type))
+    if sirepo.template.is_sim_type(path_info):
+        return _render_root_page('index', PKDict(app_name=path_info))
+    u = sirepo.uri.unchecked_root_redirect(path_info)
+    if u:
+        return http_reply.gen_redirect(u)
+    sirepo.util.raise_not_found(f'uknown path={path_info}')
 
 
 @api_perm.require_user
--- a/sirepo/uri.py
+++ b/sirepo/uri.py
@@ -21,6 +21,12 @@ except ImportError:
 #: allows us to search more easily for use of http root
 ROOT = '/'
 
+#: route parsing
+PARAM_RE = r'([\?\*]?)<{}>'
+
+#: optional parameter that consumes rest of parameters
+PATH_INFO_CHAR = '*'
+
 
 def api(*args, **kwargs):
     """Alias for `uri_router.uri_for_api`"""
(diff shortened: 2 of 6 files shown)


sirepo/uri_router.py
```diff
@@ -212,7 +212,7 @@ def _dispatch(path):
                     raise sirepo.util.raise_not_found('{}: uri missing parameter ({})', path, p.name)
                 break
             if p.is_path_info:
-                kwargs[p.name] = '/'.join(parts)
+                kwargs[p.name] = '/'.join([x.lstrip(_PATH_INFO_CHAR) for x in parts])
```
robnagler, line 215: I don't understand this change. Why would parts contain _PATH_INFO_CHAR?

sirepo/uri_router.py
```diff
@@ -257,6 +257,10 @@ def _init_uris(app, simulation_db):
     assert _default_route, \
         'missing default route'
     _empty_route = _uri_to_route.en
+    i = set(_uri_to_route.keys()).union(sim_types).intersection(
```
robnagler, line 260: Nice. Need the intersection of _uri_to_route with sim_types, just as a safety feature so this might be easiest:

```py
u = set(_uri_to_route.keys())
r = simulation_db.SCHEMA_COMMON.rootRedirectUri.keys()
t = sim_types
i = u & r | u & t | r & t
assert not i,rootRedirectUri, sim_types, and routes have overlapping uris={i}'
```

sirepo/uri.py
```diff
@@ -80,6 +81,10 @@ def local_route(sim_type, route_name=None, params=None, query=None):
     return app_root(t) + '#' + u + _query(query)
 
 
+def root_redirect(path):
+    return simulation_db.SCHEMA_COMMON.rootRedirectUri.get(path)
```
robnagler, line 85: Document what this is, since other methods are documented. I would call the arg "uri", not "path". "path" implies file system value. "uri" in this case is a part of a uri.

sirepo/uri_router.py
```diff
@@ -116,7 +116,7 @@ def init(app, simulation_db):
 
     for n in _REQUIRED_MODULES + tuple(sorted(feature_config.cfg().api_modules)):
         register_api_module(importlib.import_module('sirepo.' + n))
-    _init_uris(app, simulation_db)
+    _init_uris(app, simulation_db, feature_config.cfg().sim_types)
```
robnagler, line 119: no need to complicate this interface by passing sim_types. Rather either pass feature_config or nothing and have _init_uris import feature_config, which can be imported more easily than simulation_db, which has import loop issues.

sirepo/uri_router.py
```diff
@@ -257,6 +257,10 @@ def _init_uris(app, simulation_db):
     assert _default_route, \
         'missing default route'
     _empty_route = _uri_to_route.en
+    i = set(_uri_to_route.keys()).union(sim_types).intersection(
+        simulation_db.SCHEMA_COMMON.rootRedirectUri.keys(),
+    )
+    assert not i, f'duplicate uris={i}'
```
robnagler, line 263: I would add one more assertion,

```py
for x in r:
    assert re.search(r'^[a-z]+$'), \
        f'rootRedirectUri={x} must consist of letters only'
```

I think it is good to be restrictive right now, because the purpose is top level marketing links. By restricting, we ensure the namespace is narrow. The act of widening it should be conscious.
 
I would put this in a separate function just because the current function is getting long.

I would

sirepo/server.py
```diff
@@ -402,15 +402,14 @@ def api_robotsTxt():
 
 
 @api_perm.allow_visitor
-def api_root(simulation_type):
+def api_root(path_info):
     try:
-        req = http_request.parse_params(type=simulation_type)
+        req = http_request.parse_params(type=path_info)
```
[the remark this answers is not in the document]
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/server.py
```diff
@@ -402,15 +402,14 @@ def api_robotsTxt():
...
     except AssertionError:
-        if simulation_type == 'warp':
-            return http_reply.gen_redirect(sirepo.uri.app_root('warppba'))
-        if simulation_type == 'fete':
-            return http_reply.gen_redirect(sirepo.uri.app_root('warpvnd'))
-        sirepo.util.raise_not_found('Invalid simulation_type={}', simulation_type)
+        u = sirepo.uri.root_redirect(path_info)
```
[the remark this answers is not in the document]
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/uri_router.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/uri.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/uri_router.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/uri_router.py
e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/1912882dc322db1f33962e87148aeddc9c6ef2d7

sirepo/uri_router.py
e-carlin: @robnagler the [GUI doesn't strip the _PATH_INFO_CHAR](https://github.com/radiasoft/sirepo/blob/1912882dc322db1f33962e87148aeddc9c6ef2d7/sirepo/package_data/static/js/sirepo.js#L1821) from the url when it does a redirect. So it is sent to the server. It needs to be stripped so we can find the simulation type (ex *elegant -> elegant). Should the GUI strip the char before sending? This will also solve my comment above:
>@robnagler one possible issue: the URLS are formatted like v.radia.run:8000/*elegant in the address bar. Is this what you want or should the * be removed?

robnagler: Yes, that's a bug in the GUI.

e-carlin: https://github.com/radiasoft/sirepo/pull/2786/commits/d8ea1bab46d53e05ed40c59d87e3acd76912f310

PR #2799, opened 2020-08-18
fix #2792 fix #2793 centralize attachment download in sirepo.http_reply

### LINKED ISSUE #2792: should api_simulationData return safe_attachment when pretty?
This seems like a bug. Either it returns the value or it shouldn't be called:

```py
        if pretty:
            _safe_attachment(
                resp,
                d.models.simulation.name,
                'json',
            )
```

### LINKED ISSUE #2793: _safe_attachment should be implicit in gen_file_as_attachment
This should be moved into http_reply. If the attachment filename has to get fixed up, gen_file_as_attachment should do that:

```py
def _safe_attachment(resp, base, suffix):
    return http_reply.as_attachment(
        resp,
        http_reply.MIME_TYPE[suffix],
        '{}.{}'.format(
            re.sub(r'[^\w]+', '-', base).strip('-') or 'download',
            suffix,
        ).lower(),
    )
```

moellep:


sirepo/server.py
```diff
@@ -427,34 +428,24 @@ def api_saveSimulationData():
...
 
 @api_perm.require_user
-def api_simulationData(simulation_type, simulation_id, pretty, section=None):
+def api_simulationData(simulation_type, simulation_id, pretty=False, section=None):
     """First entry point for a simulation
 
     Might be non-session simulation copy (see `simulation_db.CopyRedirect`).
     We have to allow a non-user to get data.
     """
+    #TODO(pjm): pretty is an unused argument
```
robnagler, line 441: Do we ever want pretty in this case? If not, we should create another issue that removes pretty and fixes tests.

moellep: Several APIs have unused args at this point, downloadFile, listFiles and simulationData. Removing arguments from the API in the past caused really bad behavior like client loops in the past. Removing the args needs to be done carefully, or the API should get renamed at the same time.

PR #2800, opened 2020-08-18
Fix #2787. Add knn classifier to ml app.

### LINKED ISSUE #2787: ML KNN classifier
Impelement the KNN classifier from [this notebook](https://github.com/radiasoft/rscon/blob/master/webcon/sirepo_classifier_notebooks/sirepo_classifiers_singleoutput.ipynb)  as a new "classifier" tab on the ml app.

e-carlin:
This PR uses ES6. The test fail right now but will pass once #2745 is merged.

Classification vs regression are two diffferent prediction strategies
that depend on the type of data. Classification is picking some labels
(ex has diabetes or not). Regression is calculating some float (ex
median home price). So, the user must now select whether they want
to do classification or regression and then the correct tab will
be displayed.


moellep:
When I create a new simulation and use the boston-housing.csv data set, and select the Classification application mode, the Classification tab initially shows several empty panels "KNN Error Rate", "KNN Classification Metrics" and "KNN Confusion Matrix". Those should be hidden unless there are output frames available.

When I select "Start New SImulation" on the Classification tab, it shows the error "Unknown label type: 'continuous'.

e-carlin:
I'll fix the output frames.

The Boston Housing data can't be used with classification. The output is a float (median housing price) not a classification (has diabetes or not). If you change appmode to 'regression' then the classification tab will disappear. The user must select one or the other and they should know what type of prediction is relevant for their data.

moellep:
Could you add some data checking at the start of the simulation and have an assertion or similar failure if it looks like the selected output column is not appropriate? Just a hint to the user rather than the eventual "unknown label type" error.

e-carlin:
> Could you add some data checking at the start of the simulation and have an assertion or similar failure if it looks like the selected output column is not appropriate? Just a hint to the user rather than the eventual "unknown label type" error.

https://github.com/radiasoft/sirepo/pull/2800/commits/917e7c66e43fb53bf9f2959b9bcdbb32ff2797d5 thoughts? A bit hacky to get the right message back to them.

PR #2808, opened 2020-08-19
Fix #2806 install karma

### LINKED ISSUE #2806: karma tests should always be run on dev
In test.sh it has the line in test_main():
```
    if [[ -x ./node_modules/karma/bin/karma ]]; then
        npm run test
    fi
```
If someone doesn't have karma installed on dev the javascript doesn't get tested and you wouldn't know it. If possible test.sh should know it is on dev as always run the karma tests.

robnagler:
requires chromeheadless so need latest beamsim


e-carlin:
This looks good. I think in the future we may want to refactor the test_jshint too. It installs jshint by hand which is unnecessary with npm install. Also when would node not be installed?

PR #2809, opened 2020-08-19
Fix #2779. Remove fastcgi sock file in terminate.

### LINKED ISSUE #2779: Remove _fastcgi_file in terminate
Since the socket is in a public name space (/tmp), we should remove in terminate.

e-carlin:
The file is now in a public namespace (/tmp) so we need to cleanup
after ourselves.

### DIFF
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -333,18 +333,18 @@ class _Dispatcher(PKDict):
         if not self.fastcgi_cmd:
             m = msg.copy()
             m.jobCmd = 'fastcgi'
-            self._fastcgi_file = cfg.fastcgi_sock_dir.join(
+            self.fastcgi_file = cfg.fastcgi_sock_dir.join(
                 f'sirepo_job_cmd-{cfg.agent_id:8}.sock',
             )
             self._fastcgi_msg_q = sirepo.tornado.Queue(1)
-            pkio.unchecked_remove(self._fastcgi_file)
-            m.fastcgiFile = self._fastcgi_file
+            pkio.unchecked_remove(self.fastcgi_file)
+            m.fastcgiFile = self.fastcgi_file
             # Runs in a agent's directory, but chdir's to real runDirs
             m.runDir = pkio.py_path()
             # Kind of backwards, but it makes sense since we need to listen
             # so _do_fastcgi can connect
             self._fastcgi_remove_handler = tornado.netutil.add_accept_handler(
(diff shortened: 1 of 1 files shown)


sirepo/pkcli/job_agent.py
```diff
@@ -536,6 +536,8 @@ def _create_in_file(self):
 
 class _FastCgiCmd(_Cmd):
     def destroy(self):
+        pkio.unchecked_remove(self.dispatcher.fastcgi_file)
```
robnagler, line 539: This function should be in the dispatcher, e.g. self.fastcgi_destroy.

Also the file may be none, which I don't think is supported by unchecked_remove so check to see if it is not none.

e-carlin: https://github.com/radiasoft/sirepo/pull/2809/commits/e0b28e875b68a055a197fc4c738bd3f3107fe01e

sirepo/pkcli/job_agent.py
```diff
@@ -137,6 +137,11 @@ def __init__(self):
             fastcgi_error_count=0,
         )
 
+    def fastcgi_destroy(self):
+        self._fastcgi_file and pkio.unchecked_remove(self._fastcgi_file)
```
robnagler, line 141: This is an interesting question. I'll let it go. I personally don't mind the style, but would be curious what @moellep and @mkeilman (heebee geebees?)

e-carlin: I've [used it elsewhere](https://github.com/radiasoft/sirepo/blob/eeba8cfb8743fefaf31034b24f6ff80ad4e2035e/sirepo/job_supervisor.py#L923). I don't have a strong feeling towards it. I like fewer if's but maybe this ins't worth it. In the linked example I think it could be confusing because one might read it and think self.op_slot is a bool when in fact it is the slot_proxy object. This one we don't care about the "return" value so maybe it matters less here.

moellep: Personally, I prefer the "if" and in python it is just an extra line:
```python
if self._fastcgi_file:
```
In perl it was annoying to write:
```perl
if ($my_condition) {
    do_something();
}
```
and instead you could write:
```perl
do_something()
    if $my_condition;
```
which I didn't like because it is backwards, but at least it didn't require the curly brackets.

e-carlin: That makes sense. I'm in agreement. Unless @mkeilman or @robnagler have strong thoughts if's from now on.

mkeilman: Guess I'm an "if" guy

PR #2829, opened 2020-08-20
Fix #2827. Strip * from url without optional params

### LINKED ISSUE #2827: Login not working due to * in url

e-carlin:
@robnagler in https://github.com/radiasoft/sirepo/commit/d8ea1bab46d53e05ed40c59d87e3acd76912f310#r41568979 we talked about changing the regex to strip '*' (and '?') from urls. That regex doesn't work. The url coming for login before any manipulation looks like /*elegant. So I propose this.


robnagler:
This shouldn't have been merged.

PR #2834, opened 2020-08-20
fix #331 formatUrl parses params properly

### LINKED ISSUE #331: Normalize formatUrl templating
The local map uses ':foo' and the global map uses '<foo>'. The caller for formatUrl needs to know which map its using. 

Rather, caller should specify 'foo' as a keyword, and the formatter will take care of putting in the prefix/postfix chars, e.g. 

``` javascript
formatUrl('simulationData', {
    simulationId: simulationId(),
    simulationType: SIREPO.APP_SCHEMA.simulationType,
})
```

The route template should also check for the existence of the specified keyword. In this sense, the `<>` syntax is better, because it avoids false positives, e.g. 'simulation' would not match 'simulationId'.

robnagler:
generate two maps like in uri_router
formatUrl uses maps to render uri properly

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1673,7 +1673,7 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
     return self;
 });
 
-SIREPO.app.factory('requestSender', function(cookieService, errorService, localRoutes, $http, $location, $interval, $q, $rootScope, $window) {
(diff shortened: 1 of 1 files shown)


e-carlin:
Using a new user logging as guest I get a 404 with the route `http://v.radia.run:8000/*%3Cpath_info%3E`

e-carlin:
Clicking on the elegant visualization tab I don't go to the new page and I see `Error: param=simulationId param missing map=local route=lattice` in the console.

robnagler:
> Using a new user logging as guest I get a 404 with the route `http://v.radia.run:8000/*<path_info>`

https://github.com/radiasoft/sirepo/issues/2835

Issue #2624, opened 2020-06-26
Create new API for runSimulation/runAnalysis


robnagler:
We need to separate compute from analysis so that a single compute result can be analyzed multiple times. This has done in the past with tricks, e.g. bunchReport[1234] runs the computeJob bunchReport but there is no model bunchReport. There are several problems:

* Analysis often creates a static file in the computeJob run_dir. That file name (not the contents) is reused across all analyses. This is a synchronization and caching problem.

* If an analysis takes a long time, it probably should have a button, but there is no direct API to run an analysis, just a simulation.

* An analysis should validate the in.json to make sure it's in agreement, but there is no analysisJobHash or analysisJobSerial. They just reuse the computeJob values.

See https://github.com/radiasoft/sirepo/pull/2611 for more discussion.

We can create three new APIs: 

* computeSim - starts a job for a computeModel if not already run or always run if forceRun is true. Similar to runSimulation but does not implicitly call sequential_result. Will return the output of checkJob ("pending", "error", etc.).

* analyzeSim - runs an analysisModel if not already run or always run if forceRun is true. The computeModel is mapped server-side. Will return the result 

* checkJob - returns status of a computeJob or analysisJob similar to runStatus. In the case of an analysisJob, also returns the data if available.

Job identifiers have to be idempotent and unique to compute or analysis so we may need to expand them or change the name of the underlying computeModels. Need to think about this.

e-carlin:
Thanks for writing this up.

What is the difference between checkJob and runStatus? Just a separate function for the transition period where we have the old and new interface both running? Will it return the data in the case of a parallel compute or just sequential (runStatus has this distinction currently)?

While thinking about this topic do we want to consider any changes to simulationFrame (ex getting rid of it and using analyzeSim isntead)? I know it is a separate get api for caching purposes. If we are going to create a hash for analysis model we are implementing some form of caching ourselves. I'm not sure why caching was implemented in the first place (analysis comptue was slow, transmission of result was slow, both) but maybe that can be reconsidered while doing this work?

robnagler:
> What is the difference between checkJob and runStatus? 

runStatus translates the report/model to computeModel/Job and then will run an analysis on report/model if computeJob is valid. This is too many steps. The app would now explicitly ask for computeJob or analysisJob status. It's cleaner in that you know you'll get back a result from analysisJob but not from computeJob. 

[...]

PR #2876, opened 2020-08-28
Fix #2872. Add decision tree classifer.

### LINKED ISSUE #2872: Implement decision tree classifier in ml app

e-carlin:


sirepo/template/ml.py
```diff
@@ -32,23 +36,36 @@
...
         percentComplete=0,
         frameCount=0,
     )
     if report == 'classificationAnimation' and not is_running:
         return PKDict(
-            percentComplete=100,
+            # TODO(e-carlin): is this the best I can do? Existing sims will have
+            # to be re-run because this is a cached value in supervisor db and
+            # won't be present in old sims
+            framesForClassifier=_classifier_from_files(),
```
e-carlin, line 60: @moellep what do you think of this? I've added it because when a user has reports shown of the classifier page and then changes the classifier the reports were being retrieved for the new classifier and coming back with `report not generated` (because the output files are different based on the classifier selected). With this change users will have to re-run their simulations before being able to see any reports (framesForClassifier is "cached" in the supervisor db and prior to this commit it isn't present so the ng-if in the template will be false for both classifiers). Can you think of a better way to solve this problem?

e-carlin: To be more specific, users who had existing output prior to this change will have to run their simulations again. After this change they won't. Knn takes just a few seconds to run (with the datasets I've used it can probably take a while with large datasets) so I figure it isn't that big of deal but lmk.

sirepo/template/template_common.py
```diff
@@ -346,8 +346,8 @@ def sim_frame_dispatch(frame_args):
         ),
     )
     t = sirepo.template.import_module(frame_args.simulationType)
-    o = getattr(t, 'sim_frame', None) \
-        or getattr(t, 'sim_frame_' + frame_args.frameReport)
+    o = getattr(t, 'sim_frame_' + frame_args.frameReport, None) \
```
e-carlin, line 349: @moellep I want to call you attention to this. I changed the logic (try to use sim_frame_* before using generic sim_frame). The only existing code it should impact is Opal (that was the only code with sim_frame and sim_frame_*) which I changed. I tested all Opal examples and it seems fine but double check my logic if you don't mind.

sirepo/template/ml.py
robnagler: Do we need to cached this in the job structure? Wouldn't be better to store it on disk and just request it as a report?

e-carlin: To do that we would need to request this report then use it's output to decide if we should request the other reports. I don't immediately now how I would do that (or if we do that anywhere else). Are you aware of any other places where we do that right now? My understanding of the reports is that they are somewhat isolated from the rest of the code and there aren't good hooks to use their results for further business logic (but my understanding of the GUI is limited). Obviously that could change if we need it to.

Thinking about this more we could just change the logic in the html template, if we have frames but no framesForClassifier then assume knn since that is the only classifier implemented previously. That would get around the problem of users having to run again.

I think you are getting at a larger issue of putting "random" information in background_percent_complete as a way to communicate with the GUI. Maybe we could think about getting rid of the pattern all together...

robnagler: #2624 addresses the problem. I think there is a problem with putting random information in background_percent_complete: it's a hack. The hack then gets stored in the supervisor. It's bad style and confusing.

e-carlin: I agree. How should we proceed? Should I try to implement #2624?

robnagler: We're in a bit of a trap. #2624 is an idea. Getting there is going to be tricky. I don't have a particular answer. How big is the data _classifier_from_files?

e-carlin: It is very small. Just the name of the classifier ("dt", "knn", "svn", etc). It gets used [in the html template](https://github.com/radiasoft/sirepo/pull/2876/files#diff-3351e056502223a9461e52dd70e439bbR13) to decide which reports to request/display.

robnagler: Just leave it then. Not solvable right now

moellep: You have the selected classifier already in the in.json which is read at the beginning of background_percent_complete() to get the state when the simulation was started. No need to check for output files  using _classifier_from_files(). I checked in a fix to the branch for this.

Issue #2851, opened 2020-08-25
Objects do not always appear in 2d view after browser refresh


mkeilman:
Zooming etc. in the view causes them to appear

PR #2884, opened 2020-09-01
Fix #2797 adding objects to new group leaves copies in the ungrouped section

### LINKED ISSUE #2797: Adding item to group while creating group leaves a duplicate ungrouped item
If you save the group first and then open it to add items, it behaves properly

### LINKED ISSUE #2798: Saving transform when creating object closes all dialogs and fails to create object
Related to #2715 I imagine

mkeilman:
This also fixes #2798, #2850, and #2851.  One note: the fix to #2798 involves closing a modal dialog only when the "cancel" message comes with the matching model name, similar to handling the "modelChanged" message; currently the modal closes regardless.  There may be side effects but I have not seen any so far.


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -1213,11 +1213,16 @@ SIREPO.app.directive('modalEditor', function(appState, panelState) {
                 $scope.modalTitle = viewInfo.title;
             }
             $scope.$on('modelChanged', function (e, name) {
-                if (name == $scope.modelKey) {
+                if (name == $scope.modelName) {
```
moellep, line 1216: This change breaks places which have multiple copies of the same model (same modelName but unique modelKeys). For example, in elegant on the source panel, select a report, change a value and select "Save Changes". The modal dialog doesn't close in that case. The same problem exists in SRW on the beamline watchpoint reports.

mkeilman: 4a52fad

I can't remember why I changed it or even that I changed it!

PR #2910, opened 2020-09-03
Fix #2902. Add blues color map and make default for ml app.

### LINKED ISSUE #2902: add "blues" colormap and use as default map for ml app
Create a new color string in sirepo-plotting using the values from matplotlib:

https://github.com/matplotlib/matplotlib/blob/d492a658c4d1d270cd132a74f49c61caec58ace3/lib/matplotlib/_cm.py#L507

e-carlin:


robnagler:
I'm concerned that we picked a color blue without forming a committee to establish the hue and saturation for effective machine learning.

e-carlin:
I propose the name "blue man group".

PR #2918, opened 2020-09-04
Fix #2865 add write_files to sirepo.lib

### LINKED ISSUE #2865: decouple export_native/text from generate_parameters_file
@cchall wants to be able to get native formats from `data`.

Related to #2796. I think we want to consider encapsulating in an object.

### LINKED ISSUE #2914: many deprecated unescaped re.search() calls
Many of the re.search(), re.match() and re.sub() calls in sirepo/template aren't escaping the expression with r'' which results in deprecated warnings.

robnagler:
- Fix #2914 template_common, zgoubi, jspec, synergia all had invalid escapes
- refactored template.elegant to use nested functions for better encapsulation.  Also use class for generate_paramaters_file which is reused for sirepo.lib. Significant refactoring that makes it easier to find which functions are used where.
- animation_test includes bunchComp - four Dipole CSR, because it tests background_percent_complete. This slowed down animation test, but we need more paths tested.
- elegant_import_test calls sirepo.lib for full simulation imports (.ele); also marked deviance tests more clearly.


moellep:
Running the twiss report on the Lattice tab for the Backtracking elegant example fails with an error message "27-1".

robnagler:
i'll take a look

robnagler:
The filename_map has to be built much later.

robnagler:
_setup_backtracking makes a deepcopy of data while the other parts of the code modify data in place. Is there a reason for this? It's opaque when `data` can be modified globally or not.

The lazy evaluation of util and filename_map are implemented as properties. Maybe could have been functions, but would have had to have been stored so easier to do it this way. LMK if this makes sense or not.

PR #2920, opened 2020-09-07
Fix #2912. Correct confusion_matrix heatmap labels

### LINKED ISSUE #2912: ML wrong label position on heatmap
It looks like the confusion matrix heatmap labels are not quite right - the plot below shows the "1" value with the "0" color and one of the "0" values with a "1" color.

Screen Shot 2020-09-03 at 4 51 08 PM

e-carlin:
The x dimension on the heatmap corresponds to the i (row) in the
confusion_matrix output. The y corresponds to the j (column).

### DIFF
--- a/sirepo/template/ml.py
+++ b/sirepo/template/ml.py
@@ -273,9 +273,9 @@ def _compute_numpy_info(path):
 def _confusion_matrix_to_heatmap_report(frame_args, filename, title):
     r = pkjson.load_any(frame_args.run_dir.join(filename))
     a = None
-    for i, _ in enumerate(r.matrix):
-        for j, v in enumerate(r.matrix[i]):
-            t = np.repeat([[i, j]], v, axis=0)
+    for y, _ in enumerate(r.matrix):
+        for x, v in enumerate(r.matrix[y]):
+            t = np.repeat([[x, y]], v, axis=0)
             a = t if a is None else np.vstack([t, a])
     return template_common.heatmap(
         a,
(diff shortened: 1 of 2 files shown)


sirepo/template/ml.py
```diff
@@ -273,9 +273,9 @@ def _compute_numpy_info(path):
 def _confusion_matrix_to_heatmap_report(frame_args, filename, title):
     r = pkjson.load_any(frame_args.run_dir.join(filename))
     a = None
-    for i, _ in enumerate(r.matrix):
-        for j, v in enumerate(r.matrix[i]):
-            t = np.repeat([[i, j]], v, axis=0)
+    for y, _ in enumerate(r.matrix):
```
e-carlin, line 276: @moellep this isn't totally necessary. template_common.heatmap() needs the points to create the output. So, we construct them from the confusion_matrix output (r.matrix). But what it produces for the z_matrix field is the same as r.matrix. So this could be written like:
```python
r = pkjson.load_any(frame_args.run_dir.join(filename))
# OPTIMIZATION: Build the heatmap data sctructure by hand instead constructing 
# the points and then calling template_common.heatmap()
a = [0, len(r.matrix) - 1, len(r.matrix)]
return PKDict(
    x_range=a,
    y_range=a,
    z_matrix=r.matrix,
    labels=r.labels,
    title=title.format(**r),
    x_label='Predicted',
    y_label='True',
)
```

This is gets rid of the need to build the points to pass to template_common but spreads around the knowledge of what fields are needed for heatmaps.

moellep: Yes, it doesn't really need to call template_common.heatmap() because it is already has the correct cells created, so that would be fine. What you have already is fine too.

PR #2973, opened 2020-09-22
Fix #2946 convert floats, ints, and rpnvalues to native

### LINKED ISSUE #2946: sirepo.lib.SimData should convert to native values
See https://github.com/radiasoft/rsopt/issues/20#issuecomment-692231409

robnagler:


sirepo/template/elegant.py
```diff
@@ -166,6 +166,58 @@ def _lattice_filename(self, value):
...
+            def _rpn(rpn_type, py_type):
+                nonlocal t
+                if (
+                    t != rpn_type or
+                    not code_variable.CodeVar.is_var_value(v)
+                ):
+                    return False
+                model[k] = c.eval_var_with_assert(v)
+                return True
+
```
moellep, line 189: The embedded functions make the logic hard to follow and requires hunting for the source values like "v". I much prefer passing values to the function. Two levels of embedded functions makes this worse.

robnagler: I removed the `_rpn()` because it was only used in the `RPNValue` case. This eliminated the second nesting.  

I think `nonlocal` wasn't working right so after making that change, values of `RPNValue` were getting converted to floats, and this lead to some interesting issues like defaults don't match. Also, clearly integer values like `bins` are now floats.

Issue #2975, opened 2020-09-23
Moving jupyter.radiasoft.org to sirepo.com/jupyter


e-carlin:
Outstanding issues to be completed before jupyterhub is moved under sirepo.com

- [ ] Pattern for user names. @robnagler sez: 
    - We can try the local part of the email. If that fails, we throw the second level domain with an underscore. If that fails, just add a random number.  Users can't rename themselves so this shouldn't be a problem. With merging existing users, this should be ok, because there will be a collision at the time, and we'll follow the same algorithm.
- [ ] https://github.com/radiasoft/sirepo/issues/2979 merge users
- [ ] https://github.com/radiasoft/rsconf/issues/124
- [ ] handling jupyterhub logout. Currently we just clear all cookies (see https://github.com/radiasoft/sirepo/commit/4eb1cad16b3e89c2235f213a85dcb5983034192c#r42662986)
- [x] allowing guest users? @robnagler says no, because we don't have a name for them. This may require another api_perm or just ignore the issue, since we won't have guests.

robnagler:
Answered some questions above

PR #2976, opened 2020-09-23
Add jupyterhub service

e-carlin:
This is the beginning of moving jupyter.radiasoft.org to sirepo.com/jupyter. Outstanding issues that must be completed before ready for deployment will be kept track of here https://github.com/radiasoft/sirepo/issues/2975.


e-carlin:
Closing until all of jupyterhub migration is completed.

PR #3008, opened 2020-10-02
Fix #2990. Show display name with adm jobs

### LINKED ISSUE #2990: adm-jobs should show display name
The jobs list should do a better job of identifying the user. This may only be a problem in GitHub. At least want the display_name and possibly github handle or email.

e-carlin:
@robnagler now showing UserRegistration.display_name. Lmk if you want more/other names.

### DIFF
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -493,6 +493,7 @@ class _ComputeJob(PKDict):
                 h.insert(l, ['Name', 'String'])
             else:
                 h.insert(l, ['User id', 'String'])
+                h.insert(l, ['Display name', 'String'])
                 h.extend([
                     ['Queued', 'Time'],
                     ['Driver details', 'String'],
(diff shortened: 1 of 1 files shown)


robnagler:
what about email?

e-carlin:
Can you be a bit more specific? Do you want email in all cases? Just cases where the uid has a record in the AuthEmailUser table? Do you want an abstraction that get's the most relevant "name" (ex email, github handle, etc).

robnagler:
We are adding a "user name" function to auth that allows you to get it for any auth mode. it should return the uid if there is no user_name and return the user_name if there is a user_name. display_name shows up in the user_regristartion record. i think we display that as well. i would actually put all the values in one column (just concat with newline), since the list is relatively short. the idea is that you can see as much information as we have about the person to contact them about a runaway job or whatever.

e-carlin:
Let's talk about this over the phone. Some problems:
- We don't have access to the cookies in the supervisor so our current auth.user_name method won't work.
- sirepo.auth is not initialized so we don't know what methods are available.

e-carlin:
Closing for now. Getting access to the users/cookies in the supervisor is more work than it is worth. Someday Flask will go away and we can revisit.

PR #3018, opened 2020-10-09
Fix #3017. Use other_sim_types to enable jupyterhublogin and simplify…

### LINKED ISSUE #3017: rename proprietary_codes to other_codes

e-carlin:
… etc/run-jupyterhub

### DIFF
--- a/sirepo/pkcli/service.py
+++ b/sirepo/pkcli/service.py
@@ -101,8 +101,7 @@ def jupyterhub():
     import sirepo.feature_config
 
     assert pkconfig.channel_in('dev')
-    assert 'jupyterhublogin' in sirepo.feature_config.cfg().sim_types, \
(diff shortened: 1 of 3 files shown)


sirepo/pkcli/service.py
```diff
@@ -101,8 +101,7 @@ def jupyterhub():
     import sirepo.feature_config
 
     assert pkconfig.channel_in('dev')
-    assert 'jupyterhublogin' in sirepo.feature_config.cfg().sim_types, \
-        'you must set SIREPO_FEATURE_CONFIG_SIM_TYPES to contain the code "jupyterhublogin"'
+    assert 'jupyterhublogin' in sirepo.feature_config.cfg().other_sim_types
```
robnagler, line 104: I don't think this should check other_sim_types. It should use template.assert_sim_type.

e-carlin: 2b96cea4

PR #3019, opened 2020-10-09
Fix #3016. Create new user jupyterhub dir when rs_jupyter_migrate not…

### LINKED ISSUE #3016: Jupyterhub create user dir when github not enabled

e-carlin:
… enabled


sirepo/sim_api/jupyterhublogin.py
```diff
@@ -49,6 +49,8 @@ def api_migrateJupyterhub():
 def api_redirectJupyterHub():
     is_new_user = _create_user()
```
robnagler, line 50: _create_user doesn't always create the user always. I think we need to document that in the name. _hub_user does something similar. I think it's important to know these things when creating the code.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -49,6 +49,8 @@ def api_migrateJupyterhub():
...
     if not cfg.rs_jupyter_migrate or not is_new_user:
+        if is_new_user:
+            pkio.mkdir_parent(_user_dir())
```
robnagler, line 53: seems like this code should be in _create_user, since it is creating the user here. It's confusing to see so many conditions here. Rather test for migration in create_user.

robnagler (changes requested):

sirepo/sim_api/jupyterhublogin.py
e-carlin: 59f2d698 . I realized there was another case I didn't cover where the user dir wasn't created (migration enabled but the user says they don't want to migrate). I refactored so the user dir is always create. In the event of a migration the dir will be overwritten with the dir from the migration.

sirepo/sim_api/jupyterhublogin.py
e-carlin: 59f2d698 renamed fn.

_hub_user doesn't create anything so I don't see the similarity. Maybe _hub_user should be renamed to _unchecked_hub_user?

robnagler: _hub_user is conditional. Think of "open", if it fails, it throws an exception. In general, we model on exceptions in Python, not Go-style return values. The general idea is a function name is a statement, e.g. [jupyterhub_user_name](https://github.com/radiasoft/sirepo/blob/774129e73acda22a46f0313b6c66db3c31448f2e/sirepo/sim_api/jupyterhublogin.py#L58) (also needs to be fixed), which gives you something. Clearly, this is broken many times, e.g. `getattr` raises and `dict.get` does not, but of course `os.remove` raises, which it shouldn't. :) 

Be that as it may, I think we have to make a choice about naming. In BOP we used `unsafe_` as a prefix for such things, because we were very fail-fast focused, and we felt that not fail-fast was unsafe. Python uses `check` in a few places (`check_call` and `check_output`), which means it is going to raise if the operation fails. I grabbed on to this, and used `unchecked` to mean we aren't going to raise, but normally we raise (fail-fast focus).

If we are going to adopt a fewer-exception oriented model, then we need a name for the thing, because for the most part, functions should raise, and they do in Python (because of EAFP, which is a red herring imiho, since thread safety is done badly in Python, not to mention threading/coroutine confusion...). I don't like unchecked. If you have a better name, that's fine.

I'm writing this, because I am concerned about your recent code which has not differentiated between these two cases so as a reader, I don't know if a routine is going to return None or raise. I think it's a style thing that can get muddied very quickly unless there's a clear naming policy.

e-carlin: > for the most part, functions should raise, and they do in Python

Prefixing some functions with `check_` when they are going to raise is confusing to me then. You obviously didn't write Python so I'm barking up the wrong tree but do you agree? It seems that the default should be assume raise (so no need to add `check_`) and `unchecked_` should be the only prefix that appears.

I think I understand what you are looking for. You mention places in Python where this convention is broken but I think there are places in our existing code where we don't follow this style too (ex job_agent._cmd, sim_data._lib_file_abspath, auth.user_if_logged_in, job_supervisor._status_reply,  etc). In addition we have simulation_db.verify_app_directory which sometimes creates lib and examples but the name doesn't tell you so. To help me understand better, are those places different or should they have some prefix too?

robnagler: You are correct that `check_` is not the right approach for Python. 

I am questioning my choice of `unchecked_` vs some other name.

You are right about sim_data._lib_file_abspath. At first I looked at it, and somehow discounted that it was an internal routine, but that's not right. That's why I think we need something possibly shorter than `unchecked_` as a prefix.

user_if_logged_in has a conditional in its name. I think that's like is_sim_type vs assert_sim_type. You know it's a question versus a statement. 

One could argue that names are just hints, and you really have to know what things do. Calling job_agent._cmd is a complicated thing, and it returns an error or None, which is confusing, but I discount the API, because it is internal.

simulation_db.verify_app_directory is interesting, because it isn't verifying anything, it's ensuring. In py.path, at least, `ensure` means something like this so being completely pedantic about this it's ensure_app_dir_setup_for_user. Here's where I think there's little value to have "directory" and more value to having the other information. Adding the modifier `for` is useful, I think, instead of ensure_user_app_dir, which is valid but a lot ot unpack.

e-carlin: Make sense. Thanks for taking the time to explain. I don't have a better word than unchecked and we are already using it so stuck with it. c6bbbc2e

robnagler:
@e-carlin is this still relevant?

e-carlin:
@robnagler yes it is. It is going to conflict with #3160 but if you merge this then I'll fix #3160.

e-carlin:
@robnagler all comments addressed. I changed the logic so the user is renamed instead of the dir 8fb71ae9.

PR #3073, opened 2020-10-23
Fix #2811. The base of the controls app.

### LINKED ISSUE #2811: create new controls sirepo app
New controls app will take features from the webcon app and make it generally useful.

- Allow selecting any existing MAD-X sirepo simulation.
- Display the rendered beamline and provide editors for the kickers and quads values.
- Show the H/V monitors and BPMs as plots
- Select success criteria and run an optimization, updating plots after each simulation step.

e-carlin:
This is a basic outline of the controls app. The beamline is rendered and
editors are shown for some of the elements.

Questions:
- What elements do we want to support? For those elements what values should
the user be able to adjust?
- What reports do we want?
- How do we want to organize the editors for the beamline elements?


sirepo/package_data/template/controls/examples/beamline.json
```diff
@@ -0,0 +1,1416 @@
...
+            }
+        ],
+        "simulation": {
+            "activeBeamlineId": 125,
+            "commandTemplate": "particle",
+            "documentationUrl": "",
+            "enableParticleTracking": "1",
+            "folder": "/Examples",
+            "isExample": true,
+            "name": "TODO(e-carlin)",
```
e-carlin, line 1395: @moellep what do we want to name this thing?

sirepo/template/controls.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+u"""Controls execution template.
+
+:copyright: Copyright (c) 2020 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+from sirepo.template.madx import *
```
e-carlin, line 8: @robnagler this is the cheapest of inheritance. Do you have thoughts on what might be better?

My ideas:
- There aren't that many methods that need to be implemented. I could just implement them and call the madx equivalents in the method body.
- Move to a class base system and inherit from the madx class.

moellep: I have a better idea for this, rather than bringing in the mad-x schema and deriving the templates. Instead, just the simple editors need to be in the schema, and the whole mad-x beamline is treated as a json blob within the sirepo-data.json. I'll make some changes to this branch to demonstrate.

sirepo/template/controls.py
```diff
@@ -5,4 +5,22 @@
...
+        )
+    return PKDict(
+        percentComplete=100,
+        frameCount=1,
+    )
+
+
+def write_parameters(data, run_dir, is_parallel):
+    data.models.externalLattice.report = ''
+    sirepo.template.madx.write_parameters(data.models.externalLattice, run_dir, is_parallel)
```
moellep, line 26: The whole mad-x dataset is now contained in data.models.externalLattice. That way the controls app doesn't need to know about all the fields and the controls-schema.json has only the models specific to the app. The "externalLattice" could be swapped out with an elegant lattice or an epics controller without much additional work.

e-carlin: Makes sense. Thanks.

robnagler: how do fixups work on this?

moellep: In sim_data.controls, it runs the fixups on the embedded lattice:
```
    def fixup_old_data(cls, data):
        sirepo.sim_data.get_class('madx').fixup_old_data(data.models.externalLattice)
```

PR #3100, opened 2020-10-31
Issue/3099 Check for sdds_beam input during write

### LINKED ISSUE #3099: sdds_beam.input renamed by elegant LibAdapter.write_files
It looks like there is still some name mangling done when writing elegant run files with LibAdapter.write_files. For the attached  example I think everything is put into the new working directory correctly except for the sdds_beams which has its input file name modified and written into the new .ele as:

```
&sdds_beam
  center_arrival_time = 1,
  input = "bunchFile-sourceFile.PCG.sdds.twiss",
  reuse_bunch = 1,
&end
```
 PCG.sdds.twiss isn't symlinked in (nor bunchFile-sourceFile.PCG.sdds.twiss)

[tessamatch_PT.txt](https://github.com/radiasoft/sirepo/files/5468262/tessamatch_PT.txt)

cchall:
This fixes #3099
Looks like the issue is that `BeamInputFile` is used as the description for sdds_beam.input in the elegant schema instead of just `InputFile`. I wasn't sure of the significance of `BeamInputFile` so I just added it into `InputFileIterator` as a fix.

### DIFF
--- a/sirepo/template/lattice.py
+++ b/sirepo/template/lattice.py
@@ -83,7 +83,7 @@ class InputFileIterator(ModelIterator):
         self.sim_data = sim_data
 
     def field(self, model, field_schema, field):
-        if model[field] and field_schema[1].startswith('InputFile'):
+        if model[field] and (field_schema[1].startswith('InputFile') or field_schema[1].startswith('BeamInputFile')):
             self.result.append(self.sim_data.lib_file_name_with_model_field(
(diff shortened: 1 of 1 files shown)


e-carlin:
@cchall FYI, this caused some tests to fail (see #3103) so I reverted it on master. I'm looking into finding a fix.

PR #3102, opened 2020-11-02
WIP: Fix #3088. Add bpm plots as well as twiss and beam commands.

### LINKED ISSUE #3088: controls bpm plots
The beam position plots for the controls app can use the output from the twiss command at each of the element positions. The twiss file has a row for each element, and the position at that point. Create a report for just the watch points (element MONITOR).

This example shows the x,y position of the beam on the visualization tab.

https://www.sirepo.com/madx#/lattice/k76qFOOx

We'll want a way for the user to specify the initial beam parameters (beam energy and twiss parameters and x,y offsets). All the required fields are in the beam or twiss commands.

e-carlin:
@moellep I have some questions. I left comments with them below.

### DIFF
--- a/sirepo/package_data/static/js/controls.js
+++ b/sirepo/package_data/static/js/controls.js
@@ -33,6 +33,8 @@ SIREPO.app.config(function() {
     };
     SIREPO.appReportTypes = [
         '<div data-ng-switch-when="bpmMonitor" data-bpm-monitor-plot="" class="sr-plot" data-model-name="{{ modelKey }}"></div>',
+        '<div data-ng-switch-when="bpmHMonitor" data-bpm-monitor-plot="Horizontal" class="sr-plot" data-model-name="{{ modelKey }}"></div>',
+        '<div data-ng-switch-when="bpmVMonitor" data-bpm-monitor-plot="Vertical" class="sr-plot" data-model-name="{{ modelKey }}"></div>',
     ].join('');
 });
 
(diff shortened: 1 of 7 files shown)


sirepo/package_data/static/js/controls.js
```diff
@@ -103,13 +109,63 @@ SIREPO.app.controller('ControlsController', function(appState, controlsService,
...
+            frameCache.setFrameCount(1);
+            updateFromMonitorValues(data.monitorValues);
+        }
+    };
+
     appState.whenModelsLoaded($scope, function() {
+        self.commandEditors = [];
         self.editorColumns = [];
+        self.watches = [];
+        // TODO(e-carlin): twiss_command the fields aren't all populated correctly
```
e-carlin, line 139: @moellep there are some problems with the model being nested under externalLattice. How should I resolve? Should I make the concept of externalLattice known to sirepo-components.js and sirepo.js so it looks for the models under it if it isn't found at the top level?

sirepo/package_data/static/js/controls.js
```diff
@@ -103,13 +109,63 @@ SIREPO.app.controller('ControlsController', function(appState, controlsService,
         };
     }
 
+    function updateFromMonitorValues(monitorValues) {
+        monitorValues.forEach((value) => {
+            // TODO(e-carlin): How can I eliminate the timeout. Needed because
+            // on first load self.watches needs to be populated before there is
+            // anyone listening to the broadcast
+            $timeout(function () {
```
e-carlin, line 117: @moellep any idea how to eliminate this timeout? The sr-point-data-* listeners needs to be registered so the broadcast has somewhere to go. That is done by populating self.watches and rendering the bpm plots. But, the handleStatus comes in before models are loaded which leads to a race condition.

sirepo/package_data/static/js/controls.js
moellep: We don't want to push the externalLattice info outside of the controls app. I changed the code so the beam and twiss models get edited as a copy in appState.models and then the externalLattice gets updated whenever the models change.

sirepo/package_data/static/js/controls.js
moellep: This is tricky because the UI is building itself based on the runStatus results, so if it broadcasts right away, the directives won't be listening. For now I added a call to panelState.waitForUI() which is really just a $timeout call eventually.

moellep:
I changed the UI a bit because having all the beam and twiss fields available would be too confusing.
Screen Shot 2020-11-03 at 4 19 27 PM

e-carlin:
Thanks for the help @moellep!

PR #3105, opened 2020-11-03
Fix #3103. BeamInputFile is an input file but conforms to a different…

### LINKED ISSUE #3103: Tests failing: Simulation library file "command_sdds_beam-input.forward-beam-output.sdds" does not exist
job_concurrency1_test.py and report_test.py are both failing on master due to missing lib file.

e-carlin:
… shape than other InputFile's.

@moellep the knowledge of how to handle  BeamInputFile's is spread around the js, template, and lattice files. If you have any thoughts on how to bring it all into one place lmk. 

@cchall the filename is no longer mangled and is symlinked to the destination dir (at least in my test). Give it a shot and lmk if it works for you.


cchall:
This fix works for me. Thanks for fixing this up @e-carlin !

Issue #3097, opened 2020-10-29
Post-angular topics


mkeilman:
Not sure whether this has been explicitly laid out, so here are some possible items for discussion, at a higher level than specific issues like #2716 for example.  Once we have a plan we can close this issue.

- html5
- js 6
- jquery
- generate html etc. from classes
- apps become classes?
- schemas define not just views but as much of the UI as possible (app->[pages->[views]] ?). 

Other topics as they arise...

robnagler:
Relegating to ngui

PR #3108, opened 2020-11-04
sirepo dom classes

mkeilman:
Presenting a preliminary set of classes for generating DOM strings (see #3097) for discussion.  Notes:

- I named the branch ```rnd/sirepo-dom```, implicitly suggesting this convention for branches with experimental stuff that may or may not be merged.  Comments?
- Eschewing modules and import/export in favor of putting the classes on the SIREPO object
- This is a transitional form containing angular stuff
- Includes a couple of test enums in radia.  The "buttons" form is not quite right but works OK
- Other discussion...


robnagler:
Is this the DOM classes we say yesterday or should this be closed?

mkeilman:
It was the starting point for discussion last year but is outdated.  OK to close I think

PR #3111, opened 2020-11-06
Fix #3063. Add sirepo.lib support for OPAL.

### LINKED ISSUE #3063: Add OPAL to sirepo.lib.Importer
For fastfelo we are going to need to have OPAL parsing capability in rsopt (as was implemented in #2864 for elegant)

e-carlin:
I'm using mixins to eliminate copied code. I know we don't use them elsewhere in our code base but I think as a temporary solution they make some sense here. They allow for inheritance and eliminate repeated code. Once we have a better sense of what inheritance structure makes sense then we can move the mixins into their proper class (or get rid of inheritance all together).


sirepo/template/opal.py
```diff
@@ -448,10 +489,79 @@ def sim_frame_plot2Animation(frame_args):
 def write_parameters(data, run_dir, is_parallel):
     pkio.write_text(
         run_dir.join(OPAL_INPUT_FILE),
-        _generate_parameters_file(data),
+        _generate_parameters_file(data, is_parallel),
     )
 
 
+class _Generate(template_common.LibLatticeUtilMixin):
```
robnagler, line 496: base class unless there's a reason to do mixin, which I don't see. Also, this doesn't belong in tempate_common. Put it in sirepo.lib since nobody else would use this.

sirepo/template/opal.py
```diff
@@ -60,6 +61,45 @@
...
+        data, input_files = opal_parser.parse_file(
+            pkio.read_text(path),
+            filename=path.basename,
+        )
+        self._verify_files(path, [f.filename for f in input_files])
+        return self._convert(data, code_var, _SCHEMA)
+
+    def write_files(self, data, source_path, dest_dir):
+        """writes files for the simulation
+
```
robnagler, line 78: extra blank line

sirepo/template/opal.py
```diff
@@ -60,6 +61,45 @@
...
+        g = _Generate(data)
+        r = PKDict(commands=dest_dir.join(source_path.basename))
+        pkio.write_text(r.commands, g.sim())
+        for f in set(
+            LatticeUtil(data, _SCHEMA).iterate_models(
+                lattice.InputFileIterator(_SIM_DATA),
+            ).result,
+        ):
+            dest_dir.join(f).mksymlinkto(source_path.new(
```
robnagler, line 92: formatting could be better imho:

```py
            dest_dir.join(f).mksymlinkto(
                 source_path.new(
                     basename=_SIM_DATA.lib_file_name_without_type(f),
                 ),
                 absolute=False,
          )
```

tests/lib_test.py
```diff
@@ -25,7 +34,11 @@ def test_elegant():
         pkunit.file_eq(s.join('out.json'), d2)
         w = pkunit.work_dir().join(s.basename)
         r = d.write_files(w)
-        #TODO(robnagler) may not exist in all cases
-        pkunit.pkeq('run_setup.acceptance.sdds', r.output_files[0])
+        pkunit.pkok(
```
robnagler, line 37: could this test be in test_elegant?

robnagler (changes requested):
cursory review. Most important: no mixins, please.

sirepo/template/opal.py
```diff
@@ -60,6 +61,45 @@
 )
 
 
+class LibAdapter(template_common.LibFileMixin, template_common.LibCodeVarMixin):
```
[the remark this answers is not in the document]
e-carlin: c4d0d0bf

sirepo/template/opal.py
e-carlin: c4d0d0bf

sirepo/template/opal.py
e-carlin: fc7b4879

sirepo/template/opal.py
e-carlin: fc7b4879

tests/lib_test.py
e-carlin: 05d8978b

sirepo/template/opal.py
```diff
@@ -345,7 +386,7 @@ def prepare_sequential_output_file(run_dir, data):
 def python_source_for_model(data, model):
     if model == 'madx':
         return OpalMadxConverter().to_madx_text(data)
-    return _generate_parameters_file(data)
+    return _generate_parameters_file(data, is_parallel=True)
```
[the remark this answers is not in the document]
e-carlin: No, I was copying from elegant and not thinking. fc42fef5

PR #3131, opened 2020-11-13
Fix #3093. Allow user to select madx lattice to use in controls app.

### LINKED ISSUE #3093: Import madx lattice into controls app
Allow users to select one of their existing madx lattices and import it into the controls app.

There is some cleanup that needs to be done when the export happens:
- Make all target elements (QUAD, HMON, VMON, MON) unique. Ex. if quad with id 65 is used in multiple places in the beamline make it so it is used in only one places and create new quads (with unique ids) for all of the other places
- replace RPNValues with their evaluated Floats
- Remove commands that won't be used. We just need to generate twiss params so command like ptc_create_universe won't be needed.

e-carlin:
@moellep See comment below regarding sirepo-lattice.js rpnService. It needs to change to work with externalLattice.


sirepo/package_data/static/js/sirepo-lattice.js
```diff
@@ -507,8 +507,12 @@ SIREPO.app.service('rpnService', function(appState, requestSender, $rootScope) {
         if (self.isCaseInsensitive) {
             v = v.toLowerCase ? v.toLowerCase() : v;
         }
-        if (appState.models.rpnCache && v in appState.models.rpnCache) {
-            return appState.models.rpnCache[v];
+        // TODO(e-carlin): This is the crux. rpnCache is now under externalLattice
```
e-carlin, line 510: @moellep the whole rpnService needs to change to support the rpnCache/variables being under externalLattice. Any advice on how best to go about this?

moellep: I don't think we want any variable values in the external lattice. Part of the fixups which remove the command entries should also go through and evaluate every field and replace any rpn expressions with a float value.

e-carlin:
@moellep I implemented evaluating the rpn expressions with their values. The whole PR is ready for your review.

PR #3135, opened 2020-11-13
Fix #3129. Enable particle reflection from emitter and collector plates.

### LINKED ISSUE #3129: warpvnd: Enable particle reflection from emitter and collector plates
A common run configuration for our benchmarks are to adjust the reflection properties of the emitter and collector. Currently, the emitter/collector parameters are specified within the “Beam” tab, with no way of specifying reflection.

Screen Shot 2020-11-12 at 11 42 05 AM

Conversely, the additional conductors are explicitly specified within the “Conductor Types” window, and adding a new conductor type enables one to check “Add Reflector”.

Screen Shot 2020-11-12 at 11 39 29 AM

I would like the interface to permit users to specify the emitter and collector as particle reflectors, and choose a diffuse reflection probability as is currently allowed for these other conductors.

Commensurate with this change, I suggest changing the “Beam” tab to be titled “Emitter and Reflector Properties”,
and to list the emitter and reflector explicitly, as is done for the other conductors. This would mean that users won't need to click the "edit" icon in the right corner to view and change the temperature or work function of the cathode and anode.

Of course, the simulation cannot run without the cathode and anode being installed, so these elements do not need to be removable/renamable/movable as the other conductors are, but their temperature, work function, and reflection properties should be listed, and editable by users.

The code to create these conductors is similar to what’s done for the others. Because the emission routines are maintained separately, the specification of temperature/work function/reflectivity does not change the conductor installation procedure. For example, the Python file from the [Gridded TEC 2D Opt](https://www.sirepo.com/warpvnd#/source/alaNGnNc) example specifies on line 200

```python
# Create source conductors 
source_condid = max_c_id + 1 
source = wp.ZPlane(zcent=wp.w3d.zmmin, zsign=-1., voltage=0., condid=source_condid) 
wp.installconductor(source, dfill=wp.largepos)

# Create ground plate
plate_condid = max_c_id + 2
plate = wp.ZPlane(voltage=extractor_voltage, zcent=zplate, condid=plate_condid)
wp.installconductor(plate, dfill=wp.largepos)
scraper = wp.ParticleScraper([source, plate] + conductors, lcollectlpdata=True, lsaveintercept=False)
```

The location/orientation of these collectors (`zcent` and `zsign`) should be fixed and not viewable/editable by users, and they should not be permitted to be labelled as dielectrics.

e-carlin:
I believe this covers everything outlined in #3129. @moellep and @ncook882 I don't know how to verify that these changes are working right. The simulation runs with them but I don't know if it is correct. Lmk if there is something I can do (ex demo, send python sources) to help verify the changes are correct.

@ncook882 this is the new "Emitter and Reflector Properties
Screen Shot 2020-11-13 at 2 41 01 PM
" panel:

When the user double clicks on the cathode or anode on the Conductor Grid they will now be able to adjust the properties of the conductor. Ex for the cathode:
Screen Shot 2020-11-13 at 2 41 12 PM

### DIFF
--- a/sirepo/package_data/static/css/warpvnd.css
+++ b/sirepo/package_data/static/css/warpvnd.css
@@ -11,8 +11,8 @@
   cursor: pointer;
 }
 .warpvnd-plate-no-voltage {
-  stroke: rgb(228, 176, 95);
   fill: rgba(255, 146, 105, 0.3);
+  stroke: rgb(228, 176, 95);
 }
 .warpvnd-plate-voltage {
   fill: rgba(105, 146, 255, 0.3);
--- a/sirepo/package_data/static/html/conductor-grid.html
+++ b/sirepo/package_data/static/html/conductor-grid.html
@@ -23,12 +23,10 @@
               <svg class="plot-viewport" data-ng-attr-width="{{ width }}" data-ng-attr-height="{{ height }}" data-ng-drop="true" data-ng-drag-move="dragMove($data, $event)" data-ng-drop-success="dropSuccess($data, $event)">
                   <defs>
                       <pattern id="reflectionPattern-anode" x="0" y="0" width="20" height="20" patternUnits="userSpaceOnUse">
-                          <path d="M0,0 l20,20" stroke="{{  source.conductorColors.nonZero }}" stroke-width="1.5" stroke-opacity="{{ source.getReflectOpacity('anode', 'd') }}"></path>
-                          <!--<circle cx="2" cy="12" r="2" fill="{{  source.conductorColors.nonZero }}" fill-opacity="{{ source.getReflectOpacity('anode', 'd') }}"></circle>-->
+                          <path d="M0,0 l20,20" stroke="{{  source.conductorColors.nonZero }}" stroke-width="1.5" stroke-opacity="{{ source.getReflectOpacity('anode') }}"></path>
                       </pattern>
                       <pattern data-ng-repeat="conductorType in source.conductorTypes() track by conductorType.id" id="reflectionPattern-{{ conductorType.id }}" x="0" y="0" width="20" height="20" patternUnits="userSpaceOnUse">
-                          <path d="M0,0 l20,20" stroke="{{ conductorType.color }}" stroke-width="1.5" stroke-opacity="{{ source.getReflectOpacity(conductorType.id, 'd') }}"></path>
-                          <!--<circle cx="2" cy="12" r="2" fill="{{ conductorType.color }}" fill-opacity="{{ source.getReflectOpacity(conductorType.id, 'd') }}"></circle>-->
+                          <path d="M0,0 l20,20" stroke="{{ conductorType.color }}" stroke-width="1.5" stroke-opacity="{{ source.getReflectOpacity(conductorType.id) }}"></path>
                       </pattern>
                   </defs>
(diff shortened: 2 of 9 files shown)


ncook882:
@e-carlin thanks very much for the great update. I have a few requests for some text modifications. 

Can you change the "Add Reflector" text to read "Particle Reflector"?

Can you also change the "Emitter and Reflector" panel title to be "Emitter and Collector"?

Can you remove the "Beam X Radius" parameter from the base level panel? It would be fine to keep it in the sub panel that pops up when users select the edit icon.

ncook882:
@e-carlin I have an additional request for adjustments.

Can you make the "Particle Reflector" switch for the cathode and anode available from the "Emitter and Collector Properties" panel? Subsequently, can you make the reflection probabilities accessible from the edit pop-up panel?

Thanks.

e-carlin:
@ncook here are the changes:

Ignore the fields overflowing the border, I'm working on a fix.
Screen Shot 2020-11-16 at 11 02 17 AM

Screen Shot 2020-11-16 at 11 31 47 AM

One question for you, should specular and diffuse reflection probability be editable for cathode, anode, and any user defined elements? Specular probability wasn't editable in any form previously and diffuse probability wasn't editable for the cathode previously.

ncook882:
@e-carlin it would be great if users could specify the reflection probability for all conductor elements that can be reflectors, and especially so for the cathode and anode.

However, a conductor should either a specular or a diffuse reflector (and not both). Would it be possible to make it such that if one of them is specified, then the other is made non-editable? I realize there may be several strategies here, and it's not clear to me which is cleaner/easier from a UI/backend perspective.

e-carlin:
That should be possible. What field controls whether or not the conductor is specular or diffuse?

ncook882:
@e-carlin the reflector WarpVND implementation follows the [`ParticleReflector`](https://github.com/radiasoft/rswarp/blob/master/rswarp/particlecollector/particlereflector.py) class, and the instantiation follows:

```python
 if r['isReflector']:
        reflector = particlereflector.ParticleReflector(scraper=scraper, conductor=c, spref=wp.Species(type=wp.Electron, name='Electron'), srefprob=0., drefprob=r['diffProb'], refscheme='uniform')
        particlereflector.installparticlereflector(reflector)
```

The constructor permits the specification of specular reflection via the `srefprob` argument, and diffuse reflection via the `drefprob` argument. These arguments are templated in the WarpVND data model as `"specProb"` and `"diffProb"`, respectively. 

Technically, the ParticleReflector constructor permits both to be specified at once, so there's no internal check against both being specified. However, it's somewhat unphysical, and adds needless complexity to the simulations, so I'd rather prevent it at the level of the GUI, and permit more advanced users to change it via scripts on their own.

Let me know if this helps.

e-carlin:
@ncook882. User's now have the option to set whether or not the conductor is a specular reflector. Based on whether or not we enable/disable the probability fields.

Screen Shot 2020-11-17 at 10 23 18 AM

ncook882:
This looks great. Thanks very much for making so many updates.

e-carlin:
Sure thing. Thanks for your guidance!

@moellep  ready for your review.

moellep:
I made a few changes, mainly reducing the schema fields (isReflector, isSpecular, specProb, diffProb) into (reflectorType, reflectorProbability). I put the probability values on the main "Emitter and Collector Properties" panel because I thought the multiple toggle buttons were confusing.

Screen Shot 2020-11-17 at 7 49 23 PM

PR #3136, opened 2020-11-13
Fix #3121 added submodule for ml/lib

### LINKED ISSUE #3121: example datasets outside of sirepo repo
We need more examples for the ML app, but the input datafiles are usually quite large which prevents them from being added to the sirepo repo. We either need a separate sirepo examples repo, or add a way for examples to download an external datafile when needed.

robnagler:
I didn't do anything fancy so the submodule repos have to empty except for the lib files. Could put a filter for *.md files, for example, but that would require touching code in a few places.

Good intro to submodules (ignore the ssh discussion) https://www.activestate.com/blog/getting-git-submodule-track-branch/

Commands you need to know. First time, you need to:

```
cd ~/src/radiasoft/sirepo
git submodule update --init
```

After that, you can update the submodules with:

```
cd ~/src/radiasoft/sirepo
git submodule update --remote
```

I'll add this to the wiki once this PR is on master.


e-carlin:
Can you give the sw team permission to push to sirepo-ml-lib? I get a 403 right now.

robnagler:
radiasoft/sirepo has access now, sorry.

e-carlin (approved):

e-carlin:
LGTM. Just some thoughts on submodules which are new to me.

One not so minor annoyance: merges create new commits so things can get out of wack. Here is what I tried:
1. Create a branch `test` in each repo. Add commits and create a PR for each.
2. Merge the two PR's.
3. In Sirepo `git checkout master && git pull && git submodule update --remote`
4. In Sirepo `git status` shows `modified: sirepo-ml-lib (new commits)`. I think a merge creates a new commit so the SHA for the HEAD of master in sirepo-ml-lib is a different SHA than what the merged test branch in Sirepo thought HEAD was. So after we merge both branches, in Sirepo we will have to add another commit to update our submodule SHA to point to the latest for the submodule. Another options is to merge the submodule PR and then in sirepo ` git checkout branch && submodule update --remote && git commit -am 'updated submodule'` before merging the sirepo PR. Both seem kind of strange so maybe I'm missing something.

One insignificant annoyance: git stash doesn't work how I expect it to. If I make a change in sirepo-ml-lib and then in the sirepo if I do `git status` I see `modified: submodule (new commits)`. If I do `git stash` and then `git status` I still see the `modified ...` message. I would've expected it to go away. I have to go to the submodule and stash the changes there.

robnagler:
How about [tracking the submodule to master](https://stackoverflow.com/a/18797720)?

```txt
[submodule "SubmoduleTestRepo"]
    path = SubmoduleTestRepo
    url = https://github.com/jzaccone/SubmoduleTestRepo.git
    branch = master
```

e-carlin:
The problem persists. The SHA for HEAD of the submodule has changed and sirepo needs to be updated to reflect that. 

It seems we could use `git submodule update` (no `--remote`). If we do this then sirepo will point to the SHA outlined in the PR and not the HEAD of master of the submodule. But then were not working on the HEAD of master we are working probably one commit behind (the commit before the merge commit). I have a feeling this could cause trouble if there are two PRs in Sirepo and corresponding PRs in the submodule repo. We may have some code in Sirepo that expects SHA abc in the submodule and another that expects SHA xyz. Maybe this would always lead to a merge conflict and a human would have to resolve, not sure.

robnagler:
I don't think this is a good approach. We can pull the data automatically in dev, and when we build the docker image, we can pull then, too. There aren't a lot of big examples. The data are small, mostly.

PR #3160, opened 2020-11-20
Fix #3159. src and dst jupyterhub db root may be the same.

### LINKED ISSUE #3159: jupyterhub: Same src and dst directories for user migration
We need to add a check of existing directory names when creating sirepo jupyterhub user's because the src and dst directories for migrating users is going to be the same.

e-carlin:
In the case of the src and dst db root being the same we need to add another
check of the directories to make sure a new username doesn't conflict.

### DIFF
--- a/sirepo/sim_api/jupyterhublogin.py
+++ b/sirepo/sim_api/jupyterhublogin.py
@@ -60,17 +60,18 @@ def jupyterhub_user_name(have_simulation_db=True):
 def init_apis(*args, **kwargs):
     global cfg
 
+    d = pkio.py_path(sirepo.srdb.root()).join('jupyterhub')
     cfg = pkconfig.init(
         dst_db_root=(
-            pkio.py_path(sirepo.srdb.root()).join('jupyterhub'),
+            d,
             pkio.py_path,
             'new jupyter user db',
         ),
         rs_jupyter_migrate=(False, bool, 'give user option to migrate data from jupyter.radiasoft.org'),
         src_db_root=(
-            pkio.py_path('/var/empty'),
+            d,
             pkio.py_path,
-            'existing jupyter user db (ex /srv/jupyterhub)',
+            'existing jupyter user db (ex /srv/jupyterhub/user)',
         ),
         uri_root=('jupyter', str, 'the root uri of jupyterhub'),
     )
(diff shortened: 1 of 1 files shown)


sirepo/sim_api/jupyterhublogin.py
```diff
@@ -94,10 +95,13 @@ def __user_name(logged_in_user_name):
             logged_in_user_name.split('@')[0],
         )
         u = JupyterhubUser.search_by(user_name=n)
-        if u:
+        if u or pkio.sorted_glob(_user_dir(n)):
```
robnagler, line 98: `_user_dir(n).exists()`

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -94,10 +95,13 @@ def __user_name(logged_in_user_name):
...
             # The username already exists. Add a random letter to try and create
             # a unique user name.
             n += _HUB_USER_SEP + random.choice(string.ascii_lowercase)
+
+        assert not pkio.sorted_glob(_user_dir(n)), \
```
robnagler, line 103: `_user_dir(n).exists()`

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -94,10 +95,13 @@ def __user_name(logged_in_user_name):
             logged_in_user_name.split('@')[0],
         )
         u = JupyterhubUser.search_by(user_name=n)
-        if u:
+        if u or pkio.sorted_glob(_user_dir(n)):
             # The username already exists. Add a random letter to try and create
             # a unique user name.
             n += _HUB_USER_SEP + random.choice(string.ascii_lowercase)
```
robnagler, line 101: `sirepo.util.random_base62(3).lower()` -- gives a better shot at uniqueness.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -94,10 +95,13 @@ def __user_name(logged_in_user_name):
...
+
+        assert not pkio.sorted_glob(_user_dir(n)), \
+            f'conflict with existing username={n}'
```
robnagler, line 104: f'conflict with existing user_dir={n}'

robnagler (changes requested):
Simplify a bit and provide better chance on uniqueness test.

sirepo/sim_api/jupyterhublogin.py
e-carlin: 69b0fa64

sirepo/sim_api/jupyterhublogin.py
e-carlin: 69b0fa64

sirepo/sim_api/jupyterhublogin.py
e-carlin: 69b0fa64

sirepo/sim_api/jupyterhublogin.py
e-carlin: 69b0fa64

robnagler (approved):

PR #3172, opened 2020-11-24
Fix #3148. Remove name mangling in elegant for sirepo.lib.

### LINKED ISSUE #3148: Remove name mangling in elegant
Don't mangle any names, including the lattice from the original elegant input file.

e-carlin:
Removed all name mangling (lattice and output files I believe was all that was left).

I'm passing around `update_filenames` all over the place which is kind of ugly. If you think a polymorphic structure would be better lmk.


robnagler (approved):
Thanks for updating the tests. :)

PR #3180, opened 2020-12-01
Fix #3150. Remove rpnl.

### LINKED ISSUE #3150: Evaluate rpn in python, not using rpnl
rpnl has some insecure features so we shouldn't run it

e-carlin:
@moellep this technically works and could be merged but I have some question that I left below.


tests/lib_data/elegant_1/PClinac_TESSA_LSC.lte.out
```diff
@@ -12,15 +12,15 @@
 "TC_BC1.1": CSRCSBEND,angle=-0.0165,bins=500.0,l=0.146,n_kicks=100.0,sg_halfwidth=1.0,slice_analysis_interval=50.0
 "TC_BC1.2": CSRCSBEND,angle=0.033,bins=500.0,l=0.146,n_kicks=100.0,sg_halfwidth=1.0,slice_analysis_interval=50.0
 "TC_BC1.3": CSRCSBEND,angle=-0.0165,bins=500.0,l=0.146,n_kicks=100.0,sg_halfwidth=1.0,slice_analysis_interval=50.0
-"L3BM1L": CSRDRIFT,dz=0.01,l=0.120565622789519,lsc_bins=1024,slice_analysis_interval=50.0,use_stupakov=1
+"L3BM1L": CSRDRIFT,dz=0.01,l=0.12056562278951899,lsc_bins=1024,slice_analysis_interval=50.0,use_stupakov=1
```
e-carlin, line 15: Does the change in precision matter?

tests/template/elegant_import_data/deviance-rpn.lte.txt
```diff
@@ -1 +1 @@
-invalid rpn: "48e6 10 /x"
+invalid rpn="48e6 10 /x" error="unknown token: /x"
```
e-carlin, line 1: I think having the error adds more context. But, maybe the old message was formatted the way it was for a reason?

tests/template/elegant_test.py
```diff
@@ -14,11 +14,12 @@ def test_get_application_data():
     _expr('1 1 +', 2.0)
     _expr('3 1 1 + *', 6.0)
     _expr('2.72 ln', 1)
-    _expr('1 2 < ? 3 : 4 $', 3)
+    # TODO(e-carlin): Discuss w pjm. ? is hard to impl
+    # _expr('1 2 < ? 3 : 4 $', 3)
```
e-carlin, line 18: I'm not very clever but I think`?` is a hard operator to implement. Is it required?

sirepo/template/elegant_lattice_importer.py
```diff
@@ -33,34 +34,43 @@
...
-            return float(out.strip()), None
-        return None, 'empty'
-
-
 def elegant_code_var(variables):
-    return code_variable.CodeVar(variables, ElegantRPNEval())
+    _PI = 4 * math.atan(1)
+
+    class _P(code_variable.PurePythonEval):
+        _OPS = PKDict({
```
e-carlin, line 41: This is a fraction of all of the operators defined in defns.rpn. This was enough to cover all test cases and all of the example elegant files you sent. I'm inclined to go through and implement more of the simple defns. Thoughts?

robnagler:
Would we not want to just eval defns.rpn in the interpreter, cache that state, and then operate on new inputs?

e-carlin:
@robnagler that's possible and would be considerably more robust. The test for this work was that it can evaluate all of the example elegant runs. This can do that and didn't require changing the interpreter so I thought it was simpler. If you think it isn't enough I'm happy to go your route.

robnagler:
I think the test should also include anything @cchall has for rsopt runs.

If it's done, then it's done. 

I just thought it would be easy to eval defns.rpn since we already have the interpreter. Perhaps there are too many complications there.

e-carlin:
I think the interpreter would have to change a fair bit. It doesn't understand much of the grammar in defns.rpn (udf, pop, =, sto, etc).

@cchall is it possible for you to share your rsopt elegant runs so I can test with them?

cchall:
I went looking for some simulations I've used recently with RPN evaluations and didn't turn up much. Most cases are either RPN expressions that are can only be evaluated during a run or the separate issue of shell commands.

So I modified an existing file that I was trying to use with rsopt to have a few simple RPN expressions and one in link_elements.equation which should be ignored (I think this is handled correctly currently). Hopefully its the sort of thing you had in mind.

There are also various examples we could pull from elegant's example package. I took a quick look through and it seemed like most cases were very simple or used features that I don't know if we support in Sirepo (&rpn_load or variables filled at run time from &rf_setup).
[elegant.ele.txt](https://github.com/radiasoft/sirepo/files/5631101/elegant.ele.txt)
[elegant.lte.txt](https://github.com/radiasoft/sirepo/files/5631103/elegant.lte.txt)

e-carlin:
@cchall those are good. Thank you.

tests/template/elegant_test.py
e-carlin: In example I've seen it is not used. Removed https://github.com/radiasoft/sirepo/pull/3180/commits/51bf4f092217f7592969834b2f12f727e674add4

sirepo/template/elegant_lattice_importer.py
e-carlin: Added more defns. I think it is enough for now. https://github.com/radiasoft/sirepo/pull/3180/commits/367de305b57cca4e8a999cfb1cbdce77410755ac

e-carlin:
This is ready for another reivew. I've added a few more defns and all of the example elegant runs work as well ash Chris' examples. There are a couple lingering question above ([this](https://github.com/radiasoft/sirepo/pull/3180#discussion_r532990260) and [this](https://github.com/radiasoft/sirepo/pull/3180#discussion_r532991461)).

PR #3184, opened 2020-12-01
Fix #1351. Add role for jupyterhublogin.

### LINKED ISSUE #1351: Warp VND: set stops for zoom in/out

e-carlin:
jupyterhublogin is now a "proprietary" code. All new users get the
role for it. This will allow us to revoke the role later which will
prevent the user from logging in.

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -612,10 +612,13 @@ def _auth_state():
     return v
 
 def _create_roles_for_user(uid, method):
+    r = []
     if not (pkconfig.channel_in('dev') and method == METHOD_GUEST):
-        return
-
-    auth_db.UserRole.add_roles(uid, get_all_roles())
+        if not sirepo.template.is_sim_type('jupyterhublogin'):
+            return
+        # Every user gets the jupyterhub role initially
+        r.append(role_for_sim_type('jupyterhublogin'))
+    auth_db.UserRole.add_roles(uid, r or get_all_roles())
 
 
 def _get_user():
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -613,12 +613,10 @@ def _auth_state():
 
 def _create_roles_for_user(uid, method):
     r = []
-    if not (pkconfig.channel_in('dev') and method == METHOD_GUEST):
-        if not sirepo.template.is_sim_type('jupyterhublogin'):
-            return
-        # Every user gets the jupyterhub role initially
-        r.append(role_for_sim_type('jupyterhublogin'))
-    auth_db.UserRole.add_roles(uid, r or get_all_roles())
+    if pkconfig.channel_in('dev') and method == METHOD_GUEST:
+        auth_db.UserRole.add_roles(uid, get_all_roles())
+    elif sirepo.template.is_sim_type('jupyterhublogin'):
+        auth_db.UserRole.add_roles(uid, [role_for_sim_type('jupyterhublogin')])
 
 
 def _get_user():
--- a/sirepo/auth_db.py
+++ b/sirepo/auth_db.py
@@ -53,19 +53,23 @@ def audit_proprietary_lib_files(uid):
     import pykern.pkio
     import sirepo.auth
     import sirepo.feature_config
+    import sirepo.sim_data
     import sirepo.simulation_db
 
     x = sirepo.feature_config.cfg().proprietary_sim_types
     if not x:
         return
     for t in x:
+        if not sirepo.sim_data.get_class(t).proprietary_code_rpm():
+            return
+        d = sirepo.srdb.proprietary_code_dir(t)
+        assert d.exists(), \
+            f'{d} proprietary_code_dir must exist' \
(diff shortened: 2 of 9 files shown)


sirepo/auth/__init__.py
```diff
@@ -612,10 +612,13 @@ def _auth_state():
...
 def _create_roles_for_user(uid, method):
+    r = []
     if not (pkconfig.channel_in('dev') and method == METHOD_GUEST):
-        return
-
-    auth_db.UserRole.add_roles(uid, get_all_roles())
+        if not sirepo.template.is_sim_type('jupyterhublogin'):
+            return
+        # Every user gets the jupyterhub role initially
+        r.append(role_for_sim_type('jupyterhublogin'))
```
robnagler, line 620: only add the role if not a guest dev user? is this for testing? probably want to documnt

sirepo/auth_db.py
```diff
@@ -64,12 +64,9 @@ def audit_proprietary_lib_files(uid):
             sirepo.auth.role_for_sim_type(t),
         )
         d = sirepo.srdb.proprietary_code_dir(t)
-        assert d.exists(), \
-            f'{d} proprietary_code_dir must exist' \
-            + ('; run: sirepo setup_dev' if pykern.pkconfig.channel_in('dev') else '')
-
+        if not d.exists():
```
robnagler, line 67: we need configuration for this, perhaps. or at least do this check for flash or codes that need it. this can be hardwired in the sim_data so that we can ask the question of the code.

sirepo/jupyterhub.py
```diff
@@ -34,6 +35,9 @@ async def authenticate(self, handler, data):
         _set_cookie(handler)
         try:
             sirepo.auth.require_user()
+            sirepo.auth.require_sim_type('jupyterhublogin')
+        except werkzeug.exceptions.Forbidden:
+            return None
```
robnagler, line 40: this needs explanation. what does None mean here?

sirepo/pkcli/admin.py
```diff
@@ -24,6 +24,27 @@
 import shutil
 
 
+def add_jupyterhub_role():
```
robnagler, line 27: This is not something we will execute manually so I don't think it should be here. It needs to be a db upgrade so that it gets run on all implementations that implement jupyterhublogin similar to _migrate_db_file. I think we would check to see if juptyerhublogin is enabled and if there are any users with that role. If the answer is none, then we need to run the ugprade, atomically at startup.

sirepo/pkcli/admin.py
```diff
@@ -24,6 +24,27 @@
...
+    import sirepo.template
+    import sqlalchemy.exc
+
+    sirepo.template.assert_sim_type('jupyterhublogin')
+    r = sirepo.auth.role_for_sim_type('jupyterhublogin')
+    # TODO(e-carlin): locking
+    for u in sirepo.auth_db.all_uids():
+        try:
+            sirepo.pkcli.roles.add(u, r)
+        except sqlalchemy.exc.IntegrityError:
```
robnagler, line 44: This shouldn't happen with a db upgrade so let the exception fail. This type of "fail ignore" is problematic, and why I don't particularly like EAFP.

sirepo/pkcli/admin.py
```diff
@@ -24,6 +24,27 @@
...
+
+    sirepo.server.init()
+    import sirepo.auth
+    import sirepo.auth_db
+    import sirepo.pkcli.roles
+    import sirepo.template
+    import sqlalchemy.exc
+
+    sirepo.template.assert_sim_type('jupyterhublogin')
+    r = sirepo.auth.role_for_sim_type('jupyterhublogin')
```
robnagler, line 39: is the assert above necessary if role_for_sim_type is called?

robnagler (changes requested):

sirepo/auth/__init__.py
e-carlin: 6044c7cb. Is this clearer?

robnagler: Yes, but I get so confused with the fact that it is still showing the old code above....

sirepo/auth_db.py
e-carlin: 4c19d124

sirepo/jupyterhub.py
e-carlin: 83ca09c4

sirepo/pkcli/admin.py
e-carlin: 252b9c1d

sirepo/pkcli/admin.py
e-carlin: Ok. We are now doing the migration when the server is started 252b9c1d.

sirepo/pkcli/admin.py
e-carlin: Yes, role_for_sim_type doesn't check if the sim_type is valid. It probably should. It is moot though because this code is removed 252b9c1d

e-carlin:
@robnagler  all comments addressed.

PR #3187, opened 2020-12-02
Fix #2697. Recompute when out.json is missing.

### LINKED ISSUE #2697: missing out.json should recompute report
If the out.json file is missing, the whole directory should be deleted and the report should be re-run. Instead, the application gets stuck showing a "No such file or directory" error on the report panel.

To reproduce, open the SRW "Undulator Radiation" example and go to the beamline tab. Let the reports complete, and then remove the out.json file from the simulation's initialIntensityReport directory. Reload the web page and it will show the error message rather than re-running the report.

e-carlin:


sirepo/job_supervisor.py
```diff
@@ -536,6 +536,13 @@ def _get_queued_time(db):
     def _is_running_pending(self):
         return self.db.status in (job.RUNNING, job.PENDING)
 
+    def _init_db_missing_repsonse(self, req):
```
robnagler, line 539: typo: response

sirepo/job_supervisor.py
```diff
@@ -738,11 +745,14 @@ async def _receive_api_runStatus(self, req):
         r = self._status_reply(req)
         if r:
             return r
-        return await self._send_with_single_reply(
+        r = await self._send_with_single_reply(
             job.OP_ANALYSIS,
             req,
             jobCmd='sequential_result',
         )
+        if r.state == job.ERROR:
```
robnagler, line 753: is this the reason always for an error. Could it be something other than an init_db?

sirepo/job_supervisor.py
```diff
@@ -848,13 +858,9 @@ async def _send_with_single_reply(self, opName, req, **kwargs):
...
                     # will call _send_with_single_reply() and this will
                     # properly format the reply
-                    if not r.get('runDirNotFound'):
-                        return r
-                    self.__db_init(req, prev_db=self.db)
-                    self.__db_write()
-                    assert self.db.status == job.MISSING, \
-                        'expecting missing status={}'.format(self.db.status)
-                    return PKDict(state=self.db.status)
+                    if r.get('runDirNotFound'):
```
robnagler, line 861: what if runDir is there, but no out.json.

robnagler (changes requested):

sirepo/job_supervisor.py
e-carlin: I believe any time we do a runStatus and get an error back from the agent then we want to init_db. If we get back error from the agent then we can't read the results so the compute needs to be run again. There are some edge cases (ex fastcgi socket dies) where we might get back error but we could try again and maybe read the results. It isn't great to recompute in those cases but trickier to decipher them from an error reading the actual results. 

I interpreted [your comment here](https://github.com/radiasoft/sirepo/issues/2697#issuecomment-664667382) to mean what I implemented. Did I get it right?

sirepo/job_supervisor.py
e-carlin: Then the reading of the results will fail and the the `r.state == job.ERROR` check in runStatus will catch the error. 

Thinking about this more I think it would be best to eliminate runDirNotFound and move the `r.state == job.ERROR` check and handling that I added to runStatus into here. All callers of send_with_single_reply I think would benefit from this logic. I think every caller of send_with_single_reply will need the compute to be re-run if they get back an error from the agent. 

My assumption in all of this is that our code "works" and that when we get back error it really is an error with job. Not some transient error in our system (ex fastcgi socket dropped or we didn't send properly formatted json). Is this a safe assumption to make? If not should we create two classes of error (1 for error in our system and 1 for error with the actual compute/results)?

sirepo/job_supervisor.py
e-carlin: 5c770e74

sirepo/job_supervisor.py
robnagler: I don't think we need to assume one error or the other or have two classes of error. I do think we should interpret the errors in the GUI (or perhaps server-side) to be make more sense to the end user. We obviously can't do this always. 

I think we should also try to prevent errors percolating so if we know an out.json doesn't exist, we can just translate that to "missing", since it is probably not an end user serviceable error. We just try to do our best.

sirepo/job_supervisor.py
```diff
@@ -738,11 +745,14 @@ async def _receive_api_runStatus(self, req):
...
         if r:
             return r
-        return await self._send_with_single_reply(
+        r = await self._send_with_single_reply(
             job.OP_ANALYSIS,
             req,
             jobCmd='sequential_result',
         )
+        if r.state == job.ERROR:
+            return self._init_db_missing_repsonse(req)
```
robnagler: typo response

sirepo/job_supervisor.py
robnagler: I think if r.state error then recompute is fine.

sirepo/job_supervisor.py
robnagler: I think runDirNotFound is fine to do this way.

robnagler (approved):

PR #3268, opened 2020-12-29
Fix 3267. Case issue with jupyterhub migration.

RSSREC:
I've spent a fair bit of time thinking through this but I have trouble reasoning about this code so please take a close look.

We now format github handles (make lowercase and make all non-word
characters -). This means we now must rename directories (github
handles that are different after formatting).

This solves two bugs:
1. GitHub handles may have uppercase letters in them. This was causing
a bug where users were registering with github handles with
capitalized letters and a system downstream was making the names
lowercase.
2. Underscores were possibly being encoded/decoded wrong in
usernames. We had two directories (foo_8f2 and foo_5F8f2). This made
it look like we sometimes had an randomness of 5 characters added. We
believe that 5F (hex for ascii _) was a possible encoding/decoding
issue in a downstream system.

These changes lead to a behavior change in the app:
If a user registers, selects migrate, and they have no existing data previously they would no be alerted. Now they will be alerted that we could not migrate. They will have to select to proceed without migrating or to sign into a different account.

### DIFF
--- a/sirepo/package_data/static/html/jupyter-name-conflict.html
+++ b/sirepo/package_data/static/html/jupyter-name-conflict.html
@@ -1,10 +1,6 @@
 <div class="container-fluid">
   <div class="row">
-    <div class="col-sm-6 col-sm-offset-2" data-ng-if="! nameConflict.isMigration">
-      <h1>Error</h1>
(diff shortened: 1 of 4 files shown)


e-carlin:
@robnagler I've updated with comments from our discussion. Ready for another review.

robnagler (approved):

PR #3281, opened 2021-01-11
Fix #3123. Read csv with quoted values

### LINKED ISSUE #3123: ml: csv files with quoted values
Python's csv module correctly parses quoted values in a csv column but numpy.genfromtxt() does not. In some cases we use csv, in other numpy.

One solution would be to process the original csv file using the csv module and write out a new one which removes the quoted after removing embedded commas or newlines. Then numpy.genfromtxt() could be used as is.

e-carlin:


e-carlin:
get_application_data is still broken. Closing while I fix it.

PR #3287, opened 2021-01-14
Fix #3125: Only refresh plots when dataFile.file and dataFile.inputsS…

### LINKED ISSUE #3125: ml: don't refresh data source plots unless data file changes
On the Data Source page, changing items such as Application Mode or selected Input/Output column causes the plots to recalculate. Only changes to Data File or Inputs Scaler should cause a recalculation.

e-carlin:
…caler change

### DIFF
--- a/sirepo/sim_data/ml.py
+++ b/sirepo/sim_data/ml.py
@@ -34,7 +34,22 @@ class SimData(sirepo.sim_data.SimDataBase):
 
     @classmethod
     def _compute_job_fields(cls, data, r, compute_model):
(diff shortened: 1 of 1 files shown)


e-carlin:
@moellep I'm using _compute_job_fields now. Much simpler. I force pushed so you'll have to `git checkout issue/3125-ml-plots-refresh && git reset --hard origin/issue/3125-ml-plots-refresh # Any local changes to issue/3125-ml-plots-refresh will be lost `

PR #3288, opened 2021-01-15
Fix #3192: paging of datasets with a large number of columns

### LINKED ISSUE #3192: ml: paging columns for large datasets
For datasets with a large number of columns, allow paging through the columns. When paging, change the plots shown to match the current page's columns.

e-carlin:
@moellep [here](https://github.com/radiasoft/sirepo/files/5822169/adverts.csv.txt) is a dataset with a large number of cols you can use to test.

### DIFF
--- a/sirepo/package_data/static/js/ml.js
+++ b/sirepo/package_data/static/js/ml.js
@@ -1048,13 +1048,14 @@ SIREPO.app.directive('columnSelector', function(appState, mlService, panelState,
                 if (! appState.models.columnInfo.selected) {
                     appState.models.columnInfo.selected = [];
                 }
+                let p = 0;
+                $scope.pages = [[]];
                 for (let i = 0; i < c.header.length; i++) {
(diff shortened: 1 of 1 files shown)


moellep:
This looks good - I like that the show/hide links are no longer needed.

The change does have problems with this example dataset:
https://www.sirepo.com/ml#/data/6kekF9DQ

The dataset includes many columns which have 0 values and should be omitted. I think the pager has problems with them.

e-carlin:
> This looks good - I like that the show/hide links are no longer needed.
> 
> The change does have problems with this example dataset:
> https://www.sirepo.com/ml#/data/6kekF9DQ
> 
> The dataset includes many columns which have 0 values and should be omitted. I think the pager has problems with them.

@moellep fixed https://github.com/radiasoft/sirepo/pull/3288/commits/791f67a66c033359be7f5d7d5fdfeafc94488010

PR #3296, opened 2021-01-20
Fix #3123. Read csv with quoted values

### LINKED ISSUE #3123: ml: csv files with quoted values

e-carlin:
I added parsing in get_application_data as well as scale.py.jinja. Also, now the comments arg in numpy.genfromtxt is set to None which means the default (`#`) won't be parsed as a comment. This seems to probably be the right default for us when parsing csv's.

### DIFF
--- a/sirepo/numpy.py
+++ b/sirepo/numpy.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+u"""Wrappers for numpy
+
+:copyright: Copyright (c) 2021 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+import numpy as np
+
+def ndarray_from_csv(path, skip_header, **kwargs):
+    def _read():
+        import csv
+        import re
+
+        with open(path) as f:
+            for r in csv.reader(f):
+                yield ','.join(map(lambda v: re.sub(r'"|\n|\r|,', '', v), r))
+    return np.genfromtxt(
+        _read(),
+        comments=None,
+        delimiter=',',
+        skip_header=skip_header,
+        **kwargs
+    )
--- a/sirepo/numpy.py
+++ b/sirepo/numpy.py
@@ -14,7 +14,7 @@ def ndarray_from_csv(path, skip_header, **kwargs):
 
         with open(path) as f:
             for r in csv.reader(f):
-                yield ','.join(map(lambda v: re.sub(r'"|\n|\r|,', '', v), r))
+                yield ','.join(map(lambda v: re.sub(r'["\n\r,]', '', v), r))
     return np.genfromtxt(
         _read(),
         comments=None,
--- a/sirepo/numpy.py
+++ b/sirepo/numpy.py
@@ -12,7 +12,7 @@ def ndarray_from_csv(path, skip_header, **kwargs):
         import csv
         import re
 
-        with open(path) as f:
+        with open(path, "rt") as f:
             for r in csv.reader(f):
                 yield ','.join(map(lambda v: re.sub(r'["\n\r,]', '', v), r))
     return np.genfromtxt(
--- a/sirepo/package_data/template/ml/scale.py.jinja
+++ b/sirepo/package_data/template/ml/scale.py.jinja
@@ -12,7 +12,7 @@ from sklearn.preprocessing import {{ dataFile_outputsScaler }}
 def read_data(path, **kwargs):
     return sirepo.numpy.ndarray_from_csv(
         path,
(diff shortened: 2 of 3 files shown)


sirepo/numpy.py
```diff
@@ -0,0 +1,24 @@
...
+import numpy as np
+
+def ndarray_from_csv(path, skip_header, **kwargs):
+    def _read():
+        import csv
+        import re
+
+        with open(path) as f:
+            for r in csv.reader(f):
+                yield ','.join(map(lambda v: re.sub(r'"|\n|\r|,', '', v), r))
```
robnagler, line 17: The following is more traditional: `r'["\n\r,]'`

Is this guaranteed to work generally, that is, just stripping newlines and quotes?

sirepo/package_data/template/ml/scale.py.jinja
```diff
@@ -9,11 +10,10 @@ from sklearn.preprocessing import {{ dataFile_outputsScaler }}
 {% endif %}
 
 def read_data(path, **kwargs):
-    return np.genfromtxt(
+    return sirepo.numpy.ndarray_from_csv(
         path,
-        delimiter=',',
-        skip_header={{ 1 if columnInfo_hasHeaderRow else 0 }},
-        **kwargs
+        {{ 1 if columnInfo_hasHeaderRow else 0 }},
```
robnagler, line 15: `int(columnInfo_hasHeaderRow)` is proper Python. `bool(True) == 1`.

sirepo/numpy.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+from __future__ import absolute_import, division, print_function
+import numpy as np
+
+def ndarray_from_csv(path, skip_header, **kwargs):
+    def _read():
+        import csv
+        import re
+
+        with open(path) as f:
```
robnagler, line 15: We should either use py.path or `open(...,"rt")`.

robnagler (changes requested):

sirepo/numpy.py
e-carlin: 5ec2f51d .

I'm not exactly sure on the efficacy of the regex. @moellep was the [one to propose it](https://github.com/radiasoft/sirepo/issues/3123#issuecomment-725592205).  I know it solves the quoted values problem that we are trying to fix but I haven't seen any datasets with newlines in them.

sirepo/package_data/template/ml/scale.py.jinja
e-carlin: bool is just fine de563001.

What's up with folks using 1 and 0 in context where a truthy or falsey value is what is need. I know py.path and numpy do this in their docs. Just shorter? Do you have a preference?

sirepo/numpy.py
e-carlin: 83cb6542

sirepo/package_data/template/ml/scale.py.jinja
e-carlin: Well looking at numpy I was wrong. An int is required. `int()` is undefined in jinja. Reverted f5798c81

Question still holds for py.path (ex the rec kwarg in chmod())

e-carlin: Playing around I guess an int isn't technically required in this context, at least in what I tried. Python implicitly casts it (ex `1 + True == 2` and `for x in range(True): print(x) # Prints 0`.

robnagler: Whoops int(True) == 1. :)

Whoops on not noticing jinja which is not python.

robnagler: > What's up with folks using 1 and 0 in context where a truthy or falsey value is what is need. I know py.path and numpy do this in their docs. Just shorter? Do you have a preference?

It's messed up is what it is. :)  I think if a language has True, use it. Many people define TRUE in C, including [numpy](https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/npy_common.h#L358) (actually  NPY_TRUE and [TRUE](https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/noprefix.h#L147)).

There are reasons to use True, instead of 1, e.g. it maps properly in JSON.

e-carlin: Ok:) JSON seems like a pretty good reason to use bool, especially in our context. Thanks.

e-carlin:
ping

PR #3303, opened 2021-01-22
Fix #3130: Add monitor if none exists

### LINKED ISSUE #3130: controls: Add watch point to start and end of beamline
Upon importing a madx lattice, if the lattice doesn't have any watchpoints (ex the FODO PTC example) then add a watchpoint to the start of the beamline and one to the end.

e-carlin:


e-carlin:
ping

sirepo/template/controls.py
```diff
@@ -67,6 +67,23 @@ def write_parameters(data, run_dir, is_parallel):
...
+    b = data.models.beamlines[0]
+    if util.id_map[b['items'][-1]].type != 'MONITOR':
+        m = PKDict(
+            _id=LatticeUtil.max_id(data) + 1,
+            name='M1',
+            type='MONITOR',
+        )
+        data.models.elements.append(m)
+        b['items'].append(m._id)
+
```
moellep, line 85: I simplified _add_monitor() a bit. The import process creates a single beamline, so all you need to do is check if the last item is a MONITOR.

PR #3325, opened 2021-01-29
radiasoft/sirepoflash#16: Allow user to configure setup command args

e-carlin:
This provides the "commonly used setup arguments" from
http://flash.uchicago.edu/site/flashcode/user_support/flash4_ug_4p62/node31.html
It also provides a few less commonly used arguments that are required for
CapLaserBELLA and CapLaser3D.

This branch is based on radiasoft/sirepoflash/issue/16 . Please review and merge #3266 before reviewing this.

PR #3332, opened 2021-02-03
radiasoft/sirepoflash#16: allow user to upload FLASH problem files

e-carlin:
~~This branch is __not__ ready to be merged. There is work to be done in the installer and possibly rsconf to handle the changes in how the distribution of FLASH is performed. But, these changes can be reviewed while I’m working on them.~~
The pr's are open: radiasoft/download#149 and radiasoft/rsconf#138

This branch is base off radiasoft/sirepoflash/issue/16-setup-command. Please review and merge #3325 before reviewing this.

Users can now upload a ZIP archive containing FLASH problem files
(F90, Config, and Makefile). These will be unpacked in the proper location
in the FLASH source code and used when compiling.

Some notable changes:
- The FLASH source code is no longer distributed as an rpm. It is
distributed as a tarball that lives in the user's lib dir.
- CapLaser3D and CapLaserBELLA are no longer distributed inside
of the FLASH source code. They are now files in the lib dir. They are
distributed alongside the FLASH source code. Upon getting the FLASH role
they are unpacked in the user's lib dir.

### DIFF
--- a/sirepo/auth_db.py
+++ b/sirepo/auth_db.py
@@ -48,6 +48,7 @@ def audit_proprietary_lib_files(uid):
     Args:
         uid: user to audit
     """
+    import contextlib
     import py
     import pykern.pkconfig
     import pykern.pkio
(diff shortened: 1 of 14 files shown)


robnagler (approved):
some comments to fix, but not important

PR #3352, opened 2021-02-18
Fix #3350: wrap calls to flask.g

### LINKED ISSUE #3348: Opal TypeError: list.sort is not a function
I was clicking around the Opal CSR Bend Drift app on alpha with the dev tools open and saw this when I went to the source tab:
```
TypeError: list.sort is not a function
    sortList sirepo-components.js:773
    loadAuxiliaryData sirepo.js:2038
<Snip>
Error: [ngRepeat:dupes] http://errors.angularjs.org/1.7.8/ngRepeat/dupes?p0=item%20in%20itemList()&p1=string%3A%3C&p2=%3C
```

### LINKED ISSUE #3350: Wrap flask.g
Alpha is currently broken due to a bug involving flask.g. The new db_upgrade code must call `auth.set_user()` which ends up calling `cookie.set_cookie_for_utils()` which sets `flask.g = PKDict()`. Setting flask.g to a new value breaks the flask proxy object for all threads and caused failures once the server was running and handling requests. This highlighted some other places (ex jupyterhub auth) where we were not careful about messing with flask.g.

So, we are going to wrap the setting of flask.g. A few important things to keep in mind:
1. Wrap all setting of flask.g inside of a contextmanager (synchronous context manager, not `async with`). This will ensure we set the value in a manner that works in a threaded and non-threaded environment (see below).
2. In the case of Flask (multi-threaded) we will set on flask.g. Using flask.g will mean we use the flask request context system and thus work in a multi threaded environment. We must be sure that we are inside of a flask request not just that there is a flask application context.
3. In the case of Tornado (single-threaded) we will use our own global variable which will be set upon entry to the context manager and cleared upon exit and there will be no async code in between. This ensure that we set and clear the global while we have the thread of control.

e-carlin:
Inside of a flask request should be the only time we touch
flask.g.  We need to use flask.g here because flask is threaded
and flask.g handles getting the right object for the thread for
us. Outside of flask (tornado, jupyterhub) we can use a global
variable because it is single-threaded. All calls must be wrapped
in a `with` to ensure we maintain the thread of control (no
`await`'s).

Fix #3348

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -17,6 +17,7 @@ from sirepo import events
 from sirepo import http_reply
 from sirepo import http_request
 from sirepo import job
+from sirepo import simulation_db
 from sirepo import util
 import contextlib
 import datetime
@@ -162,13 +163,21 @@ def get_module(name):
 
 
 def init_apis(*args, **kwargs):
-    global uri_router, simulation_db, visible_methods, valid_methods, non_guest_methods
+    global uri_router, visible_methods, valid_methods, non_guest_methods
     assert not _METHOD_MODULES
 
     assert not cfg.logged_in_user, \
         'Do not set $SIREPO_AUTH_LOGGED_IN_USER in server'
     uri_router = importlib.import_module('sirepo.uri_router')
-    simulation_db = importlib.import_module('sirepo.simulation_db')
+    # TODO(e-carlin): why is simulation_db imported like this?
+    # It is a problem because we call
+    # job_supervisor.init()
+    #  auth_db.init()
+    #    _migrate_role_jupyterhub
+    #      UserRole.add_roles()
+    #        simulation_db.user_path() # simulation_db is undef because we haven't called init_apis
+    # Maybe move to _init()? I've just moved to regular import rn
+    # simulation_db = importlib.import_module('sirepo.simulation_db')
     auth_db.init()
     p = pkinspect.this_module().__name__
     visible_methods = []
@@ -327,9 +336,11 @@ def need_complete_registration(model):
     return not auth_db.UserRegistration.search_by(uid=model.uid).display_name
 
 
+@contextlib.contextmanager
 def process_request(unit_test=None):
-    cookie.process_header(unit_test)
-    _set_log_user()
+    with cookie.process_header(unit_test):
+        _set_log_user()
+        yield
 
 
 def require_auth_basic():
@@ -415,27 +426,18 @@ def role_for_sim_type(sim_type):
     return 'sim_type_' + sim_type
 
 
-def set_user_for_utils(uid=None):
-    """A mock user for utilities"""
-    cookie.set_cookie_for_utils()
-    import sirepo.auth.guest
-    if uid:
-        _login_user(sirepo.auth.guest, uid)
-    else:
-        login(sirepo.auth.guest, is_mock=True)
-
-
 @contextlib.contextmanager
-def set_user(uid):
-    """Set the user (uid) for the context"""
-
-    assert not util.flask_app(), \
-        'Flask sets the user on the request'
-    try:
-        set_user_for_utils(uid=uid)
+def set_user_outside_of_flask_request(uid=None):
+    """A user set explicitly outside of flask request cycle"""
+    assert not util.in_flask_request(), \
+        'Only call from outside a flask request context'
+    with cookie.set_cookie_for_utils():
+        import sirepo.auth.guest
+        if uid:
+            _login_user(sirepo.auth.guest, uid)
+        else:
+            login(sirepo.auth.guest, is_mock=True)
         yield
-    finally:
-        reset_state()
 
 
 def unchecked_get_user(uid):
@@ -635,6 +637,7 @@ def _init():
     if not cfg.logged_in_user:
         return
     global logged_in_user, user_dir_not_found
+
 
     def logged_in_user(*args, **kwargs):
         return cfg.logged_in_user
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -612,7 +612,6 @@ def _init():
         return
     global logged_in_user, user_dir_not_found
 
-
     def logged_in_user(*args, **kwargs):
         return cfg.logged_in_user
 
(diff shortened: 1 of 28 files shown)


sirepo/auth_db.py
```diff
@@ -61,8 +64,8 @@ def audit_proprietary_lib_files():
 
     def _add(proprietary_code_dir, sim_type, sim_data_class):
         p = proprietary_code_dir.join(sim_data_class.proprietary_code_tarball())
-        sirepo.simulation_db.verify_app_directory(sim_type)
-        with sirepo.simulation_db.tmp_dir(chdir=True) as t:
+        sirepo.simulation_db.verify_app_directory(sim_type, uid=uid)
```
robnagler, line 67: This really shouldn't be here. We should have a registration interface so auth_db doesn't have to import all this stuff, which is irrelevant to the service that auth_db provides.

sirepo/auth_db.py
```diff
@@ -303,13 +301,12 @@ def _migrate_db_file(fn):
 
 
 def _migrate_role_jupyterhub():
```
robnagler, line 303: This should get moved out, too. It should be in db_ugprade, and it should be a registered call.

sirepo/auth_role.py
```diff
@@ -0,0 +1,37 @@
...
+import sirepo.util
+
+ROLE_ADM = 'adm'
+ROLE_PAYMENT_PLAN_ENTERPRISE = 'enterprise'
+ROLE_PAYMENT_PLAN_PREMIUM = 'premium'
+
+PAID_USER_ROLES = (ROLE_PAYMENT_PLAN_PREMIUM, ROLE_PAYMENT_PLAN_ENTERPRISE)
+
+
+def check_user_has_role(uid, role, raise_forbidden=True):
```
robnagler, line 18: see above about leaving in auth.

sirepo/cookie.py
```diff
@@ -28,6 +31,8 @@
 
 _SERIALIZER_SEP = ' '
 
+_SRCONTEXT_COOKIE_STATE_KEY = 'cookie_state'
```
robnagler, line 34: `_SRCONTEXT_KEY = __name__` is better: locally and globally unique. Also less problematic on module/package renames.

sirepo/cookie.py
```diff
@@ -103,13 +113,24 @@ def unchecked_remove(key):
         return None
 
 
+@contextlib.contextmanager
+def _set_cookie(header):
```
robnagler, line 117: `def _set_state():`

sirepo/cookie.py
```diff
@@ -203,7 +224,7 @@ def _cfg_http_name(value):
 
 
 def _state():
-    return flask.g.sirepo_cookie
+    return sirepo.srcontext.get(_SRCONTEXT_COOKIE_STATE_KEY)
```
robnagler, line 227: This makes me feel better. :)

sirepo/db_upgrade.py
```diff
@@ -36,13 +36,13 @@ def _20210211_add_flash_proprietary_lib_files():
...
     for u in sirepo.auth_db.all_uids():
-        with sirepo.auth.set_user(u):
-            # Remove the existing rpm
-            pkio.unchecked_remove(sirepo.simulation_db.simulation_lib_dir(
-                'flash',
-            ).join('flash.rpm'))
-            # Add's the flash proprietary lib files (unpacks flash.tar.gz)
-            sirepo.auth_db.audit_proprietary_lib_files()
+        # Remove the existing rpm
+        pkio.unchecked_remove(sirepo.simulation_db.simulation_lib_dir(
```
robnagler, line 40: Formatting:

```py
        pkio.unchecked_remove(
            sirepo.simulation_db.simulation_lib_dir(
                'flash',
                uid=u,
            ).join('flash.rpm'),
       )
```

This style supports global name changes better.

sirepo/http_reply.py
```diff
@@ -75,8 +75,8 @@ def f():
         if isinstance(content_or_path, pkconst.PY_PATH_LOCAL_TYPE):
             return flask.send_file(str(content_or_path))
         if content_type == 'application/json':
-            return flask.current_app.response_class(pkjson.dump_pretty(content_or_path))
-        return flask.current_app.response_class(content_or_path)
+            return sirepo.util.flask_app().response_class(pkjson.dump_pretty(content_or_path))
```
robnagler, line 78: Add and use liberally:

```py
def gen_response(*args, **kwargs):
    return sirepo.util.flask_app().response_class(*args, **kwargs)
```

There are some places that call `flask.Response` directly, which should be changed to this. Then we have one more encapsulation to allow migration to tornado without the need to remove "flask" from N places.

sirepo/http_request.py
```diff
@@ -134,16 +134,20 @@ def t(v):
...
     """Interface for uri_router"""
-    return _set_flask_g(_POST_ATTR, data)
+    # Always remove data (if there)
+    res = sirepo.srcontext.pop(_POST_ATTR, None)
+    if data is not None:
+        sirepo.srcontext.set(_POST_ATTR, data)
+    return res
 
 
 def set_sim_type(sim_type):
```
robnagler, line 144: This should be a contextlib. It should stay in http_request, but be called from uri_router appropriately (perhaps call_api needs to be split into to functions?). http_request should call an internal function that assumes (asserts) that srcontext is already initialized by a prior call. This is similar to set_post, I think. I didn't think about this stuff when designing uri_router. Learning Python still. :(

sirepo/auth/__init__.py
```diff
@@ -411,31 +396,19 @@ def reset_state():
...
 
 @contextlib.contextmanager
-def set_user(uid):
-    """Set the user (uid) for the context"""
-
-    assert not util.flask_app(), \
-        'Flask sets the user on the request'
-    try:
-        set_user_for_utils(uid=uid)
+def set_user_outside_of_flask_request(uid=None):
```
robnagler, line 401: Let's call it `set_user_outside_of_http_request`. This will be true in tornado, too, because it's really the concept of cookies that is important here.

robnagler (approved):
I'm going to merge, because I think it works, but I have a lot of comments that we should address after we stabilize this release.

PR #3356, opened 2021-02-19
Fix #3349: backup db on upgrade and log running upgrade name

### LINKED ISSUE #3349: db_upgrade backup db before applying upgrade

e-carlin:
This branch is based on issue/3354-audit-force. Please review and merge  #3355 before reviewing this.


sirepo/db_upgrade.py
```diff
@@ -23,26 +24,29 @@ def do_all():
     a = sirepo.auth_db.DbUpgrade.search_all_for_column('name')
     f = pkinspect.module_functions('_2')
     for n in sorted(set(f.keys()) - set(a)):
-        f[n]()
-        sirepo.auth_db.DbUpgrade(
-            name=n,
-            created=sirepo.srtime.utc_now(),
-        ).save()
+        with _backup_db_and_restore_on_error():
```
e-carlin, line 27: @robnagler how do you feel about the backup/restore happening for each upgrade? My thinking is that if we have 3 upgrades and the 3rd one fails we still may have updated aspects of sirepo outside of the auth_db in the 2 that worked so the restored db should reflect that.

robnagler: I don't think auto-restore is a good idea. It should exit the server, and even if it is restarted, it shouldn't. Maybe the way to get around the systemd restart problem is to put a sentinel in the start directory or the database that prevents restart.

robnagler (changes requested):
Let's wait on this one.

sirepo/db_upgrade.py
e-carlin: Ok, auto restore was removed and setting a sentinel file that prevents server restart was added https://github.com/radiasoft/sirepo/pull/3356/commits/76a07c3ffe7977607d2b5b43334ece21d31238d0

PR #3372, opened 2021-03-02
Fix #3307: running multiple jobs at NERSC

### LINKED ISSUE #3307: srw: running multiple concurrent jobs at NERSC from a single user account
Up to 5 NESC simulations should be allowed from a single user's account. Currently, it appears that only one simulation can be run at once per user.

### LINKED ISSUE #3371: sbatch: job in queue (pending) cannot be cancelled until it starts running
To reproduce:
```sh
cat << EOF > test.sbatch
#!/bin/bash
while true; do
    echo 'running...'
    sleep 5
done
EOF
```
- `sbatch test.sbatch`
- Run srw young's double slit
- `squeue` # Should see test.sbatch running and srw pending
- cancel srw in gui
- `squeue` # Run this multiple times, srw won't be cancelled
- `scancel <id-of-test.sbatch>`
-  `squeue` # srw will quickly run and then be cancelled

The agent is in a loop waiting for the job to transition to RUNNING. It won't read from the incoming socket until the job transitions so it won't read the cancel until it does.

e-carlin:
Make the number of concurrent OP_RUN per user configurable.

Also, make it so jobs in squeue (pending) can be cancelled (fix #3371).

~~This PR lead to a change in behavior. computeJobStart will now be the time the job is submitted to the queue (roughly) not the time the job actually starts. To make it the time the job actually starts running would involve more of a refactor and I think is worth a discussion.~~ With https://github.com/radiasoft/sirepo/pull/3372/commits/3f3de31188f07f2daf8e5ac68d3dd746d2d8967a this is no longer true.

### DIFF
--- a/sirepo/job_driver/sbatch.py
+++ b/sirepo/job_driver/sbatch.py
@@ -35,20 +35,20 @@ class SbatchDriver(job_driver.DriverBase):
 
     def __init__(self, op):
         def _op_queue_size(op_kind):
-            return self.cfg.num_concurrent_op_run if op_kind == job.OP_RUN else 1
+            return self.cfg.run_slots if op_kind == job.OP_RUN else 1
 
         super().__init__(op)
         self.pkupdate(
             # before it is overwritten by prepare_send
             _local_user_dir=pkio.py_path(op.msg.userDir),
             _srdb_root=None,
-            # Allow num_concurrent_op_run and every other op type to
+            # Allow self.cfg.run_slots and every other op type to
             # run (assume OP_RUN is one of OPS_THAT_NEED_SLOTS). This
             # is essentially a no-op (sbatch constrains its own cpu
             # resources) but makes it easier to code the other cases.
             cpu_slot_q=sirepo.job_supervisor.SlotQueue(
                 len(job_driver.OPS_THAT_NEED_SLOTS)
-                + self.cfg.num_concurrent_op_run
+                + self.cfg.run_slots
                 - 1,
             ),
             op_slot_q={
@@ -89,7 +89,7 @@ class SbatchDriver(job_driver.DriverBase):
             cores=(None, int, 'dev cores config'),
             host=pkconfig.Required(str, 'host name for slum controller'),
             host_key=pkconfig.Required(str, 'host key'),
-            num_concurrent_op_run=(1, int, 'number of concurrent OP_RUN for each user'),
+            run_slots=(1, int, 'number of concurrent OP_RUN for each user'),
             shifter_image=(None, str, 'needed if using Shifter'),
             sirepo_cmd=pkconfig.Required(str, 'how to run sirepo'),
             srdb_root=pkconfig.Required(_cfg_srdb_root, 'where to run job_agent, must include {sbatch_user}'),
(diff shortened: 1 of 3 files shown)


sirepo/job_driver/sbatch.py
```diff
@@ -33,14 +34,27 @@ class SbatchDriver(job_driver.DriverBase):
     __instances = PKDict()
 
     def __init__(self, op):
+        def _op_queue_size(op_kind):
+            return self.cfg.num_concurrent_op_run if op_kind == job.OP_RUN else 1
```
robnagler, line 38: num_concurrent_op_run => run_slots?

sirepo/job_driver/sbatch.py
```diff
@@ -33,14 +34,27 @@ class SbatchDriver(job_driver.DriverBase):
...
-            # but makes it easier to code the other cases.
-            cpu_slot_q=job_supervisor.SlotQueue(len(job_driver.OPS_THAT_NEED_SLOTS)),
+            # Allow num_concurrent_op_run and every other op type to
+            # run (assume OP_RUN is one of OPS_THAT_NEED_SLOTS). This
+            # is essentially a no-op (sbatch constrains its own cpu
+            # resources) but makes it easier to code the other cases.
+            cpu_slot_q=sirepo.job_supervisor.SlotQueue(
+                len(job_driver.OPS_THAT_NEED_SLOTS)
+                + self.cfg.num_concurrent_op_run
+                - 1,
```
robnagler, line 52: `-1` is unclear. Is this subtracting RUN from OPS_THAN_NEED_SLOTS? Document that, e.g. `-1 accounts for run having num_concurrent_op_run and not needing a separate slot`. I'm not sure about the coupling here. Do we even run analysis?

sirepo/pkcli/job_agent.py
```diff
@@ -754,14 +754,11 @@ async def _sbatch_status(self):
             )
             return
         self._status = r.group()
-        if self._status in ('PENDING', 'CONFIGURING'):
+        if not self._start_ready.is_set():
```
robnagler, line 757: Was this a bug? Or is this a consequence of having a real slot q? I don't understand.

robnagler (changes requested):
Maybe we need to talk about this.

sirepo/job_driver/sbatch.py
e-carlin: My hope was that the comment above served the purpose. Maybe I should move it above the -1?

sirepo/job_driver/sbatch.py
e-carlin: https://github.com/radiasoft/sirepo/pull/3372/commits/279e8616938a349b54e812dccfdb05a6da4daa82

sirepo/pkcli/job_agent.py
e-carlin: https://github.com/radiasoft/sirepo/pull/3372/commits/3f3de31188f07f2daf8e5ac68d3dd746d2d8967a I've reverted back to the old behavior per the line you commented on. In addition, we now wait for _start_ready in a callback instead of in the agent dispatcher loop. Implementing waiting for transitioning to run was considerably easier than I thought. One of these days I won't be afraid of code:/

robnagler (approved):

Issue #3384, opened 2021-03-10
katex in svg labels


e-carlin:
Investigate using katex for rendering in svg labels.

moellep:
I found the html2canvas library does both html and svg and can render KaTeX math as well. And it is as easy as supplying the element and receiving a canvas:
```javascript
    html2canvas(el).then(canvas => {
        ....
    });
```

moellep:
This issue is important for plot improvements. I'll take a look. I found the png download was displaying the 2 different renderings of the label, so the code below hides the ```katex-html``` version.

``` javascript
         html2canvas(el, {
             scale: outputHeight / $(el).height(),
             backgroundColor: '#ffffff',
             ignoreElements: (element) => {
                 return element.matches("path.pointer.axis") || element.classList.contains('katex-html');
             }
         }).
```

Also, the global css doesn't make it into the ```foreignObject```, so the label css needs to get added to the svg within a ```style``` tag.

moellep:
This was implemented for parameterPlots with 59b71d0, but should be expanded to include plot3d and heatmaps.

PR #3385, opened 2021-03-10
WIP #3384: rudimentary foreignObject

e-carlin:
@moellep here is a rudimentary way of rendering html for the svg labels. I know next to nothing about svg/foreignObject so opening up to see if you have any thoughts. You can go to an elegant parameter plot and see it in action for the y-axis-label. If you think this approach might work I'll close and implement fully.


moellep:
Unfortunately foreignObject doesn't work with the Download Report as SVG feature which is important.

moellep (changes requested):
foreignObject with SVG to PNG support

e-carlin:
I'll fix this when I have time. Closing for now...

PR #3416, opened 2021-03-22
Fix radiasoft/sirepoflash#18: 1D profile plots

e-carlin:
Depends on radiasoft/download#159

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ pykern.pksetup.setup(
         'user-agents',
         'uwsgi',
         'werkzeug',
+        'yt',
     ],
     license='http://www.apache.org/licenses/LICENSE-2.0.html',
     name='sirepo',
--- a/sirepo/package_data/static/html/flash-visualization.html
+++ b/sirepo/package_data/static/html/flash-visualization.html
@@ -11,7 +11,7 @@
     </div>
     <div class="col-md-6" data-ng-hide="visualization.panelState.isHidden('simulationStatus')">
       <div data-simple-panel="simulationStatus">
-        <div data-sim-status-panel="visualization.simState"></div>
+        <div data-sim-status-panel="visualization.simState" data-start-function="visualization.startSimulation()"></div>
       </div>
     </div>
     <div data-ng-if="visualization.simState.hasFrames()">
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -31,6 +31,7 @@
         'user-agents',
         'uwsgi',
         'werkzeug',
+        'yt',
```
e-carlin, line 34: @robnagler is this a problem? I could work with the physicist on rsflash to make it so consumers of rsflash don't need to use yt.

moellep:
I've experimented with this pull request a bit - and I think I led you astray with a bad example. Rather than showing a list of plot files to choose, it should be an animation of plots similar to the elegant simulation "bunchComp - fourDipoleCSR" plot B1:

Screen Shot 2021-04-01 at 1 49 48 PM

Each frame would be a from a single *_plt_cnt_* file. That way you can plot multiple variables at once and have an easy way to navigate between files. Playing the animation allows you to see how the variables change over time. Each plot should have a timestamp in the plot title.

e-carlin:
Thanks for taking a look @moellep. I could be misunderstanding things but I don't think that is what Nathan wants. If you look at [feature number 2 in the issue](https://github.com/radiasoft/flashcap/issues/11) you'll see he wants field and axis to stay the same and to be able to select different plot files to plot against each other which I believe this feature accomplishes. Thoughts?

e-carlin:
Reading over it again I see I missed converting the filenames to sim time. Implementing now...

moellep:
OK - for a single variable comparison, it would probably be better to have something similar to the "Turn Comparison" plot from synergia - it allows selecting a single variable and then selecting which turns to plot. In this case, you could have 3 drop downs which show a list of times. I think showing a whole list of filenames is confusing. 

If you would like to keep the checkbox list for now, rather than showing "flash_hdf5_plt_cnt_0070" in the list you would show something like  "2.3e-2 seconds".

moellep:
When running a new simulation, it shouldn't try to plot the files which are no longer available. For example, if I select file 50 to plot after running a simulation, then start a new simulation, file 50 isn't available and the plot shows "report not generated". Selecting the pencil while the simulation is running seems to cause the modal dialog to be hidden each time the other plots refresh.

e-carlin:
I'm still working on your comments. The times returned by rsflash and those calculated from the filenames are different and I'm not sure why. Also, need to clear out the selected plot files when starting a new sim and figure out how to not hide the modal dialog when the data is updated.

e-carlin:
@moellep all comments addressed. Ready for another review.

PR #3424, opened 2021-03-23
Fix #3422: validate authState in status test

### LINKED ISSUE #3422: sirepo.status should check authState for validity

e-carlin:


sirepo/status.py
```diff
@@ -5,9 +5,10 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
-from pykern import pkconfig
 from pykern import pkcompat
+from pykern import pkconfig
 from pykern.pkdebug import pkdc, pkdexc, pkdlog, pkdp
+from pykern.pkunit import pkre
```
robnagler, line 11: This is not a unit test so let's not use pkunit.

sirepo/status.py
```diff
@@ -52,6 +53,10 @@ def init_apis(*args, **kwargs):
 
 def _run_tests():
     """Runs the SRW "Undulator Radiation" simulation's initialIntensityReport"""
+    pkre(
+        r'"isLoggedIn": true',
```
robnagler, line 57: i think it would be better to parse the json to be sure it is completely valid then assert x.isLoggedIn == True (in this case we literally want this, because it should be a boolean.

robnagler (changes requested):

sirepo/status.py
e-carlin: 8f3de227

sirepo/status.py
e-carlin: 8f3de227

e-carlin:
@robnagler all comments addressed

sirepo/status.py
```diff
@@ -99,3 +101,11 @@ def _run_tests():
             uri_router.call_api('runCancel', data=d)
         except Exception:
             pass
+
+
+def _validate_auth_state():
+    r = pkcompat.from_bytes(uri_router.call_api('authState').data)
+    m = re.search(r'SIREPO.authState = (.*?);', r)
+    assert m, f'no authState in response: {r}'
```
robnagler, line 109: Here's where things get tricky and we want to avoid logging what is returned by the server with an fstring. We'll want to truncate it with pkdformat, I think, e.g. `assert m, pkdformat('no authState in response={}', r)`

sirepo/status.py
```diff
@@ -99,3 +101,11 @@ def _run_tests():
...
+    assert pkjson.load_any(m.group(1)).isLoggedIn, \
+        'expecting isLoggedIn: {m.group(1)}'
```
robnagler, line 111: ditto with m.group(1).

sirepo/status.py
```diff
@@ -99,3 +101,11 @@ def _run_tests():
             uri_router.call_api('runCancel', data=d)
         except Exception:
             pass
+
+
+def _validate_auth_state():
+    r = pkcompat.from_bytes(uri_router.call_api('authState').data)
+    m = re.search(r'SIREPO.authState = (.*?);', r)
```
robnagler, line 108: make the regex a bit looser: `r'SIREPO.authState\s*=\s*(.*?);'`. I know it's picky, but it won't break if we ever change the format.

robnagler (changes requested):

sirepo/status.py
e-carlin: 8e5eee4e

sirepo/status.py
e-carlin: 8e5eee4e

sirepo/status.py
e-carlin: 8e5eee4e

e-carlin:

sirepo/status.py
```diff
@@ -99,3 +101,11 @@ def _run_tests():
...
         except Exception:
             pass
+
+
+def _validate_auth_state():
+    r = pkcompat.from_bytes(uri_router.call_api('authState').data)
+    m = re.search(r'SIREPO.authState\s*=\s*(.*?);', r)
+    assert m, pkdformat('no authState in response={}', r)
+    assert pkjson.load_any(m.group(1)).isLoggedIn, \
+        pkdformat('expecting isLoggedIn=', m.group(1))
```
robnagler, line 111: missing `{}` after =

robnagler (changes requested):

sirepo/status.py
e-carlin: b6d4f145

robnagler (approved):

PR #3448, opened 2021-03-30
Fix #3407: pkcli for creating jupyterhub user

### LINKED ISSUE #3407: Register jupyter users with pkcli
Supply an email address and it returns the jupyterhub user name created. If the user is already registered, just returns the jupyterhub user. This will allow us to register groups of people for classes in advance.

e-carlin:


sirepo/auth/__init__.py
```diff
@@ -73,6 +74,11 @@
 #: visible_methods excluding guest
 non_guest_methods = None
 
+#: sirepo.uri_router module
```
robnagler, line 77: i think it is clear it's a module by it's name. rather, say "avoid circular import issues by importing in init_apis".

sirepo/auth/__init__.py
```diff
@@ -146,32 +152,18 @@ def get_module(name):
 
 
 def init_apis(*args, **kwargs):
-    global uri_router, simulation_db, visible_methods, valid_methods, non_guest_methods
-    assert not _METHOD_MODULES
-
+    global uri_router
+    assert not uri_router
```
robnagler, line 156: init_apis is guaranteed to not be called twice so no need for an assertion even though it seems like a good idea. The reason is that a contract is a contract and if init_apis were called twice anywhere, other things would break so that's a contract we can trust, not assert, and advertise that by not asserting.

sirepo/auth/__init__.py
```diff
@@ -23,6 +23,7 @@
 import importlib
 import sirepo.auth_role
 import sirepo.feature_config
+import sirepo.simulation_db
```
robnagler, line 26: Would it not be better to defer initialization until init_apis? Is there a reason to import sooner?

sirepo/auth/basic.py
```diff
@@ -38,9 +38,13 @@ def _check(v):
     return cfg.uid == v.username and cfg.password == v.password
 
 
-def init_apis(*args, **kwargs):
+def _init():
```
robnagler, line 41: this idiom is probably not useful here since this is just initializing cfg so could remove the function and just cfg = pkconfig. By having an `_init`, it begs the question where is it called. Without it, there's no question.

sirepo/auth/bluesky.py
```diff
@@ -119,11 +119,14 @@ def auth_hash(req, verify=False):
         )
 
 
-def init_apis(*args, **kwargs):
+def _init():
```
robnagler, line 122: ditto

sirepo/auth/email.py
```diff
@@ -128,6 +125,10 @@ def unchecked_user_by_user_name(user_name):
        return None
 
 
+def _init():
+    auth_db.init_model(_init_model)
```
robnagler, line 129: ditto

sirepo/auth/guest.py
```diff
@@ -108,3 +101,13 @@ def _cfg_login_days(value):
         return None
     assert value >= 1
     return datetime.timedelta(days=value)
+
+
+def _init():
```
robnagler, line 106: ditto

sirepo/job_driver/__init__.py
```diff
@@ -213,6 +213,7 @@ def _agent_env(self, env=None):
         return job.agent_env(
             env=(env or PKDict()).pksetdefault(
                 PYKERN_PKDEBUG_WANT_PID_TIME='1',
+                SIREPO_AUTH_METHODS='guest',
```
robnagler, line 216: Why would this need to be set?

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -0,0 +1,45 @@
...
+                        if email already exists.
+    """
+    import pyisemail
+    import sirepo.auth
+    import sirepo.server
+    import sirepo.sim_api.jupyterhublogin
+    import sirepo.template
+
+
+    assert pyisemail.is_email(email), \
```
robnagler, line 28: use pkcli.command_error.

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -0,0 +1,45 @@
...
+        f'invalid email={email}'
+    sirepo.server.init()
```
robnagler, line 30: This is very annoying. Perhaps pkcli needs a function that gets called from sirepo_console.py?

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -0,0 +1,45 @@
...
+    if not u:
+        pykern.pkcli.command_error('no sirepo user with email={}', email)
+    with sirepo.auth.set_user_outside_of_http_request(u, method='email'):
+        # TODO(e-carlin): locking
+        n = sirepo.sim_api.jupyterhublogin.unchecked_jupyterhub_user_name(
+            have_simulation_db=False,
+        )
+        if n:
+            return n
+        assert not sirepo.sim_api.jupyterhublogin.user_dir(
```
robnagler, line 42: This should be inside create_user, perhaps with a flag, so that the abstraction of how local parts are created are not leaked.

sirepo/pkcli/roles.py
```diff
@@ -70,10 +70,9 @@ def _parse_args(uid_or_email, roles):
 
     # POSIT: Uid's are from the base62 charset so an '@' implies an email.
     if '@' in uid_or_email:
-        m = sirepo.auth.get_module('email')
-        assert m, \
-            f'uid_or_email={uid_or_email} contains "@" but email not configured'
-        u = m.unchecked_user_by_user_name(uid_or_email)
+        u = sirepo.auth.get_module(
```
robnagler, line 73: why remove the assert?

tests/pkcli/jupyterhublogin_test.py
```diff
@@ -0,0 +1,30 @@
...
+    import sirepo.srdb
+
+    fc = auth_fc
+    e = 'e-t@b.c'
+    with pkunit.pkexcept(f'no sirepo user with email={e}'):
+        sirepo.pkcli.jupyterhublogin.create_user(e)
+    fc.sr_email_register(e, sim_type='myapp')
+    r = e.split('@')[0]
+    pkunit.pkeq(r, sirepo.pkcli.jupyterhublogin.create_user(e))
+    pkunit.pkeq(r, sirepo.pkcli.jupyterhublogin.create_user(e))
```
robnagler, line 30: maybe document that it is reentrant (just returning the user).

tests/pkcli/jupyterhublogin_test.py
```diff
@@ -0,0 +1,30 @@
...
+def test_create_user(auth_fc):
+    from pykern import pkunit
+    import sirepo.pkcli.jupyterhublogin
+    import sirepo.srdb
+
+    fc = auth_fc
+    e = 'e-t@b.c'
+    with pkunit.pkexcept(f'no sirepo user with email={e}'):
+        sirepo.pkcli.jupyterhublogin.create_user(e)
+    fc.sr_email_register(e, sim_type='myapp')
```
robnagler, line 27: test the case where there is no sirepo user? I will need this to work, because not everybody will be registered in advance.

robnagler (changes requested):

sirepo/auth/__init__.py
robnagler: It's still [a circular import](https://github.com/radiasoft/sirepo/blob/d2624a6728449a9217c991016f81eac872e94b2d/sirepo/simulation_db.py#L29) which is a bad thing. I think auth should defer, since simulation_db is more complex and "lower down" the food chain.

sirepo/pkcli/jupyterhublogin.py
e-carlin: Agreed. I'll think about it some https://github.com/radiasoft/sirepo/issues/3462

sirepo/pkcli/roles.py
e-carlin: A [KeyError will be raised](https://github.com/radiasoft/sirepo/blob/d2624a6728449a9217c991016f81eac872e94b2d/sirepo/auth/__init__.py#L151) if the module isn't found

tests/pkcli/jupyterhublogin_test.py
e-carlin: That's tested [two lines above](https://github.com/radiasoft/sirepo/pull/3448/files/d2624a6728449a9217c991016f81eac872e94b2d#diff-ed944e3330b3f0de21257dddb2ac55299a7d1c70232ac4bf17ef9d7947e28148R25)

sirepo/auth/__init__.py
e-carlin: c1482e38

sirepo/auth/__init__.py
e-carlin: c1482e38

sirepo/auth/basic.py
e-carlin: c1482e38

sirepo/auth/bluesky.py
e-carlin: c1482e38

sirepo/auth/email.py
e-carlin: c1482e38

sirepo/auth/guest.py
e-carlin: c1482e38

sirepo/pkcli/jupyterhublogin.py
e-carlin: c1482e38

tests/pkcli/jupyterhublogin_test.py
e-carlin: c1482e38

sirepo/auth/__init__.py
e-carlin: Urgh, of course. Thanks. 85119cb7

e-carlin:

robnagler (approved):

Issue #3475, opened 2021-04-06
running "sirepo admin create-examples" fails on alpha


moellep:
Running admin create-examples on alpha fails with a python exception:
```
[py3; sirepo]$ sirepo admin create-examples
...
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/sirepo/auth/__init__.py", line 150, in get_module
    return _METHOD_MODULES[name]
KeyError: 'guest'
```

e-carlin:
`SIREPO_AUTH_METHODS='guest' admin create-examples` will fix for now

moellep:
Fixed with #3481

PR #3481, opened 2021-04-06
Fix #3479: Guess auth method do not default to guest

### LINKED ISSUE #3479: sirepo.auth.get_module shold not default to guest
We can't assume that guest is configured. Need to guess the method

e-carlin:


sirepo/auth/__init__.py
```diff
@@ -402,13 +402,26 @@ def reset_state():
...
-    """A user set explicitly outside of flask request cycle"""
+def set_user_outside_of_http_request(uid):
+    """A user set explicitly outside of flask request cycle
+
+    This will guess the auth method the user used to authenticate.
+    If the method cannot be guessed it will default to guest if configured
+    otherwise raise.
+    """
+    def _auth_module():
+        for m in _METHOD_MODULES.values():
```
robnagler, line 413: this list should be cfg.methods, that is, the methods that are configured but not deprecated and possibly hidden (basic is not visible).

sirepo/auth/__init__.py
```diff
@@ -402,13 +402,26 @@ def reset_state():
...
+            if _method_user_model(m, uid):
+                return m
+        assert 'guest' in _METHOD_MODULES, \
```
robnagler, line 416: 'guest' should be METHOD_GUEST. Why is defaulting to guest necessary? shouldn't it be in _METHOD_MODULES?

sirepo/auth/__init__.py
```diff
@@ -462,15 +475,17 @@ def user_if_logged_in(method):
 
 
 def user_name():
+    m = cookie.unchecked_get_value(_COOKIE_METHOD)
```
robnagler, line 478: I don't understand why this is changed.

robnagler (changes requested):

sirepo/auth/__init__.py
e-carlin: Wouldn't we want `visible_methods` not `cfg.methods`? `cfg.methods` may contain not visible methods.

sirepo/auth/__init__.py
e-carlin: 934d0c8c. It is in _METHOD_MODULES but it doesn't have a `_method_user_model`

sirepo/auth/__init__.py
e-carlin: This was just me cleaning up the code. I thought it was strange that this method wasn't prefixed with "unchecked" but could fall through and return none. Agree?

sirepo/auth/__init__.py
e-carlin: nvm I see what your are saying. We should allow non-visible methods for this. Implementing...

e-carlin: 1f427b40

e-carlin:

sirepo/auth/__init__.py
robnagler: Sometimes routines like this are not called unchecked. I think in the case of hot fixes, we should only change what's absolutely necessary. I thought the change was semantically different.

e-carlin: Makes sense. Will do.

sirepo/auth/__init__.py
robnagler: The _method_user_model is the wrong idiom. What we want to know is if the user is a valid user and what its method is. The valid user question can be answered by checking UserRegistration if there is no user model. It is insufficient to simply return guest.

e-carlin: 5e753cc4 I think this does it. Every user will be checked to see whether or not they are in UserRegistration. We try all methods with UserModel first because they are more "specific" (they have more info like user_name; which is required for jupyterhublogin). If they all fail then we try to find a method w/o a UserModel and use it.

e-carlin:
This will also fix #3475. It will pickup the auth methods in the environment or use guest since it [is the default](https://github.com/radiasoft/sirepo/blob/b878a4ba9af2f441460cef12a4e2faf201367216/sirepo/auth/__init__.py#L632)

sirepo/auth/__init__.py
```diff
@@ -402,13 +402,34 @@ def reset_state():
...
+def set_user_outside_of_http_request(uid):
+    """A user set explicitly outside of flask request cycle
+
+    This will try to guess the auth method the user used to authenticate.
+    """
+    def _auth_module():
+        for m in cfg.methods:
+            a = _METHOD_MODULES[m]
+            if _method_user_model(a, uid):
+                return a
```
robnagler, line 414: Not important to fix now, but to me this is clearer:

```py
      x = None
      for m in cfg.methods:
            a = _METHOD_MODULES[m]
            if _method_user_model(a, uid):
                return a
            elif not x:
                x = a
    assert x,  f'no module found for uid={uid} in cfg.methods={cfg.methods}'
    return x
```

sirepo/auth/__init__.py
```diff
@@ -402,13 +402,34 @@ def reset_state():
...
+        )
+
     assert not util.in_flask_request(), \
         'Only call from outside a flask request context'
+    assert auth_db.UserRegistration.search_by(uid=uid), \
+        f'no registered user with uid={uid}'
     with cookie.set_cookie_outside_of_flask_request():
         _login_user(
-            get_module(method),
+            _auth_module(),
```
robnagler, line 432: Again, not important to fix now, but I agree with @moellep that _auth_module should take `uid` as a parameter as opposed to implicitly copied. I would also not make this a local method, since it is generally useful. I would name it `_module_for_user`.

sirepo/auth/__init__.py
robnagler: I also see that the semantics are better. I didn't understand that it wasn't returning anything, which is bad. I think it would have blown up, because getattr would blow up so there would always be a `u`. Therefore the assert should be on _COOKIE_METHOD. It is "unchecked" because that was the only way when this was written (me thinks).

robnagler (approved):

PR #3485, opened 2021-04-07
Fix radiasoft/flashcap#10: use yt for variable animation

e-carlin:
This moves to using yt to parse the hdf5 output files instead of parsing
them ourselves. In addition it allows specification of an aixs to perform
the slice on. Finally, it allows specifying if limits for the color bar.


e-carlin:
Need to make some changes.
- zmax/min -> v max/min
- when a different axis is specified (ex x) need to change the labels on the var animation plot to reflect this (ex y,z). Also, need to handle the axis for other types of geoms (cylindrical etc).

Issue #1608, opened 2019-04-17
rework javascript application code to remove appState.isLoaded()


moellep:
Currently the javascript applications have a lot of checks like this:
```
self.updateSimulationGridFields = function() {
        if (! appState.isLoaded()) {
            return;
        }
        // do something with appState.models here
    }
```
Components are sometime active before the appState.loadModels() has received data. 

If possible, the app should make sure the data is available before the route change completes. This would simplify the application business logic, similar to improvements made to require the d3 library to be present before plotting logic is run.

moellep:
This would simplify the client business logic. This doesn't apply to cases where there is a callback from a server request and appState.models may have been unloaded.

moellep:
Bumping the priority of this.

PR #3495, opened 2021-04-13
Fix #3474 radia wiggler solve fails

### LINKED ISSUE #3474: Radia: Wiggler example fails on solve
I discarded changes to example.
After pressing, "Start new solve", I get the error:
"Application Error: Error: Solver failed: error exit(1). A report was sent to the server."

Here's a link to the simulation
https://www.sirepo.com/radia#/visualization/hNMlEm54

mkeilman:
Notes:

- As commented in #3474, this geometry is not solvable.  Although we don't know whether any given geometry is solvable, we do know this example is not, so a flag has been added to keep the solver panel from showing
- The solver panel is now up at the top so other panels don't displace it
- This fix also displays Radia errors to the user instead of the unhelpful ```exit(1)```

### DIFF
--- a/sirepo/package_data/static/js/radia.js
+++ b/sirepo/package_data/static/js/radia.js
@@ -1058,11 +1058,11 @@ SIREPO.app.controller('RadiaVisualizationController', function (appState, errorS
     self.solution = null;
 
     self.enableKickMaps = function() {
-        return (appState.models.simulation || {}).enableKickMaps === '1';
+        return appState.isLoaded() && appState.models.simulation.enableKickMaps === '1';
     };
 
     self.isSolvable = function() {
-        return (appState.models.geometry || {}).isSolvable == '1';
+        return appState.isLoaded() && appState.models.geometry.isSolvable == '1';
     };
 
     self.simHandleStatus = function(data) {
--- a/sirepo/package_data/static/js/radia.js
+++ b/sirepo/package_data/static/js/radia.js
@@ -1061,11 +1061,14 @@ SIREPO.app.controller('RadiaVisualizationController', function (appState, errorS
         return (appState.models.simulation || {}).enableKickMaps === '1';
     };
 
+    self.isSolvable = function() {
+        return (appState.models.geometry || {}).isSolvable == '1';
+    };
+
     self.simHandleStatus = function(data) {
         if (data.error) {
             throw new Error('Solver failed: ' + data.error);
         }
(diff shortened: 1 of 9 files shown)


sirepo/template/template_common.py
```diff
@@ -461,6 +461,20 @@ def parse_enums(enum_schema):
     return res
 
 
+def parse_mpi_log(run_dir):
```
e-carlin, line 464: nit: use this in template.synergia

sirepo/package_data/template/radia/examples/wiggler.json
```diff
@@ -5,8 +5,9 @@
             "paths": []
         },
         "geometry": {
+            "isSolvable": "0",
```
e-carlin, line 8: Not sure if this example is out in the wild but if so - maybe modify sim_data.radia.fixup_old_data to check if the data to fixup is the wiggler example and if isSolvable isn't found set it to 0. Currently it will be set to the default which is 1.

sirepo/package_data/static/js/radia.js
```diff
@@ -1061,11 +1061,14 @@ SIREPO.app.controller('RadiaVisualizationController', function (appState, errorS
         return (appState.models.simulation || {}).enableKickMaps === '1';
     };
 
+    self.isSolvable = function() {
+        return (appState.models.geometry || {}).isSolvable == '1';
```
e-carlin, line 1065: I've noticed you use this construct `(appState.models.<model> || {}).<field>` berfore. Out of curiosity, what is the use case? Is it to handle models not being loaded yet? If so, I think `appState.isLoaded() && appState.models.geometry.isSolvable` is clearer.

sirepo/template/template_common.py
mkeilman: I'd prefer to make a separate issue for that once this PR is in place

sirepo/package_data/template/radia/examples/wiggler.json
mkeilman: It almost certainly is not in the wild, but it's probably best practice to assume it is

sirepo/package_data/static/js/radia.js
mkeilman: 57f8898

I won't hunt down all the other cases of this, deferring to #1608

sirepo/package_data/template/radia/examples/wiggler.json
mkeilman: 1a3ae83

PR #3506, opened 2021-04-14
Fix #3504: user may already exist when create_user is called

### LINKED ISSUE #3504: Handle uid already exists when creating jupyterhub user
From #3501:

> api_migrateJupyterhub needs to be protected from this (uid to create exists). Either the user should be taken straight to jupyterhub or alerted that they have already migrated. Need to think about this more and what fits with all of the other cases we are handling.

e-carlin:


sirepo/sim_api/jupyterhublogin.py
```diff
@@ -118,11 +118,16 @@ def __user_name():
             n += _HUB_USER_SEP + sirepo.util.random_base62(3).lower()
         return n
 
-    if check_dir:
-        n = __handle_or_name_sanitized()
-        assert not _user_dir(n).exists(), \
-            f'existing user dir with same name={n}'
     with sirepo.auth_db.thread_lock:
+        n = unchecked_jupyterhub_user_name()
```
robnagler, line 122: There's a double thread lock here. I don't think _unchecked_hub_user should lock. It's too low level so the lock should occur at an outer level. For the pkcli, there doesn't need to be a lock (it's wrong, in fact). Only APIs can thread lock.

robnagler (changes requested):

sirepo/sim_api/jupyterhublogin.py
e-carlin: 27b7d0fe removes the lock.

For the pkcli, should I supply a flag and not lock when it is supplied? Is this a new development in thinking around locking? Nowhere else (as far as I know) do we not lock when coming from a pkcli. I know the lock doesn't do anything in the pkcli but what is wrong about having it?

robnagler: Threadlocks and sessions are coupled. There's no need for a lock without a session, and vice-versa.

We may need to make them reentrant. Thread locks already are. The question is how to make sessions lockable as the state should be on the stack, and not coupled to a thread. Need to think about this, because nested locks/sessions usually indicates a problem with the design.

e-carlin: 580d2d6e this commit is in PR #3512 since that is where I added the session concept.

e-carlin:

robnagler (approved):

Issue #3505, opened 2021-04-14
sqlalchemy - need to rollback failed commits


e-carlin:
from #3501:

> We need to call rollback() in sqlalchemy if the commit fails. This is easy (I think) to do with a try/except around every commit. If we wanted to get more involved we could create a real transaction system (#3395)

Issue #3462, opened 2021-03-31
eliminate server.init() from pkcli


e-carlin:
@robnagler [says it is annoying[(https://github.com/radiasoft/sirepo/pull/3448#discussion_r604416314). Find a way to factor it out.

robnagler:
What's tricky is that not all commands want to do this. One fun project would be adding the ability to have classes for pkcli, and then the commands could be subclassable.

e-carlin:
@robnagler I think this should be addressed with #3505. We are adding a `with srcontext` that will need to be either added to every pkcli that needs it or handled in some general way as this issue describes. Thoughts?

PR #3508, opened 2021-04-15
WIP #3505: srcontext

e-carlin:
This is not ready to be merged but I want to discuss some of the tricky bits. It works clicking around the GUI and fixes the session.rollback bug from #3505.

Things not working:
- pkcli: pkcli's that interact with the db will need to have an srcontext created. We can do this by hand in each pkcli that needs it (see example in pkcli/roles.py) or we could abstract away the details by fixing #3462.
- tests: same thing. Some tests need to interact with the db from within the test function so we will need to create a context. We can do this by hand in each relevant test function (see example in adm_own_jobs_test.py). We could also abstract the details away in conftest. pytest allows fixtures to yield so we can create the context within the fixture. But, srcontext.create will need to be reentrant for this to work (ex adm_own_jobs_test will need a context and it calls pkcli.roles.add which creates a context). Not a big deal to make it reentrant but before going down that road wanted to discuss.
- db_upgrade: If we made srcontext.create() reentrant then it would work as is. If not we will need to move the with to protect just the calls to auth_db.

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -23,6 +23,7 @@ import datetime
 import importlib
 import sirepo.auth_role
 import sirepo.feature_config
+import sirepo.srcontext
 import sirepo.template
 import sirepo.uri
 import werkzeug.exceptions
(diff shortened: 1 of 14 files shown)


e-carlin:
We decided to srcontext.create in sirepo_console to handle pkcli and create a context by hand for tests and db_upgrade. Closing while i implement.

PR #3510, opened 2021-04-16
fix #1608 simulation detail pages wait for appState.isLoaded()

### LINKED ISSUE #1608: rework javascript application code to remove appState.isLoaded()

moellep:
- controllers and directives on simulation detail pages no longer need to check
  appState.isLoaded() or appState.whenModelsLoaded()
- any server callbacks (such as requestSender.getApplicationData()) still need the
  checks in case the models unloaded before the request completed

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -15,6 +15,30 @@ SIREPO.INFO_INDEX_MAX = 5;
 SIREPO.ENUM_INDEX_VALUE = 0;
 SIREPO.ENUM_INDEX_LABEL = 1;
 
+SIREPO.app.directive('simulationDetailPage', function(appState, $compile) {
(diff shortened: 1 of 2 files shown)


robnagler (approved):
I assume it works. It looks ugly but if it is hidden like this then it's fine.

moellep:
Yes, it is ugly because it has to delay the normal route controller/template handling which usually happens immediately. The upside is non of the existing code needs to change.

PR #3512, opened 2021-04-16
Fix #3505: Put sqlalchemy session on srcontext

### LINKED ISSUE #3505: sqlalchemy - need to rollback failed commits

e-carlin:
In #3505 the sqlalchemy session was not rolled back after a failed
commit. We were using a global session across the app which brought
Sirepo down.

This puts the sqlalchemy session on srcontext. The seesion is
instantiated and destroyed/rolled back with each request (in
uri_router._dispatch).

Not using a global session means that all code that needs to use the
auth_db needs to have an instantiated session. In Flask it is created
and destroyed with each request (_dispatch). For pkcli, tests, and in
the supervisor we must create the context by hand everywhere we need
it.

For pkcli we considered creating the context in
sirepo_console.py. This works fine in Flask where `sirepo service
http` ends up creating a new process so the context in
sirepo_console.py doesn't conflict with the one created in _dispatch
But, in the case of the job_supervisor we are in the same process as
the "spawning" console process. srcontext is not reentrant so we would
have to use one global context for the whole supervisor. This would
mean if the supervisor had one failed commit then the session is
ruined for the whole process which is a problem.

### DIFF
--- a/sirepo/auth/email.py
+++ b/sirepo/auth/email.py
@@ -150,11 +150,11 @@ def _init_model(base):
         @classmethod
         def delete_changed_email(cls, user):
             with auth_db.thread_lock:
-                cls._session.query(cls).filter(
+                cls._session().query(cls).filter(
                     (cls.user_name == user.unverified_email),
                     cls.unverified_email != user.unverified_email
                 ).delete()
-                cls._session.commit()
+                cls._session().commit()
 
     UserModel = AuthEmailUser
 
--- a/sirepo/auth_db.py
+++ b/sirepo/auth_db.py
@@ -263,7 +263,7 @@ def init_model(callback):
 
 
 @contextlib.contextmanager
-def session_context():
+def session():
     init()
     with sirepo.srcontext.create():
         yield
--- a/sirepo/auth_db.py
+++ b/sirepo/auth_db.py
@@ -15,7 +15,6 @@ import threading
 # limit imports here
 import sirepo.auth_role
 import sirepo.srcontext
-import sirepo.events
 import sirepo.srdb
 
 
@@ -129,10 +128,6 @@ def init():
         # we access a single connection across threads
         connect_args={'check_same_thread': False},
     )
-    sirepo.events.register(PKDict(
-        srcontext_created=_create_session,
-        srcontext_destroyed=_destroy_session,
-    ))
 
     @sqlalchemy.ext.declarative.as_declarative()
     class UserDbBase(object):
@@ -265,8 +260,12 @@ def init_model(callback):
 @contextlib.contextmanager
 def session():
     init()
-    with sirepo.srcontext.create():
-        yield
+    with sirepo.srcontext.create() as c:
+        try:
+            _create_session(c)
+            yield
+        finally:
+            _destroy_session(c)
 
 
 @contextlib.contextmanager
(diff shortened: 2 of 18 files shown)


sirepo/auth_db.py
```diff
@@ -252,6 +274,13 @@ def init_model(callback):
         UserDbBase.metadata.create_all(_engine)
 
 
+@contextlib.contextmanager
+def session_context():
```
robnagler, line 278: This should just be called session(), not session_context(). The implementation is a detail to the clients of session(). They need to call it in a with, but that will raise an exception when it's used outside a with.

sirepo/util.py
```diff
@@ -187,8 +187,10 @@ def flask_app():
     return flask.current_app or None
 
 def in_flask_request():
+    import sys
+    if 'flask' not in sys.modules:
+        return None
     import flask
```
robnagler, line 193: f = sys.modules.get('flask')
return f and f.request or None

sirepo/srunit.py
```diff
@@ -99,6 +99,13 @@ def sim_data(sim_name=None, sim_type=None, sim_types=CONFTEST_DEFAULT_CODES, cfg
     return fc.sr_sim_data(sim_name=sim_name, sim_type=sim_type), fc
 
 
+@contextlib.contextmanager
+def srcontext():
```
robnagler, line 103: I think it is problematic naming this the same name as a module. That aside, I think this is now `auth_db_session()`.

sirepo/auth/email.py
```diff
@@ -150,11 +150,11 @@ def create_token(self):
         @classmethod
         def delete_changed_email(cls, user):
             with auth_db.thread_lock:
-                cls._session.query(cls).filter(
+                cls._session().query(cls).filter(
```
robnagler, line 153: I don't think this is right. I think this should be:

```py
with auth_db.session() as s:
    s.query(class).filter(...).delete()
```

This then also does the commit implicitly. Any time a session() call exits, it either rolls back (finally) or commits (normal flow).

Implicit in the session is an auth_db.thread_lock, too.

sirepo/auth_db.py
```diff
@@ -126,59 +141,66 @@ def init():
...
 
         @classmethod
         def delete_all(cls):
             with thread_lock:
-                cls._session.query(cls).delete()
-                cls._session.commit()
+                cls._session().query(cls).delete()
+                cls._session().commit()
 
         def save(self):
```
robnagler, line 169: I think we should do away with save. I think the "add" should be implicit in the operation.

robnagler (changes requested):
We probably need to talk about this.

sirepo/auth_db.py
```diff
@@ -114,9 +117,21 @@ def db_filename():
 
 
 def init():
-    global _session, _engine, DbUpgrade, UserDbBase, UserRegistration, UserRole
+    def _create_session(_):
```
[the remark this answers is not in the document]
e-carlin: 65fc17e9

sirepo/util.py
e-carlin: a3911a26

sirepo/srunit.py
e-carlin: fc84edd7

sirepo/auth/email.py
e-carlin: I'm confused, doing this would create/destroy the session for every call. We discussed doing that but you said that we should have one session per request.

e-carlin:
@robnagler I addressed the simple stuff. Ping when you have time and we can talk about the trickier things.

sirepo/auth_db.py
```diff
@@ -126,59 +141,66 @@ def init():
         # we access a single connection across threads
         connect_args={'check_same_thread': False},
     )
-    _session = sqlalchemy.orm.Session(bind=_engine)
+    sirepo.events.register(PKDict(
+        srcontext_created=_create_session,
```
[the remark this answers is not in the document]
e-carlin: Events remove d3658cbc

sirepo/auth/email.py
e-carlin: We discussed leaving the auth_db.session() in uri_router._dispatch (every request) and then manually calling it when needed outside of flask (pkcli, tests, supervisor)

sirepo/auth_db.py
e-carlin: We decided not to implement

e-carlin:

sirepo/auth_db.py
```diff
@@ -252,6 +257,35 @@ def init_model(callback):
         UserDbBase.metadata.create_all(_engine)
 
 
+@contextlib.contextmanager
+def session():
+    init()
+    with sirepo.srcontext.create() as c:
```
robnagler, line 263: do we always want to create a context here? it seems like this is a distinct operation.

robnagler (approved):
one comment but merging anyway.

PR #3540, opened 2021-04-27
hydro and flame custom views

e-carlin:


sirepo/template/flash_views.py
```diff
@@ -100,11 +111,24 @@ class SpecializedViews():
             ],
         ),
         physics_Gravity=PKDict(
+            advanced=[],
             basic=[
                 'useGravity',
-                'grav_boundary_type',
-                'physics_Gravity_Constant.gconst',
-                'physics_Gravity_Constant.gdirec',
```
e-carlin, line 107: @moellep I removev this because the physics_Gravity_Constant model doesn't exist in the sedovchamber example. Should we create a way to have models referenced that may not exist?

e-carlin:
Not ready.

sirepo/template/flash_views.py
moellep: Yes, good point - the SpecializedViews view definition in should have all the possible fields, but when the view is added to the schema in __update_views(), it should omit fields which don't exist in the simulation.

PR #3554, opened 2021-04-29
Fix #2347 mpmd parallel

### LINKED ISSUE #2347: mpmd: parallel processing
Not an immediate priority, but new parallel Radia solver and file writing has to be done correctly

mkeilman:
This rearranges the code and adds calls to support general-purpose and Radia-specific parallel execution.

Notes:

- I added ```invoke_single()``` in ```mpi.py``` as an attempt to centralize code found in various ```setup_magnetic_measurement_files()```
- Most of the relevant code changes are in ```geometry.py.jinja```.  Radia appears to want calls of UtiMPI('in' | 'out' | 'barrier') to do its thing
- I expect potential problems with output files but want to see everything _in situ_


sirepo/package_data/template/radia/kickMap.py.jinja
```diff
@@ -1,19 +1,30 @@
 from __future__ import absolute_import, division, print_function
 
 import h5py
+import sirepo.mpi
 from pykern.pkcollections import PKDict
 from sirepo.template import radia_tk
 from sirepo.template import template_common
 
+def _write_h5_data(data, h5_path):
```
robnagler, line 9: unnecessary repetition which would be solved with suggestions above.

sirepo/template/radia_tk.py
```diff
@@ -1,9 +1,12 @@
+
```
robnagler, line 1: header? why is this called "tk"?

sirepo/template/radia_tk.py
```diff
@@ -1,9 +1,12 @@
...
 import numpy
 import radia
 import re
 import sirepo.util
 import sys
 
+from mpi4py import MPI
```
robnagler, line 8: what if there's no MPI. The whole "check for MPI" thing that's in sirepo.mpi is mooted by this. Don't import unqualified.

sirepo/template/radia_tk.py
```diff
@@ -152,6 +155,10 @@ def field_integral(g_id, f_type, p1, p2):
     return radia.FldInt(g_id, 'inf', f_type, p1, p2)
 
 
+def free_symmetries(g_id):
+    return radia.ObjDpl(g_id, 'FreeSym->True')
+
+
```
robnagler, line 161: the line below isinstance is redundant. PKDict is an instance of dict

sirepo/template/radia_tk.py
```diff
@@ -260,6 +271,10 @@ def new_geom_object():
     )
 
 
+def radia_mpi(mpi_state):
```
robnagler, line 274: I think this doesn't belong here.

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -4,6 +4,7 @@ import h5py
 import os
 import re
 from pykern.pkcollections import PKDict
+import sirepo.mpi
```
robnagler, line 7: radia_tk name (below) is confusing to me. I don't think this has anything to do with "tk", and it should be named differently from other modules (jupyter).

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
...
+        template_common.dict_to_h5(data, hf, path=h5_path)
+
+
+def _write_dmp(g_id, f_path):
+    with open(f_path, 'wb') as f:
+        f.write(radia_tk.dump_bin(g_id))
+
+
+{% if isParallel %}
+radia_tk.radia_mpi('in')
```
robnagler, line 94: This needs to be encapsulated like restrict_op_to_first_rank. I would create a module radia_uti, and have an object which is a [Context Manager](https://docs.python.org/3/library/stdtypes.html#context-manager-types), e.g.

```py
with sirepo.template.radia_uti.MPI() as m:
      m.barrier()
```

The `in` and `off` would be hidden in the context manager, and would be protected against mpi running or not. barrier would be a nop when mpi was not running. That's the reason for the object.

sirepo/template/radia_tk.py
```diff
@@ -260,6 +271,10 @@ def new_geom_object():
...
+    return radia.UtiMPI(mpi_state)
+
+
 def reset():
```
robnagler, line 278: move this to radia_uti (see above)

robnagler (changes requested):
I think the work on dict_to_h5 would be useful to put in this PR.

sirepo/package_data/static/json/radia-schema.json
```diff
@@ -465,12 +465,13 @@
...
         },
         "solver":  {
+            "jobRunMode": ["Execution Mode", "JobRunMode", "parallel"],
             "lastSolved": ["", "Integer", "-1"],
             "maxIterations": ["Max iterations", "Integer", 1500, "", 1],
             "method": ["Method", "SolveMethod", "0"],
             "precision": ["Precision (T)", "Float", 0.00010, "", ""],
             "result": ["Result", "String", ""],
-            "sbatchCores": ["Cores", "Integer", 1],
+            "sbatchCores": ["Cores", "Integer", 4],
```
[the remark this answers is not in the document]
mkeilman: 82b1096

sirepo/mpi.py
```diff
@@ -13,6 +13,30 @@
 import sys
 
 
+def invoke_single(fn, *args):
```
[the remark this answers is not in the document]
mkeilman: 59802b4

sirepo/mpi.py
```diff
@@ -13,6 +13,30 @@
...
+        c = mpi4py.MPI.COMM_WORLD
+        if c.Get_size() > 1:
+            r = c.Get_rank()
+    except Exception:
+        pass
+    if r == 0:
+        try:
+            res = fn(*args)
+        except Exception as e:
+            pkdc(f'Exception in invoked function {fn.__name__}: {e}')
```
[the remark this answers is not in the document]
mkeilman: 365fb87

sirepo/mpi.py
```diff
@@ -13,6 +13,30 @@
...
+    r = 0
+    res = None
+    try:
+        import mpi4py.MPI
+        c = mpi4py.MPI.COMM_WORLD
+        if c.Get_size() > 1:
+            r = c.Get_rank()
+    except Exception:
+        pass
+    if r == 0:
```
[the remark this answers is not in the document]
mkeilman: 78ba2ef

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
...
+    for obj in sirepo_objs:
+        o = PKDict(obj)
+        if o.model not in _MODELS or o.id in id_map.values():
+            continue
+        o_id = _add_object(o, radia_objs, id_map)
+        id_map[o_id] = o.id
+    g_id = radia_tk.build_container(radia_objs)
+{% endif %}
+    # free symmetries to test parallel?
+    #g_id = radia_tk.free_symmetries(g_id)
```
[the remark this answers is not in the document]
mkeilman: eb8c2c2

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
...
+{% if doReset %}
+radia_tk.reset()
+{% endif %}
 
+g_id, id_map = _build_radia_geom()
 g_data = radia_tk.geom_to_data(g_id, name='{{ geomName }}')
-with h5py.File('{{ dataFile }}', 'a') as hf:
-    template_common.dict_to_h5(g_data, hf, path='{{ h5ObjPath }}')
-    template_common.dict_to_h5(id_map, hf, path='idMap')
+sirepo.mpi.invoke_single(_write_h5_data, g_data, '{{ h5ObjPath }}')
```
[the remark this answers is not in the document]
mkeilman: Also used with ```_write_dmp()```

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
         radia_objs.append(g_id)
     return g_id
 
-id_map = PKDict()
+def _build_radia_geom():
+    id_map = PKDict()
+    radia_objs = []
 {% if isRaw %}
```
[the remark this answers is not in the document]
mkeilman: 7cff407

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
...
+        id_map[o_id] = o.id
+    g_id = radia_tk.build_container(radia_objs)
+{% endif %}
+    # free symmetries to test parallel?
+    #g_id = radia_tk.free_symmetries(g_id)
+    return g_id, id_map
+
+
+def _write_h5_data(data, h5_path):
+    with h5py.File('{{ dataFile }}', 'a') as hf:
```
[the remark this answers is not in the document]
mkeilman: c1f0ea7

sirepo/package_data/template/radia/geometry.py.jinja
```diff
@@ -58,35 +57,58 @@ def _add_object(o):
...
+        template_common.dict_to_h5(data, hf, path=h5_path)
```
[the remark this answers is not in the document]
mkeilman: c1f0ea7

Incomplete, because I'm not sure what to do with what is now ```write_dict_to_h5()```.  Is ```with h5py.File``` to move there?  If so, is it better to have once at the top or every time we call ```create_dataset```?  Any issues for a recursive function?

sirepo/package_data/template/radia/geometry.py.jinja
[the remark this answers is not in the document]
mkeilman: d03aa7a

sirepo/template/radia_tk.py
mkeilman: 20e2ea5

```import MPI``` was left over from messing around with timing the solve, but point taken

File: sirepo/template/radia_tk.py (hunk 2 above)
Comment (mkeilman) [reply sirepo/template/radia_tk.py:3]: 20e2ea5

(just a skeleton atm)

File: sirepo/package_data/template/radia/geometry.py.jinja (hunk 6 above)
Comment (mkeilman) [reply sirepo/package_data/template/radia/geometry.py.jinja:7]: Was originally a direct copy from ```jupyter-rs-radia``` with an eye toward a single library for both.  I can't remember why I preferred a "toolkit" to "utilities".  So is the idea to rename to radia_util and put the new stuff in there as well?

File: sirepo/template/radia_tk.py (hunk 2 above)
Comment (mkeilman) [reply sirepo/template/radia_tk.py:3]: a537825

File: sirepo/package_data/template/radia/geometry.py.jinja (hunk 6 above)
Comment (mkeilman) [reply sirepo/package_data/template/radia/geometry.py.jinja:7]: a537825

File: sirepo/mpi.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (mkeilman) [reply sirepo/mpi.py]: 8e2680b

File: sirepo/package_data/template/radia/geometry.py.jinja [hunk 18]
```diff
@@ -97,11 +119,15 @@ f = radia_tk.get_field(g_id, '{{ fieldType }}', {{ fieldPoints }})
...
-    with h5py.File('{{ dataFile }}', 'a') as hf:
-        template_common.dict_to_h5(g_data, hf, path='{{ h5FieldPath }}')
+    sirepo.mpi.invoke_single(_write_h5_data, g_data, '{{ h5FieldPath }}')
 {% endif %}
 
-with open('{{ dmpOutputFile }}', 'wb') as f:
-    f.write(radia_tk.dump_bin(g_id))
+sirepo.mpi.invoke_single(_write_dmp, g_id, '{{ dmpOutputFile }}')
+{% if isParallel %}
+radia_tk.radia_mpi('off')
```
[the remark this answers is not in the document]
mkeilman: a3fd7bd

sirepo/template/radia_tk.py
mkeilman: a3fd7bd

sirepo/package_data/template/radia/geometry.py.jinja
mkeilman: a3fd7bd

This works when everything is "correct" but I don't know what to do when they are not.  When I run as though parallel is not available (invoking ```run()``` instead of ```run_background()``` I get 

```Attempting to use an MPI routine before initializing MPICH```

and can't seem to catch it or otherwise guard against it. Do I still need to check isParallel in the jinja template?

I am also unclear as to how I can handle or test the absence of mpi

File: sirepo/template/radia_tk.py (hunk 8 above)
Comment (mkeilman) [reply sirepo/template/radia_tk.py]: a3fd7bd

File: sirepo/template/radia_tk.py (hunk 4 above)
Comment (mkeilman) [reply sirepo/template/radia_tk.py:183]: f853973

File: sirepo/package_data/template/radia/geometry.py.jinja (hunk 7 above)
Comment (mkeilman) [reply sirepo/package_data/template/radia/geometry.py.jinja]: bace751

needed to do this to avoid the init error above

**Comment by mkeilman** [comment]
Need some clarifications at this point...

File: sirepo/package_data/template/radia/geometry.py.jinja (hunk 6 above)
Comment (robnagler) [reply sirepo/package_data/template/radia/geometry.py.jinja:7]: "tk" didn't translate to "toolkit" in my mind. "util" maps to "uti" in radia world, which is why I thought of it. We also use util other places in sirepo. I think I was confused with VTK and radias-vtk. Anyway, thanks for changing to radia_util. :)

File: sirepo/template/radia_util.py [hunk 19]
```diff
@@ -35,6 +41,25 @@
 _ZERO = [0, 0, 0]
 
 
+class MPI:
+    def __init__(self):
+        radia.UtiMPI('in')
```
robnagler, line 46: This is what I meant:

```py
def __init__(self):
    # Null op for when not in MPI
    self._uti_mpi = lambda x: None
    try:
        import mpi4py.MPI
      	    self._uti_mpi = radia.UtiMPI
    except Exception:
	pass

def __enter__(self):
    self._uti_mpi('in')
    return self

def __exit__(self):
    self._uti_mpi('off')

def barrier(self):
    self._uti_mpi('barrier')
```

robnagler (changes requested):

sirepo/template/radia_util.py
mkeilman: 8705a6a

sirepo/package_data/template/radia/geometry.py.jinja
mkeilman: 8705a6a

sirepo/package_data/template/radia/geometry.py.jinja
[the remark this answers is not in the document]
mkeilman: 918c953 7a01b53 0c28245

 I answered my own question: ```with h5py.File()``` can live in ```write_dict_to_h5()``` as long it's done only when actually creating the dataset.  Some other points:

- The only access mode that makes sense is 'a' so I moved that in as well
- ```kickMap.py.jinja``` is not currently used but probably will be since it should be susceptible to parallel processing.  I  updated it.  The ```_write_dict_to_h5()``` there is repeated but can be moved to a separate jinja file when ready
- warpvnd uses ```template_common.write_dict_to_h5()``` in one place only, so it has been updated.  Its jinja files use a local version and I am not going to touch those for now
- I am also not going to change the corresponding ```h5_to_dict()``` yet since other codes use it and this PR needs to get folded in.  Should make a separate issue for it

sirepo/template/radia_tk.py
mkeilman: 32d399c

sirepo/package_data/template/radia/radia_export.py.jinja
```diff
@@ -14,7 +14,7 @@ _MODELS = (_MODEL_BOX, _MODEL_GROUP, _MODEL_GROUP_UND, _MODEL_OBJECT)
 radia_objs = []
 sirepo_objs = {{ objects }}
 
-{% include 'radia_tk.py' %}
+{% include 'radia_util.py' %}
```
robnagler, line 17: Does this actually work? It would seem a simple import would be good enough. radia_util.py imports sirepo.util so it seems that an import is required anyway.

sirepo/template/radia_util.py
```diff
@@ -242,10 +277,11 @@ def kick_map(
         g_id, begin, dir_long, num_periods, period_length, dir_trans, range_trans_1,
         num_pts_trans_1, range_trans_2, num_pts_trans_2
     ):
-    return radia.FldFocKickPer(
```
robnagler, line 245: Why introduce a temporary variable?

sirepo/template/template_common.py
```diff
@@ -270,23 +270,36 @@ def compute_plot_color_and_range(plots, plot_colors=None, fixed_y_range=None):
     return y_range
 
 
-def dict_to_h5(d, hf, path=None):
-    if path is None:
-        path = ''
+def write_dict_to_h5(d, file_path, h5_path=None):
+    """ Store the contents of a dict in an h5 file starting at the provided path.
+    Stores the data recursively so that
```
robnagler, line 275: In order for this to work, you need to have a `::` so that napoleon-sphinx picks it up as a code block.

sirepo/template/template_common.py
```diff
@@ -270,23 +270,36 @@ def compute_plot_color_and_range(plots, plot_colors=None, fixed_y_range=None):
...
+        {a: A, b: {c: C, d: D}}
+    maps the data to paths
+        <h5_path>/a   -> A
+        <h5_path>/b/c -> C
+        <h5_path>/b/d -> D
+
+    h5_to_dict() performs the reverse process
```
robnagler, line 282: Would be good to describe the `Args:`.

Also:

```txt
`h5_to_dict` performs the reverse process
```

sirepo/template/warpvnd.py
```diff
@@ -1085,8 +1085,11 @@ def _stl_polygon_file(filename):
...
-            template_common.dict_to_h5(data, hf, path='/')
+        template_common.write_dict_to_h5(
+            data,
+            _SIM_DATA.lib_file_write_path(_stl_polygon_file(data.file)),
+            h5_path='/'
+        )
     except Exception as e:
-        pkdlog('!save_stl_polys FAIL: {}', e)
+        pkdlog('save_stl_polys error={}', e)
         pass
```
robnagler, line 1095: `pass` is unnecessary.

robnagler (approved):
The comments are mostly about comments so merging but note them.

sirepo/package_data/template/radia/radia_export.py.jinja
mkeilman: It does work, in concert with this in ```template/radia.py```:

```
    return template_common.render_jinja(SIM_TYPE, v, j_file, jinja_env=PKDict(loader=jinja2.PackageLoader('sirepo', 'template')))
```

This is basically a sop to Oleg but incomplete as you can see...

sirepo/template/radia_util.py
mkeilman: When value involves a method call with a lot of parameters I think it's easier to see what's going on with a temp variable.  One fewer pair of parens to navigate.  Maybe it's just me, and I try not to overdo it

PR #3563, opened 2021-05-04
Implement all of the custom view that existed in old flash-schema.json

e-carlin:


sirepo/template/flash_parser.py
```diff
@@ -344,7 +344,9 @@ def generate_schema(self):
 
     @classmethod
     def model_name_from_flash_unit_name(cls, text):
-        return '_'.join(filter(lambda x: not re.search(r'Main$', x), text.split('/')))
+        # TODO(e-carlin): discuss with pjm. Main units have different vars than the non-main
```
e-carlin, line 347: @moellep what do you think of this code? From what I found there is an important distinction between the Unit and UnitMain. [Every unit is required](http://flash.uchicago.edu/site/flashcode/user_support/flash_ug_devel/node28.html) to have a Main subunit but they aren't always included together.

For example, in the SedovChamber example physics/Diffuse is included but physics/Diffuse/DiffuseMain is not. Many of the fields we had in the old flash schema were from the  DiffuseMain unit. So if we mangle the names like you were doing we can't tell whether the DiffuseMain unit is included or just the top-level Diffuse unit is. If we can't tell then we may try to include fields which aren't actually included in the sim. Separating them out like this prevents this.

You can see the difference between Diffuse and DiffuseMain [here](http://flash.uchicago.edu/site/flashcode/user_support/rpDoc_4p3.py?submit=rp_Diffuse.txt)

Issue #2201, opened 2020-01-23
getApplicationData via job_api


moellep:
When calling into an app using getApplicationData or running a fixup, any code specific calls need to be run outside of the main webserver. For example, all srwlib calls in template/srw.py should get moved into a separate module which can create a subprocess to execute. Consolidate srw fixups into one routine for clarity, possibly always run it as a subprocess.

robnagler:
To be clear, it needs to run in pkcli.job_cmd, which runs in a separate docker container so that there are no privilege issues and so that the main web server doesn't block a thread for an extended period of time.

robnagler:
I think in the end, we will need to keep fastcgi, which is "pretty fast". Application data is just another `OP_ANALYSIS` so it will run over the same channel. Perhaps we might even have simulations started this way, because you can't run a sequential simulation at the same time as a simulation frame. They all run in the same docker container which is assigned one core so there would be no point in overloading it.

e-carlin:
There has been some email and in person dicussion about this issue. Putting the notes from those conversations here.

We've decided to try and categorize the types of get_application_data calls and to use new apis to incrementally move over. 

I did an initial categorizing of all of [the existing get_application_data calls](https://github.com/radiasoft/sirepo/files/6418267/get-application-data.txt). 

@robnagler responded with:

> I think the calls need to be categorized as follows according to "level":
> A. statelesCompute - Purely functional/stateless but requiring access to the code. An agent has to be running, but it does not go to NERSC/sbatch, because it is just getting access to the code, not to something about the data/compute.
> B. simulation_db - read the simulation directory that contains sirepo-data.json or lib directory but not runDir
> C. sim_db - access to the runDir and/or runDirs of other codes. This may requiring hitting NERSC/sbatch agents
> D. analysisJob - something that reads/writes from an existing runDir(s)
> E. computeJob - something that rewrites the runDir
>
> (A) pass parameters and get results. No reference to any external data. 
>
> (B) needs to run in Flask, because it does not need access to a code, and all the files reside there. Hopefully this request is fast, but it if it isn't, we would need to optimize by using a better database. 
>
> (C) can also get information from (B), but it doesn't write anything, just read. It might get access to all runDirs, e.g. jspec's getting twiss parameters. I make the distinction on writing, but maybe that's not a good distinction. It might be that (C) and (D) are the same.

e-carlin:
@robnagler Mostly makes sense. Some questiona:
[...]

PR #3565, opened 2021-05-05
WIP: #2201: Implement statelessCompute API

e-carlin:
@robnagler here is the basic structure for statelessCompute. Lmk what you think. Once you've signed off I'll move all of the relevant get_application_data calls over and write a test.

One open question, @moellep says these calls need to be "fast" (ex for rpn evaluation in elegant). The slowest part of this code is the first call where the agent must start. We discussed starting the agent when an app is opened to help speed things up but I don't believe we came up with a firm answer. Do you have a preference on how to speed things up?


sirepo/job_api.py
```diff
@@ -174,6 +174,18 @@ def api_sbatchLogin():
     return _request(_request_content=r)
 
 
+@api_perm.require_user
+def api_statelessCompute():
+    return _request(
+        req_data=PKDict(
+            computeJobHash='unused',
+            report='statelessCompute',
+            **sirepo.http_request.parse_post().req_data
```
robnagler, line 183: i think this should be before, not after. The problem is that the external caller could override any value including computeJobHash and report, which are important to not override (well, the report at least).

sirepo/package_data/static/json/schema-common.json
```diff
@@ -150,6 +150,7 @@
         "simulationData": "/simulation/<simulation_type>/<simulation_id>/<pretty>/?<section>",
         "simulationFrame": "/simulation-frame/<frame_id>",
         "simulationSchema": "/simulation-schema",
+        "statelessCompute": "/stateless-compute",
```
robnagler, line 153: "t" comes after "r". C-x C-l in our emacs will sort a region.

sirepo/pkcli/job_agent.py
```diff
@@ -352,7 +353,8 @@ async def _fastcgi_op(self, msg):
             self._fastcgi_msg_q = sirepo.tornado.Queue(1)
             pkio.unchecked_remove(self._fastcgi_file)
             m.fastcgiFile = self._fastcgi_file
-            # Runs in a agent's directory, but chdir's to real runDirs
+            # Runs in an agent's directory and chdirs to real runDirs.
+            # Except in stateless_compute which doesn't interact with the db.
```
robnagler, line 357: is the current directory safe to execute from?

sirepo/pkcli/job_cmd.py
```diff
@@ -172,8 +169,10 @@ def _recv():
             m = _recv()
             if not m:
                 return
-            m.runDir = pkio.py_path(m.runDir)
-            with pkio.save_chdir(m.runDir):
+            m.runDir = pkio.py_path(m.runDir) if m.runDir else None
```
robnagler, line 172: Encapsulate this in a function, including setting m.runDir.

sirepo/pkcli/job_cmd.py
```diff
@@ -241,6 +233,20 @@ def _do_sequential_result(msg, template):
     return r
 
 
+def _do_stateless_compute(msg, template):
+    try:
+        return template.stateless_compute(msg.data)
```
robnagler, line 238: One of the problems with getApplicationData is that there's no API for the sub functions. Solve the dispatch so that it calls something like stateless_compute_setup_command.

robnagler (changes requested):

sirepo/pkcli/job_agent.py
e-carlin: I believe so but the truth is I'm not exactly sure what to be on the lookout for that wouldn't be safe. 

These jobs shouldn't read or wirte any files so I don't think anything will be exposed. In local and docker it is userDir/agent-local/<agent-id>. In sbatch It is /global/cscratch1/sd/{sbatch_user}/sirepo-dev .  So, I think in all cases it is just user data.

sirepo/pkcli/job_cmd.py
e-carlin: 17a94973c

sirepo/package_data/static/json/schema-common.json
e-carlin: e80129335

sirepo/job_api.py
e-carlin: 3c17ea4e1

e-carlin:
@robnagler all comments addressed. If it looks good just leave a comment and I'll move over the other get_application_data calls and write a test.

e-carlin:
Yes, thanks.

robnagler (changes requested):

sirepo/job_api.py
```diff
@@ -174,6 +174,19 @@ def api_sbatchLogin():
     return _request(_request_content=r)
 
 
+@api_perm.require_user
+def api_statelessCompute():
+    return _request(
```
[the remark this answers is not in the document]
e-carlin: 75579bd26

robnagler:
Yes, looks good.

e-carlin:
@robnagler ready for review. All of the relevant get_application_data calls have been converted to statelessCompute. A test has also been written. FWIW [this is the diff](https://github.com/radiasoft/sirepo/compare/df2780b89..a011bc763) between when you last reviewed and now

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2242,6 +2242,16 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, $http,
         );
     };
 
+    self.statelessCompute = function(appState, data, callback) {
+        const m = SIREPO.APP_SCHEMA.constants.methods.statelessCompute;
```
robnagler, line 2246: I don't think this is a good idea. Maintaining a list of methods in three places is unnecessary. It will be a programming error if the method is incorrect. An assertion on the server is all we need. (The issue of how that assertion is displayed is distinct and as you noted in our call this week, problematic currently, but that's a different bug.) Consider the other problem of maintaining a list of correct parameters, which is also important, and also a programming error that the server should assert.

sirepo/srschema.py
```diff
@@ -164,6 +165,15 @@ def validate(schema):
         for src in schema.dynamicModules[t]:
             pkresource.filename(src[1:])
     _validate_strings(schema.strings)
+    _validate_constants_methods(schema.constants.get('methods', PKDict()))
+
+
+def _validate_constants_methods(methods):
```
robnagler, line 171: Any validation should be against the template itself: do these methods exist? Which then begs the question, why not just generate the list from the template, which then begs the question: why bother?

tests/template/elegant_test.py
```diff
@@ -27,7 +27,7 @@ def test_file_iterator():
     from sirepo.template.lattice import LatticeUtil
     from pykern.pkunit import pkeq
     data = _find_example('bunchComp - fourDipoleCSR')
-    v = LatticeUtil(data, _elegant()._SCHEMA).iterate_models(
+    v = LatticeUtil(data, _elegant().SCHEMA).iterate_models(
```
robnagler, line 30: This seems not right. Shouldn't it get the information from sim_data?

robnagler (changes requested):

sirepo/package_data/static/js/sirepo.js
e-carlin: Ok. I waffled on this quite a bit. I had a feeling you might not like it:) I liked having everything enumerated but I suppose a simple grep in template will do the same.

d6fccd452

sirepo/srschema.py
e-carlin: d6fccd452

tests/template/elegant_test.py
e-carlin: Can you elaborate? Should it get the schema from sim_data (ex `_elegant()._SIM_DATA.schema()`)?

robnagler: yes, there shouldn't be a sharing of that variable unless it is necessary.

e-carlin: Thanks b24452abc

e-carlin:

robnagler:
There's a merge conflict

e-carlin:
Rebased with conflict fixed. Ready for merge.

PR #3566, opened 2021-05-06
Issue/3558 single lock

e-carlin:


sirepo/util.py
```diff
@@ -192,21 +189,6 @@ def in_flask_request():
...
-    assert not cfg
-    cfg = pkconfig.init(
-        create_token_secret=('oh so secret!', str, 'used for internal test only'),
-    )
-    if server_context:
-        SIMULATION_DB_LOCK = threading.RLock()
-        return
-    # Use nullcontext instead of an actual lock because supervisor is in tornado
-    # which is single threaded
-    SIM_DB_FILE_LOCK = contextlib.nullcontext()
```
e-carlin, line 207: I ended up removing this. Moving the single lock in here means that util needed to be initialized in many new places in the app (for auth). I don't think having these two "locks" provided enough value that it was worth the added complication of requiring that util be initialized in a special way. In db0302aa I got it all working but it seemed messy. I also think keeping the SIM_DB_FILE_LOCK added some confusion because we are moving to a one lock model.

robnagler: That's fine

robnagler (approved):
Looks good, thanks.

PR #3567, opened 2021-05-07
Fix radiasoft/rsdockerspawner#25: prevent blocked jupyterhublogin use…

e-carlin:
…rs from accessing the system

There were a couple problems that made it so blocked users could still
access the system:
1. In sirepo.jupyterhub.Authenticator.refresh_user we validated the user but we
did not validate that they still had the sim_type_jupyterhublogin role
2. In require_sim_type (which is how we validated the user has the role for the
sim type) we only checked proprietary_sim_types. jupyterhublogin is a
default_proprietary_sim_type

@robnagler when we delete the `jupyterhublogin` role from a user we need to also stop the user's server. This is the only way I've found that guranteees that they can no longer access the system.

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -334,8 +334,7 @@ def require_auth_basic():
 
 
 def require_sim_type(sim_type):
-    if sim_type not in sirepo.feature_config.cfg().proprietary_sim_types:
-        # only check role for proprietary_sim_types
+    if sim_type not in sirepo.feature_config.cfg().all_proprietary_sim_types:
         return
     if not _is_logged_in():
         # If a user is not logged in, we allow any sim_type, because
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -121,6 +121,9 @@ def _init():
     i = _cfg.proprietary_sim_types.intersection(_cfg.default_proprietary_sim_types)
     assert not i, \
         f'{i}: cannot be in proprietary_sim_types and default_proprietary_sim_types'
+    _cfg.all_proprietary_sim_types = frozenset(
+        _cfg.proprietary_sim_types.union(_cfg.default_proprietary_sim_types),
+    )
     s = set(
         _cfg.sim_types or (
             _PROD_FOSS_CODES if pkconfig.channel_in('prod') else _FOSS_CODES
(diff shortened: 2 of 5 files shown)


sirepo/auth/__init__.py
```diff
@@ -334,8 +334,7 @@ def require_auth_basic():
 
 
 def require_sim_type(sim_type):
-    if sim_type not in sirepo.feature_config.cfg().proprietary_sim_types:
-        # only check role for proprietary_sim_types
+    if sim_type not in sirepo.feature_config.cfg().all_proprietary_sim_types:
```
robnagler, line 337: maybe we need to call this "auth_controlled" instead of "all_proprietary". Something like that.

sirepo/feature_config.py
```diff
@@ -121,6 +121,9 @@ def b(msg, dev=False):
     i = _cfg.proprietary_sim_types.intersection(_cfg.default_proprietary_sim_types)
     assert not i, \
         f'{i}: cannot be in proprietary_sim_types and default_proprietary_sim_types'
+    _cfg.all_proprietary_sim_types = frozenset(
```
robnagler, line 124: `_cfg` should be restricted to pkconfig only, since the object should be treated as an abstraction. It should not be automatically assumed it is a PKDict (indeed it didn't used to be) and could just be augmented/edited to. If it is not something configurable, store elsewhere. It could be a function on feature_cfg.auth_controlled_sym_types.

sirepo/jupyterhub.py
```diff
@@ -58,14 +57,20 @@ async def authenticate(self, handler, data):
...
                 # https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.refresh_user
                 return False
             return True
 
+    def _check_permissions(self):
+        sirepo.auth.require_user()
+        sirepo.auth.require_sim_type('jupyterhublogin')
+
+
+
```
robnagler, line 73: too many lines

robnagler (approved):
Some comments for a later PR. I wanted to get this merged.

sirepo/feature_config.py
e-carlin: https://github.com/radiasoft/sirepo/pull/3573

PR #3588, opened 2021-05-14
Fix #3581: Create jupyterhub user given email that may not be Sirepo …

### LINKED ISSUE #3581: pkcli.jupyterhublogin.create_user add display_name
I realized that we'll need a display name to be added. Please add a non-optional parameter to create_user.

e-carlin:
…user

@robnagler take a close look at this one. I'm not 100% certain I got everything right regarding the edge cases of registration states a user may be in. I also had to refactor a small amount of auth code.

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -141,6 +141,14 @@ def complete_registration(name=None):
     cookie.set_value(_COOKIE_STATE, _STATE_LOGGED_IN)
 
 
+def create_new_user(uid_generated_callback, module):
+    import sirepo.simulation_db
+    u = sirepo.simulation_db.user_create()
+    uid_generated_callback(u)
+    _create_roles_for_new_user(u, module.AUTH_METHOD)
+    return u
+
+
 def guest_uids():
     """All of the uids corresponding to guest users."""
     return auth_db.UserRegistration.search_all_for_column('uid', display_name=None)
@@ -241,9 +249,7 @@ def login(module, uid=None, model=None, sim_type=None, display_name=None, is_moc
             # This handles the case where logging in as guest, creates a user every time
             _login_user(module, uid)
         else:
-            import sirepo.simulation_db
-            uid = sirepo.simulation_db.user_create(lambda u: _login_user(module, u))
-            _create_roles_for_new_user(module.AUTH_METHOD)
+            uid = create_new_user(lambda u: _login_user(module, u), module)
         if model:
             model.uid = uid
             model.save()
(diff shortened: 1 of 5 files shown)


sirepo/auth/__init__.py
```diff
@@ -495,19 +501,21 @@ def user_name():
     )
 
 
-def user_registration(uid):
+def user_registration(uid, display_name=None):
```
robnagler, line 504: Remove =None

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -25,14 +33,54 @@ def create_user(email):
...
     import sirepo.template
 
+    def maybe_create_sirepo_user(module, email, display_name):
+        u = module.unchecked_user_by_user_name(email)
+        if u:
+            # Fully registered email user
+            r = sirepo.auth_db.UserRegistration.search_by(uid=u)
+            assert r.display_name, \
+                f'uid={u} authorized AuthEmailUser record but no UserRegistration.display_name'
+            if not r.display_name == display_name:
```
robnagler, line 43: Don't change the name, just leave it. The purpose is not to ensure a name. Let the person's registration override what we do

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -25,14 +33,54 @@ def create_user(email):
...
+                )
+                r.display_name = display_name
+                r.save()
+            return u
+        m = module.AuthEmailUser.search_by(unverified_email=email)
+        if m:
+            # Email user that needs to complete registration (no display_name but have unverified_email)
+            assert sirepo.auth.need_complete_registration(m), \
+                'email={email} has no display_name but does not need to complete registration'
+            pkcli.command_error(
```
robnagler, line 58: good

robnagler (changes requested):

sirepo/auth/__init__.py
e-carlin: Why? In the non-pkcli case we don't have a display name when this is called. Should I just pass in None in that case?

sirepo/pkcli/jupyterhublogin.py
e-carlin: b5e8a8431

e-carlin:
@robnagler comments addressed. I left [one question](https://github.com/radiasoft/sirepo/pull/3588#discussion_r634485167).

sirepo/auth/__init__.py
robnagler: I was confused. I thought this was only for the pkcli. That's fine. leave as is.

PR #3595, opened 2021-05-18
Fix #3571: Create simulationDb API

### LINKED ISSUE #3571: create simulation_db API
#2201

e-carlin:
This API is for reading the simulation directory that contains
sirepo-data.json or lib directory but not runDir

This branch is based on issue/3564-statelessCompute-api. Review and merge #3565 before reviewing this.


sirepo/job_api.py
```diff
@@ -151,6 +151,29 @@ def api_runStatus():
...
+def api_sbatchLogin():
+    r = _request_content(
+        PKDict(computeJobHash='unused', jobRunMode=sirepo.job.SBATCH),
+    )
+    r.sbatchCredentials = r.pkdel('data')
+    return _request(_request_content=r)
+
+
+@api_perm.require_user
+def api_simulationDb():
```
robnagler, line 164: I think we need a different name. This is about doing something that requires the simulation_db. It's similar to statelessCompute, but one level higher, so it should be a "compute" of some sort. statefulAnalysis, statefulCompute, statefulQuery, etc.

sirepo/job_api.py
```diff
@@ -151,6 +151,29 @@ def api_runStatus():
...
+    return _request(
+        jobRunMode=sirepo.job.SEQUENTIAL,
+        req_data=PKDict(
+            **sirepo.http_request.parse_post().req_data,
+        ).pkupdate(
+            computeJobHash='unused',
```
robnagler, line 170: This is repeated.

sirepo/pkcli/job_cmd.py
```diff
@@ -238,6 +238,13 @@ def _do_sequential_result(msg, template):
     return r
 
 
+def _do_simulation_db(msg, template):
+    try:
+        return template_common.simulation_db_dispatch(msg.data)
```
robnagler, line 243: This looks like a pattern to encapsulate. Perhaps these requests are actually one API with a method attached with attributes describing what level of "compute" they require.

tests/template/elegant_test.py
```diff
@@ -27,7 +27,7 @@ def test_file_iterator():
     from sirepo.template.lattice import LatticeUtil
     from pykern.pkunit import pkeq
     data = _find_example('bunchComp - fourDipoleCSR')
-    v = LatticeUtil(data, _elegant()._SCHEMA).iterate_models(
+    v = LatticeUtil(data, _elegant().SCHEMA).iterate_models(
```
robnagler, line 30: I don't think we should be exporting this constant for this purpose. A unit test can reach inside for _SCHEMA, but really it should ask sim_data.

sirepo/job_api.py
e-carlin: l went with "statefulCompute". I think anaylsis should be left for things that need runDirs and the relationship between stateful and stateless is nice. daba63620

sirepo/pkcli/job_cmd.py
e-carlin: 43797f40e . Thoughts? I'm shying away from using one API because I think that leads us back down the road of IOCTL.

tests/template/elegant_test.py
e-carlin: It is really exported for use [here](https://github.com/radiasoft/sirepo/pull/3595/files#diff-687323f8346dfdf343f1eee4795f4492f47f05a6fae712ee4b920bd63897e68aR579). That ok?

sirepo/job_api.py
e-carlin: 27fd4c7d2

robnagler:
Is this ready to merge? Sorry I got behind, and now there are merge conflicts.

e-carlin:
Np - I've fixed the conflicts and done some testing again. Ready.

tests/template/elegant_test.py
robnagler: Do you mean [here](https://github.com/radiasoft/sirepo/pull/3595/files#diff-24496517cc6cc5369fc69a13a07158c876115ef9eec8ceb098a25b374b083f7fL748)? If so, this is not exported, it's part of an object. There's not particular reason to use the constant vs calling sim_data or simulation_db in a test. It's not a performance issue.

tests/template/elegant_test.py
```diff
@@ -42,8 +42,10 @@ def _elegant():
 
 def _expr(expr, expect, variables=None):
     from pykern.pkunit import pkok, pkfail
-    res = _elegant().get_application_data(PKDict(
+    from sirepo.template import template_common
+    res = template_common.stateful_compute_dispatch(PKDict(
         method='rpn_value',
+        simulationType='elegant',
         value=expr,
         variables=variables or {},
```
robnagler, line 50: Change to PKDict() when you touch code like this.

robnagler (changes requested):

tests/template/elegant_test.py
e-carlin: Sorry - not sure why my link didn't have a better line number. It is exported for use inside of [`template.common.stateful_compute_dispatch`](https://github.com/radiasoft/sirepo/pull/3595/files#diff-687323f8346dfdf343f1eee4795f4492f47f05a6fae712ee4b920bd63897e68aR592)

tests/template/elegant_test.py
e-carlin: e3c31a2ed

robnagler:
I thought this was merged. Shoudl it be?

robnagler:
We should merge this, but there are conflicts. I think the concept is sound.

e-carlin:
Ok, will fix today and report back when done.

e-carlin:
@robnagler all merge conflicts fixed. Ready for another review.

This has been sitting for a while and has a > 0% chance of breaking things so I squashed into 1 commit and rebased on master. I think this makes the history easier to understand and makes it easier to rollback if needed. If you have the branch checked out locally you'll need to 
```
git checkout issue/3571-simulation-db-api
git pull # This will complain about problems from upstream. Next command will fix
git reset --hard origin/issue/3571-simulation-db-api # This will blow away all of your local changes (if there are any)
```

robnagler (approved):
This is going to be one of those that we just have to try out in dev after the merge.

PR #3603, opened 2021-05-20
Fix #3593: Prevent labels from overlapping with beamline elements

### LINKED ISSUE #3593: webcon: don't overlap element labels with beamlines
Element labels shouldn't overlap with the beamline elements. This is mainly a problem with long lines. The monitor labels (which are above the beamline) are in too close. 

Screen Shot 2021-05-18 at 2 08 49 PM

e-carlin:
In testing it seemed that calling `getBoundingClientRect()` on the label
div multiple times yielded different results so I could eliminate
keeping track of `yOffset`.

### DIFF
--- a/sirepo/package_data/static/js/controls.js
+++ b/sirepo/package_data/static/js/controls.js
@@ -701,11 +701,10 @@ SIREPO.app.directive('latticeFooter', function(appState, controlsService, lattic
             }
 
             function rectanglesOverlap(pos1, pos2) {
-                if (pos1.rect.left > pos2.rect.right || pos2.rect.left > pos1.rect.right) {
+                if (pos1.left > pos2.right || pos2.left > pos1.right) {
                     return false;
                 }
-                if (pos1.rect.top + pos1.yOffset > pos2.rect.bottom + pos2.yOffset
-                    || pos2.rect.top + pos2.yOffset > pos1.rect.bottom + pos1.yOffset) {
+                if (pos1.top > pos2.bottom || pos2.top > pos1.bottom) {
                     return false;
                 }
                 return true;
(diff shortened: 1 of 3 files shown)


sirepo/package_data/static/js/sirepo-lattice.js
```diff
@@ -1263,7 +1264,7 @@ SIREPO.app.directive('lattice', function(appState, latticeService, panelState, p
             const ABSOLUTE_POSITION_TYPE = 'absolutePosition';
             $scope.plotStyle = $scope.flatten ? '' : 'cursor: zoom-in;';
             $scope.isClientOnly = true;
-            $scope.margin = $scope.flatten ? 0 : 3;
+            $scope.margin = $scope.margin || ($scope.flatten ? 0 : 3);
```
e-carlin, line 1267: @moellep I don't love this. The reason we need to add margin is because labels were being placed outside of the viewable field when zoomed all of the way out of the beamline. Since the labels aren't part of the svg I'm not sure there is a better way to handle this.

sirepo/package_data/static/js/controls.js
```diff
@@ -745,34 +748,35 @@ SIREPO.app.directive('latticeFooter', function(appState, controlsService, lattic
                         })
                         .on('dblclick', () => elementClicked(values[0]))
                         .appendTo($('.sr-lattice-holder'));
-                    let divPos = {
-                        rect: div[0].getBoundingClientRect(),
-                        // need to track offset manually
```
e-carlin, line 750: @moellep I didn't find this to be true. I tried in Chrome and FF. Will you test and see what happens?

PR #3616, opened 2021-05-25
Fix #3493: standardize reading bytes from h5

### LINKED ISSUE #3493: standard way to extract bytes to string from hdf5
When reading a string from hdf5, it is encoded as bytes. We need a standard way to read this in one place. One example in template/opal.py
```
    pkcompat.from_bytes(h5file.attrs['{}Unit'.format(field.name)])
```

e-carlin:
@moellep this was the only other place where I could identify us doing a decode. Seems like using pkcompat is abstract enough but lmk if you want it centralized more (ex in template_common)

### DIFF
--- a/sirepo/template/radia.py
+++ b/sirepo/template/radia.py
@@ -9,15 +9,16 @@ Radia "instance" goes away and references no longer have any meaning.
 """
 from __future__ import division
 from pykern import pkcollections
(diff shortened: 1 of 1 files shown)


moellep:
Thanks, using pkcompat is abstract enough.

PR #3631, opened 2021-05-28
Issue/3627 only show prod codes in list of sims

moellep:


sirepo/html.py
```diff
@@ -70,11 +70,13 @@ def _widget_supported_codes(ctx):
               </button>
               <ul class="dropdown-menu" aria-labelledby="sr-landing-supported-codes">
     '''
-    x = PKDict([(s, s) for s in sirepo.feature_config.cfg().sim_types])
+    x = PKDict([(s, s) for s in sirepo.feature_config.PROD_FOSS_CODES])
```
robnagler, line 73: I think this issue still needs to be open, because this should be an intersection at least.

moellep: I opened a new ticket as #3632

PR #3639, opened 2021-06-01
Fix #3620: pkcli delete user

### LINKED ISSUE #3620: pkcli delete user
We need to be able delete users. It's complicated to do manually.

e-carlin:


sirepo/pkcli/admin.py
```diff
@@ -64,6 +64,28 @@ def create_examples():
                     if example.models.simulation.name not in names:
                         _create_example(example)
 
+# TODO(e-carlin): more than uid (ex email)
+def delete_user(uid):
```
e-carlin, line 68: @robnagler [you mentioned](https://github.com/radiasoft/sirepo/issues/3620#issuecomment-849799813) wanting a generic concept for identifying a user. Can you elaborate a bit on what you are thinking?

sirepo/pkcli/admin.py
```diff
@@ -64,6 +64,28 @@ def create_examples():
...
+    all service (jupyterhublogin, email, etc.) that may be relevant. Once
+    this script runs all records are blown away from the db's so if you
+    forget to configure something you will have to delete manually.
+    """
+    import sirepo.server
+    import sirepo.template
+    sirepo.server.init()
+    with auth_db.session_and_lock(), \
+         auth.set_user_outside_of_http_request(uid):
+        if sirepo.template.is_sim_type('jupyterhublogin'):
```
e-carlin, line 81: @robnagler if the user doesn't exist in auth_db (UserRegistration) then this code won't work. Is that a problem?

sirepo/auth_db.py
```diff
@@ -140,6 +140,17 @@ def delete(self):
                 self._session().delete(self)
                 self._session().commit()
 
+        @classmethod
+        def delete_user(cls, uid):
+            """Delete user from all tables"""
+            for t in [x for x in cls.metadata.tables.values() if 'uid' in x.columns]:
+                cls._session().execute(
+                    sqlalchemy.delete(t).where(uid==uid).execution_options(
+                        synchronize_session='fetch',
```
robnagler, line 149: this should be encapsulated. I think there should be a context manager.

sirepo/pkcli/admin.py
```diff
@@ -64,6 +64,28 @@ def create_examples():
...
 
+# TODO(e-carlin): more than uid (ex email)
+def delete_user(uid):
+    """Delete a user and all of their data across Sirepo and Jupyter
+
+    This will delete information based on what is configured. So configure
+    all service (jupyterhublogin, email, etc.) that may be relevant. Once
+    this script runs all records are blown away from the db's so if you
+    forget to configure something you will have to delete manually.
+    """
```
robnagler, line 75: I think we should describe args like we do for other functions.

sirepo/pkcli/admin.py
```diff
@@ -64,6 +64,28 @@ def create_examples():
...
+    import sirepo.server
+    import sirepo.template
+    sirepo.server.init()
```
robnagler, line 78: we need to encapsulate this...

sirepo/pkcli/admin.py
robnagler: If the user doesn't exist, it should blow up.

sirepo/srunit.py
```diff
@@ -90,6 +90,17 @@ def init_auth_db():
...
 
 
+def setup_srdb_root(cfg=None):
+    from pykern import pkunit, pkio
+    import os
+    e = cfg
+    if not e:
+        e = os.environ
+    e.update(
+        SIREPO_SRDB_ROOT=str(pkio.mkdir_parent(pkunit.work_dir().join('db'))),
```
robnagler, line 100: Something doesn't smell right. The `db` is definitely not right. That should come from somewhere else in srunit at least.

tests/pkcli/admin_test.py
```diff
@@ -0,0 +1,50 @@
...
+    pkunit.data_dir().join('db').copy(sirepo.srdb.root())
+    sirepo.pkcli.admin.delete_user('IYgnLlSy')
+    with auth_db.session_and_lock():
+        _is_empty(jupyterhublogin.JupyterhubUser.search_all_by())
+        _is_empty(auth_db.UserRegistration.search_all_by())
+    _is_empty(pkio.sorted_glob(jupyterhublogin.cfg.user_db_root_d.join('*')))
+    _is_empty(pkio.sorted_glob(simulation_db.user_path().join('*')))
+
+
+def _is_empty(result):
```
robnagler, line 47: This encapsulation doesn't seem right to me. It seems like it should just be in pkunit and apply to anything falsey.

tests/pkcli/admin_test.py
```diff
@@ -0,0 +1,50 @@
...
+    srunit.setup_srdb_root()
+    os.environ.update(
+        SIREPO_FEATURE_CONFIG_DEFAULT_PROPRIETARY_SIM_TYPES='jupyterhublogin',
+    )
+
+
+def test_no_user():
+    import sirepo.pkcli.admin
+    from pykern import pkunit
+    with pkunit.pkexcept(AssertionError):
```
robnagler, line 23: That's probably fine but maybe want to test for a specific regex.

sirepo/pkcli/admin.py
```diff
@@ -64,6 +64,28 @@ def create_examples():
...
+    with auth_db.session_and_lock(), \
+         auth.set_user_outside_of_http_request(uid):
+        if sirepo.template.is_sim_type('jupyterhublogin'):
+            from sirepo.sim_api import jupyterhublogin
+            jupyterhublogin.delete_user_dir(uid)
+        pkio.unchecked_remove(simulation_db.user_path(uid=uid))
```
robnagler, line 84: It's really important that uid be not None. Otherwise it will delete the entire database. I think this should be a routine in simulation_db that asserts its argument. Similar to jupyterhublogin.delete_user_dir.

robnagler (changes requested):

sirepo/auth_db.py
```diff
@@ -140,6 +140,17 @@ def delete(self):
                 self._session().delete(self)
                 self._session().commit()
 
+        @classmethod
+        def delete_user(cls, uid):
+            """Delete user from all tables"""
+            for t in [x for x in cls.metadata.tables.values() if 'uid' in x.columns]:
```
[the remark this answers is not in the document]
e-carlin: 96cbb207d this changes to commit after updating all tables.

sirepo/auth_db.py
e-carlin: efb9eef49 . I'm not sure how to do this in a contextmanager. I looked to set `syncrhonize_session` when creating the session but it can't be set then...

sirepo/pkcli/admin.py
e-carlin: 37af3c624

sirepo/pkcli/admin.py
e-carlin: 37af3c624

sirepo/pkcli/admin.py
e-carlin: yes... i'll spend some time this week thinking about it more.

sirepo/srunit.py
e-carlin: 7e0ffcb1b better?

tests/pkcli/admin_test.py
e-carlin: Do you imagine something different than pkok (with a not)? I didn't want to use pkok because it requires passing in a fmt string which I didn't think was relevant for this.

tests/pkcli/admin_test.py
e-carlin: 7277606d2

sirepo/pkcli/admin.py
e-carlin: 9abe57661

sirepo/srunit.py
e-carlin: 7e0ffcb1b better?

e-carlin:

sirepo/auth_db.py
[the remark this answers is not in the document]
robnagler: That still won't work. Deletions have to be in order of constraints. For example, we might create the user_registration_t first, since that's the only table that provides a uniqueness constraint on the uid. We would then relate user_registration_t to auth_user_email_t on the field uid, which would ensure that auth_user_email_t had globally unique uids. The order of the tables is undefined in values() so it could be that auth_user_email_t was first (correct) for the delete or after user_registration_t (constraint violation).  

We should know the relationships and have a cascading delete. Barring that, we should list the tables explicitly with a reasonable order (user_registration_t last).  This type of code gets copy ("how do I delete all tables with a particular column in sqlalchemy?") so it's better not to have it as an example. If the table does not exist, then it should be checked against this values() list but otherwise not deleted. You could add an assertion if you like that all tables get visited.

robnagler (changes requested):

sirepo/auth_db.py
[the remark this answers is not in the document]
e-carlin: Thanks for the details. SQLite doesn't support `ADD CONSTRAINT` so I went with the the ordered list of tables. 46d3660dd

tests/pkcli/admin_test.py
robnagler: It's fine. Just something to keep in mind.

sirepo/auth_db.py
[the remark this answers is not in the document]
robnagler: I was thinking we'd be using a "real" DBMS when we added constraints. The point is that the code is hidden, and wouldn't obviously break.

PR #3675, opened 2021-07-05
Issue 3672 radia compute job names

mkeilman:
Notes:

- Most of this is changing "geometry" to "geometryReport" and "solver" to "solverAnimation"
- Fixed a bug where ```template_common.write_dict_to_h5()``` was still wrapped in ```with...as```
- A little renaming to align with work for #2598


sirepo/template/radia.py
```diff
@@ -553,7 +582,7 @@ def _generate_field_integrals(g_id, f_paths):
                 res[p.name][i_type] = radia_util.field_integral(g_id, i_type, p1, p2)
         return res
     except RuntimeError as e:
-        pkdc('Radia error {}', e.message)
+        pkdlog('Radia error {}', e.message)
```
robnagler, line 585: This is a pattern which we may want to encapsulate? Returning `error` should be known by something, e.g. http_reply. I think we may want more context, e.g. which simulation. "Radia" is not necessary in the error message as it it is coming out of template.radia and the line/function will be known

sirepo/template/radia.py
```diff
@@ -800,21 +795,22 @@ def _parse_input_file_arg_str(s):
...
     try:
-        with h5py.File(_get_res_file(sim_id, filename, run_dir=run_dir), 'r') as hf:
-            return template_common.h5_to_dict(hf, path=h5path)
+        p = _get_sim_file(sim_id, filename, run_dir=run_dir)
+        with h5py.File(p, 'r') as f:
+            return template_common.h5_to_dict(f, path=h5path)
     except IOError as e:
         if pkio.exception_is_not_found(e):
-            pkdc(f'{filename} not found in {run_dir}')
+            pkdlog(f'file {filename} not found in {p}')
```
robnagler, line 808: let's continue to use `pkdlog('file {} not found in {}', filename, p)`. Also, `p` may not be exist if the exception happens inside `_get_sim_file` so you should have a `p = None` outside the `try`

sirepo/template/radia.py
```diff
@@ -800,21 +795,22 @@ def _parse_input_file_arg_str(s):
...
             # need to generate file
             return None
     except KeyError:
```
robnagler, line 811: This is too course of an error. I don't see where `KeyError` is being thrown, but it might be thrown in any of the functions. Better to wrap this deeper in the code, I think.

sirepo/template/radia.py
```diff
@@ -800,21 +795,22 @@ def _parse_input_file_arg_str(s):
...
         # no such path in file
-        pkdc(f'path {h5path} not found in {run_dir}/{filename}')
+        pkdlog(f'path {h5path} not found in {p}')
```
robnagler, line 813: We've been tending to a model of messages like this now: `'h5path={} not found in file={}'. It's slightly more structured, and it uses names of variables so easier to establish context when debugging.

robnagler (changes requested):
Looks good from the rename perspective. I didn't test it, but assume you did. The comments are unrelated to the name change, just saw some code that has some stylistic and other issues. If you see other code that's similar, fix it up, too. f-strings shouldn't be used in pkdlog especially in exceptions.

sirepo/template/radia.py
mkeilman: 575512a

I wanted to distinguish between "sirepo radia" and "oleg Radia" - now using "backend" to clarify that

sirepo/template/radia.py
mkeilman: 7679250

Added ```NoH5PathError``` which seems general enough to put in ```template_common```

sirepo/template/radia.py
mkeilman: 7679250

sirepo/template/radia.py
mkeilman: 7679250

e-carlin (approved):

e-carlin:
I'll merge after I merge #3661

e-carlin:
@mkeilman there are conflict resulting from merging #3661 . Can you resolve them?

mkeilman:
> @mkeilman there are conflict resulting from merging #3661 . Can you resolve them?

All conflicts should now be resolved

mkeilman:
@e-carlin did the solve work on NERSC?

PR #3679, opened 2021-07-06
Fix #3676: enable mpi in dev

### LINKED ISSUE #3676: distributed jobs run on single MPI rank under slurm
This happens in ```mpy.restrict_op_to_first_rank()```  and its progenitor code in SRW ```parameters.py.jinja.setup_magnetic_measurement_files()```.  This can be seen with some logging:

```
...
    try:
        import mpi4py.MPI
        pkdp('MPI WORLD SIZE {}', mpi4py.MPI.COMM_WORLD.Get_size())
            c = mpi4py.MPI.COMM_WORLD
            r = c.Get_rank()
    except Exception:
        pass
    pkdp('MPI RANK {} VS SBATCH RANK {}', r, os.environ['SLURM_PROCID'])
...
```

The "sbatch rank" goes from 0 to <num tasks> - 1, but the "mpi rank" is always 0, because ```COMM_WORLD.Get_size()``` returns 1.  The env variable ```$SIREPO_MPI_CORES``` looks correct. This does not happen when running under SMP - the mpi rank varies as expected.   I don't know what happens on NERSC.

e-carlin:
NERSC is working as configured.

### DIFF
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -727,7 +727,7 @@ fi
 
 exec python {template_common.PARAMETERS_PYTHON_FILE}
 EOF
-exec srun {s} /bin/bash bash.stdin
+exec srun {'--mpi=pmi2' if pkconfig.channel_in('dev') else ''} {s} /bin/bash bash.stdin
 '''
         )
         return f


mkeilman (approved):
Appears to work -```MPI.COMM_WORLD.Get_size()``` reports the expected number

sirepo/pkcli/job_agent.py
```diff
@@ -727,7 +727,7 @@ def _assert_project():
 
 exec python {template_common.PARAMETERS_PYTHON_FILE}
 EOF
-exec srun {s} /bin/bash bash.stdin
+exec srun {'--mpi=pmi2' if pkconfig.channel_in('dev') else ''} {s} /bin/bash bash.stdin
```
e-carlin, line 730: @e-carlin move this into slurm.conf

PR #3700, opened 2021-07-15
Fix #2621: Do not reload for changes to run dir

### LINKED ISSUE #2621: inotify reloader looking into user dir on dev
It looks like the python reloader is looking into the run/user directory when it should ignore it. This can cause the dev server to fail if it reloads while processing a simulation request.

```
[py3;@v4 sirepo]$ sirepo service http
../sirepo/job_driver/__init__.py:352:init modules=['local']
../sirepo/pkcli/job_supervisor.py:63:default_command ip=127.0.0.1 port=8001
 * Serving Flask app "sirepo.server" (lazy loading)
 * Environment: development
 * Debug mode: off
 * Running on http://0.0.0.0:8000/ (Press CTRL+C to quit)
 * Restarting with inotify reloader
10.10.10.1 - - [26/Jun/2020 16:36:08] "POST /delete-simulation HTTP/1.1" 200 -
 * Detected change in '/home/vagrant/src/radiasoft/sirepo/run/user/lvYyLP14/elegant/2HV2tAFA/bunchReport/parameters.py', reloading
10.10.10.1 - - [26/Jun/2020 16:36:09] "POST /simulation-list HTTP/1.1" 200 -
 * Restarting with inotify reloader
```

e-carlin:


sirepo/pkcli/service.py
```diff
@@ -31,7 +31,8 @@ def flask():
     from sirepo import server
     import sirepo.pkcli.setup_dev
 
-    with pkio.save_chdir(_run_dir()):
+    r = _run_dir()
+    with pkio.save_chdir(r):
```
robnagler, line 35: `with pkio.save_chdir(_run_dir()) as r:`

robnagler (changes requested):

sirepo/pkcli/service.py
e-carlin: Thanks. f693a9f6

PR #3726, opened 2021-07-28
Fix #3724: make pkcli.service.jupyterhub also start jupyterlab

### LINKED ISSUE #3724: Make pkcli.service.jupyterhub start jupyterlab
Need to make sure environment for the kernel is configured properly (ex LD_LIBRARY_PATH set).

e-carlin:


e-carlin:
Merging because I want to demo and it only impacts dev. Leave comments and I'll address in a followup PR.

PR #3728, opened 2021-07-28
Fix #3723: use knl processor when debug queue selected

### LINKED ISSUE #3723: use knl for cori debug queue
We hardwire haswell for all queues, which is fine. However, someone using the debug queue wants fast turn around, and for that, KNL would be faster. We may have to adjust the limits since they may be different from Haswell.

e-carlin:


robnagler (approved):
Cool! Nice code :)

Issue #3727, opened 2021-07-28
corrupt server thread?


robnagler:
The server started not returning static files (last stack trace). This is probably the first error:

```txt
21:19:33 sirepo/uri_router.py:87:call_api api=<function api_staticFile at 0x7ff0823cef28> exception='REQUEST_METHOD' stack=Traceback (most recent call last):
21:19:33 sirepo/uri_router.py", line 79, in call_api
21:19:33 r = flask.make_response(f(**kwargs) if kwargs else f())
21:19:33 sirepo/server.py", line 570, in api_staticFile
21:19:33 return flask.send_file(p, conditional=True)
21:19:33 flask/helpers.py", line 625, in send_file
21:19:33 cache_timeout=cache_timeout,
21:19:33 werkzeug/utils.py", line 789, in send_file
21:19:33 rv = rv.make_conditional(environ, accept_ranges=True, complete_length=size)
21:19:33 werkzeug/wrappers/response.py", line 797, in make_conditional
21:19:33 if environ["REQUEST_METHOD"] in ("GET", "HEAD"):
21:19:33 KeyError: 'REQUEST_METHOD'
21:19:33 Exception was printed at:
21:19:33 KeyError: 'REQUEST_METHOD'
21:19:33 flask/app.py", line 2088, in __call__
21:19:33 return self.wsgi_app(environ, start_response)
21:19:33 flask/app.py", line 2070, in wsgi_app
21:19:33 response = self.full_dispatch_request()
21:19:33 flask/app.py", line 1513, in full_dispatch_request
21:19:33 rv = self.dispatch_request()
21:19:33 flask/app.py", line 1499, in dispatch_request
21:19:33 return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
21:19:33 sirepo/uri_router.py", line 219, in _dispatch
21:19:33 return call_api(route.func, kwargs)
21:19:33 sirepo/http_reply.py:284:_gen_exception_error unsupported exception=<class 'KeyError'> msg='REQUEST_METHOD'
```

```txt
21:19:33 sirepo/uri_router.py:221:_dispatch exception='NoneType' object has no attribute 'save_to_cookie' path=static/css/sirepo.css stack=Traceback (most recent call last):
21:19:33 sirepo/uri_router.py", line 219, in _dispatch
21:19:33 return call_api(route.func, kwargs)
21:19:33 sirepo/uri_router.py", line 93, in call_api
[...]

PR #3731, opened 2021-07-30
WIP #3727: Add assertions to help diagnose error

e-carlin:
From looking through the logs it is unclear what exactly caused the error.
These assertions and logging will hopefully narrow down places where we are in a
state we do not expect (ex beleive we are in a request but flask.request is
False).

### DIFF
--- a/sirepo/util.py
+++ b/sirepo/util.py
@@ -40,6 +40,8 @@ THREAD_LOCK = threading.RLock()
 #: length of string returned by create_token
 TOKEN_SIZE = 16
 
+_logged_in_request = _logged_outside_request = 0
+
 
 class Reply(Exception):
     """Raised to end the request.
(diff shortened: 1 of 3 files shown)


sirepo/util.py
```diff
@@ -185,7 +187,14 @@ def flask_app():
 
 def in_flask_request():
     import sys
```
robnagler, line 189: Move this globally.

sirepo/util.py
```diff
@@ -185,7 +187,14 @@ def flask_app():
...
+    global _logged_in_request, _logged_outside_request
     f = sys.modules.get('flask')
+    if _logged_in_request < 10 and f and f.request:
```
robnagler, line 192: I think it is important to note that this is a global that will be accessed in a threaded context. It's ok, but it just should be noted that the number "10" doesn't need to be exact.

I would this code this way:

```py
if not f:
    if _log_not_flask < 10:
        _log_not_flask += 1
        pkdlog('flask is not imported')
    return False
if not f.request:
    if _log_not_request < 10:
        _log_not_request += 1
        pkdlog('flask.request is False')
    return False
return True
```

robnagler (changes requested):

sirepo/util.py
e-carlin: Thanks. I like that better. 7a3d2ae8

sirepo/util.py
e-carlin: 7a3d2ae8

robnagler (approved):

PR #3744, opened 2021-08-06
Fix #3707 radia undulator params

### LINKED ISSUE #3707: radia: additional undulator parameters
Such as:

 - termination spacing
 - more pole shapes
 - other?

mkeilman:
Notes:

- The main new item is adding "terminations" to an undulator, that is extra poles and magnets bracketing the periodic body. Further goodies like extruded shapes will move to new issues
- Modified the undulator example to use a termination
- Divorced the shape button from the shape selector, the latter being omitted from the UI until the new shapes are available


mkeilman:
Ready

sirepo/template/radia.py
```diff
@@ -357,12 +357,12 @@ def _build_cuboid(
     ):
     return _update_cuboid(
         _build_geom_obj('box', obj_name=name),
-        center or [0.0, 0.0, 0.0],
-        size or [1.0, 1.0, 1.0],
-        segments or [1, 1, 1],
+        center if center is not None else [0.0, 0.0, 0.0],
```
robnagler, line 360: Seems to me that this could be simplified. If _build_cuboid took `**kwargs` instead of a list of parameters, and then the caller always supplied keywords (not currently done), then an `pksetdefault` might do what you want.

sirepo/template/radia.py
```diff
@@ -1140,6 +1145,31 @@ def _update_geom_from_undulator(geom, und, beam_axis):
     magnet_transverse_ctr = magnet_dim_half.width / 2 - \
                             (gap_offset + magnet_dim_half.height + gap_half_height)
 
+    obj_props = PKDict(
```
robnagler, line 1148: This seems repetitive and could be collapsed especially if the code above used more dicts and/or lists.

sirepo/template/radia_util.py
```diff
@@ -172,6 +147,31 @@ def _radia_material(material_type, magnetization_magnitude, h_m_curve):
...
+
+    b = numpy.array(beam_dir)
+    g = numpy.array(gap_dir)
+    x = numpy.array(trans_dir)
+    sz = numpy.array(obj_size)
+    ctr = numpy.array(obj_ctr)
+    e = int(bevel.edge)
+    half_size = sz / 2
+    corner = ctr + half_size * [-x + g + b, x + g + b, x - g + b, -x - g + b][e]
+    trans_offset = bevel.amountTrans * x * [1, -1, -1, 1][e]
```
robnagler, line 160: naming the various lists `[1, -1, -1, 1]` as constants would be good, especially if they could be looked up based on use, i.e. the lists would be part of a dict.

sirepo/template/radia_util.py
```diff
@@ -172,6 +147,31 @@ def _radia_material(material_type, magnetization_magnitude, h_m_curve):
 )
 
 
+def apply_bevel(g_id, beam_dir, gap_dir, trans_dir, obj_ctr, obj_size, bevel):
+
+    b = numpy.array(beam_dir)
```
robnagler, line 152: This seems repetitive.

sirepo/package_data/static/js/radia.js
```diff
@@ -2079,6 +2085,128 @@ SIREPO.app.directive('numberList', function() {
...
+                        $scope.field.push(appState.models[modelName]);
+                        isEditing = true;
+                    }
+                    appState.saveChanges('geometry', function () {
+                        $scope.loadItems();
+                    });
+                });
+
+                $scope.$on('cancelChanges', function(e, name) {
+                    if (watchedModels.indexOf(name) < 0) {
```
robnagler, line 2196: Shouldn't this watch be abstracted?

robnagler (commented):
I can't really say if this works or not, but I made some comments about style. It seems to me there ought to be a way to compress some of this code.

PR #3747, opened 2021-08-06
Fix #3732: Search path for resources, sim_data, pkcli, and templates outside of Sirepo

### LINKED ISSUE #3732: Search path for resources and templates
This will likely have to move to pykern, because pkresource needs to use it. We want to be able to configure a search path for loading modules, resources, and static files. This will help private cloud customers have custom home pages, for example.

e-carlin:
Resources outside of Sirepo are searched first. This allows a file with the same name
(ex landing.html) to override the Sirepo version.

It is assumed that the files in the other module are organized in the same way as Sirepo. For example,
sim_data for srw is in sirepo.sim_data.srw. For another code it would be in <package_name>.sim_data.<code_name>.

There is a dependent PR in pykern
https://github.com/radiasoft/pykern/pull/104


sirepo/feature_config.py
```diff
@@ -109,6 +119,10 @@ def b(msg, dev=False):
         # No secrets should be stored here (see sirepo.job.agent_env)
         api_modules=((), set, 'optional api modules, e.g. status'),
         default_proprietary_sim_types=(set(), set, 'codes where all users are authorized by default but that authorization can be revoked'),
+        dynamic_sim_types=(
+            '',
+            _parse_to_dict,
```
e-carlin, line 124: @robnagler I have a feeling you won't like this. How should I best supply the cfg?

sirepo/server.py
```diff
@@ -657,7 +658,8 @@ def init(uwsgi=None, use_reloader=False, is_server=False):
     _app = flask.Flask(
         __name__,
         static_folder=None,
-        template_folder=str(simulation_db.STATIC_FOLDER),
+        # TODO(e-carlin): I don't think we use templates. Talk with rn
```
e-carlin, line 661: @robnagler what was this line for? We don't use flask templates right?

tests/dynamic_sim_type_test.py
```diff
@@ -0,0 +1,59 @@
...
+        cfg=PKDict(
+            SIREPO_FEATURE_CONFIG_DYNAMIC_SIM_TYPES='code1:rsprivate',
+        ),
+        no_chdir_work=True,
+    )
+    return fc
+
+
+@contextlib.contextmanager
+def _rsprivate():
```
e-carlin, line 41: @robnagler thoughts on this? I don't know if there is a more elegant way for installing a package just for a test.

sirepo/feature_config.py
```diff
@@ -144,5 +158,20 @@ def b(msg, dev=False):
...
     assert not x, \
         'sim_type(s) invalid={} expected={}'.format(x, VALID_CODES)
+    # SECURITY: _cfg.dynamic_sim_types cannot be validated because their names
+    # cannot be publicly exposed.
+    s.update(_cfg.dynamic_sim_types.keys())
     _cfg.sim_types = frozenset(s)
     return _cfg
+
+
+def _parse_to_dict(string):
```
robnagler, line 168: share? pkconfig? But I actually don't see the need.

sirepo/simulation_db.py
```diff
@@ -28,9 +27,11 @@
 import re
 import sirepo.auth
 import sirepo.job
+import sirepo.resource
 import sirepo.srdb
 import sirepo.template
 import time
+import werkzeug.utils
```
robnagler, line 34: i didn't see this used

sirepo/simulation_db.py
```diff
@@ -245,8 +248,8 @@ def get_schema(sim_type):
         else list(feature_config.cfg().sim_types)[0]
     if t in _SCHEMA_CACHE:
         return _SCHEMA_CACHE[t]
-    schema = read_json(
-        STATIC_FOLDER.join('json/{}-schema'.format(t)))
+    schema = read_json(sirepo.resource.static('json', f'{t}-schema.json'))
+    _merge_dicts(schema.get('appInfo', PKDict()), SCHEMA_COMMON.appInfo)
```
robnagler, line 252: This merge is new. Why?

sirepo/simulation_db.py
```diff
@@ -904,14 +892,13 @@ def _init():
...
-    with open(str(fn)) as f:
+    with open(str(sirepo.resource.static('json', f'schema-common{JSON_SUFFIX}'))) as f:
         SCHEMA_COMMON = json_load(f)
     # In development, any schema update creates a new version
     if pkconfig.channel_in('dev'):
         SCHEMA_COMMON.version = max([
-            _timestamp(pkio.py_path(fn).mtime()) \
-            for fn in glob.glob(str(STATIC_FOLDER.join('json/*{}'.format(JSON_SUFFIX))))
+            _timestamp(fn.mtime()) \
+            for fn in sirepo.resource.glob_static(f'json/*{JSON_SUFFIX}')
```
robnagler, line 901: As long as we are abstracting in this way, make the interfaces parallel: glob_static('json'). There's no need to expose more of the interface.

sirepo/simulation_db.py
```diff
@@ -994,7 +981,7 @@ def _pkg_relative_path_static(file_dir, file_name):
     Returns:
         str: full relative path of the file
     """
-    return '/' + RESOURCE_FOLDER.bestrelpath(static_file_path(file_dir, file_name))
+    return '/' + sirepo.resource.static(file_dir, file_name, relpath=True)
```
robnagler, line 984: There should be documentation about how this works, because it is going to have to be found later when the client requests the file. It works, because the search path is the same.

tests/dynamic_sim_type_test.py
robnagler: I think you might want to use PYTHONPATH so you don't have to install globally, which is not the best. At a minimum, the name of the package should be something like sirepo_test_dynamics_sim_type or something like that so we know it is used for testing and for which package.

tests/dynamic_sim_type_test.py
```diff
@@ -0,0 +1,59 @@
...
+        pkunit.data_dir().join('rsprivate.tar.gz').copy(d)
+        subprocess.run('tar xzf rsprivate.tar.gz', shell=True)
+        with pkio.save_chdir('rsprivate'):
+            try:
+                pip.main(['install', '-e', '.'])
+                # Makes rsprivate importable
+                importlib.reload(site)
+                yield
+            finally:
+                subprocess.run('pip uninstall -y rsprivate', shell=True)
```
robnagler, line 59: You probably don't want this at the end, because you may want to see how things got installed or whatever. Clean up at start with an uninstall.

tests/template/srw_import_test.py
```diff
@@ -45,7 +45,6 @@ def test_srw_2():
 def _t(tests):
     from sirepo.template.srw_importer import import_python
     from pykern import pkio, pkjson
-    from pykern import pkresource
```
robnagler, line 48: thanks for cleaning up cruft

sirepo/feature_config.py
```diff
@@ -77,6 +77,16 @@ def cfg():
     return _cfg or _init()
 
 
+def dynamic_sim_type_packages():
+    """Names of packages that contain the dynamic sim types"""
+    return frozenset(cfg().dynamic_sim_types.values())
```
robnagler, line 82: This needs to be a list (tuple), not a set, because it is ordered.

sirepo/util.py
```diff
@@ -183,6 +184,19 @@ def flask_app():
...
+
+def import_sim_module(sim_type, module):
+    """Import module for sim type
+
+    sim_type will be used to find the root package (ex sirepo)
+    and module will be appended to it (ex sim_data.srw)
+    """
+    import sirepo.feature_config
+    return importlib.import_module(
+        sirepo.feature_config.cfg().dynamic_sim_types.get(sim_type, 'sirepo') + \
```
robnagler, line 196: This should just search a path.

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+        paths (str): Path components of file
+        relpath (bool): If True path is relative to package package_data dir
+
+    Returns:
+        py.path: path to file
+    """
+    p = pkresource.filename(
+        # SECURITY: We take untrusted paths (ex api_staticFile). Must use
+        # safe_join to make sure they don't escape the static dir.
+        werkzeug.utils.safe_join('static', *paths),
```
robnagler, line 59: Wrap in safe_join sirepo.util. I'm not sure this should be here, and instead, it should be used in location of user input which needs to be checked. Not all paths need to be checked, and it is clearer that the api is doing the right thing if it does the checking or if it passes a parameter (check_input=True). I like being able to audit the APIs quickly.

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Args:
+        paths (str): Names of paths to get to resource file
+
+    Returns:
+        py.path: absolute path to resource file
+    """
+    return pkio.py_path(pkresource.filename(
+        os.path.join(*paths),
+        additional_packages=sirepo.feature_config.dynamic_sim_type_packages(),
```
robnagler, line 26: This is repeated too many times. There should be a iterable (tuple) `search_path` that can be passed in to pkresource and importlib. For sirepo, it would add sirepo to the end of the search path, instead of it could be implicitly added in pkresource, but I also think there's nothing wrong with it being explicit, e.g. feature_config.package_path, if used, would have to be set to `('rsprivate','sirepo')` for rsprivate.

sirepo/feature_config.py
robnagler: I could play Socrates here  and ask "Why not?" I won't. ;-)

The functional reason this is incorrect is that a single private repo might contain multiple codes (likely) or override parts of different codes (less likely). A simpler search solution would allow us to solve both problems. 

This should just a list of root packages which are checked for existence at bootstrap time that is you should be able to import the root packages themselves. See my comment below about sirepo being implicit or explicit.

sirepo/server.py
```diff
@@ -685,7 +687,8 @@ def _handle_error(error):
     except Exception:
         error_file = DEFAULT_ERROR_FILE
     return (
-        flask.send_from_directory(static_dir('html'), error_file),
+        # SECURITY: We control the path of the file so using send_file is ok.
```
robnagler, line 690: Good comment.

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+        werkzeug.utils.safe_join('static', *paths),
+        additional_packages=sirepo.feature_config.dynamic_sim_type_packages(),
+        relpath=relpath
+    )
+    if not relpath:
+        p = pkio.py_path(p)
+    return p
+
+
+def template(common_filename_or_sim_type):
```
robnagler, line 68: Maybe template_static? The plain word template is overrused.

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+def glob_static(path):
+    """Get paths that match path
+
+    Args:
+        path (str): pattern
+
+    Returns:
+        [py.path]: paths that match pattern
+    """
+    for f in pkresource.glob(
```
robnagler, line 39: Call this `pkresource.glob_files` is a bit redundant but not really since we should only return files. It could also be called `sorted_glob`. And perhaps in this case it could return py.paths(?), which is not the normal interface, but maybe a useful one.

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+from __future__ import absolute_import, division, print_function
+from pykern import pkio
+from pykern import pkresource
+import os
+import sirepo.feature_config
+import werkzeug.utils
+
+
+def filename(*paths):
```
robnagler, line 15: This returns a path so file_path?

sirepo/resource.py
```diff
@@ -0,0 +1,69 @@
...
+    Returns:
+        py.path: absolute path to resource file
+    """
+    return pkio.py_path(pkresource.filename(
+        os.path.join(*paths),
+        additional_packages=sirepo.feature_config.dynamic_sim_type_packages(),
+    ))
+
+
+def glob_static(path):
```
robnagler, line 30: static_paths_for_type might be a better name (see comment elsewhere about specifying interface).

robnagler (changes requested):
Thanks for the test and improvements. General change is to simplify to a search path, not a dict.

sirepo/resource.py
e-carlin: 8e67c167 and 9738d4e1

tests/dynamic_sim_type_test.py
e-carlin: 6e28eb52

sirepo/simulation_db.py
e-carlin: 6e28eb52

sirepo/simulation_db.py
e-carlin: 6e28eb52

tests/dynamic_sim_type_test.py
e-carlin: 6e28eb52

sirepo/feature_config.py
e-carlin: :)

FWIW `_parse_to_dict` can handle multiple codes from one package `code1:rsprivate:code2:rsprivate`.

How should we get sim_types? We will need to update `_cfg.sim_types` with them and they can't be verified against `VALID_CODES`.

robnagler: The VALID_CODES test probably should be removed or changed. If a code isn't in VALID_CODES then when the code is imported, it will fail. That's ok. If we have a type in sim_types config, that's probably ok, too. Not great, but unlikely on prod.

sirepo/server.py
robnagler: Ah, right. Good fix, but the commit should have documented it. That's a problem with PRs. I'm only looking at that comment.

sirepo/simulation_db.py
e-carlin: Currently appInfo in schema-common.json contains the appInfo for all supported codes. We need a way to add appInfo for dynamic codes to this map.

The way I've implemented is a codes schema (ex code1-schema.json) can define an appInfo key. SCHEMA_COMMON will then be lazily updated with this additional app info when a code of that type is loaded.

Without the merge  the line below (`schema.update(SCHEMA_COMMON)`) would overwrite the codes appInfo and appInfo for the dynamic code would be lost. Also, the appInfo for the dynamic code wouldn't be present in SCHEMA_COMMON in [sirepo.exporter](https://github.com/radiasoft/sirepo/blob/12d1eb5038bb4f30e16e06c32a0d1e0946c0df31/sirepo/exporter.py#L71).

sirepo/feature_config.py
e-carlin: 58607069

sirepo/feature_config.py
e-carlin: removed 58607069

sirepo/feature_config.py
e-carlin: removed 58607069

sirepo/util.py
e-carlin: 58607069 is this what you mean by search a path?

sirepo/resource.py
e-carlin: 58607069 is this what you mean?

sirepo/server.py
e-carlin: 1b7d57ae

sirepo/resource.py
e-carlin: afe522e0

sirepo/resource.py
e-carlin: I don't like static. It doesn't live under static so that's confusing to me. Personally, I think this will be used in a qualified context `resource.template` so it is clear that this isn't "other" templates. I thought about `template_resource` or `resource_template` but `resource.resource_template` is redundant. Thoughts?

sirepo/resource.py
e-carlin: eb1b916d

sirepo/resource.py
e-carlin: eb1b916d

sirepo/simulation_db.py
e-carlin: eb1b916d

e-carlin:
@robnagler ready for another reivew.

e-carlin:
I've identified some problems in the code. Please wait until I fix to review.

robnagler (changes requested):
I stopped reviewing after your message. I just thought you might want to see these.

sirepo/codes.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+u"""List of codes
+
+:copyright: Copyright (c) 2021 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
```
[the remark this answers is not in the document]
e-carlin: File removed

sirepo/codes.py
```diff
@@ -0,0 +1,43 @@
...
+# No sirepo imports so this can be used in test setup
```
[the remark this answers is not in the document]
e-carlin: 71a09f15

sirepo/feature_config.py
```diff
@@ -113,6 +68,11 @@ def b(msg, dev=False):
             derbenevskrinsky_force_formula=b('Include Derbenev-Skrinsky force formula'),
         ),
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
+        root_packages=(
```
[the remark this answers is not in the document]
e-carlin: 94d777d2

sirepo/feature_config.py
```diff
@@ -113,6 +68,11 @@ def b(msg, dev=False):
...
+            tuple(),
```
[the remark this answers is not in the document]
e-carlin: 94d777d2

e-carlin:

sirepo/util.py
```diff
@@ -235,6 +253,10 @@ def random_base62(length=32):
     return ''.join(r.choice(numconv.BASE62) for x in range(length))
 
 
+def safe_join(*paths):
```
robnagler, line 256: Let's call this `safe_path`, because it's unclear what's being joined in safe_join, and we can also see quickly when a safe_join needs to be updated to safe_path.

sirepo/template/template_common.py
```diff
@@ -538,11 +539,11 @@ def render_jinja(sim_type, v, name=PARAMETERS_PYTHON_FILE, jinja_env=None):
...
     """
-    d = sirepo.sim_data.get_class(sim_type).resource_dir() if sim_type \
-        else sirepo.sim_data.resource_dir()
+    # append .jinja, because file may already have an extension
+    b = f'{name}.jinja'
     return pkjinja.render_file(
-        # append .jinja, because file may already have an extension
-        d.join(name) + '.jinja',
+        sirepo.sim_data.get_class(sim_type).resource_path(b) if sim_type \
+        else sirepo.sim_data.resource_path(b),
```
robnagler, line 546: indent

sirepo/resource.py
```diff
@@ -0,0 +1,70 @@
...
+        paths (str): Path components of file
+        relpath (bool): If True path is relative to package package_data dir
+        check_input (bool): Make sure that the path given does not escape the static dir
+
+    Returns:
+        py.path: path to file
+    """
+    p = pkresource.filename(
+        sirepo.util.safe_join('static', *paths) if check_input \
+        else os.path.join('static', *paths),
```
robnagler, line 44: indent

sirepo/feature_config.py
```diff
@@ -134,15 +132,21 @@ def b(msg, dev=False):
...
     s.update(_cfg.proprietary_sim_types, _cfg.default_proprietary_sim_types)
     for v in _DEPENDENT_CODES:
         if v[0] in s:
             s.add(v[1])
-    x = s.difference(VALID_CODES)
-    assert not x, \
-        'sim_type(s) invalid={} expected={}'.format(x, VALID_CODES)
     _cfg.sim_types = frozenset(s)
+    if 'sirepo' not in _cfg.package_path:
+        _cfg.package_path = (*_cfg.package_path, 'sirepo')
```
robnagler, line 144: [Previous comment](https://github.com/radiasoft/sirepo/pull/3747/files#r687068441) was not this. We don't always want to inject sirepo. Rather, if there is a package_path, it's explicitly listed. So remove this code, and add this above:

```py
package_path=(
    ('sirepo',),
```

robnagler (changes requested):

e-carlin:
Idk why but there is no box to respond to this comment https://github.com/radiasoft/sirepo/pull/3747#discussion_r688611898 so replying here. Caching on start e2130cd57

e-carlin:
https://github.com/radiasoft/sirepo/pull/3747#discussion_r688597092 resolved with 866d28547

e-carlin:
Replying to https://github.com/radiasoft/sirepo/pull/3747#discussion_r688612281 because I can't reply inline:

None so far but in the original issue you mention [overriding the home page](https://github.com/radiasoft/sirepo/issues/3732#issue-957281908) which would be a static file living outside of sirepo.

e-carlin:
https://github.com/radiasoft/sirepo/pull/3747#discussion_r688609310 

Mind if I solve this when it becomes a problem? This would require a rejigger of the code for searching templates which has already changed 2x and I'm becoming nervous that I'm introducing bugs.

e-carlin:
https://github.com/radiasoft/sirepo/pull/3747#discussion_r688724188 resolved with 09313d759

e-carlin:

I apologize for the extremely confusing presentation of my replies. The GitHub UI for PRs is either broken or has changed. Resolve comment buttons and reply boxes are not always present anymore...

sirepo/feature_config.py
```diff
@@ -9,48 +9,41 @@
...
-    'ml',
-    'opal',
-    'radia',
-    'shadow',
-    'srw',
-    'synergia',
-    'warppba',
-    'warpvnd',
-    'zgoubi',
+        'controls',
```
robnagler, line 19: indentation should be 4 not 8. It counts on the previous line. Just because there are two parens, doesn't mean you need to increase indentation, rather a good thing not to in this case.

sirepo/sim_data/__init__.py
```diff
@@ -127,6 +126,8 @@ class SimDataBase(object):
 
     _EXE_PERMISSIONS = 0o700
 
+    _LIB_REOUSRCE_DIR = 'lib'
```
robnagler, line 129: spelling error

sirepo/simulation_db.py
```diff
@@ -151,8 +145,16 @@ def delete_user(uid):
...
 def examples(app):
+    d = None
+    try:
+        d = sirepo.sim_data.get_class(app).resource_path(_EXAMPLE_DIR)
+    except Exception as e:
+        if pkio.exception_is_not_found(e):
+            # Not all sim types have examples (ex jupyterhublogin)
+            return []
+        raise
     files = pkio.walk_tree(
```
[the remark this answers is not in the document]
robnagler: Did you not understand my comment?

sirepo/template/srw_importer.py
```diff
@@ -461,9 +459,11 @@ def _get_default_drift():
     Returns:
         str: default drift propagation paramters
     """
-    c = pkio.read_text(_JS_DIR.join('srw.js'))
-    m = re.search(r'function defaultDriftPropagationParams.*?return\s*(\[[^\]]+\])', c, re.DOTALL)
-    return pkjson.load_any(m.group(1))
+    return pkjson.load_any(re.search(
```
robnagler, line 462: This is an example where naming the function, e.g. re.search, is a good idea. Even if it is a nested function, it avoids too much nesting without naming. Using intermediate variables is a way of naming, but it is better to name the function begin performed as a functional call.

robnagler (changes requested):

sirepo/template/srw_importer.py
e-carlin: I'm having a hard time understanding what you're saying. Can you say it in a different way? 

Should I make a new named inner function for just the re.search bit? Should I revert back to the original structure so `m` names the re.search? 

When you say "nested function" do you mean an inner function (ex a new `def ...` inside of '`def _get_default_drift` or do you mean that my refactoring where I nested the function inside of another function and removed the local var?

sirepo/feature_config.py
e-carlin: Sorry, copy paste bug from codes.py cf155829d

sirepo/sim_data/__init__.py
e-carlin: cf155829d

sirepo/simulation_db.py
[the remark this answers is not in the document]
e-carlin: I did - I had responded here https://github.com/radiasoft/sirepo/pull/3747#issuecomment-899726777.

Thinking about this some more I think we need to decide what the semantics are because, as you say, they are inconsistent. Making examples augment is only making it more inconsistent. The concept of `sim_data.resource_path` doesn't make any sense if some resources have a single path (ex a jinja file) whereas other resources have multiple paths (ex example files).

Here's what I think should be done:
- pkresource should be changed so it doesn't do *any* overriding. We should make new method(s) that return a list of all matching filenames in all searched packages. We never stop once we've found one match. This is a change in behavior so we can probably just leave the old `filename` method and make new method(s) for Sirepo to use with new behavior. 
- Sirepo will then decide what makes sense for certain files. Ex, in the case of example it makes sense to augment (a sim can have any number of examples). In the case of say `landing.html` it doesn't make sense so if more than one file is returned then Sirepo will have to decide what it wants to do. pkresource can provide a guarantee that files are returned in the same order as the supplied packages so Sirepo can just use the 1st file returned if it wants the current override behavior. 
- All calls to `sim_data.resource_path` will be removed in favor of a call like `sim_data.example_paths`. The different function calls will know how to handle the files (whether or not multiple files makes sense or not). 

Thoughts?

e-carlin:
@robnagler ready for another review. Need you to answer my to questions:

https://github.com/radiasoft/sirepo/pull/3747#discussion_r692256062
https://github.com/radiasoft/sirepo/pull/3747/files#r692286241

robnagler (changes requested):
We can also discuss if you like.

sirepo/simulation_db.py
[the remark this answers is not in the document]
e-carlin: > I don't understand why this would be needed and more importantly how the caller would know what to do with the list of files, when it only wants one.

My thinking was that pkresource shouldn't be concerned with knowing when to override or when to get all files. Sirepo really knows this so pkresource should always return all matching files in all supplied packages and let Sirepo decide. For example, when we get examples we use all. In contrast to static files we would pick just one.  pkresource could guarantee that the files are returned in order of supplied packages so for static files sirepo would pick just the first file.

> glob_files(EXAMPLE_DIR/*.json) would work

Almost, except that it needs to prepend `template/<sim_type>` to the `EXAMPLE_DIR`

I think I've found a simple solution. Make `sim_data` be the only one that knows about `template` and make `sim_data.resource_path` always return a path to a file instead of a relative path to the resource dir.

e8fedc0df

e-carlin:

I made change to `simulation_db.examples` https://github.com/radiasoft/sirepo/pull/3747/files/94d777d2b20a3dcffd060bd513192c9ae218d977#r693176020 . Happy to talk in real-time if my changes don't make sense.

Still need an answer to https://github.com/radiasoft/sirepo/pull/3747#discussion_r692256062

sirepo/template/srw_importer.py
robnagler: [Nested Function](https://en.wikipedia.org/wiki/Nested_function) is the generic, "computer science" term for Python's inner functions.

Yes, make an inner function or perhaps not a nested function. 

This is a bit like the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)). Some people say that when a function is longer than a screen, it needs to be broken up. Others say more than 15 lines. Yet others (and me) says when you can name the function, name it as a function, not as a temporary variable, which creates problems. A function and a variable are layers of indirection, but the function is stateless, and the variable is stateful, so it's the loser in the naming contest. If a function can be stateless, better to make it so.

sirepo/simulation_db.py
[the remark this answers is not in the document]
robnagler: Unfortunately, I disagree. The problem we have is: map an abstract name to an actual path. The mapping in Sirepo's (and I argue almost all other cases) is one to one, that is, there is one and only one path for a name. 

Consider the problem of executing a command in Bash, you don't ever get a choice. Bash executes the first command it finds in $PATH. The same is true of the loader and $LD_LIBRARY_PATH. 

I am having a hard time thinking of a case where you want multiple values for a single abstract name in Sirepo. Do you have one? 

There are times in BOP that you want to know the "next one in the path", which is tricky, because you need to know where you are in the path (the concept iterates). This is handled by the BOP loader for class inheritance. We don't have this case in Sirepo or Pykern.

> Almost, except that it needs to prepend template/<sim_type> to the EXAMPLE_DIR

This is a detail that can be handled in this particular case, because it is very specific to this case (creating a URL, not an actual file). From the path, you get basename, and then prepend `template/<sim_type>`. This doesn't change the abstraction name to actual path mapping: it's still one to one.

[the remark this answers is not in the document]
e-carlin: > Consider the problem of executing a command in Bash, you don't ever get a choice. Bash executes the first command it finds in $PATH. The same is true of the loader and $LD_LIBRARY_PATH.

I'm not sure I like that they do that:) I think tools like pyenv had to be created in part because of that behavior of `$PATH`.

I've had many run ins with `$LD_LIBRARY_PATH` where I wish it didn't pick the library for me.

Recently I was trying to compile emacs from source. The linker kept picking the wrong zlib (one with xcode not the one I installed). Supplying all sorts of args to impact `$LD_LIBRARY_PATH` seemed to have no effect. If it was designed where I had to decide which one to use I wouldn't have had this problem (although I'm sure the behavior would lead to other problems I don't see...).

> I am having a hard time thinking of a case where you want multiple values for a single abstract name in Sirepo. Do you have one?

In the [original issue](https://github.com/radiasoft/sirepo/issues/3732) you say, `private cloud customers have custom home pages`. My assumption is that would be implemented by overriding `landing.html`. This would mean we have two values for the single `static/html/landing.html` abstract name (one in Sirepo and one in the customer's repo). Another thought would be customers could override pages like forbidden.html so they could provide information specific to their organization (contact it@customer.com not support@radiasoft.net)

> This is a detail that can be handled in this particular case, because it is very specific to this case

Make sense, thanks. The change I linked to above obviates the need for prepending and instead keeps all of the details in `sim_data` (e8fedc0). Thoughts?

sirepo/template/srw_importer.py
e-carlin: Thanks:) c61058932

Is there a name for a function called from within another function? Ex what do you call the function bar in `foo(bar())`

e-carlin:

sirepo/sim_data/__init__.py
```diff
@@ -194,6 +199,16 @@ def compute_model(cls, model_or_data):
...
         return cls._delete_sim_db_file(cls._sim_file_uri(sim_id, basename))
 
+    @classmethod
+    def example_paths(cls):
+        import sirepo.simulation_db
+        return sirepo.resource.glob_paths(
+            _TEMPLATE_RESOURCE_DIR,
+            cls.sim_type(),
+            cls._EXAMPLE_RESOURCE_DIR,
+            f'*{sirepo.simulation_db.JSON_SUFFIX}',
```
robnagler, line 209: This is an example where JSON_SUFFIX should be split off from simulation_db perhaps in sirepo.const or even elsewhere (pkconst?).

sirepo/resource.py
```diff
@@ -0,0 +1,78 @@
...
+        paths (str): Path components of file
+
+    Returns:
+        [py.path]: paths that match pattern
+    """
+    for f in pkresource.glob_files(
+            os.path.join(*paths),
+            packages=sirepo.feature_config.cfg().package_path,
+    ):
+        yield pkio.py_path(f)
```
robnagler, line 44: If glob_files was turned into glob_paths, then this code would be simpler, no need for the `py.path`, hence `for` would go away, just returning the iterator.

sirepo/resource.py
```diff
@@ -0,0 +1,78 @@
...
+    Returns:
+        py.path: path to file
+    """
+    p = pkresource.filename(
+        sirepo.util.safe_path('static', *paths) if check_input \
+            else os.path.join('static', *paths),
+        packages=sirepo.feature_config.cfg().package_path,
+        relpath=relpath
+    )
+    if not relpath:
```
robnagler, line 64: if this were handled inside pkresource, then this code would not be necessary.

sirepo/resource.py
```diff
@@ -0,0 +1,78 @@
...
+        [py.path]: paths that match pattern
+    """
+    for f in pkresource.glob_files(
+            os.path.join(*paths),
+            packages=sirepo.feature_config.cfg().package_path,
+    ):
+        yield pkio.py_path(f)
+
+
+def static(*paths, relpath=False, check_input=False):
```
robnagler, line 47: I feel like check_input is in the wrong place. It complicates this routine, and it should be in the caller.

sirepo/server.py
```diff
@@ -540,17 +541,16 @@ def api_srUnit():
...
     Returns:
         flask.Response: flask.send_from_directory response
     """
     if not path_info:
-        raise sirepo.util.raise_not_found('empty path info')
-    p = pkio.py_path(flask.safe_join(str(simulation_db.STATIC_FOLDER), path_info))
+        sirepo.util.raise_not_found('empty path info')
+    # SECURITY: Use check_input to verify that the path does not escape the
+    # static dir.
+    p = sirepo.resource.static(path_info, check_input=True)
```
robnagler, line 553: Here's an example that explicitly calling safe_path here would be better. You wouldn't need to write this comment, instead it would look like:

`sirepo.resource.static(sirepo.util.safe_path(path_info))`

Much clearer to me.

sirepo/template/srw_importer.py
robnagler: Just "call" afaik. If `bar()` is called within `bar()` then it is called a recursive call. If `bar()` is called within an outer function it is called a nested call. The name is based on how the callee is defined.

sirepo/simulation_db.py
```diff
@@ -1000,7 +978,9 @@ def _pkg_relative_path_static(file_dir, file_name):
     Returns:
         str: full relative path of the file
     """
-    return '/' + RESOURCE_FOLDER.bestrelpath(static_file_path(file_dir, file_name))
+    # POSIT: static files outside of Sirepo use the same path structure relative
+    # to the static dir
+    return '/' + sirepo.resource.static(file_dir, file_name, relpath=True)
```
robnagler, line 983: This is an example where we need to think about adding a parameter. There's a lot going on here that's subtle. First, it's a URL that's being returned, not a "relpath", which implies a file. Second, resource.static is about files, not URLs. Third, this is in the context of a schema initialization. It's called in one place to do a very specific thing. The caller of this routine is very complicated. Too complicated imiho.

That's why I think relpath should be removed from pkresource and sirepo.resource, and this specific thing is to be handled in the routine here, or rather, the function itself should be unwound a bit. I can help with this. 

We should definitely have a test for compiling schemas to expected results.

robnagler (changes requested):

sirepo/resource.py
e-carlin: 6e7eede4d

sirepo/server.py
e-carlin: 6e7eede4d

sirepo/resource.py
e-carlin: bc12ff9ff

sirepo/simulation_db.py
e-carlin: bc12ff9ff

sirepo/sim_data/__init__.py
e-carlin: 53f4c531e

sirepo/resource.py
e-carlin: 282fba3b1

e-carlin:

sirepo/feature_config.py
```diff
@@ -9,8 +9,9 @@
 
 #: Codes that depend on other codes. [x][0] depends on [x][1]
 _DEPENDENT_CODES = [
-    ['jspec', 'elegant'],
-    ['controls', 'madx'],
+        ['jspec', 'elegant'],
```
robnagler, line 12: incorrect indent

sirepo/job_supervisor.py
```diff
@@ -247,7 +245,7 @@ def _get_uids_and_files():
             u = None
             p = sirepo.auth_db.UserRole.uids_of_paid_users()
             for f in pkio.sorted_glob(_DB_DIR.join('*{}'.format(
-                    sirepo.simulation_db.JSON_SUFFIX,
+                    sirepo.const.JSON_SUFFIX,
```
robnagler, line 248: incorrect indent

sirepo/resource.py
```diff
@@ -0,0 +1,84 @@
...
+    """Get matching paths
+
+    Args:
+        paths (str): Path components of file
+
+    Returns:
+        [py.path]: paths that match pattern
+    """
+    return pkresource.glob_paths(
+            _join_paths(paths),
```
robnagler, line 40: incorrect indent

sirepo/sim_data/__init__.py
```diff
@@ -640,14 +654,24 @@ def _lib_file_list(cls, pat, want_user_lib_dir=True):
...
+            )
+        )
         if want_user_lib_dir:
             # lib_dir overwrites resource_dir
-            x.append(simulation_db.simulation_lib_dir(cls.sim_type()))
-        for d in x:
-            for f in pkio.sorted_glob(d.join(pat)):
-                res[f.basename] = f
+            res.update(
+                (f.basename, f) for f in \
```
robnagler, line 670: not sure you need "\".

sirepo/simulation_db.py
```diff
@@ -929,6 +859,51 @@ def _init():
...
+        _merge_dicts(s.get('appInfo', PKDict()), a)
+        s.update(SCHEMA_COMMON)
+        s.feature_config = feature_config.for_sim_type(t)
+        s.simulationType = t
+
+        #TODO(mvk): improve merging common and local schema
+        _merge_dicts(s.common.dynamicFiles, s.dynamicFiles)
+        s.dynamicModules = _files_in_schema(s.dynamicFiles)
+        for i in [
+                'appDefaults',
```
robnagler, line 877: incorrect indent

robnagler (approved):
Some formatting comments. Not important to fix,  just next time you are in the code.

PR #3749, opened 2021-08-09
Fix #3740: set alfy to alfy not alfx

### LINKED ISSUE #3740: webcon: alpha_x for alpha_y in match beam to lattice
Jon writes:
>It looks like there is a bug in the match beam to lattice function in the source tab on madx. Seems to be taking the wrong alpha_y
>It looks like it is just pulling alpha_x for both actually.

e-carlin:


e-carlin:
@j-edelen FYI

PR #3785, opened 2021-08-26
Fix #3745 radia extruded shapes

### LINKED ISSUE #3745: radia: interface for extruded shapes
Continuing the work that began in #3707, finish the UI for adding extrusion of simple built-in polygons. This will facilitate #3519, #3659, and #3667

mkeilman:
This is still somewhat clunky but important to get in place to start parameterizing more complex magnets such as dipoles. I plan to clean things up as that work progresses.

Notes:

- adds a variety of shapes to extrude (C, L, J) and massages the coordinates to pass to Radia
- the extrusion points are calculated on the client but should move to the server (undulators do this with L shapes as a first step in this process)
- objects created in the client now get uuids instead of integer ids, to match objects created on the server


mkeilman:
Ping for review - I'll need this in place to get dipole examples ready

e-carlin:
I'll review starting at 10am.

sirepo/package_data/static/html/radia-source.html
```diff
@@ -8,7 +8,7 @@
       </div>
       </div>
   <div class="col-md-12" data-ng-if="src.showParams()">
-    <!--<div data-advanced-editor-pane="" data-view-name="src.getMagnetType()" data-parent-controller="src"></div>-->
-    <div data-basic-editor-panel="" data-view-name="hybridUndulator" data-parent-controller="src" data-panel-title="Undulator"></div>
+    <div data-ng-if="src.getMagnetType() == 'undulator'" data-basic-editor-panel="" data-view-name="hybridUndulator" data-parent-controller="src" data-panel-title="Undulator"></div>
```
e-carlin, line 11: use `====`

sirepo/package_data/static/html/radia-source.html
```diff
@@ -8,7 +8,7 @@
...
+    <!--<div data-ng-if="src.getMagnetType() == 'dipole'" data-basic-editor-panel="" data-view-name="dipole" data-parent-controller="src"></div>-->
```
e-carlin, line 12: Can the code be removed since it is commented out?

sirepo/package_data/static/js/radia.js
```diff
@@ -169,6 +179,13 @@ SIREPO.app.factory('radiaService', function(appState, fileUpload, panelState, re
         self.showPathPicker(true, true);
     };
 
+    self.nextId = function() {
```
e-carlin, line 182: This code seems particularly confusing. I think a comment saying why we need a uuid of this shape would be helpful.

sirepo/package_data/static/js/radia.js
```diff
@@ -3758,36 +3742,113 @@ SIREPO.app.directive('shapeSelector', function(appState, panelState, plotting, r
 SIREPO.viewLogic('objectShapeView', function(appState, panelState, radiaService, $scope) {
     let ctr = [];
     let modelType = null;
+    let parent = $scope.$parent;
```
e-carlin, line 3745: `const parent`

sirepo/package_data/static/js/radia.js
```diff
@@ -3758,36 +3742,113 @@ SIREPO.app.directive('shapeSelector', function(appState, panelState, plotting, r
...
     let size = [];
-    let geomIndices = {};
 
     $scope.modelData = appState.models[$scope.modelName];
+    size = $scope.modelData.size.split(/\s*,\s*/).map((x) => {
```
e-carlin, line 3749: I think the code would be clearer if you use a named function for the split, map, parseFloat and passed in the string.

sirepo/template/radia.py
```diff
@@ -700,19 +703,26 @@ def _generate_parameters_file(data, is_parallel, for_export=False, run_dir=None)
...
+            data.models.hybridUndulator,
+            v.beam_axis,
+            v.height_axis,
+        )
     v.objects = g.get('objects', [])
     _validate_objects(v.objects)
-    # read in h-m curves if applicable
+
     for o in v.objects:
+        if o.get('type'):
```
e-carlin, line 723: I think `if 'type' in o` is more "pythonic".

sirepo/template/radia.py
```diff
@@ -1047,7 +1057,7 @@ def _validate_objects(objects):
         if 'material' in o and o.material in _SCHEMA.constants.anisotropicMaterials:
             if numpy.linalg.norm(sirepo.util.split_comma_delimited_string(o.magnetization, float)) == 0:
                 raise ValueError(
-                    '{}: anisotropic material {} requires non-0 magnetization'.format(
+                    'name={}, : material={}: anisotropic material requires non-0 magnetization'.format(
```
e-carlin, line 1060: I think we've settled on `f''` strings in stead of `.format()`

sirepo/template/radia.py
```diff
@@ -1086,42 +1096,58 @@ def _undulator_termination_name(index, term_type):
...
+    ell.points = [[2 * c[i] * k[i] + -1**k[i] * v for (i, v) in enumerate(p)] for p in ell.points]
+
+    # Radia's extrusion expects points in permutation order based on the extrusion
+    # axis (x -> [y, z], y -> [z, x], z -> [x, y])
+    if w.tolist().index(1) != (b.tolist().index(1) + 1) % 3:
+        for p in ell.points:
+            p.reverse()
+
+
+def _update_geom_from_undulator(geom, und, beam_axis, height_axis):
```
e-carlin, line 1141: This method is pretty long. I'd consider splitting it up.

e-carlin (approved):
Looks good. Mostly style nitpicks. You can merge when you're ready.

sirepo/package_data/static/js/radia.js
e-carlin: To be clear I mean something like

```js
function myNewFunctionWithAGreatName(str) {
  return str.split(/\s*,\s*/).map(x => parseFloat);
}
```

sirepo/package_data/static/html/radia-source.html
mkeilman: 738b063

sirepo/package_data/static/html/radia-source.html
mkeilman: 738b063

Agreed, I'll put it back when needed

sirepo/package_data/static/js/radia.js
mkeilman: 9c10bd2

sirepo/package_data/static/js/radia.js
```diff
@@ -3685,41 +3667,43 @@ SIREPO.app.directive('shapeButton', function(appState, panelState, plotting, rad
...
+            function updateShape() {
+                const o = appState.models[$scope.modelName];
+                const size = o.size.split(/\s*,\s*/).map((x) => {
                     return parseFloat(x);
                 });
-                let s = shapes[obj.type];
-                s.setFill(obj.color);
-                const ar = size[radiaService.axes.indexOf(appState.models.simulation.widthAxis)] /
-                    size[radiaService.axes.indexOf(appState.models.simulation.heightAxis)];
+                let s = shapes[o.type];
```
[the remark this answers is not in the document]
mkeilman: 9c10bd2

sirepo/template/radia.py
mkeilman: That and other cleanup is indeed planned

sirepo/template/radia.py
mkeilman: My understanding was to avoid f strings in error messages but maybe I misconstrued that. I'll leave it as-is until I get clarity on the matter

robnagler: format and f-strings are identical.

We avoid f-strings and format in log messages. The assumption is that log messages could contain large values that need to be truncated by pkdebug whereas error messages are not that way. If you expect an error message to be large, then you will want to truncate manually (possible with f-strings or format, e.g. `{big_thing:.100s}`).

e-carlin: You don't want to use f-strings in pkdebug functions (pkdp, pkdlog, etc). In those functions we [modify args and kwargs](https://github.com/radiasoft/pykern/blob/master/pykern/pkdebug.py#L562) (ex redact secrets). So, we need directs access to args and kwargs. In ValueError (or any other stdlib error) you can use them since we don't do any manipulation.

sirepo/package_data/static/js/radia.js
mkeilman: I'll make a separate issue for it

sirepo/package_data/static/js/radia.js
mkeilman: uuid is appropriate so I'll document the reasoning (```nextId()``` is historical)

PR #3812, opened 2021-09-08
Fix #2917 global hide guest warning

### LINKED ISSUE #2917: rename sirepo.feature_config.srw.hide_guest_warning without srw
It applies to all codes so should be globally available.

### LINKED ISSUE #3811: remove rs4pi_dose_calc from feature_config
As long as you are futzing with feature_config in #2917 please remove this too. has to get removed from pkcli/rs4pi.py. Assume it is always false in that code.

mkeilman:
This stows the guest warning config in a ```sim_global``` dict to keep such things apart from ```proprietary_sim_types``` and the like.

Also fixes #3811, removing rs4pi_dose_calc and thus always using the fake dose calculation. ```generate_rtdose_file``` and the associated bash script remain in place. Note: I could not figure out how to test this as the doseCalculation report appears to be unreachable...???


sirepo/feature_config.py
```diff
@@ -109,12 +118,12 @@ def b(msg, dev=False):
             'Names of root packages that should be checked for codes and resources. Order is important, the first package with a matching code/resource will be used. sirepo added automatically.',
         ),
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
-        #TODO(robnagler) make this a sim_type config like srw and warpvnd
-        rs4pi_dose_calc=(False, bool, 'run the real dose calculator'),
+        sim_global=dict(
+            hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
```
e-carlin, line 122: What's the advantage to adding the `sim_global` dict instead of making `hide_guest_warning` at the top-level? Should things like `package_path` be added to this?

mkeilman: The sim-specific cfg goes into the app schema (which is where ```sirepo.js``` looks for ```hide_guest_warning```) - my thought was to keep top-level stuff and not add it piecemeal. I guess anything that the client needs to know can go there

e-carlin: Ah makes sense. Thanks.

robnagler:
@e-carlin please review and merge

sirepo/feature_config.py
```diff
@@ -85,6 +85,15 @@ def for_sim_type(sim_type):
     )
 
 
+def global_sim_cfg():
+    """Get cfg to use across all simulations. Separate from global cfg
+
+    Returns:
+        dict: global sim config
+    """
+    return cfg()['sim_global']
```
e-carlin, line 94: Convert to a PKDict so the interface is the same as `for_sim_type`.

sirepo/feature_config.py
```diff
@@ -85,6 +85,15 @@ def for_sim_type(sim_type):
     )
 
 
+def global_sim_cfg():
```
e-carlin, line 88: Nits: 
- I don't think `cfg` is needed in the name. Everything from this module is cfg.
- What do you think about renaming to be `gui_global` or something similar? All of the cfg values are "global" on the backend so sim confused me. Another option would be to just say YAGNI and put `hide_guest_warning` at the top-level. Then the next person who has a better idea of what the abstraction should be can solve it.

e-carlin (approved):

e-carlin:
@mkeilman I'm signing off for vacation shortly. I've approved the PR. Do what you will with my comments. Merge when you're ready.

sirepo/feature_config.py
mkeilman: > @mkeilman I'm signing off for vacation shortly. I've approved the PR. Do what you will with my comments. Merge when you're ready.

This got past me somehow - I have kept the flag in its own dict but am using ```sim_common``` - we use "common" in the schema e.g. so there is precedent.

PR #3845, opened 2021-09-21
Fix radiasoft/flashcap#42: flash has no template/flash/lib dir so do …

e-carlin:
…not assume it exists

### DIFF
--- a/sirepo/sim_data/__init__.py
+++ b/sirepo/sim_data/__init__.py
@@ -354,10 +354,20 @@ class SimDataBase(object):
 
     @classmethod
     def lib_file_resource_path(cls, path):
-        return sirepo.resource.file_path(
-            _TEMPLATE_RESOURCE_DIR,
-            cls.sim_type(),
-        ).join(cls._LIB_RESOURCE_DIR, path)
+        try:
+            return sirepo.resource.file_path(
+                _TEMPLATE_RESOURCE_DIR,
+                cls.sim_type(),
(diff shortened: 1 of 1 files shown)


sirepo/sim_data/__init__.py
```diff
@@ -357,8 +357,7 @@ def lib_file_resource_path(cls, path):
         return sirepo.resource.file_path(
             _TEMPLATE_RESOURCE_DIR,
             cls.sim_type(),
-            cls._LIB_RESOURCE_DIR,
-        ).join(path)
+        ).join(cls._LIB_RESOURCE_DIR, path)
```
robnagler, line 360: Concerned about this solution. I'm not sure it fixes the problem. If someone creates, say, `template/<some-type>` and only wants to have the examples augmented, then it will be found, and there won't be any lib files in it (because the lib dir doesn't exist).

robnagler (changes requested):

sirepo/sim_data/__init__.py
e-carlin: Shoot, that's true. Need to think some more...

e-carlin: 91e5a986f thoughts?

e-carlin:
f4beabdfc

e-carlin:
@robnagler all tests passing. ready for another review.

robnagler (approved):

PR #3846, opened 2021-09-22
Fix #3827 bot detect

### LINKED ISSUE #3827: detect bot when throwing errors
Use https://pypi.org/project/robot-detection/ to detect if user-agent is a bot, and then throw a real HTTP status instead of javascript redirect in sirepo.http_reply.

mkeilman:
Notes:

- this uses the existing ```user_agents``` instead of ```robot_detection```. I tested both and got the same result
- I think there is a "cuter" way to do getApplicationData in the test, but this works
- "werkzueg*" user agent is NOT considered a bot (by either ```user_agents``` or ```robot_detection```)
- I left the redirect section of srunit in place

I presume the real test will be when bluesky is in the picture...


sirepo/http_reply.py
```diff
@@ -283,6 +283,8 @@ def _as_attachment(resp, content_type, filename):
 
 def _gen_exception_error(exc):
     pkdlog('unsupported exception={} msg={}', type(exc), exc)
+    if sirepo.http_request.is_spider():
+        raise werkzeug.exceptions.InternalServerError
```
robnagler, line 287: I think this needs to be `raise werkzeug.exceptions.InternalServerError()`.

tests/uri_router_test.py
```diff
@@ -9,6 +9,24 @@
...
+    from sirepo import srunit
+
+    fc = srunit.flask_client()
+    fc.sr_login_as_guest()
+
+    uri = '/get-application-data'
+    d = PKDict(simulationType='srw', method='NO SUCH METHOD')
+
+    r = fc.post(uri, json=d)
+    pkeq(200, r.status_code)
```
robnagler, line 24: I wouldn't expect this to succeed. Document it by perhaps checking the return result redirect to /#not-found or whatever we expect to happen.

tests/uri_router_test.py
```diff
@@ -9,6 +9,24 @@
...
+
+    r = fc.post(uri, environ_base={'HTTP_USER_AGENT': 'SPIDER'}, json=d)
```
robnagler, line 26: Prefer PKDict and set it to python-requests, i.e. `PKDict(HTTP_USER_AGENT='python-requests/2.26.0')`. python-requests should be considered a spider. If it isn't modify is_spider to recognize it and return True.

It's probably good that werkzeug/* is not a spider, because that would break all our tests, I think.

robnagler (changes requested):

sirepo/http_reply.py
e-carlin: I was curious. Looks like it can be either https://docs.python.org/3.7/tutorial/errors.html#raising-exceptions

mkeilman: 85a40e1

We use ```raise Whatever()``` everywhere else so I changed it

robnagler: Consistency is better. Looks unfinished otherwise. Semantics would be different btw with:

```py
def f():
   return ValueError
```

and

```py
def f():
    return ValueError
```

In the first case, the exception is probably created inside the call to `f`. In the second, the exception would be created at the time `raise` was called with the return value of `f`.

Python is very strange...

tests/uri_router_test.py
mkeilman: f04f83c

Oddly, python-requests is a bot according to ```robot_detection``` but not ```user_agents```.  I added it to a list for ```is_spider()``` to check

tests/uri_router_test.py
mkeilman: f04f83c

I added a known "good" user agent to this test, and a check for the error page in the result data

tests/uri_router_test.py
```diff
@@ -9,6 +9,39 @@
...
+    fc = srunit.flask_client()
+    fc.sr_login_as_guest()
+
+    uri = '/get-application-data'
+    d = PKDict(simulationType='srw', method='NO SUCH METHOD')
+
+    # "Real" browsers get redirected to an error page with a 200 status
+    r = fc.post(
+        uri,
+        environ_base=PKDict(HTTP_USER_AGENT=http_request.USER_AGENT_MOZILLA),
```
robnagler, line 28: This type of testing is problematic. We want to decouple the meaning, since the test is supposed to be testing something implicit, not explicit. Rather, you would really want to bring in some actual browser UA values here as literals.

tests/uri_router_test.py
```diff
@@ -9,6 +9,39 @@
...
+        json=d
+    )
+    pkeq(200, r.status_code)
+    pkok(
```
robnagler, line 32: Use pkre

tests/uri_router_test.py
```diff
@@ -9,6 +9,39 @@
...
+        '/srw#/error' in pkcompat.from_bytes(r.data),
```
robnagler, line 33: Again the rule is parsing loose. Here we want to know that error appears, but we do not want to create an implicit coupling that `/srw` is a URL we expect. The implicit part is that applications begin with `/<app>`. While this is "known", the fewer tests that rely on this, the more easily it is changed. Therefore, you simply want to test for `/error` in the data, e.g.

```py
pkunit.pkre('/error', r.data)
```

If pkre doesn't support this, then we should fix pkre to support the conversion of the data. Again, parsing loose here: we can assume that pkre is going to do a reasonable job of matching the result. I noticed that it doesn't actually implement what it says it is implementing (action is type object). It should stringify its result like it does in other cases. So, in your never ending question please also fix https://github.com/radiasoft/pykern/issues/106

tests/uri_router_test.py
```diff
@@ -9,6 +9,39 @@
...
+        'Unknown data method should redirect to error page'
+    )
+
+    r = fc.post(
+        uri,
+        environ_base=PKDict(HTTP_USER_AGENT=http_request.USER_AGENT_PYTHON),
```
robnagler, line 39: Here we want a specific version, e.g. python-requests/2.0 and maybe another version python-requests/1.3.

robnagler (changes requested):

sirepo/http_request.py
```diff
@@ -15,6 +15,10 @@
 import sirepo.srschema
 import user_agents
 
+USER_AGENT_MOZILLA = 'Mozilla/5.0'
+USER_AGENT_PYTHON = 'python-requests/2.26.0'
```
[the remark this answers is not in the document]
mkeilman: 1164018

sirepo/http_request.py
```diff
@@ -27,7 +31,8 @@ def init(**imports):
 
 
 def is_spider():
-    return user_agents.parse(flask.request.headers.get('User-Agent')).is_bot
+    a = flask.request.headers.get('User-Agent')
+    return user_agents.parse(a).is_bot or a in _BOTS
```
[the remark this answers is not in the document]
mkeilman: 1164018

tests/uri_router_test.py
mkeilman: 1164018

tests/uri_router_test.py
mkeilman: 1164018

tests/uri_router_test.py
mkeilman: 1164018

tests/uri_router_test.py
mkeilman: 1164018

mkeilman:
Ready - note that commit 1164018 depends on the pykern fix being ready first

sirepo/http_request.py
```diff
@@ -27,7 +29,8 @@ def init(**imports):
 
 
 def is_spider():
-    return user_agents.parse(flask.request.headers.get('User-Agent')).is_bot
+    a = flask.request.headers.get('User-Agent')
+    return user_agents.parse(a).is_bot or any([a in b or b in a for b in _BOTS])
```
robnagler, line 33: This code is unclear: `a in b or b in a for b in _BOTS`. It looks as if you don't actually know what's going on. I do not think this needs to be generalized with a constant. It's a special case due to a missing definition in user_agents. For this specific case, it's better to document a solution:

```py
if 'python-requests' in a:
    # user_agent doesn't see Python's requests module as a bot.
    # The package robot_detection does see it, but we don't want to introduce another dependency.
    return True
return user_agents.parse(a).is_bot
```

robnagler (changes requested):
The test looks very good, thanks.

sirepo/http_request.py
mkeilman: db24678

robnagler:
Are we ready to go with this?

mkeilman:
I think so -  @e-carlin is tagged to review but I don't see any comments

robnagler:
Are we ready to go here?

PR #3935, opened 2021-11-03
WIP radiasoft/ops#645: ENOENT when trying to rename flash.tar.gz

e-carlin:
https://github.com/radiasoft/sirepo/blob/ce452bff87ce8f3e44894c96bcb5d8ea4d95ca6b/sirepo/simulation_db.py#L780
`verify_app_directory` ensures that the app dir exists but not that
the lib dir exists. Maybe we were in a state where the app dir existed
but not the lib dir.  This should alert us if we encounter this again.


robnagler (changes requested):
Yes, it was an empty directory (no lib). I think it should create the lib.

e-carlin:
d0274815e @robnagler ready for another review.

robnagler (approved):

PR #3952, opened 2021-11-11
fix #3950 ModelArray delete bug

### LINKED ISSUE #3950: ModelArray component delete row always removes first row
The ModelArray component has a delete bug.

moellep:


mkeilman:
Is there an example somewhere that implements this?

moellep:
No, it isn't in any public repo yet.

PR #3969, opened 2021-11-23
Fix #3909: Raydata run multiple analyses

### LINKED ISSUE #3909: raydata: initial queuing of scans
- User searches for scans
- They select some from the results
- A job is started (parallel job for now) that run analysis notebook for each scan
- Another user searches and selects overlapping scans
- They start a job and they can see the results from the other user and will only run scans that haven't been run

Notes:
- locking will happen on the filesystem. If a directory (scan id for now) exists then that means that uid is running/completed and the user can just hook into that status

e-carlin:
Add the 'analysisStatusPanel' to Raydata which shows the status of
selected analysis jobs and allows one to start analysis for missing
jobs. In addition, each row in the table can be clicked if the job is
running or completed. A modal will appear with the png's output from
the notebook.

Two new API's:
- runMulti: Run multiple requests in one API call. In addition, these
requests can be "fire and forget". Normall runSimulation blocks until
the simulation starts. This can take a long while if there is resource
contention. This allows starting the job, not waiting for the
response, and then polling runStatus for updates.
- analysisJob: Call template code on the agent with access to the run
dir.

### DIFF
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -120,6 +120,9 @@ def _init():
             'Names of root packages that should be checked for codes and resources. Order is important, the first package with a matching code/resource will be used. sirepo added automatically.',
         ),
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
+        raydata=dict(
+            data_dir=('raydata', str, 'dir to store raydata output under sirepo.srdb.root')
+        ),
         sim_common=dict(
             hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
         ),
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -35,7 +35,6 @@ _NON_PROD_FOSS_CODES = frozenset((
     'genesis',
     'irad',
     'myapp',
-    'raydata',
     'rcscon',
     'rs4pi',
     'silas',
@@ -98,6 +97,7 @@ def sim_common():
 
 def _init():
     from pykern import pkconfig
+    from pykern import pkio
     global _cfg
 
     def b(msg, dev=False):
@@ -106,6 +106,10 @@ def _init():
             bool,
             msg,
         )
+
+    def _raydata_data_dir(path):
+        from sirepo import srdb
+        return srdb.root().join(path)
 
     _cfg = pkconfig.init(
         # No secrets should be stored here (see sirepo.job.agent_env)
@@ -121,7 +125,8 @@ def _init():
         ),
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
         raydata=dict(
-            data_dir=('raydata', str, 'dir to store raydata output under sirepo.srdb.root')
+            # POSIT: Included in sirepo.job_driver.docker.cfg.aux_volumes
+            data_dir=(None, _raydata_data_dir, 'dir to store raydata output under sirepo.srdb.root')
         ),
         sim_common=dict(
             hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
@@ -151,6 +156,9 @@ def _init():
         if v[0] in s:
             s.add(v[1])
     _cfg.sim_types = frozenset(s)
+    if 'raydata' in _cfg.sim_types:
+        assert _cfg.raydata.data_dir, \
+            'raydata is a sim type but no cfg.raydata.data_dir (also check job_driver.cfg.aux_volumes)'
     _check_packages(_cfg.package_path)
     return _cfg
 
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -107,10 +107,6 @@ def _init():
             msg,
         )
 
-    def _raydata_data_dir(path):
-        from sirepo import srdb
-        return srdb.root().join(path)
-
     _cfg = pkconfig.init(
         # No secrets should be stored here (see sirepo.job.agent_env)
         api_modules=((), set, 'optional api modules, e.g. status'),
@@ -126,7 +122,7 @@ def _init():
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
         raydata=dict(
             # POSIT: Included in sirepo.job_driver.docker.cfg.aux_volumes
-            data_dir=(None, _raydata_data_dir, 'dir to store raydata output under sirepo.srdb.root')
+            data_dir=(None, pkio.py_path, 'abspath of dir to store raydata analysis output')
         ),
         sim_common=dict(
             hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
--- a/sirepo/job_api.py
+++ b/sirepo/job_api.py
@@ -22,6 +22,7 @@ import sirepo.http_request
 import sirepo.job
 import sirepo.mpi
 import sirepo.sim_data
+import sirepo.uri_router
 import sirepo.util
 
 
@@ -33,7 +34,7 @@ def adjust_supervisor_srtime(days):
     return _request(
         api_name='not used',
         _request_content=PKDict(days=days),
-        _request_uri=cfg.supervisor_uri + sirepo.job.SERVER_SRTIME_URI,
+        _request_uri=_supervisor_uri(sirepo.job.SERVER_SRTIME_URI),
     )
 
 
@@ -46,6 +47,11 @@ def api_admJobs():
     return _request(
         _request_content=PKDict(**sirepo.http_request.parse_post()),
     )
+
+
+@api_perm.require_user
+def api_analysisJob():
+    return _request()
 
 
 @api_perm.require_user
@@ -98,7 +104,7 @@ def api_jobSupervisorPing():
         k = sirepo.job.unique_key()
         r = _request(
             _request_content=PKDict(ping=k),
-            _request_uri=cfg.supervisor_uri + sirepo.job.SERVER_PING_URI,
+            _request_uri=_supervisor_uri(sirepo.job.SERVER_PING_URI),
         )
         if r.get('state') != 'ok':
             return r
@@ -136,6 +142,29 @@ def api_runCancel():
         pkdlog('ignoring exception={} stack={}', e, pkdexc())
     # Always true from the client's perspective
     return sirepo.http_reply.gen_json({'state': 'canceled'})
+
+
+@api_perm.require_user
+def api_runMulti():
+    def _api(api):
+        # SECURITY: Make sure we have permission to call API
+        a = sirepo.uri_router.check_api_call(api).__name__
+        # SECURITY: Only allow these two API's for now. Certain API's
+        # (ex api_admJobs) have more security checks in the method (ex
+        # check_user_has_role) in the Flask server that could be
+        # circumvented since we don't call the Falsk server method.
+        assert a in ('api_runSimulation', 'api_runStatus')
+        return a
+
+    r = []
+    for m in sirepo.http_request.parse_json():
+        c = _request_content(PKDict(req_data=m))
+        c.data.pkupdate(api=_api(c.data.api), awaitReply=m.awaitReply)
+        r.append(c)
+    return _request(
+        _request_content=PKDict(data=r),
+        _request_uri=_supervisor_uri(sirepo.job.SERVER_RUN_MULTI_URI),
+    )
 
 
 @api_perm.require_user
@@ -206,7 +235,7 @@ def _request(**kwargs):
                 '{}: max frame search depth reached'.format(f.f_code)
             )
     k = PKDict(kwargs)
-    u = k.pkdel('_request_uri') or cfg.supervisor_uri + sirepo.job.SERVER_URI
+    u = k.pkdel('_request_uri') or _supervisor_uri(sirepo.job.SERVER_URI)
     c = k.pkdel('_request_content') if '_request_content' in k else _request_content(k)
     c.pkupdate(
         api=get_api_name(),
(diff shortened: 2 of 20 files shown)


sirepo/job_api.py
```diff
@@ -138,6 +144,29 @@ def api_runCancel():
     return sirepo.http_reply.gen_json({'state': 'canceled'})
 
 
+@api_perm.require_user
+def api_runMulti():
+    def _api(api):
```
robnagler, line 149: nice doc

sirepo/package_data/static/js/raydata.js
```diff
@@ -151,13 +413,43 @@ SIREPO.app.directive('appHeader', function(appState, panelState) {
...
+	    let currentElement = $element;
+	    function compile() {
+		$compile($scope.html)($scope , function(compiledContent){
+		    const r = angular.element(compiledContent);
+		    currentElement.replaceWith(r);
+		    currentElement = r;
+		});
+	    }
+	    $scope.$on($scope.changeEvent, function(){
+		// Need to wait one digest cycle before compile. The
```
robnagler, line 443: could we name this "callNextDigestCycle" or something? I don't like 0 timeouts, because it's an implementation detail that should be hidden.

sirepo/package_data/template/raydata/parameters.py.jinja
```diff
@@ -1,11 +1,28 @@
...
-    'papermill',
-    # POSIT: same kernel isntalled by rscode-ipykernel
-    '--kernel=py3',
-    '--log-output',
-    '{{ input_name }}',
-    '{{ output_name }}',
-))
+def _main():
+    try:
+       os.makedirs('{{ scan_dir }}')
```
robnagler, line 9: doc locking

sirepo/pkcli/job_cmd.py
```diff
@@ -250,7 +255,6 @@ def _do_stateful_compute(msg, template):
 def _do_stateless_compute(msg, template):
     return _dispatch_compute(msg)
 
-
```
robnagler, line 253: pep8: two lines?

sirepo/pkcli/job_supervisor.py
```diff
@@ -138,12 +140,42 @@ def on_connection_close(self):
...
+    def post(self):
+        assert pkconfig.channel_in_internal_test(), \
+            'You can only adjust time in internal test'
+        sirepo.srtime.adjust_time(pkjson.load_any(self.request.body).days)
+        self.write(PKDict())
+
+
+class _ServerReqRunMulti(_ServerReq):
+
+    async def post(self):
```
robnagler, line 160: nice

sirepo/template/raydata.py
```diff
@@ -63,38 +62,29 @@
...
+                ),
             )
-        ]
+        )
 
-    def _sanitized_name(filename):
-        return sirepo.util.sanitize_string(filename) + 'Animation'
+    def _paths():
+        d = _dir_for_scan_uuid(_parse_scan_uuid(data))
+        for f in sorted(glob.glob(str(d.join('*.png'))), key=os.path.getmtime):
```
robnagler, line 78: add key to `pkio.sorted_glob` and this becomes:

```py
for f pkio.sorted_glob(d.join('*.png'), key='mtime'):
    yield f
```

I'm not sure `mtime` as a string is the best way to handle this, but `d.mtime` would not work. What you want is the method, or `key=lambda x: x.mtime()`, which is awkward. I think the string is ok, just have to call the method.

sirepo/template/raydata.py
```diff
@@ -63,38 +62,29 @@
...
-    res = PKDict(
-        pngOutputFiles=[
-            PKDict(name=_sanitized_name(f), filename=f) for f in _png_filenames()
-        ],
-    )
-    res.pkupdate(frameCount=len(res.pngOutputFiles))
-    if is_running:
-        return res.pkupdate(percentComplete=0)
-    return res.pkupdate(percentComplete=100)
+    return PKDict(data=list(map(_filename_and_image, _paths())))
```
robnagler, line 81: Why not use a comprehension: `data=[_filename_and_image(x) for x in _paths()]`

sirepo/template/raydata.py
```diff
@@ -63,38 +62,29 @@
...
-def sim_frame(frame_args):
-    return PKDict(image=pkcompat.from_bytes(
-        base64.b64encode(
-            pkio.read_binary(
-                sirepo.util.safe_path(frame_args.run_dir, frame_args.filename),
-            ),
-        ),
-    ))
+def background_percent_complete(report, run_dir, is_running):
+    if is_running:
```
robnagler, line 85: `return PKDict(percentComplete=0 if is_running else 100)`

sirepo/template/raydata.py
```diff
@@ -141,12 +131,23 @@ def _catalog():
...
+    ))
+
+
+def _generate_parameters_file(data, run_dir):
+    s = _parse_scan_uuid(data)
     return template_common.render_jinja(
         SIM_TYPE,
         PKDict(
-            input_name=data.models.analysisAnimation.notebook,
+            input_name=str(run_dir.join(data.models.analysisAnimation.notebook)),
```
robnagler, line 147: i don't think you should need `str()`, because jinja will render py.path as str.

sirepo/uri_router.py
```diff
@@ -70,9 +70,7 @@ def call_api(func_or_name, kwargs=None, data=None):
             # Any (GET) uri will have simulation_type in uri if it is application
             # specific.
             s = sirepo.http_request.set_sim_type(kwargs.get('simulation_type'))
-        f = func_or_name if callable(func_or_name) \
```
robnagler, line 73: nice

robnagler (approved):
I didn't see any bugs so I approved, but I'll let you decide to merge or fix comments and merge.

sirepo/package_data/static/js/raydata.js
e-carlin: be69ce9e9

sirepo/pkcli/job_cmd.py
e-carlin: b117dcbcc

sirepo/template/raydata.py
e-carlin: b9df0c573

sirepo/template/raydata.py
e-carlin: f2abd827b

sirepo/template/raydata.py
e-carlin: 267666695

sirepo/template/raydata.py
e-carlin: 7dc92d008

e-carlin:
@robnagler  I realized there is one outstanding question. How to mount `feature_config.raydata.data_dir` in the docker container? [This is the code](https://github.com/radiasoft/sirepo/blob/sirepo-3909-raydata-initial-queuing/sirepo/job_driver/docker.py#L260) for attaching volumes to the container. I know you want to avoid using `ifs` in the supervisor code. Any thoughts on how to implement?

robnagler:
I think we need to consider:

* does the user have privs for data_dir?
* where is it being mounted from (nfs or not)?
* where does the config come from?

I think we can do a limited implementation that doesn't use NFS and that assumes all users have access to data_dir. Since this is a very narrow scope, having a denormalized "sirepo.job_driver.docker.aux_volumes" that gets mounted inside the container is probably fine. Docker will create the parent directories so I don't think it should be a problem inside the agent container to just add it to the list of volumes. We could do something more clever, but straightforward is probably better.

robnagler (changes requested):
I have a question which we can discuss perhaps tomorrow.

sirepo/feature_config.py
```diff
@@ -107,6 +107,10 @@ def b(msg, dev=False):
             msg,
         )
 
+    def _raydata_data_dir(path):
+        from sirepo import srdb
+        return srdb.root().join(path)
```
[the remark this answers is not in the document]
e-carlin: aee9543ac

Can you define `data directory` and `global output directory`?

sirepo/feature_config.py
```diff
@@ -98,6 +97,7 @@ def sim_common():
 
 def _init():
     from pykern import pkconfig
+    from pykern import pkio
```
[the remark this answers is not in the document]
e-carlin: Sorry was leftover from a previous iteration. Now it is needed in _init aee9543ac

sirepo/feature_config.py
```diff
@@ -151,6 +156,9 @@ def b(msg, dev=False):
         if v[0] in s:
             s.add(v[1])
     _cfg.sim_types = frozenset(s)
+    if 'raydata' in _cfg.sim_types:
+        assert _cfg.raydata.data_dir, \
```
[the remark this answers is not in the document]
e-carlin: I've been using https://github.com/radiasoft/raydata/blob/f46da6a9d89c6cc14be8c342051c8de9afa51bb1/etc/sirepo-dev.sh which follows our pattern of using scripts to start the server with interesting configuration. But it breaks the pattern by having the script in raydata not in sirepo. It could be moved into sirepo/etc. I think even with defaulting we are going to need a script because raydata needs to be put on sim_types and possibly added to package_path. So, my feeling is better to have less code in feature_config and push it out to the script.

sirepo/template/raydata.py
```diff
@@ -131,8 +131,7 @@ def _catalog():
 
 def _dir_for_scan_uuid(scan_uuid):
     return pkio.py_path(sirepo.util.safe_path(
-        sirepo.srdb.root(),
-        sirepo.feature_config.for_sim_type(SIM_TYPE).data_dir,
+        str(sirepo.feature_config.for_sim_type(SIM_TYPE).data_dir),
```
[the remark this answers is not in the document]
e-carlin: aee9543ac

sirepo/package_data/template/raydata/parameters.py.jinja
```diff
@@ -1,11 +1,32 @@
...
-))
+def _main():
+    try:
+        # Creating the directory acts as a lock. If we are able to
+        # create the dir then we get the "lock". If the dir already
+        # exists then another agent has it locked or the analysis is
+        # completed.
+        os.makedirs('{{ scan_dir }}')
+    except FileExistsError:
+        return
```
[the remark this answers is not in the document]
e-carlin: Ideally we'd want to never get here (lock in postgres) and for the user to then get the state in the db. W/o postgres we could use state files and we could read it here but that adds new overhead I'm not sure is worth it. With the code as it is, the state will always be reported back completed. That is a bug because the state may not be completed (running or error). I'm not sure what the best implementation is. Thoughts?

sirepo/feature_config.py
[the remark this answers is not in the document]
e-carlin: If we are working under the model that the system for the phase 1 really only works with one user then we will never get here (we will get the sate from the supervisor) and this is good enough. We really don't even need any locking in that case...

sirepo/package_data/template/raydata/parameters.py.jinja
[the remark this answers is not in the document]
e-carlin: If we are working under the model that the system for the phase 1 really only works with one user then we will never get here (we will get the sate from the supervisor) and this is good enough. We really don't even need any locking in that case...

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2335,32 +2355,21 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, $http,
     };
 
     self.sendStatelessCompute = function(appState, successCallback, data, options) {
+	// TODO(e-carlin): This is a hack to get around dependency
+	// cycle. Comment out and use normal dependency injection (add
+	// 'panelState' to parameter list of 'requestSender') to see
+	// the cycle.
+	const panelState = $injector.get('panelState');
```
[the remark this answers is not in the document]
e-carlin: Ok. I'll be in the office tomorrow if you're around otherwise ping when free.

sirepo/feature_config.py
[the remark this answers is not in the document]
robnagler: The data directory is "input", that is, the XPCS files. Maybe they don't exist, and it's Bluesky that delivers all the input files directly to the notebook. My impress

The global output directory is where the notebook outputs its results. We are using those newly created directories as a mutex. These directories may be large, I believe, so they might not fit in srdb.root. We cannot symlink the directory  into srdb.root, because we have to mount it inside a container so the actual directory must be specified in the config (or I guess read from the symlinks, but this is annoying code to have write).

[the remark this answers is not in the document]
e-carlin: The only input files currently are the mask files. Those are in the run_dir and unzipped into the output directory 377f2efca

sirepo/package_data/template/raydata/parameters.py.jinja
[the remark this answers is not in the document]
e-carlin: c926e4820

sirepo/feature_config.py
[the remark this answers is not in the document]
e-carlin: Decided script was fine.

sirepo/package_data/static/js/sirepo.js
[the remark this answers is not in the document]
e-carlin: Reverted back to passing in panelState. It makes sense since appState is also passed in 03acd031d.

e-carlin:
Running test then will merge.

PR #3985, opened 2021-12-04
Fix #3981 blicon ml tabulated und fails

### LINKED ISSUE #3981: blicon: SRW ML data generator with tabulated undulator
The "export ML script" feature works OK for idealized undulator simulations, but fails with a "Magnetic Field is not defined" for simulations which use a tabulated undulator.

To reproduce, use the Export ML script with the "NSLS-II CHX beamline (tabulated)" simulation.

mkeilman:
Tabulated undulators are only "activated" when a ```plot_reports``` param is true. We don't want to produce plots as part of the rsopt runs (at least by default) so this adds a check for rsopt runs in the appropriate places.

Also cleans up the readme and shell script

### DIFF
--- a/sirepo/package_data/template/srw/rsoptExport.sh.jinja
+++ b/sirepo/package_data/template/srw/rsoptExport.sh.jinja
@@ -14,8 +14,13 @@ then
     if [[ $confirm == [yY] ]]; then
         echo Running SRW...
         eval "$run_py"
-        echo Propagated single electron intensities are in datasets/beam_intensities.npy
-        echo Corresponding parameter values are in datasets/parameters.npy
+        if [ $? -eq 0 ]
+        then
+            echo Propagated single electron intensities are in datasets/beam_intensities.npy
+            echo Corresponding parameter values are in datasets/parameters.npy
+        else
+            echo "ERROR: SRW failed with return code $?"
+        fi
     else
         echo "NOTE: to run SRW with generated data, use
         $run_py"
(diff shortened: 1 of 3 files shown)


sirepo/package_data/template/srw/rsoptExport.sh.jinja
```diff
@@ -14,8 +14,13 @@ then
     if [[ $confirm == [yY] ]]; then
         echo Running SRW...
         eval "$run_py"
-        echo Propagated single electron intensities are in datasets/beam_intensities.npy
-        echo Corresponding parameter values are in datasets/parameters.npy
+        if [ $? -eq 0 ]
```
e-carlin, line 17: I think this can be simplified to
```bash
if eval "run_py"; then
...
else
...
fi
```

sirepo/template/srw.py
```diff
@@ -1991,6 +1993,7 @@ def _set_magnetic_measurement_parameters(run_dir, v):
 
 def _set_parameters(v, data, plot_reports, run_dir):
     report = data.report
+    for_rsopt = report == 'rsoptExport'
```
e-carlin, line 1996: May be clearer to make this a function since it is used across functions.

mkeilman:
I was just working on the comments :). I will incorporate them into the next pr

moellep:
Yes, I thought this was good enough for the alpha release today.

PR #3989, opened 2021-12-08
Fix #3980: Poll bluesky for new scans and start analysis for them

### LINKED ISSUE #3980: raydata: Stream new scans from Bluesky
Be able to get new scans from bluesky (callback or streaming), start running analysis for them, and update the GUI

e-carlin:


sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,82 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+	if (s.every((e) => {
 	    // POSIT: If start is present so are the other fields we need
 	    return e in (simulationDataCache.scans || {}) && simulationDataCache.scans[e].start;
 	})) {
 	    successCallback(
-		maybeGetOnlyOneScan(
-		    scanUuids.map((u) => simulationDataCache.scans[u]
-		).sort((a, b) => a.start > b.start ? 1 : -1)),
+		s.map(
+		    (u) => simulationDataCache.scans[u]
```
mkeilman, line 44: Can omit parens around single parameter

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,82 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+		).sort((a, b) => a.start > b.start ? 1 : -1),
```
mkeilman, line 45: more conventional is to return the difference of the values

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,82 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
 	);
     };
 
+    self.getScansRequestPayload = function(scanUuids) {
+	return (scanUuids || Object.keys(appState.models.scans.selected)).map(s => {
+	    return {
+		models: appState.models,
+		report: s,
+		simulationType: SIREPO.APP_SCHEMA.simulationType,
+		simulationId: appState.models.simulation.simulationId
```
mkeilman, line 73: add trailing comma

sirepo/package_data/static/js/raydata.js
```diff
@@ -322,25 +376,34 @@ SIREPO.app.directive('analysisStatusPanel', function() {
 		    appState,
 		    (data) => $scope.images = data.data,
 		    {
-			computeModel: $scope.selectedScan.computeModel,
-			method: 'output_files'
+			method: 'output_files',
+			models: appState.models,
+			report: $scope.selectedScan.uid
```
mkeilman, line 381: trailing comma

sirepo/package_data/static/js/raydata.js
mkeilman: (or should we keep it for consistency?)

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,82 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
-    };
-
-    self.removeScanFromCache = function(scan) {
-	if (!simulationDataCache.scans) {
-	    return;
+	scan.selected = !scan.selected;
+	if (scan.selected) {
+	    appState.models.scans.selected[scan.uid] = true;
+	    self.updateScansInCache([scan]);
+	} else {
```
mkeilman, line 90: not sure about this ```} else {``` style...

mkeilman (commented):
I only have a handful of minor style comments

sirepo/package_data/static/js/raydata.js
e-carlin: 6d5085885

sirepo/package_data/static/js/raydata.js
e-carlin: 6d5085885

sirepo/package_data/static/js/raydata.js
e-carlin: 6d5085885

sirepo/package_data/static/js/raydata.js
e-carlin: I think omitting is good. 6d5085885

sirepo/package_data/static/js/raydata.js
e-carlin: Thanks didn't know about that. Also, caught a bug where sort didn't happen after the request to the server. e46a18f00

mkeilman (approved):
Looks cool to me!

sirepo/package_data/static/js/raydata.js
```diff
@@ -11,13 +11,16 @@ SIREPO.app.config(() => {
...
-SIREPO.app.factory('raydataService', function(appState, requestSender, simulationDataCache, timeService) {
+SIREPO.app.factory('raydataService', function(appState, requestSender, runMulti, simulationDataCache, timeService, $rootScope) {
     const self = {};
     let id = 0;
 
-    self.computeModel = function(scanUuid) {
-	return 'animation' + scanUuid;
-    };
+    function removeScanFromCache(scan) {
+	if (!simulationDataCache.scans) {
```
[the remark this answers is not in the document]
e-carlin: 6d5085885

sirepo/package_data/static/js/raydata.js
```diff
@@ -122,10 +143,40 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+	// still being sent from the backend (in parallelStatus).
+	if (data.state !== 'running' || ! data.scans) {
+	    return;
+	}
+	const u = [];
+	data.scans.forEach((s) => {
+	    if (appState.models.scans.selected[s.uid]) {
+		return;
+	    }
+	    raydataService.selectOrDeselectScan(s);
```
robnagler, line 167: I found selectOrDeselect to be a bit confusing, and here, it is used just to select, and the abstraction has to be broken to know that (line 164 above). Seems to me that you might want the toggle code to be simpler for the checkbox, and to have a select and a deleselect, e.g.

```js
if (s.selected) {
    select(s)
}
else {
    deselect(s)
}
```

And then the select and delesect functions would be simpler, too. And the use here would be same.

Not sure. Just noting.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1028,8 +1028,12 @@ SIREPO.app.factory('timeService', function() {
     const UNIX_TIMESTAMP_SCALE = 1000;
     const self = {};
 
+    self.getCurrentUnixTime = () => {
```
robnagler, line 1031: unixTimeNow()?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1028,8 +1028,12 @@ SIREPO.app.factory('timeService', function() {
...
+	return self.getUnixTime(new Date());
+    };
+
     self.getUnixTime = (date) => {
```
robnagler, line 1035: This is converting to unixTime so I would call it "toUnixTime" or just "unixTime".

sirepo/package_data/static/json/raydata-schema.json
```diff
@@ -58,14 +58,19 @@
         "planMetadata": {
             "notes": ["Notes", "Text", ""]
 	},
+	"pollBlueskyForScansAnimation": {
+	    "minutes": ["Minutes to poll for", "Integer", 1, "Number of minutes to poll for new scans", 1, 1440]
```
robnagler, line 62: Perhaps we want use use units like we do elsewhere eg. `Poll Interval [min]`

sirepo/package_data/template/raydata/poll_bluesky.py.jinja
```diff
@@ -0,0 +1,3 @@
+# -*- python -*-
+import time
+time.sleep({{ time }})
```
robnagler, line 3: Too many "time"s. Maybe `poll_secs`?

sirepo/pkcli/job_supervisor.py
```diff
@@ -158,11 +159,22 @@ def post(self):
 class _ServerReqRunMulti(_ServerReq):
 
     async def post(self):
+        async def _await_reply(content, handler):
+            r = copy.deepcopy(content.data)
+            del r['models']
```
robnagler, line 164: Doesn't this delete models.simulation? I don't know if that's used.

sirepo/pkcli/raydata.py
```diff
@@ -4,16 +4,48 @@
...
+    template_common.write_sequential_result(PKDict())
+
 def run_background(cfg_dir):
-    pksubprocess.check_call_with_signals(
-        [sys.executable, template_common.PARAMETERS_PYTHON_FILE],
-    )
+    _run()
+
+
+def create_scans(num_scans, delay=True):
```
robnagler, line 20: alphabetical order

sirepo/pkcli/raydata.py
```diff
@@ -4,16 +4,48 @@
...
+        RE(bluesky.plans.count(
+            [ophyd.sim.det1, ophyd.sim.det2],
+            num=5,
+            md={
+                'T_sample_': i,
+                'owner': 'foo',
+                'sequence id': i,
+            },
+        ))
+        a = set(raydata.catalog())
```
robnagler, line 45: Why is `a` needed?

sirepo/template/raydata.py
```diff
@@ -78,7 +78,28 @@ def _paths():
...
+    except Exception as e:
+        if not pkio.exception_is_not_found(e):
+            raise
+        t = sirepo.simulation_db.read_json(
+            run_dir.join(template_common.INPUT_BASE_NAME),
+        ).models.pollBlueskyForScansAnimation.start
+
+    s = []
+    for k, v in catalog().search({'time': {'$gte': t}}).items():
+        t = max(t, v.metadata['start']['time'])
```
robnagler, line 95: v.metadata.start.time?

sirepo/template/raydata.py
```diff
@@ -78,7 +78,28 @@ def _paths():
 
 
 def background_percent_complete(report, run_dir, is_running):
-    return PKDict(percentComplete=0 if is_running else 100)
+    r = PKDict(percentComplete=0 if is_running else 100)
+    if report != 'pollBlueskyForScansAnimation':
+        return r
+    try:
+        t = float(pkio.read_text(_BLUESKY_POLL_TIME_FILE).strip())
```
robnagler, line 85: run_dir.join(_BLUESKY_POLL_TIME_FILE)? I don't know what directory this is in.

robnagler (changes requested):
Seems fine. I'm a little unsure of the whole flow, but that's probably mostly me not spending the time to remember runMulti and how bluesky works.

sirepo/feature_config.py
```diff
@@ -120,10 +120,9 @@ def b(msg, dev=False):
             'Names of root packages that should be checked for codes and resources. Order is important, the first package with a matching code/resource will be used. sirepo added automatically.',
         ),
         proprietary_sim_types=(set(), set, 'codes that require authorization'),
-        raydata=dict(
-            # POSIT: Included in sirepo.job_driver.docker.cfg.aux_volumes
-            data_dir=(None, pkio.py_path, 'abspath of dir to store raydata analysis output')
-        ),
+        # SECURITY: This path should not be exposed to the user so do
```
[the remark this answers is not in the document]
e-carlin: I think how we configure paths shouldn't be exposed to users. It gives them information about how our system is organized that they don't need to know and that could help them attack us. OTOH maybe so much is exposed through rsconf that it doesn't really matter. What do you think?

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
-    self.getScansInfo = function(scanUuids, successCallback, options) {
-	function maybeGetOnlyOneScan(scans) {
-	    return s ? scans[0] : scans;
-	}
+    self.getScansInfo = function(successCallback, options) {
+	function helper(successcallback, options, recursionDepth) {
+	    const s = Object.keys(appState.models.scans.selected);
+	    if (s.every((e) => {
+		// POSIT: If start is present so are the other fields we need
+		return e in (simulationDataCache.scans || {}) && simulationDataCache.scans[e].start;
```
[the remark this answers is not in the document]
e-carlin: I think a utility could work.

Creating something like PKDict seems problematic because we have so much code that wouldn't be using it and from just looking at code I don't think you would be able to know whether the code was a plain JS object or one of our new objects. I think that would lead to bugs. Am I right to be concerned about this?

Here is how this code would possibly look using the utility:
```js
function jkunchecked_nested_get(obj, dottedKey) {
    let o = obj;
    for (const k of dottedKey.split('.')) {
        if (o === undefined) {
            return null;
        }
        o = o[k];
    }
    return o;
}


return jkunchecked_nested_get(simulationDataCache, `scans.${e}.start`)
```

I think that is a big improvement.

Thoughts?

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
-	if (scanUuids.every((e) => {
-	    // POSIT: If start is present so are the other fields we need
-	    return e in (simulationDataCache.scans || {}) && simulationDataCache.scans[e].start;
-	})) {
-	    successCallback(
-		maybeGetOnlyOneScan(
-		    scanUuids.map((u) => simulationDataCache.scans[u]
-		).sort((a, b) => a.start > b.start ? 1 : -1)),
-		simulationDataCache.scanInfoTableCols
+	    if (recursionDepth > 0) {
```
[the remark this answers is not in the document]
e-carlin: 94fabea51 better?

FWIW this was the same pattern we use in the supervisor https://github.com/radiasoft/sirepo/blob/94fabea51d56a78f2d7ad5ad4b1c60bb9bdc20e5/sirepo/job_supervisor.py#L632

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+		throw new Error(`infinite recursion detected scans=${JSON.stringify(s)} cache=${JSON.stringify(simulationDataCache.scans)}`);
+	    }
+	    requestSender.sendStatelessCompute(
+		appState,
+		(json) => {
```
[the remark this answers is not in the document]
e-carlin: I think it is a well used pattern in the frontend code:

```
js (3980-raydata-streaming)$ g '(json)' ~/src/radiasoft/sirepo/sirepo/package_data/static/js
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/silas.js:506:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:1134:        $scope.updatePlot = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:2338:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:2367:                $scope.updatePlot(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:2773:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:3076:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:3493:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:3670:                $scope.updatePlot(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:3801:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:3841:                $scope.updatePlot(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ml.js:588:                analysisReport.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ml.js:591:                    oldLoad(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ml.js:1374:                analysisReport.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ml.js:1822:            $scope.load = (json) => {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/jspec.js:325:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/webcon.js:630:                analysisReport.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/webcon.js:633:                    oldLoad(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/warpvnd.js:2656:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/warpvnd.js:2668:                $scope.updatePlot(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/warpvnd.js:2885:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/warpvnd.js:2934:                $scope.updatePlot(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/cookieconsent.js:1027:                return json.error ? toError(json) : {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/cookieconsent.js:1045:                return json.error ? toError(json) : {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:1472:function fromJson(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:1473:  return isString(json)
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:1474:      ? JSON.parse(json)
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:248:            plotScope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:249:                loadFunction(json);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/srw.js:3138:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/zgoubi.js:1030:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/zgoubi.js:1240:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/radia.js:3481:    self.objBounds = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/radia.js:3589:    self.vectorsToPolyData = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/raydata.js:57:         (json) => {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/raydata.js:678:                    (json) => {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/madx.js:446:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:1568:            $scope.load = function(json) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/opal.js:807:            $scope.load = (json) => {
```


I think JS prefers that "json" is an object and "json string" is a string that could be parsed as json.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#static_methods
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

I could be hindered by availability bias and I don't have a strong opinion either way. You've made this comment before so let's make a final decision.

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
 
     self.nextPngImageId = function() {
 	return 'raydata-png-image-' + (++id);
     };
 
-    self.parseComputeModel = function(computeModel) {
-	const p = computeModel.split('animation');
-	if (p.length !== 2) {
-	    throw new Error(`cannot parse computeModel=${computeModel}`);
+    self.selectOrDeselectScan = function(scan) {
```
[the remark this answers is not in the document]
e-carlin: ff872d0af

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+	if (! ('selected' in scan)) {
+	    scan.selected = false;
```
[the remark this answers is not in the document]
e-carlin: Can you give me some insight into how I could make it clearer?

What a scan get's selected for the first time (either through user clicking the checkbox or because we are notified of it through polling) then it won't have the 'selected' key so we add it and set it to false. But the following line will set it to true which is what we really want.

sirepo/package_data/static/js/raydata.js
```diff
@@ -30,70 +33,88 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
     self.setPngDataUrl = function (element, png) {
 	element.src = 'data:image/png;base64,' + png;
     };
 
+    self.startAnalysis = function(modelKey, scanUuids) {
+	runMulti.simulation(
+	    self.getScansRequestPayload(scanUuids),
+	    {modelName: modelKey}
+	);
+	$rootScope.$broadcast('runMultiSimulationStarted');
```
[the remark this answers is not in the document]
e-carlin: We don't use constants elsewhere. 

I still struggle to know when you like constants vs not. My impression was that you thought a reasonably unique string was as good as a constant?

```
js (3980-raydata-streaming)$ g 'rootscope..broadcast' ~/src/radiasoft/sirepo/sirepo/package_data/static/js
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:42:        $rootScope.$broadcast(name + '.summaryData', summaryData);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-plotting.js:43:        $rootScope.$broadcast('summaryData', name, summaryData);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/elegant.js:590:        $rootScope.$broadcast('elementAnimation.outputInfo', outputInfo);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/hellweg.js:295:            $rootScope.$broadcast('animation.summaryData', data.summaryData);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-components.js:451:                $rootScope.$broadcast('sr-clearDisableAfterClick');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-components.js:3679:                $rootScope.$broadcast('sbatchLoginDone');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-components.js:4359:        $rootScope.$broadcast('sbatchLoginSuccess');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular-route.js:712:        if ($rootScope.$broadcast('$routeChangeStart', preparedRoute, lastRoute).defaultPrevented) {
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular-route.js:727:        $rootScope.$broadcast('$routeUpdate', lastRoute);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular-route.js:749:                  $rootScope.$broadcast('$routeChangeSuccess', nextRoute, lastRoute);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular-route.js:754:              $rootScope.$broadcast('$routeChangeError', nextRoute, lastRoute, error);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/ngDraggable.js:193:                            $rootScope.$broadcast('draggable:start', {x:_mx, y:_my, tx:_tx, ty:_ty, event:evt, element:element, data:_data});
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/ngDraggable.js:209:                        $rootScope.$broadcast('draggable:move', { x: _mx, y: _my, tx: _tx, ty: _ty, event: evt, element: element, data: _data, uid: _myid });
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/ngDraggable.js:217:                        $rootScope.$broadcast('draggable:end', {x:_mx, y:_my, tx:_tx, ty:_ty, event:evt, element:element, data:_data, callback:onDragComplete, uid: _myid});
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:15331:        defaultPrevented = $rootScope.$broadcast('$locationChangeStart', newUrl, oldUrl,
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:15367:            var defaultPrevented = $rootScope.$broadcast('$locationChangeStart', newUrl, oldUrl,
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/ext/angular.js:15397:      $rootScope.$broadcast('$locationChangeSuccess', $location.absUrl(), oldUrl,
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:62:                    $rootScope.$broadcast('irad-roi-available');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:73:                $rootScope.$broadcast('irad-compute-dose-comparison', plane, doseFrames);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:95:                $rootScope.$broadcast('irad-vti-available', frame);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:171:        $rootScope.$broadcast('irad-roi-activated', roiNumber);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:1485:                        $rootScope.$broadcast('irad-dicom-zoom', $scope.model.dicomPlane, xAxisScale, yAxisScale);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/irad.js:1622:                $rootScope.$broadcast('irad-dicom-point', info);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/zgoubi.js:442:            $rootScope.$broadcast('activeBeamlineChanged');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/radia.js:3317:                $rootScope.$broadcast('radiaViewer.loaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/radia.js:3318:                $rootScope.$broadcast('vtk.hideLoader');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/radia.js:3366:                    $rootScope.$broadcast('vtk.showLoader');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/raydata.js:117:        $rootScope.$broadcast('runMultiSimulationStarted');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:138:        $rootScope.$broadcast('sr-addToBeamline', item);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:199:                $rootScope.$broadcast('elementDeleted', type, element);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:250:        $rootScope.$broadcast('activeBeamlineChanged', updateNoWait);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:957:                $rootScope.$broadcast('sr-beamlineDropItem', data, $scope.beamlineItems);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:988:                        $rootScope.$broadcast('sr-beamlineDropItem', item, $scope.beamlineItems);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo-lattice.js:1996:                $rootScope.$broadcast('sr-beamlineItemSelected', item.beamlineIndex);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:325:        $rootScope.$broadcast('clearCache');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:329:        $rootScope.$broadcast(name + '.changed');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:330:        $rootScope.$broadcast('modelChanged', name);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:334:        $rootScope.$broadcast('modelsLoaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:338:        $rootScope.$broadcast(name + '.saved');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:339:        $rootScope.$broadcast('modelSaved', name);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:460:            $rootScope.$broadcast('cancelChanges', name[i]);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:472:        $rootScope.$broadcast('modelsUnloaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:1338:            $rootScope.$broadcast('framesCleared');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:1343:            $rootScope.$broadcast('framesLoaded', oldFrameCount);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:1413:        $rootScope.$broadcast('sr-window-resize');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:1813:            $rootScope.$broadcast(showEvent);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:1830:                $rootScope.$broadcast(showEvent);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:2151:            $rootScope.$broadcast('showSbatchLoginModal', e.params);
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/sirepo.js:3591:    $rootScope.$broadcast('simulationUnloaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:144:            $rootScope.$broadcast('roiPointsLoaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:159:                $rootScope.$broadcast('roiPointsLoaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:165:        $rootScope.$broadcast('roiActivated');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:180:            $rootScope.$broadcast('planeCoordChanged');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:186:        $rootScope.$broadcast('refreshDicomPanels');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:271:            $rootScope.$broadcast('roiPointsLoaded');
/home/vagrant/src/radiasoft/sirepo/sirepo/package_data/static/js/rs4pi.js:1517:                    $rootScope.$broadcast('updatePlaneFrameIndex', targetPlane, newIndex);
```

sirepo/package_data/static/js/raydata.js
```diff
@@ -104,15 +125,15 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
 	    simulationDataCache.scans = {};
 	}
 	return scans.map((s) => {
-	    if (!s.uid) {
-		if (!s.computeModel) {
-		    throw new Error(`scan=${JSON.stringify(s)} must have uid or computeModel.`);
-		}
-		s.uid = self.parseComputeModel(s.computeModel);
+	    // Requests with awaitReply (ex runStatus). Will have a
```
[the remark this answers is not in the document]
e-carlin: I made a function for this specific case which I think the makes the comment clearer. e748917f7 . Better?

sirepo/package_data/static/js/raydata.js
```diff
@@ -104,15 +125,15 @@ SIREPO.app.factory('raydataService', function(appState, requestSender, simulatio
...
+	    // request and a response part of the body so we know
+	    // which response goes which with request
+	    if ('request' in s) {
+		s.uid = s.request.report;
 	    }
 	    simulationDataCache.scans[s.uid] = angular.extend(
 		simulationDataCache.scans[s.uid] || {},
-		s
+		s.response || s
```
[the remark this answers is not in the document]
e-carlin: e81123257 the new function now passes the response in. I think this function was doing too much before.

sirepo/feature_config.py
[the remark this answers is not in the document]
robnagler: RSConf config or it's output (e.g. `env`) is not visible to the user so I think that's a non-issue.

In this particularly case, I guess the config hits the javascript(?) so you're concerned about that?

While I appreciate hiding config (in general), I think we are in the gray area of hiding too much at the expense of added complexity (I like the modularized cfg in sub-dict `raydata`). One of the things we don't do well is protect error messages or output from code. If the notebook outputs something about the images (eg their full path), then the information will be exposed. Since we have moved to (almost :) full execution in agents, I've been less concerned about path security. Once arbitrary code can be executed (uploaded notebooks), the entire container's file system is visible, and therefore modifiable unless specifically mounted read-only (which data_dir can't be). 

In this particular application, we have to assume the users are trusted. We do this anyway, since there is no concept of user in Bluesky. If there was, we would only mount that user's files in the container, and we'd buffer all accesses through controlling credentials to Bluesky in the server. If this happens in a future version, the root of the directory can be known, because it would not be mounted in the container, just the user's directory.

sirepo/package_data/static/js/raydata.js
e-carlin: I agree that it is confusing. There is a fair bit of state (and caching) that needs to be organized. I think spreading that across two functions may lead to bugs. It is clearer to me with an if/else that you need to do one thing in the select case and reverse it in the deselect case. If it is across functions one of the functions may be outside of my view and I would forget to update it. 

I think your code also breaks the abstraction some knowing that there is a "selected" field on the scan (although the checkbox code needs to know this so the cat is already out of the bag). I tried 0d4a6a78a which may be just more confusing. Not sure.

sirepo/package_data/static/js/raydata.js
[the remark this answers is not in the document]
robnagler: I agree this is a big improvement. I don't like the name (in pkcollections either :). The "jk" sounds like just kidding. I think this is a sirepo specific so something like "srxxx". 

Do we have functions that just float around? Perhaps we need something like sr.nestedGet where `sr` is a global with utility functions. Perhaps it could also be a factory. Maybe something relating to classes. I think the argument could also be a list, e.g. sr.nestedGet(simulationDataCache, ['scans', e, 'start']). It's a bit ugly, but it would work.

I think an object might be better. I think it wouldn't lead to bugs, because calling a function (nestedGet) will fail with an exception if the Object's attribute doesn't exist or isn't a function. 

@moellep?

sirepo/package_data/static/js/sirepo.js
e-carlin: 2fbd304ef

sirepo/package_data/static/js/sirepo.js
e-carlin: 2fbd304ef

sirepo/package_data/static/js/raydata.js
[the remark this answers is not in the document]
robnagler: Yes, clearer to use a boolean.

sirepo/package_data/static/js/raydata.js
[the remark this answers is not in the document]
robnagler: I don't remember making the comment before. :)  If it's in widespread use, leave it. Not important enough. I'll try to refrain from not remember that I made this comment before. :)

sirepo/package_data/static/js/raydata.js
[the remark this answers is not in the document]
robnagler:
```
if (selected !== undefined) {
    scan.selected = selected;
}
else if ('selected' in scan) {
    scan.selected = !scan.selected;
}
else {
    scan.selected = true;
}
```

robnagler (approved):
These comments got lost due to GitHub's outdated "feature":

> I made a function for this specific case which I think the makes the comment clearer. e748917 . Better?

Yes.

> We don't use constants elsewhere.
> I still struggle to know when you like constants vs not. My impression was that you thought a reasonably unique string was as good as a constant?

I like enums for cases like this, because it provides a central place for a decentralized concept. We don't always use enums, e.g. job.OP_ANALYSIS, and that's fine, too.

There are many places I don't bother with constants, e.g. when it's the only definition or when it will be asserted at the referenced location in the code. For example, config (pkconfig) will raise if the key doesn't exist, so it's fine to use a string.

So a simple rule is: if it's fail fast, strings are fine.

robnagler:
I will let you merge, because I think you are still working on it.

sirepo/package_data/static/json/raydata-schema.json
e-carlin: 2eacb899b I don't like interval because to me that means "poll for new scans once every n mins" which isn't what is going on.

sirepo/package_data/template/raydata/poll_bluesky.py.jinja
e-carlin: 2eacb899b

sirepo/pkcli/job_supervisor.py
e-carlin: Yes, it deletes all models. Models aren't useful for the GUI to know which request the response is to. All the GUI needs is the report (scan uuid). Really that is a problem because the server uses computeJid/hash/serial to identify which are things the GUI doesn't know much about....

sirepo/pkcli/raydata.py
e-carlin: 9abb7c424

sirepo/pkcli/raydata.py
e-carlin: 9abb7c424 not needed. leftover from testing

sirepo/template/raydata.py
e-carlin: v isn't a pkdict. It is a bluesky document

sirepo/template/raydata.py
e-carlin: f5d600154

sirepo/feature_config.py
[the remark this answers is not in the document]
e-carlin: > RSConf config or it's output (e.g. env) is not visible to the user so I think that's a non-issue.

True. My thinking was that rsconf still exposes some info. How much is too much I have no clue (other than the obvious like a password). For example, rsconf exposes that proprietary codes live in a `proprietary_codes` dir that are at one level down under the run_dir. I don't know that this info is all that useful but it may be enough for an attacker to say "well I just need to be able to reach *this* level in the filesystem and I can get at useful info".

> In this particularly case, I guess the config hits the javascript(?) so you're concerned about that?

Yes. If we do the way of cfg in a sub-dict then the user can type `SIREPO.APP_SCHEMA.feature_config` in the console and see the path.

> If the notebook outputs something about the images (eg their full path), then the information will be exposed

Their full path inside of the docker container will be exposed. With the raydata.data_dir the path on the host machine will be exposed.

sirepo/package_data/static/js/raydata.js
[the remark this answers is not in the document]
e-carlin: a2236021f thanks

sirepo/feature_config.py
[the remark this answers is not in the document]
robnagler: > For example, rsconf exposes that proprietary codes live in a proprietary_codes dir that are at one level down under the run_dir.

The point you are making is subtle. If someone wants to hack Sirepo, you are saying they can extract a lot of information from Sirepo, Pykern, and RSConf, because we write open source software, and we don't try to hide or configure everything on the assumption that the host system can be secured reasonably. This is, perhaps, over-reliance on Docker. Knowing a host path doesn't do much, if you can't inject code on the host. That's the premise here.

> "well I just need to be able to reach this level in the filesystem and I can get at useful info"

The question is what is useful. RSConf exposes `env` in the source code of RSConf. If we use Vault or something else, it's really just a bit of an abstraction that defers access. I don't think it fixes anything when, say, someone can inject code. Rather, knowing app uses Vault gives a lot of information. Using microservices can further obfuscate things, but I don't think it ensures much if someone can penetrate the system.

Originally I wanted to hide all paths, but given the nature of being a scientific gateway for complex codes, it's become harder. Scientists need all the info they can get to debug. 

> Their full path inside of the docker container will be exposed. With the raydata.data_dir the path on the host machine will be exposed.

I think that's generally true, because we almost always mount the same name inside the agent (except NERSC where we don't control things). This helps debugging. There's a difference between exposing a path and having access to the path, even if it does give a clue to how to hack the system. 

In this particular case, this is a behind-the-firewall implementation for a particular class of database that we can't protect. I don't think knowing paths is going to change the level of security of the whole system or even in leaving a "bad example" for others to follow.

That said, this is pretty abstract so make a choice, and I'll approve. :)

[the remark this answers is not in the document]
e-carlin: Thanks for all the details:) 

> Originally I wanted to hide all paths, but given the nature of being a scientific gateway for complex codes, it's become harder. Scientists need all the info they can get to debug.

I guess I was still working under the "hide all paths" model. If you're cool exposing the path then I'm happy to work that way.

> In this particular case, this is a behind-the-firewall implementation for a particular class of database that we can't protect.
That makes sense. In cases for other codes should I be concerned about exposing paths? What about paths that contain the user's sirepo uid?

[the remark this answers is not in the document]
e-carlin: bf96aecc7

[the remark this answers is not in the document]
robnagler: > If you're cool exposing the path then I'm happy to work that way.

I think each time we do this, we need to look at it. Security is a process, not a rule.

> In cases for other codes should I be concerned about exposing paths?

I don't think so in general, because run_dir gets exposed. However, see note above. :)

> What about paths that contain the user's sirepo uid?

That's an interesting one. run_dir contains uid and sid. I don't know what we can do about this than stripping all output. That's hard to do, because it comes in many forms. I think in general we should be careful where we can, but until we fix the general output filtering problem, there will be leakage.

e-carlin:
Merged. We can continue talking about https://github.com/radiasoft/sirepo/pull/3989#discussion_r766897105 and I'll fix.

sirepo/feature_config.py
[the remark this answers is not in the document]
e-carlin: "process not a rule" I like it.

What you said sounds good.

PR #3992, opened 2021-12-09
Fix #3659 radia parameterized dipole

### LINKED ISSUE #3659: radia: parameterize dipole example
Similar to the undulator, allow the user to adjust features of the dipole.  Will need input as to the most useful parameters

mkeilman:
Using deepcopy

### DIFF
--- a/sirepo/sim_data/__init__.py
+++ b/sirepo/sim_data/__init__.py
@@ -414,10 +414,11 @@ class SimDataBase(object):
     @classmethod
     def model_defaults(cls, name):
         """Returns a set of default model values from the schema."""
+        import copy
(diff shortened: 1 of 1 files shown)


sirepo/sim_data/__init__.py
```diff
@@ -414,10 +414,11 @@ def lib_files_to_run_dir(cls, data, run_dir):
     @classmethod
     def model_defaults(cls, name):
         """Returns a set of default model values from the schema."""
+        import copy
```
robnagler, line 417: Just FYI, low level modules like this can be imported at the top, but this is fine.

robnagler (approved):

PR #3993, opened 2021-12-10
Fix #3659 radia parameterized dipole

### LINKED ISSUE #3659: radia: parameterize dipole example

mkeilman:
This update allows the user to create three different kinds of parameterized dipoles (basic -- just two permanent magnets -- c-bend (one coil), and h-bend (two coils)). Once the default values get settled they can be turned into examples (so the current dipole example will be replaced). The interesting bits are:

- field definitions in the schema can refer to models, e.g. ```"coil": ["Coil", "model.racetrack", {...}]```. Then a view can include the  sub-object fields as ```"dipoleC.coil.color"```
- more use of schema-based "sub-classes" on the client side so their fields can be populated properly
- some simplification of building geometries of complex magnets. c.f. ```template.radia._update_geom_from_dipole()``` and the older ```template.radia._update_geom_from_undulator()```. I will update the latter after this pr is approved.

One thing I have left in is the "smelly" code in ```simulation_db._merge_dicts(base, derived, depth=-1, extend_arrays=True)``` - I can't see a ready alternative but perhaps the hive-mind...


sirepo/package_data/static/js/radia.js
```diff
@@ -1030,6 +1069,19 @@ SIREPO.app.controller('RadiaSourceController', function (appState, geometry, pan
...
                     }
                 }
+                if (o.type === 'racetrack') {
+                    // calculate the size
+                    let s = [0, 0, 0];
+                    const sides = utilities.splitCommaDelimitedString(o.sides, parseFloat);
+                    const radii = utilities.splitCommaDelimitedString(o.radii, parseFloat);
+                    const i = geometry.basis.indexOf(o.axis);
+                    s[i] = o.height;
+                    for (let j of [0, 1]) {
```
e-carlin, line 1079: j can be const.

sirepo/package_data/static/js/radia.js
```diff
@@ -1030,6 +1069,19 @@ SIREPO.app.controller('RadiaSourceController', function (appState, geometry, pan
                         self.selectedObject = null;
                     }
                 }
+                if (o.type === 'racetrack') {
```
e-carlin, line 1072: I don't know radia so take this with a grain of salt. But, I'm struggling to understand what is going on here. I think this whole on model changed function is quite large and could benefit from more functions. At the very least it would be helpful to me to bundle all of the work for o.type === racetrack into a named function. I think since sides and radii aren't single letter variable names it is a tip to me that too much is going on.

sirepo/package_data/static/js/radia.js
```diff
@@ -1330,8 +1393,17 @@ SIREPO.app.directive('bevelTable', function(appState, panelState, radiaService)
                         return;
                     }
                     $scope.selectedItem = null;
+                    let m = appState.models[modelName];
```
e-carlin, line 1396: m can be const.

sirepo/package_data/static/js/radia.js
```diff
@@ -3723,89 +3780,25 @@ SIREPO.viewLogic('objectShapeView', function(appState, panelState, radiaService,
 
     $scope.whenSelected = function() {
         modelType = appState.models.geomObject.type;
+        $scope.modelData = appState.models[$scope.modelName];
+        radiaService.updateModelAndSuperClasses(modelType, $scope.modelData);
```
e-carlin, line 3784: I don't fully understand the super class stuff yet so excuse my ignorance but is there a reason the state has to be denormalized across the model and all super classes? I ask because there is something off to me about having to call this method.

sirepo/package_data/static/js/radia.js
```diff
@@ -3723,89 +3780,25 @@ SIREPO.viewLogic('objectShapeView', function(appState, panelState, radiaService,
...
         return m ? m : [parent.modelName, f];
     }
 
-
     function updateObjectEditor() {
         modelType = appState.models.geomObject.type;
-        calcExtrusionPoints();
         parent.activePage.items.forEach((f) => {
             const m = modelField(f);
+            let hasField = SIREPO.APP_SCHEMA.model[modelType][m[1]] !== undefined;
```
e-carlin, line 3797: const

sirepo/package_data/static/js/radia.js
```diff
@@ -3822,6 +3815,55 @@ SIREPO.viewLogic('geomObjectView', function(appState, panelState, radiaService,
     };
 });
 
+for (let d of SIREPO.APP_SCHEMA.enum.DipoleType) {
+    SIREPO.viewLogic(d[0] + 'View', function(appState, panelState, radiaService, $scope) {
+        let editedModels = [];
+        let models = {};
```
e-carlin, line 3821: const

sirepo/package_data/static/js/radia.js
e-carlin: Also, does a saveChanges have to happen after?

sirepo/package_data/static/js/radia.js
```diff
@@ -3909,10 +3966,16 @@ SIREPO.viewLogic('simulationView', function(activeSection, appState, panelState,
...
 
     $scope.watchFields = [
-        ['simulation.beamAxis'], updateSimEditor,
+        ['simulation.beamAxis', 'simulation.magnetType'], updateSimEditor,
         ['simulation.heightAxis'], radiaService.setWidthAxis,
     ];
 
+    $scope.whenSelected = function() {
+        model = appState.models[$scope.modelName];
+        //$scope.modelData = model;
```
e-carlin, line 3975: remove?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -374,14 +374,15 @@ SIREPO.app.factory('appState', function(errorService, fileManager, requestQueue,
         if (! self.isLoaded()) {
             return false;
         }
-        var models = self.models;
-        for (var m in fieldsByModel) {
+        let models = self.models;
```
e-carlin, line 377: const

sirepo/package_data/static/js/sirepo.js
```diff
@@ -374,14 +374,15 @@ SIREPO.app.factory('appState', function(errorService, fileManager, requestQueue,
...
+        for (let m in fieldsByModel) {
             if (models[m]) {
                 if (! savedModelValues[m]) {
                     return true;
                 }
-                for (var i = 0; i < fieldsByModel[m].length; i++) {
-                    var f = fieldsByModel[m][i];
+                const a = Array.from(fieldsByModel[m]);
+                for (let i = 0; i < a.length; i++) {
```
e-carlin, line 384: I think you can use a for...of loop here.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -812,6 +817,16 @@ SIREPO.app.factory('appState', function(errorService, fileManager, requestQueue,
         return model;
     };
 
+    self.superClasses = (modelName) => {
```
e-carlin, line 820: Is there a reason this is classes plural? Looking through schemas I only see that things have one superClass.

sirepo/srschema.py
```diff
@@ -233,6 +233,9 @@ def _validate_number(val, sch_field_info):
     # used for other purposes, so we return rather than fail for non-numeric values
     except ValueError:
         return
+    # ignore object-valued fields
```
e-carlin, line 236: `except (ValueError, TypeError)`

e-carlin (commented):
I'm submitting this to checkpoint but I still have more to review. 

I didn't call out all of the places but all for...of loops can use const instead of let.

sirepo/package_data/static/js/radia.js
```diff
@@ -240,6 +253,16 @@ SIREPO.app.factory('radiaService', function(appState, fileUpload, panelState, re
             });
     };
 
+    // update models so that editors see the correct values
+    // for now assign the entire object
+    self.updateModelAndSuperClasses = (modelName, model) => {
+        const s = [modelName, ...appState.superClasses(modelName)];
+        for (let c of s) {
```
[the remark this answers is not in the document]
mkeilman: a4fb637

sirepo/package_data/static/js/radia.js
mkeilman: a4fb637

sirepo/package_data/static/js/radia.js
mkeilman: a4fb637

sirepo/package_data/static/js/radia.js
mkeilman: a4fb637

sirepo/package_data/static/js/radia.js
mkeilman: a4fb637

sirepo/package_data/static/js/sirepo.js
mkeilman: a4fb637

sirepo/template/radia.py
```diff
@@ -538,6 +548,15 @@ def _build_field_axis(length, beam_axis):
     return f
 
 
+def _build_obj(obj_type, beam_axis, height_axis, **kwargs):
+    return globals()[f'_update_{obj_type}'](
```
e-carlin, line 552: use `pkinspect.module_functions`

sirepo/srschema.py
mkeilman: a16e4f0

sirepo/template/radia.py
```diff
@@ -546,6 +565,11 @@ def _copy_geom_obj(o):
     return o_copy
 
 
+def _delim_string(val=None, default_val=None):
+    d = default_val if default_val is not None else []
+    return sirepo.util.to_comma_delimited_string(val if val is not None else d)
```
e-carlin, line 570: I think this could all be simplified to 

`return sirepo.util.to_comma_delimited_string(val or default_val or [])`

sirepo/package_data/static/js/radia.js
mkeilman: e282207

sirepo/template/radia.py
```diff
@@ -584,6 +608,11 @@ def _field_lineout_plot(sim_id, name, f_type, f_path, beam_axis, v_axis, h_axis)
     )
 
 
+def _find_obj_by_id(obj_arr, obj_id):
```
e-carlin, line 611: For me this is easier to read. May not be the case for all

```python
def _find_obj_by_id(obj_arr, obj_id):
  return next(filter(lambda o: o.id == obj_id, obj_arr), None)
```

sirepo/package_data/static/js/sirepo.js
mkeilman: 13fbddd

sirepo/template/radia.py
```diff
@@ -1092,51 +1126,179 @@ def _save_kick_map_sdds(name, x_vals, y_vals, h_vals, v_vals, path):
...
+    return _update_cee_points(_update_geom_obj(o, **kwargs), beam_axis, height_axis)
 
 
-def _update_ell(ell, beam_axis, height_axis, **kwargs):
-    ell = _update_geom_obj(ell, **kwargs)
-    _update_ell_points(ell, beam_axis, height_axis)
-    return ell
+def _update_cee_points(o, beam_axis, height_axis):
+    g = _update_stemmed_points(o, beam_axis, height_axis)
+    w, h, b, c, s = g.geom
```
e-carlin, line 1146: h and s are unused. Make `_`

sirepo/template/radia.py
```diff
@@ -1092,51 +1126,179 @@ def _save_kick_map_sdds(name, x_vals, y_vals, h_vals, v_vals, path):
...
-def _update_ell_points(ell, beam_axis, height_axis):
-    w, h, b = _geom_directions(beam_axis, height_axis)
-    ctr = sirepo.util.split_comma_delimited_string(ell.center, float)
-    sz = sirepo.util.split_comma_delimited_string(ell.size, float)
-    c = [numpy.sum(w * ctr), numpy.sum(h * ctr)]
-    s = [numpy.sum(w * sz), numpy.sum(h * sz)]
+    o.points = _update_extrusion_points(
+        [
+            [ax1, ay1], [ax2, ay1], [ax2, ay2],
+            [sx2, ay2], [sx2, sy2], [ax2, sy2], [ax2, sy1], [sx1, sy1],
```
e-carlin, line 1154: From the looks of it for each call to _update_extrusion_points we do the exact same work except this line is different. I think more objects should be passed to the function (ex pass the whole g and o) and have it do more pulling things apart.

sirepo/package_data/static/js/sirepo.js
mkeilman: The ```_super``` array gets extended by ```simulation_db._merge_subclasses()``` so it contains the chain of supers back to the root

sirepo/template/radia.py
```diff
@@ -1092,51 +1126,179 @@ def _save_kick_map_sdds(name, x_vals, y_vals, h_vals, v_vals, path):
...
+    m_obj = _find_obj_by_id(geom.objects, m.id)
+    c = dipole.coil
+    c_obj = _find_obj_by_id(geom.objects, c.id)
+    cp_obj = _find_obj_by_id(geom.objects, dipole.corePoleGroup.id)
+    mc_obj = _find_obj_by_id(geom.objects, dipole.magnetCoilGroup.id)
+
+    mag_sz = numpy.array(sirepo.util.split_comma_delimited_string(m.size, float))
+
+    if dipole.dipoleType == 'dipoleC':
+        # resize the poles to fit between the arms (minus gap)
```
e-carlin, line 1255: Could you use a named function instead of the comment?

sirepo/template/radia.py
```diff
@@ -1092,51 +1126,179 @@ def _save_kick_map_sdds(name, x_vals, y_vals, h_vals, v_vals, path):
...
-    if w.tolist().index(1) != (b.tolist().index(1) + 1) % 3:
-        for p in ell.points:
+    if width_dir.tolist().index(1) != (length_dir.tolist().index(1) + 1) % 3:
+        for p in pts:
             p.reverse()
 
+    return pts
+
+
+def _update_geom_from_dipole(geom, dipole, beam_axis, height_axis):
```
e-carlin, line 1228: This function is quite large and doing a lot of work. I think it could benefit from calling more functions to make it clearer to the untrained user (me) what the math operations are doing.

sirepo/template/radia.py
```diff
@@ -1236,6 +1398,8 @@ def _update_geom_from_undulator(geom, und, beam_axis, height_axis):
     else:
         half_pole = _update_cuboid(
             half_pole,
+            None,
```
e-carlin, line 1401: What about making all of the `_update_*` functions take just kwargs and then you wouldn't have to do this? Also, there are many `def _update_...` functions and I don't think you dispatch to all of them. The ones that you do dispatch to should have a different name so it is clearer they are called through dispatch.

sirepo/template/radia.py
```diff
@@ -1389,9 +1564,32 @@ def _update_geom_obj(o, delim_fields=None, **kwargs):
...
+
+
+def _update_racetrack(o, _, __, **kwargs):
+    return _update_geom_obj(o)
+
+
+def _update_stemmed_points(o, beam_axis, height_axis):
+    w, h, b, c, s = _stemmed_geometry(o, beam_axis, height_axis)
+
+    # start with arm top, stem left - then reflect across centroid axes as needed
```
e-carlin, line 1578: I think using named functions would be clearer than a comment and could reduce some of the repeated code.

sirepo/template/radia.py
```diff
@@ -1413,6 +1611,18 @@ def _update_kickmap(km, und, beam_axis):
     km.periodLength = und.periodLength
 
 
+def _validate_objects(objects):
+    from numpy import linalg
```
e-carlin, line 1615: Use fully qualified import `import numpy.linalg`

sirepo/template/radia_util.py
```diff
@@ -193,6 +193,11 @@ def build_container(g_ids):
     return radia.ObjCnt(g_ids)
 
 
+def build_racetrack(center=[0, 0, 0], size=[1, 1, 1], radii=[1, 2], sides=[1, 1], height=1, axis='x', calc='a', num_segs=3, curr_density=-1.0):
```
e-carlin, line 196: I think it would be easier to just pass the object in and have this method pull out the bits it needs.

e-carlin (commented):
Finished just looking over the code. When you've made changes lmk and I'll run it and see how it works.

sirepo/template/radia.py
mkeilman: 0054c06

sirepo/template/radia.py
mkeilman: 565d86c

sirepo/template/radia.py
mkeilman: That does not work because you get ```ValueError: The truth value of an array with more than one element is ambiguous```

sirepo/package_data/static/js/radia.js
mkeilman: The modelField directive needs the model and field, so if we have
```
baseModel: {
   baseField: ["Base Field", "BaseType", "Base Default"]
},
derivedModel: {
    "_super": ["_", "model", "baseModel"],
    "derivedField": ["Derived Field", "DerivedType", "Derived Default"]
}
```
and we want to edit ```baseField``` on the ```derivedModel```, we have to supply ```baseModel``` to modelField. We should be editing the object of interest  instead of its ancestors, but this gets the job done for now.

Also, in this case we can't save because it will instantly close a dialog. The user has not modified anything yet.

sirepo/template/radia.py
mkeilman: Hm, I like the list comprehension better. With your indulgence I'd like to table this for discussion in the dev meeting

sirepo/template/radia.py
mkeilman: d044d5f

sirepo/template/radia.py
e-carlin: Sure thing. 

I also noticed that _find_obj_by_id() and _find_obj_by_name() consist of the exact same code. Maybe create _find_obj_by_key() and pass in name or id

sirepo/template/radia_util.py
mkeilman: This may be misguided, but I want to keep ```radia_util``` as agnostic as possible as to the details of sirepo objects. This is partly so that the ```geometryReport.py``` generated from the template can remain "pure" for use in notebooks (there is work to do in this regard)

e-carlin: Ok, I think we should add this to the SW agenda too. I think in other circumstances (ex rsopt) we lean heavily on using sirepo objects outside of sirepo.

mkeilman:
The remaining comments relate to clarifying and streamlining the code. I think these are important to address but do not affect the functionality, so I propose making issues for them to deal with after this pr is approved. We need feedback from @cchall et al on the UI etc.

sirepo/template/radia.py
mkeilman: agree, see recent comment

sirepo/template/radia.py
mkeilman: agree, see recent comment

sirepo/template/radia.py
mkeilman: agree, see recent comment

sirepo/template/radia.py
mkeilman: agree, see recent comment

e-carlin:
Sounds good. I tried importing this [Dipole.dat.txt](https://github.com/radiasoft/sirepo/files/7713550/Dipole.dat.txt)(export of the dipole example). And I think am stuck on this screen
Screen Shot 2021-12-14 at 10 23 13 AM

mkeilman:
Looks like imports don't work at all even in production, so not related to these changes.  I'll open a new issue

e-carlin:
Ok, is there a way I can test the code w/o an import?

mkeilman:
Yes, make a new simulation and choose a dipole of whatever type. Incidentally the dipole example is not parameterized - it will be replaced with the new model once it's ready

sirepo/package_data/static/js/radia.js
```diff
@@ -1030,6 +1069,19 @@ SIREPO.app.controller('RadiaSourceController', function (appState, geometry, pan
                         self.selectedObject = null;
                     }
                 }
+                if (o.type === 'racetrack') {
+                    // calculate the size
```
e-carlin, line 1073: How does one create an object of this type (racetrack)? It isn't in shapeSelector.availableTypes so I'm not sure how to exercise this code.

mkeilman: create a  new freehand simulation and go to the design tab. There is a racetrack icon in the toolbar you can drag and drop into the workspace:

Screen Shot 2021-12-15 at 9 43 03 AM

(there is a bug where the shape does not immediately show up when you drop it (issue forthcoming). Go back to the viz tab to see the coil)

e-carlin: ah thanks!

e-carlin (approved):
LGTM! I'll leave merging to you.

PR #3994, opened 2021-12-10
Fix #3983 blicon set rnd seed

### LINKED ISSUE #3983: blicon: set random seed for ML data generation
When generating the SRW ML script, an option should be provided to set the random seed so the data is reproducible.

### LINKED ISSUE #3984: blicon: column names with ML data
When the ML output files are created, it would be nice to have the column names associated with the varied values within the parameters file. If this could be done directly in numpy that would be good, or else using something like pandas.DataFrame

### LINKED ISSUE #3987: blicon: unused reports can be turned off for ML generator
Only the v.ws flag (single electron wavefront propagation) needs to be turned on for the ML reports. All the other reports should be disabled. 

```python
    op = set_optics(v, names, True)
    v.ws = True
    # v.ss = True                                                                                                                      
    # v.sm = True                                                                                                                      
    # v.pw = True                                                                                                                      
    # v.si = True                                                                                                                      
    # v.tr = True                                                                                                                      
```

mkeilman:
Notes:

- adds a randomSeed field, along with buttons to generate a seed randomly or based on time (modulo the max value set in the schema)
- fixes #3984 by just putting everything into an hdf5 file. There did not seem to be a reason to separate the data. The shell script and README reflect this change
- fixes #3987, explicitly turning off unused reports

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -544,6 +544,41 @@ SIREPO.app.directive('exportPythonLink', function(appState, panelState) {
     };
 });
 
+SIREPO.app.directive('randomSeed', function() {
+    return {
+        restrict: 'A',
+        scope: {
+            field: '=',
+            form: '=',
+            max: '=',
+            model: '=',
+            modelName: '=',
+            viewName: '=',
+        },
+        template: `
+            <div class="col-sm-3">
+                <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true" required/>
+            </div>
+            <button class="btn btn-default" data-ng-click="setSeedRandom()" title="generate random seed"><span class="glyphicon glyphicon-random"></span></button>
+            <button class="btn btn-default" data-ng-click="setSeedTime()" title="use current time"><span class="glyphicon glyphicon-time"></span></button>
+        `,
+        controller: function($scope) {
+
+            $scope.setSeedRandom = () => {
+                $scope.model[$scope.field] = Math.floor(Math.random() * $scope.max - 1);
+            };
+
+            $scope.setSeedTime = () => {
+                $scope.model[$scope.field] = (new Date()).getTime() % $scope.max;
+            };
+
+            if (! $scope.model[$scope.field] && $scope.model[$scope.field] !== 0) {
+                $scope.setSeedTime();
+            }
+        },
+    };
+});
+
 SIREPO.app.directive('srTooltip', function(appState, mathRendering, $interpolate) {
     return {
         restrict: 'A',
@@ -564,6 +599,7 @@ SIREPO.app.directive('srTooltip', function(appState, mathRendering, $interpolate
                         // evaluate angular text first if {{ }} is present
                         if (/\{\{.*?\}\}/.test(res)) {
                             $scope.appState = appState;
+                            $scope.SIREPO = SIREPO;
                             res = $interpolate(res)($scope);
                         }
                         if (mathRendering.textContainsMath(res)) {
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -557,7 +557,7 @@ SIREPO.app.directive('randomSeed', function() {
         },
         template: `
             <div class="col-sm-3">
-                <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true"/>
+              <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true"/>
             </div>
             <button class="btn btn-default" data-ng-click="setSeedRandom()" title="generate random seed"><span class="glyphicon glyphicon-random"></span></button>
             <button class="btn btn-default" data-ng-click="setSeedTime()" title="use current time"><span class="glyphicon glyphicon-time"></span></button>
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -557,7 +557,7 @@ SIREPO.app.directive('randomSeed', function() {
         },
         template: `
             <div class="col-sm-3">
-                <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true" required/>
+                <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true"/>
             </div>
             <button class="btn btn-default" data-ng-click="setSeedRandom()" title="generate random seed"><span class="glyphicon glyphicon-random"></span></button>
             <button class="btn btn-default" data-ng-click="setSeedTime()" title="use current time"><span class="glyphicon glyphicon-time"></span></button>
@@ -571,10 +571,6 @@ SIREPO.app.directive('randomSeed', function() {
             $scope.setSeedTime = () => {
                 $scope.model[$scope.field] = (new Date()).getTime() % $scope.max;
             };
-
-            if (! $scope.model[$scope.field] && $scope.model[$scope.field] !== 0) {
-                $scope.setSeedTime();
-            }
         },
     };
 });
--- a/sirepo/package_data/static/json/srw-schema.json
+++ b/sirepo/package_data/static/json/srw-schema.json
@@ -651,7 +651,7 @@
             "numCores": ["Number of Cores", "Integer", 4],
             "numSamples": ["Samples per Parameter", "Integer", 100, "", 1],
             "numWorkers": ["Number of Workers", "Integer", 1],
-            "randomSeed": ["Random Seed", "RandomSeed", null, "Explicit seed for random numbers"],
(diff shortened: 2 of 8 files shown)


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -544,6 +544,37 @@ SIREPO.app.directive('exportPythonLink', function(appState, panelState) {
...
+            field: '=',
+            form: '=',
+            max: '=',
+            model: '=',
+            modelName: '=',
+            viewName: '=',
+        },
+        template: `
+            <div class="col-sm-3">
+                <input data-string-to-number="integer" data-ng-model="model[field]" data-min="0" data-max="max" class="form-control" style="text-align: right" data-lpignore="true"/>
```
e-carlin, line 560: 2 spaces

e-carlin: Also, when should we use lpignore? This is the first time I've noticed it and I see it is on other input tags.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -544,6 +544,37 @@ SIREPO.app.directive('exportPythonLink', function(appState, panelState) {
...
+            <button class="btn btn-default" data-ng-click="setSeedTime()" title="use current time"><span class="glyphicon glyphicon-time"></span></button>
+        `,
+        controller: function($scope) {
+
+            $scope.setSeedRandom = () => {
+                $scope.model[$scope.field] = Math.floor(Math.random() * $scope.max - 1);
+            };
+
+            $scope.setSeedTime = () => {
+                $scope.model[$scope.field] = (new Date()).getTime() % $scope.max;
```
e-carlin, line 572: We now have a [`timeService`](https://github.com/radiasoft/sirepo/blob/03acd031d8586233a98a27eeff38b810a4777ae0/sirepo/package_data/static/js/sirepo.js#L1027) for this kind of thing. #3989 has the method you want. Feel free to fix up later once this and that PR have been merged.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -564,6 +595,7 @@ SIREPO.app.directive('srTooltip', function(appState, mathRendering, $interpolate
                         // evaluate angular text first if {{ }} is present
                         if (/\{\{.*?\}\}/.test(res)) {
                             $scope.appState = appState;
+                            $scope.SIREPO = SIREPO;
```
e-carlin, line 598: Is this needed?

sirepo/package_data/static/json/srw-schema.json
```diff
@@ -651,6 +651,7 @@
             "numCores": ["Number of Cores", "Integer", 4],
             "numSamples": ["Samples per Parameter", "Integer", 100, "", 1],
             "numWorkers": ["Number of Workers", "Integer", 1],
+            "randomSeed": ["Random Seed", "RandomSeed", null, "Explicit seed for random numbers in the range [{{ SIREPO.APP_SCHEMA.model.exportRsOpt.randomSeed[SIREPO.INFO_INDEX_MIN] }}, {{ SIREPO.APP_SCHEMA.model.exportRsOpt.randomSeed[SIREPO.INFO_INDEX_MAX] }}). Leave empty to allow server to decide the seed.", 0, 4294967296],
```
e-carlin, line 654: The interpolation is cool!

e-carlin (approved):

e-carlin:
Minor nits. LGTM. 

One thing when I ran it it said
`Continue? (** will run SRW 0 times **)? (Y/N): y`

It then seemed to run something even though it was going to run **0** times. I don't know much about SRW so maybe running 0 times makes sense to a real user.

sirepo/package_data/static/js/sirepo-components.js
mkeilman: 9abe9da

As for lpignore, all I know is that it blocks LastPass from auto-filling fields. I just copy-pasted it in; I don't think we have any guidelines as to when we actually need it

sirepo/package_data/static/js/sirepo-components.js
mkeilman: This or some other way to get the schema info on the scope so that the interpolation works. This was quickest

mkeilman:
> Minor nits. LGTM.
> 
> One thing when I ran it it said `Continue? (** will run SRW 0 times **)? (Y/N): y`
> 
> It then seemed to run something even though it was going to run **0** times. I don't know much about SRW so maybe running 0 times makes sense to a real user.

Looks like a bug - you should not be able to generate a script if you have not enabled any elements

mkeilman:
Requesting a look at the fix just added - nervous about ```rm -rf```...

mkeilman:
> I am concerned about rm -rf, too, especially as the first thing in a shell script. Removing a directory with rm -rf called "datasets", etc. is not a good idea.
> 
> I would create a temporary directory with a unique name, e.g. rsopt-$RANDOM, and then cd into that directory to do the operations, then remove the entire directory.
> 
> If the program gets run again without the delete happening, then it will not collide ($(date +%Y%m%d%H%M%S) or $RANDOM). Frankly, I don't think the delete is needed at all. It would be up to the user to remove them.
> 
> In this case, I would write the code more simply:
> 
> ```shell
> for d in runDir beams data_files datasets parameters; do
>     rm -rf --preserve-root "$d"
> done
> ```
> 
> The double quotes is problematic if you are substituting jinja variables, because the shell will interpolate the value, which is not correct.

This section started as an admonition to empty the scan directory.  I'll go back to that

robnagler:
I think it would be better to create a temporary directory and print the name at the start. Creating lots of arbitrary directories is going to be problematic. We should copy in the files into that directory along with a comment of how the script was invoked. This would go a long ways towards reproducible research.

mkeilman:
Do you then also recommend doing away with 'beams' directory etc.?

robnagler:
I am recommending something like this:

```sh
#!/bin/bash
set -euo pipefail
d=rsopt-srw-$(date +%Y%m%d%H%M%S)
mkdir "$d"
cp name.yml name.py "$d"
cd "$d"
echo "Entering: $PWD"
```

There are UI issues with this that I don't understand. For example, name.sh might actually contain all the files internally and unpack them. This would be clearer from a UI perspective, because you don't know what to run when there are three files dumped into the directory. The program should also print a usage on how to run it.

mkeilman:
> I don't think this eval is necessary. Rather use:
> 
> run_rsopt=( rsopt sample configuration '{{ ymlFileName }}' )
> 
> then
> 
> "${run_rsopt[@]}" > '{{ outFileName }}' 2>&1

 It works...but I don't understand that syntax. All I can find is stuff like ```for i in "${LIST[@]}"; do```.

Also, why is ```eval "$run_py"``` desirable but not ```eval "$run_rsopt"``` ?

mkeilman:
> For example, name.sh might actually contain all the files internally and unpack them.

This I don't get. Can you point to an example of how to do that?

robnagler:
> It works...but I don't understand that syntax. All I can find is stuff like for i in "${LIST[@]}"; do.

REPL is your friend. If you don't understand a syntax in any language, try it. Play with it until you understand it. 

I have never been a big fan of polylingualism, but here we are: javascript, python, jinja, bash, etc. 

Given that we are wrapping codes, it's important to learn bash, which has the way we are doing a lot of wrapping. It's the way SLURM wraps sbatch scripts.

Learn the terms: [parameter substitution](https://tldp.org/LDP/abs/html/parameter-substitution.html), [variable substitution](https://tldp.org/LDP/abs/html/varsubn.html), and [command substitution](https://tldp.org/LDP/abs/html/commandsub.html). 

There are many cheatsheets on the web, e.g. https://devhints.io/bash.

> Also, why is eval "$run_py" desirable but not eval "$run_rsopt" ?

It's not. I didn't reply to that, 1) because I thought you'd get the pattern, and 2) because my review is a mess because I'm in "comment mode" not in review mode... Anyway, continuing in this mode...

The problem is the level of interpretation. eval does two levels of interpretation:

```sh
$ x='$y'
$ y=whoops
$ echo $x
$y
z=$( echo $x )
$ "${z[@]}"
$y
$ eval "${z[@]}"
whoops
```

Note that `$y` is evaluated in the `eval` case but not in the non-eval case. This is important, because you don't control the contents of run_opt.sh necessarily, that is, you can't see it in the jinja script. It might contain something funky. 

With bash, sh, etc. you need to understand when values are evaluated. Tcl is similar, because it came up recently since elegant uses it as part of its build system. Even Python has this now with f-strings. Anyway, it's important to understand when a value should be in double quotes (evaluated, aka interpolated, substituted) and when in single quotes (not-evaluated). `eval` is obvious, because that's what the function does, but when you see single quotes vs double quotes vs no-quotes in bash, sh, Perl, and Tcl. Each language behaves differently, even sh v bash. Just like parameter passing behaves differently in each of these languages.

robnagler:
> This I don't get. Can you point to an example of how to do that?

The concept is similar to how elegant's run.py works. You write a bunch of files and then execute them. It might look something like:

```sh
cd "$run_dir"
cat > x.yml <<'EOF'
{{ yml_contents }}
EOF
cat > x.py <<'EOF'
{{ py_contents }}
EOF
```

yml_contents and py_contents are jinja variables. They can be read in to the template and then assigned so you can keep the foo.yml.jinja and foo.py.jinja files that are in package_data now. 

Note the single quotes on the `<<'EOF'`. This protects the "here document" from getting substituted. Since we know there is not going to be a standalone `EOF` in the yml or py, we don't have to do anything more complicated than this. If the contents were unknown types, then we'd use something like SHAR which was in the [first volume of comp.sources.unix](http://ftp.sunet.se/mirror/archive/ftp.sunet.se/pub/usenet/ftp.uu.net/comp.sources.unix/volume1/).

mkeilman:
if the latest changes look OK I will do the merge

PR #4013, opened 2021-12-21
Sbatch agent logging

### LINKED ISSUE #2136: Slurm node state down
Leaving this here for myself or anyone else who runs across this.

Jobs submitted to sbatch were stuck in the 'pending' state because the localhost node was down on my vm. (`sinfo` reported: debug*       up   infinite      1   down localhost)

`systemctl status slurmd` reported slurmd was running but there was an error:
error: Node configuration differs from hardware: Procs=2:4(hw) Boards=1:1(hw) SocketsPerBoard=2:1(hw) CoresPerSocket=1:4(hw) ThreadsPerCore=1:1(hw)

The solution:
`sudo su -`
`scontrol`
`update NodeName=debug State=DOWN Reason="undraining"`
`update NodeName=debug State=RESUME`

### LINKED ISSUE #3658: Better logging for sbatch agents
When starting an sbatch agent we should read the existing job_agent.log (if it exists) ][before overwriting it](https://github.com/radiasoft/sirepo/blob/6fbf8035c63e4c332270913011aa35f33331ede5/sirepo/job_driver/sbatch.py#L150).This will allow devs some chance to see what went on with the job_agent. Currently, you have to know that the file exists (in dev) or we lose it all (in alpha and beyond).

e-carlin:
Fix #3658: log job_agent.log before overwriting when starting a new agent

    In addition in dev write all log messages read from the job_agent to
    the job_supervisor log stream.

    Fix #2136: reset slurm node state at every start

    The node on dev machines seemes to go down randomly. It is quick to
    reset it so do so every time the supervisor is started in an sbatch
    configuration.

### DIFF
--- a/sirepo/job_driver/sbatch.py
+++ b/sirepo/job_driver/sbatch.py
@@ -154,10 +154,9 @@ disown
             p = pkio.py_path(self._local_user_dir).join('agent-sbatch', self.cfg.host)
             pkio.mkdir_parent(p)
             f = p.join(f'{datetime.datetime.now().strftime("%Y%m%d%H%M%S")}-{filename}.log')
-            d = PKDict(stdout=stdout, stderr=stderr)
-            pkjson.dump_pretty(d, f)
+            r = pkjson.dump_pretty(PKDict(stdout=stdout, stderr=stderr, filename=f), f)
             if pkconfig.channel_in('dev'):
-                pkdlog(d.pkupdate(filename=f))
+                pkdlog(r)
 
         async def get_agent_log(connection, before_start=True):
             try:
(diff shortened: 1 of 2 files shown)


sirepo/job_driver/sbatch.py
```diff
@@ -153,19 +153,30 @@ async def _do_agent_start(self, op):
...
-            async with connection.create_process(
-                f'/bin/cat {agent_start_dir}/{log_file}'
-            ) as p:
-                o, e = await p.communicate()
-                write_to_log(o, e, 'remote')
+            f = p.join(f'{datetime.datetime.now().strftime("%Y%m%d%H%M%S")}-{filename}.log')
+            d = PKDict(stdout=stdout, stderr=stderr)
+            pkjson.dump_pretty(d, f)
+            if pkconfig.channel_in('dev'):
+                pkdlog(d.pkupdate(filename=f))
```
robnagler, line 160: do we not want to log the file name always?

robnagler (approved):

sirepo/job_driver/sbatch.py
e-carlin: I think it is redundant. In cases where we don't log it then we'll be reading from the file so we'll know the name. But, it doesn't hurt 9d59f2800

PR #4032, opened 2021-12-31
Fix #4030 radia replace uuids

### LINKED ISSUE #4030: radia: replace uuid for object ids
Do this on client and server side. This is to simplify the code and reduce computation, since these ids don't need to be globally unique or cryptographically strong.  I will also better document their purpose.

mkeilman:
Notes:

- I left the "UUID" field handling in place in ```SimDataBase.model_defaults```  since it could be useful, but am not opposed to ditching it
- ctrl-click of objects was not working in the radia viewer - fixed since this is a primary use of the id mapping

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -4599,13 +4599,12 @@ SIREPO.app.service('utilities', function($window, $interval) {
         };
     };
 
-    // create a non-cryptographic baseN string, with N <= 64
-    this.randomString = function(base=62, length=32) {
-        const BASE64 = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz+/';
-        const s = BASE64.substring(0, base);
+    // create a non-cryptographic base62 string
+    this.randomString = function(length=32) {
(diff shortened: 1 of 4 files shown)


sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4599,6 +4599,16 @@ SIREPO.app.service('utilities', function($window, $interval) {
         };
     };
 
+    // create a non-cryptographic baseN string, with N <= 64
+    this.randomString = function(base=62, length=32) {
```
e-carlin, line 4603: YAGNI on specifying the base? Always use base62

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4599,6 +4599,16 @@ SIREPO.app.service('utilities', function($window, $interval) {
...
+        const BASE64 = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz+/';
+        const s = BASE64.substring(0, base);
+        return new Array(length)
+            .fill('')
```
e-carlin, line 4607: Geez. I didn't know you had to do this. JS can be so strange.

e-carlin (approved):
LGTM one nit.

sirepo/package_data/static/js/sirepo-components.js
mkeilman: a4dd0cf

PR #4044, opened 2022-01-14
Fix #4038: Always call python parameters.py from slurm

### LINKED ISSUE #4038: NERSC: exec sirepo not found
sirepo command is not found on NERSC. Likely related to changes in https://github.com/radiasoft/sirepo/pull/4031/commits/8508dd7960859f35295d70a720c5c6989cdd2ad1

e-carlin:
In #4031 the code was changed to get the cmd output by prepare_simulation
and use it to initiate sirepo from slurm (ex srun sirepo srw run_background /path/to/run).

This doesn't work at NERSC because the sirepo command is not installed in the
user's environment. It is only avaiable once we are inside shifter.

So, we reverted to the old behavior of calling `python parameters.py`. But,
with the addition that all prologue/epilogue after calling a code (ex deleting large
intermediate files in srw coherentModesAnimation) must be done in the parameters.py.
So, if a code is going to be run at NERSC there can't be a pkcli run_background command
for it. This forces us to make sure all prologue/epilogue is done in parameters.py.

This means that any epilogue/prologue code will run in a parallel environment so it
must have proper protections (ex ensuring only one mpi process runs it by calling
`sirepo.mpi.restrict_op_to_first_rank`).


sirepo/package_data/template/common-header.py.jinja
```diff
@@ -7,3 +7,12 @@
 {% endfor %}
 """
 {% endif %}
+{% if mpi %}
+#TODO(e-carlin): discuss with rn if this is needed? It exists in sirepo.mpi but
+# it is guarded by a re of '^from __future.*' so is it only for py2?
+#TODO(e-carlin): share with sirepo.mpi
+from mpi4py import MPI
+if MPI.COMM_WORLD.Get_rank():
```
e-carlin, line 15: @robnagler is this code needed in py3? The regex for a future import is throwing me off but I think it probably is needed in py3. Should it be used in slurm too? Probably can't hurt...

tests/template/srw_generate_data/gaussian-x-ray-beam-through-a-beamline-containing-imperfect-mirrors.py
```diff
@@ -389,6 +389,21 @@ def set_optics(v, names=None, want_final_propagation=True):
 
 
 
+def epilogue():
```
robnagler, line 392: Epilogues only run in the server, not when user is running manually. I think we want to insert a simple code in the parameters.py only when run on our server:

```py
import sirepo.template
sirepo.template.run_epilogue('srw')
```

This will have access to in.json, for example, which run.py would not.

sirepo/pkcli/job_agent.py
```diff
@@ -700,18 +709,27 @@ async def _prepare_simulation(self):
...
-        if 'cmd' not in c:
-            raise AssertionError(pkdformat('No cmd returned from {}', c))
-        return ' '.join(c.cmd)
 
-    def _sbatch_script(self, start_cmd):
+    def _sbatch_script(self):
+        def _assert_no_run_background():
+            try:
+                from sirepo import pkcli
+                getattr(pkcli.import_module(self.msg), 'run_background')
```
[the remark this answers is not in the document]
e-carlin: 5e95a50e3

sirepo/pkcli/job_agent.py
```diff
@@ -738,12 +757,15 @@ def _processor():
...
+cat > bash.stdin <<'EOF'
 {self.job_cmd_source_bashrc()}
 {self.job_cmd_env()}
 if [[ ! $LD_LIBRARY_PATH =~ /usr/lib64/mpich/lib ]]; then
     export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib64/mpich/lib
 fi
-exec {start_cmd}
+exec python {template_common.PARAMETERS_PYTHON_FILE}
+EOF
+exec srun {'--mpi=pmi2' if pkconfig.channel_in('dev') else ''} {s} /bin/bash bash.stdin
```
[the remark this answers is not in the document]
e-carlin: 5e95a50e3

tests/template/srw_generate_data/gaussian-x-ray-beam-through-a-beamline-containing-imperfect-mirrors.py
e-carlin: 4b8ceee19

sirepo/template/srw.py
```diff
@@ -1965,6 +1968,10 @@ def _rsopt_main():
     ]
 
 
+def _run_with_mpi(data):
+   return _SIM_DATA.is_parallel(data) and data.report != 'beamlineAnimation'
```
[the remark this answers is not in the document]
e-carlin: 64f97aac7

sirepo/package_data/template/common-header.py.jinja
e-carlin: a565074a0

sirepo/package_data/template/srw/parameters.py.jinja
```diff
@@ -502,6 +501,15 @@ def _rsopt_set_params({{ rsOptFuctionSignature() }}):
 
 {% endif %}
 
+def epilogue():
```
robnagler, line 504: i don't see the point of epilogue function if it's just two lines in a template for server side only. Just inline it. The srwMain could we in an if __name__ == main, I think, so the code could be simplified for rsOpt, too.

robnagler (approved):

PR #4057, opened 2022-01-19
Fix #4050 rsopt zip security

### LINKED ISSUE #4050: rsOpt files written without directory
I think there are a few things here. The zip file should be generated in the agent, not the server. This would prevent files from getting written in the wrong place. The second problem is the files are getting written to an arbitrary location, when they should be in the user's tmpdir.

mkeilman:
Notes:

- The ```api_exportRSOptConfig``` method now lives in ```job_api```
- The zip file is created in a temp directory
- Accordingly, the zip operation needs to happen in the temp dir context. The template returns the items


sirepo/job_api.py
```diff
@@ -95,6 +95,36 @@ def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=No
...
+def api_exportRSOptConfig(simulation_type, simulation_id, filename):
+    import zipfile
+    t = sirepo.template.import_module(simulation_type)
+    assert hasattr(t, 'export_rsopt_config'), 'Export rsopt unavailable'
+    e = t.export_rsopt_config(
+        simulation_db.read_simulation_json(simulation_type, sid=simulation_id),
+        filename
+    )
+    with simulation_db.tmp_dir() as tmp:
+        fz = tmp.join(filename)
```
e-carlin, line 108: `sirepo.util.safe_path` instead of `join`. The filename could be malicious. 

One better option to consider - Use a non-user supplied filename and then only return the user supplied filename in the final `return http_reply.gen_file_as_attachment()`. I don't really know how zipfile and gen_file_as_attachment work so maybe this won't work.

sirepo/job_api.py
```diff
@@ -95,6 +95,36 @@ def api_downloadDataFile(simulation_type, simulation_id, model, frame, suffix=No
         )
 
 
+@api_perm.require_user
+def api_exportRSOptConfig(simulation_type, simulation_id, filename):
+    import zipfile
```
e-carlin, line 100: We are trying to move away from running code inside the server to inside the agent. I think this is a good example of code that should be run within the agent. From the looks of the code I think it should be a call to [statefulCompute](https://github.com/radiasoft/sirepo/wiki/Interacting-with-Simulations-Architecture#b-stateful-compute). You can see an example of this type of call [here](https://github.com/radiasoft/sirepo/blob/f45197e925b068846a96f6fa4b15c35640ad07e3/sirepo/package_data/static/js/controls.js#L197). 

One thing that will need to change is that the call to [`lib_files_for_export`](https://github.com/radiasoft/sirepo/blob/ec7ef734012eda4937139efc63fa367df48252c7/sirepo/template/srw.py#L400) will fail because that call is protected by [`sim_data._assert_server_side()`](https://github.com/radiasoft/sirepo/blob/ec7ef734012eda4937139efc63fa367df48252c7/sirepo/sim_data/__init__.py#L368) and we will be in the agent not in the server. @robnagler can you remind me why we have calls to assert_server_side()? Was it before we had ability to pull over lib files?

mkeilman: I thought job apis were run in the agent.  What distinguishes "agent" from "server"?

e-carlin: `job_api` just contains api endpoints in the flask server that _should_ be passed on to the supervisor (who will then pass onto the agent if the api is destined for the agent). The api function defined in job_api needs to call the [proper code](https://github.com/radiasoft/sirepo/blob/f45197e925b068846a96f6fa4b15c35640ad07e3/sirepo/job_api.py#L245) to get the request forward onto the supervisor.

That help? 

is the entrypoint from calls to the server to be passed to the supervisor who will then pass to the agent.

e-carlin: So a request destined for an agent goes on a journey like: GUI -> flask server (sirepo.job_api) -> supervisor (pkcli.job_supervisor -> sirepo.job_supervisor) -> job_driver (whichever type is configured / needed) -> job_agent

mkeilman:
Ready for another look.  Notes:

- Now uses statefulCompute and returns the zip data as base64
- Note this means the client has to convert that back to binary, which takes a little massaging.  See ```sirepo-components.utilities.base64ToBlob()```
- ```lib_files_for_export``` does not seem to be a problem. I've run the server and supervisor separately in both local and sbatch mode and it works...???
- ZipFile writes to a buffer in memory instead of a file

robnagler:
Do you think you could use data files to return the data?

e-carlin:
I was curious about how the download file code worked. It has been long enough that I had mostly forgotten about it. From the looks of it it should work. The sim id might need to [be passed in](https://github.com/radiasoft/sirepo/blob/be5ea66d2d2c4cf8916142cf91c756f0bf6bf65c/sirepo/pkcli/job_cmd.py#L128) unless there is an in.json in the run_dir and that could be read instead.

@mkeilman  - I'm sorry that I said statelessCompute earlier. I think @robnagler is right and downloadDataFile is a better match. As always, happy to meet to talk about how to use it.

mkeilman:
Update: there actually is no run_dir in this case. Exporting the rsopt file is neither a simulation nor a report and so downloadDataFile fails (no ```job_supervisor._run()``` -> no ```job_supervisor.__db_write()``` etc.). It is analogous to exportArchive which I note is on the sever, not the agent.  Taking a further look...

robnagler:
Look at the SRW import. It's similar in that it is a quasi-report (reports are terrible names for simulation runs, we should say "sim_run" or something similar).

robnagler:
@mkeilman please complete 2 or 3 sentences on what gets done for the silas report.

mkeilman:
Calling ```lib_file_basenames``` takes care of the server-side issue

mkeilman:
Converted to draft until blicon gets going again

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+    f = re.sub(r'[^\w.]+', '-', pkio.py_path(filename).purebasename).strip('-')
+    fz = pkio.py_path(filename)
+    v.runDir = f'{f}_scan'
+    v.fileBase = f
+    tf = {k: PKDict(filename=f'{f}.{k}') for k in ['py', 'sh', 'yml']}
+    for t in tf:
+        v[f'{t}FileName'] = tf[t].filename
+    v.outFileName = f'{f}.out'
+    v.readmeFileName = 'README.txt'
+    lf = _SIM_DATA.lib_file_basenames(data)
```
robnagler, line 1270: temporary unnecessary

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
     return ''
 
 
+def _export_rsopt_config(data):
+    v = _rsopt_jinja_context(data.models.exportRsOpt)
+    filename = 'rsOptExport.zip'
```
robnagler, line 1260: Not sure if this temporary is necessary. Why not: p = pkio.py_path('rsOptExport.zip')?

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+        allowZip64=True,
+    ) as z:
+        for f in tf:
+            z.writestr(tf[f].filename, tf[f].content)
+        z.writestr(
+            v.readmeFileName,
+            template_common.render_jinja(SIM_TYPE, v, v.readmeFileName)
+        )
+        for f in lf:
+            z.write(pkio.py_path(f), f)
```
robnagler, line 1296: I don't think pkio.py_path is necessary here.

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+    v.fileBase = f
+    tf = {k: PKDict(filename=f'{f}.{k}') for k in ['py', 'sh', 'yml']}
+    for t in tf:
+        v[f'{t}FileName'] = tf[t].filename
+    v.outFileName = f'{f}.out'
+    v.readmeFileName = 'README.txt'
+    lf = _SIM_DATA.lib_file_basenames(data)
+    v.libFiles = lf
+    v.hasLibFiles = len(v.libFiles) > 0
+    v.randomSeed = data.models.exportRsOpt.randomSeed if \
```
robnagler, line 1273: This seems like this should be somewhere else. Why is randomSeed the only data.model value needed?

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
 
+def _export_rsopt_config(data):
+    v = _rsopt_jinja_context(data.models.exportRsOpt)
+    filename = 'rsOptExport.zip'
+    f = re.sub(r'[^\w.]+', '-', pkio.py_path(filename).purebasename).strip('-')
+    fz = pkio.py_path(filename)
+    v.runDir = f'{f}_scan'
+    v.fileBase = f
+    tf = {k: PKDict(filename=f'{f}.{k}') for k in ['py', 'sh', 'yml']}
+    for t in tf:
```
robnagler, line 1266: items would work better

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
 
 
+def _export_rsopt_config(data):
+    v = _rsopt_jinja_context(data.models.exportRsOpt)
+    filename = 'rsOptExport.zip'
+    f = re.sub(r'[^\w.]+', '-', pkio.py_path(filename).purebasename).strip('-')
+    fz = pkio.py_path(filename)
+    v.runDir = f'{f}_scan'
+    v.fileBase = f
+    tf = {k: PKDict(filename=f'{f}.{k}') for k in ['py', 'sh', 'yml']}
```
robnagler, line 1265: I don't know what "tf" means. I think there are too many temporaries in this function. Better to modularize.
This for loop and the next could be combined.

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+    lf = _SIM_DATA.lib_file_basenames(data)
+    v.libFiles = lf
+    v.hasLibFiles = len(v.libFiles) > 0
+    v.randomSeed = data.models.exportRsOpt.randomSeed if \
+        data.models.exportRsOpt.randomSeed is not None else ''
+
+    # do this in a second loop so v is fully updated
+    # note that the rsopt context is regenerated in python_source_for_model()
+    for t in tf:
+        tf[t].content = python_source_for_model(data, 'rsoptExport', plot_reports=False) \
```
robnagler, line 1279: The temporary here seems unnecessary. Why not make a function that is called that gets the content below.

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+        z.writestr(
+            v.readmeFileName,
+            template_common.render_jinja(SIM_TYPE, v, v.readmeFileName)
+        )
+        for f in lf:
+            z.write(pkio.py_path(f), f)
+
+    return PKDict(
+        content_type='application/zip',
+        filename=fz
```
robnagler, line 1300: trailing comma

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
...
+    ) as z:
+        for f in tf:
+            z.writestr(tf[f].filename, tf[f].content)
+        z.writestr(
+            v.readmeFileName,
+            template_common.render_jinja(SIM_TYPE, v, v.readmeFileName)
+        )
+        for f in lf:
+            z.write(pkio.py_path(f), f)
+
```
robnagler, line 1297: blank unnecessary. Probably an indication to modularize more.

sirepo/template/srw.py
```diff
@@ -1286,6 +1255,52 @@ def _enum_text(name, model, field):
     return ''
 
 
+def _export_rsopt_config(data):
+    v = _rsopt_jinja_context(data.models.exportRsOpt)
+    filename = 'rsOptExport.zip'
+    f = re.sub(r'[^\w.]+', '-', pkio.py_path(filename).purebasename).strip('-')
```
robnagler, line 1261: I think although don't know, that `.` should be `\.`, since dot means "anything" so the regex does nothing afaict. What is f supposed to be?

sirepo/template/srw.py
```diff
@@ -946,9 +909,15 @@ def write_parameters(data, run_dir, is_parallel):
...
     """
+    if data.report == 'exportRsOpt':
+        p = ''
+        _export_rsopt_config(data)
+    else:
+        p = _trim(_generate_parameters_file(data, run_dir=run_dir))
     pkio.write_text(
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
-        _trim(_generate_parameters_file(data, run_dir=run_dir))
+        #_trim(_generate_parameters_file(data, run_dir=run_dir))
```
robnagler, line 919: remove?

robnagler (changes requested):

sirepo/template/srw.py
mkeilman: [2d2f299](https://github.com/radiasoft/sirepo/pull/4057/commits/2d2f29970658b5957369b785f089edbf86b88e86)

sirepo/template/srw.py
mkeilman: moot due to [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: They cannot be combined - the shell script jinja needs the other filenames to be in the context. I could reorder the list but that introduces fragility. Anyway [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

sirepo/template/srw.py
mkeilman: [075b377](https://github.com/radiasoft/sirepo/pull/4057/commits/075b377ec4e22e26dac2d429ee1d02791ddfeaf1)

mkeilman:
Ping - some refactoring for clarity I hope

sirepo/template/srw.py
```diff
@@ -1286,6 +1254,13 @@ def _enum_text(name, model, field):
     return ''
 
 
+def _export_rsopt_config(data):
+    r = 'exportRsOpt'
+    v = _rsopt_jinja_context(r, data.models[r])
+    v.libFiles = _SIM_DATA.lib_file_basenames(data)
```
robnagler, line 1260: Simplify:

```py
return _write_rsopt_zip(data, _rsopt_jinja_context(data))
```

`_rsopt_jinja_context` can know the constant `exportRsOpt`, which actually could be a constant in the module and used in _OUTPUT_FOR_MODEL and write_parameters.

sirepo/template/srw.py
```diff
@@ -1628,7 +1603,7 @@ def _generate_parameters_file(data, plot_reports=False, run_dir=None):
     dm = data.models
     # do this before validation or arrays get turned into strings
     if is_for_rsopt:
-        rsopt_ctx = _rsopt_jinja_context(dm.exportRsOpt)
+        rsopt_ctx = _rsopt_jinja_context(data.report, dm.exportRsOpt)
```
robnagler, line 1606: see above. _SIM_DATA.lib_file_basenames(data) needed here too? If not, pass in a conditional to not add it to the context.

sirepo/template/srw.py
```diff
@@ -2270,6 +2250,38 @@ def _wavefront_pickle_filename(el_id):
     return 'initial.pkl'
 
 
+def _write_rsopt_zip(data, ctx):
+    filename = f'{ctx.fileBase}.zip'
+    files = []
```
robnagler, line 2255: make this a function, e.g. `_files()` to eliminate the temporary by the same name.

sirepo/template/srw.py
```diff
@@ -2270,6 +2250,38 @@ def _wavefront_pickle_filename(el_id):
...
+        ctx[f'{t}FileName'] = f
+        files.append(f)
+    with zipfile.ZipFile(
+        filename,
+        mode='w',
+        compression=zipfile.ZIP_DEFLATED,
+        allowZip64=True,
+    ) as z:
+        # the shell script depends on the other filenames being defined
+        for f in files:
```
robnagler, line 2267: call _files()

sirepo/template/srw.py
```diff
@@ -2270,6 +2250,38 @@ def _wavefront_pickle_filename(el_id):
     return 'initial.pkl'
 
 
+def _write_rsopt_zip(data, ctx):
+    filename = f'{ctx.fileBase}.zip'
```
robnagler, line 2254: Add this to  _rsopt_jinja_context() so not needed as a constant here.

sirepo/sim_data/srw.py
```diff
@@ -76,6 +76,8 @@ def _compute_model(cls, analysis_model, *args, **kwargs):
             return 'multiElectronAnimation'
         if 'beamlineAnimation' in analysis_model:
             return 'beamlineAnimation'
+        if 'exportRsOpt' in analysis_model:
```
robnagler, line 79: Why is this an `in` and not `==`?

sirepo/template/srw.py
```diff
@@ -128,6 +127,7 @@
         labels=['Horizontal Position', 'Vertical Position', 'Intensity'],
         units=['m', 'm', '{intensity_units}'],
     ),
+    exportRsOpt=PKDict(filename='exportRsOpt.zip')
```
robnagler, line 130: Add a trailing comma

sirepo/template/srw.py
```diff
@@ -946,9 +909,14 @@ def write_parameters(data, run_dir, is_parallel):
...
     """
+    if data.report == 'exportRsOpt':
+        p = ''
+        _export_rsopt_config(data)
+    else:
+        p = _trim(_generate_parameters_file(data, run_dir=run_dir))
     pkio.write_text(
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
-        _trim(_generate_parameters_file(data, run_dir=run_dir))
+        p
```
robnagler, line 919: add trailing comma

robnagler (changes requested):

sirepo/sim_data/srw.py
mkeilman: copy-paste

mkeilman: [815be41](https://github.com/radiasoft/sirepo/pull/4057/commits/815be4174fee8f9ee268720b7145f65971567670)

sirepo/template/srw.py
mkeilman: [769a46c](https://github.com/radiasoft/sirepo/pull/4057/commits/769a46c017dbadf2eabb2c06d7a7b9fe545915c4)

sirepo/template/srw.py
mkeilman: [210b274](https://github.com/radiasoft/sirepo/pull/4057/commits/210b27494aeea9efe147b4e3ea574847edabba0d)

sirepo/template/srw.py
mkeilman: [210b274](https://github.com/radiasoft/sirepo/pull/4057/commits/210b27494aeea9efe147b4e3ea574847edabba0d)

sirepo/template/srw.py
mkeilman: [210b274](https://github.com/radiasoft/sirepo/pull/4057/commits/210b27494aeea9efe147b4e3ea574847edabba0d)

sirepo/template/srw.py
mkeilman: Add a comma where? `pkio.write_text()` takes 2 arguments, or am I overlooking something?

mkeilman:
Ping - I had one question about a trailing comma

sirepo/template/srw.py
robnagler: Our style is this:

```py
 pkio.write_text(
        _trim(_generate_parameters_file(data, run_dir=run_dir))
        p,
)
```

Trailing commas on all lines even when they aren't necessary means that you can reorder or add/delete lines without knowing or disturbing the context of the other lines.

mkeilman: I understand it for lists and the like, but for function args it seems misleading. Anyway, [f7b95c4](https://github.com/radiasoft/sirepo/pull/4057/commits/f7b95c4ecd35be76f5c3171076435f21698cb7a9)

sirepo/template/srw.py
```diff
@@ -40,6 +39,8 @@
 
 _CANVAS_MAX_SIZE = 65535
 
+_RSOPT_EXPORT_BASE_NAME = 'exportRsOpt'
```
robnagler, line 42: This is used in sim_data, and it seems like it should be defined there so it can be reused consistently throughout the code. I think EXPORT_RSOPT is probably a sufficient name for this purpose.

sirepo/template/srw.py
```diff
@@ -128,6 +129,7 @@
         labels=['Horizontal Position', 'Vertical Position', 'Intensity'],
         units=['m', 'm', '{intensity_units}'],
     ),
+    exportRsOpt=PKDict(filename=f'{_RSOPT_EXPORT_BASE_NAME}.zip',)
```
robnagler, line 132: This label is also _RSOPT_EXPORT_BASE_NAME

sirepo/template/srw.py
```diff
@@ -1967,18 +1941,25 @@ def _rotate_report(report, ar2d, x_range, y_range, info):
...
+        libFiles=_SIM_DATA.lib_file_basenames(data),
         numCores=int(model.numCores),
         numWorkers=max(1, multiprocessing.cpu_count() - 1),
         numSamples=int(model.numSamples),
+        outFileName=f'{_RSOPT_EXPORT_BASE_NAME}.out',
+        randomSeed=model.randomSeed if model.randomSeed is not None else '',
+        readmeFileName='README.txt',
         rsOptElements=e,
         rsOptParams=_RSOPT_PARAMS,
         rsOptParamsNoRot=_RSOPT_PARAMS_NO_ROTATION,
```
robnagler, line 1960: If you use an abbreviation, use is conistently: `_RSOPT_PARAMS_NO_ROT`

sirepo/template/srw.py
```diff
@@ -1967,18 +1941,25 @@ def _rotate_report(report, ar2d, x_range, y_range, info):
...
         rsOptOutFileName='scan_results',
+        runDir=f'{_RSOPT_EXPORT_BASE_NAME}_scan',
```
robnagler, line 1962: Why isn't this just `scan`? Why does _RSOPT_EXPORT_BASE_NAME need to be part of the name? Isn't it in the rsopt directory?

sirepo/template/srw.py
```diff
@@ -2270,6 +2251,42 @@ def _wavefront_pickle_filename(el_id):
...
+    filename = f'{_RSOPT_EXPORT_BASE_NAME}.zip'
+    with zipfile.ZipFile(
+        filename,
+        mode='w',
+        compression=zipfile.ZIP_DEFLATED,
+        allowZip64=True,
+    ) as z:
+        # the shell script depends on the other filenames being defined
+        for f in _files():
+            z.writestr(
```
robnagler, line 2273: Encapsulate in a function:

```py
def _write(zip, path):
    zip.writestr(
        path,
        python_source_for_model(data, ctx.fileBase, plot_reports=False) if path.endswith('.py')
            else template_common.render_jinja(SIM_TYPE, ctx, path)
     )
```

The loop will be simpler and easier to understand.
         
        d =  python_source_for_model(data, ctx.fileBase, plot_reports=False)

sirepo/template/srw.py
mkeilman: [8ddc012](https://github.com/radiasoft/sirepo/pull/4057/commits/8ddc012d751d5458e94639decc0f52140fba1dec)

sirepo/package_data/static/js/srw.js
```diff
@@ -1897,9 +1899,7 @@ SIREPO.app.directive('rsOptElements', function(appState, panelState, requestSend
                         numParams += e[srwService.rsOptElementOffsetField(p)]
                             .split(',')
                             .map(parseFloat)
-                            .filter((x) => {
-                                return x != 0;
-                            })
+                            .filter(x => x !== 0)
```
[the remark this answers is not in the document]
mkeilman: [6df8441](https://github.com/radiasoft/sirepo/pull/4057/commits/6df84415a2a3a4cbd06ea136900a4202288ebb70)

sirepo/template/srw.py
mkeilman: To finish my thought on this: to me `fn(a, b,)` looks like the function could take more parameters even if it can't. It blurs the meta-information. It is also visually ragged. I accept that this is a "me" problem.

PR #4074, opened 2022-01-24
Fix #4061: sirepo command not found at NERSC

### LINKED ISSUE #4061: nersc: Simulation Error: no reply Elapsed time: 00:00:15
Starting Young's on NERSC but just get "no reply". Discarding changes to example didn't fix it.

no-reply

e-carlin:
At NERSC we start the job_agent by supplying the sirepo command
through env (sirepo.job_driver.sbatch.sirepo_cmd). This means that
sirepo commands called from the job agent do not have the proper env.
This was causing the sim_type pkcli lookup to fail.

Modify so the assert for no `run_background` happens at server init.

This has the side effect that all sim_type pkcli's that support sbatch
must be able to be imported without being fully initialized in the
server. This is because we must assert for no run background in the
sim type initialization code (srschema.validate).

I also updated the NERSC development wiki with details on how to setup one's NERSC env to support development https://github.com/radiasoft/sirepo/wiki/Development/_compare/4b3dee65ad2f732a64ce77bbecbbda3de87fba00...29c2bcf0d1d775f618188c1fbc7590575ac15255

### DIFF
--- a/sirepo/srschema.py
+++ b/sirepo/srschema.py
@@ -216,8 +216,8 @@ def _validate_job_run_mode(field_name, schema):
     if hasattr(m, 'run_background'):
         raise AssertionError(
             f'simulation_type={t} cannot have'
-            ' pkcli.run_background because it supports slurm. Slurm only'
-            ' supports running a code through `python parameters.py` ',
+            + ' pkcli.run_background because it supports slurm. Slurm only'
+            + ' supports running a code through `python parameters.py` ',
         )
 
 
--- a/sirepo/srschema.py
+++ b/sirepo/srschema.py
@@ -148,6 +148,7 @@ def validate(schema):
         _validate_model_name(model_name)
         sch_model = sch_models[model_name]
         for field_name in sch_model:
+            _validate_job_run_mode(field_name, schema)
             sch_field_info = sch_model[field_name]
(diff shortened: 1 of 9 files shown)


sirepo/srschema.py
```diff
@@ -206,6 +207,20 @@ def _validate_enum(val, sch_field_info, sch_enums):
...
+def _validate_job_run_mode(field_name, schema):
+    if field_name != 'jobRunMode':
+        return
+    from sirepo import pkcli
+    t = schema.simulationType
+    m = pkcli.import_module(t)
+    if hasattr(m, 'run_background'):
+        raise AssertionError(
+            f'simulation_type={t} cannot have'
+            ' pkcli.run_background because it supports slurm. Slurm only'
```
robnagler, line 219: Pet peeve: the idea of concatenating strings without an operator in between is problematic. I know this is valid Python, but it seems to me, it's also strange. Rather, just add a leading plus to the continued strings.

robnagler (approved):
I was concerned about the dependencies problem, but it seems that isn't a real problem. We already protect dependencies by putting them in function calls.

sirepo/srschema.py
e-carlin: f13e9db85

PR #4085, opened 2022-01-27
Fix #4080 - fixed extrusion bug

### LINKED ISSUE #4080: radia: h-bend dipole L shape does not resize properly
For example, if you double the size of the magnet, the pole updates properly but the core does not

mkeilman:
Notes:

- moved the extruded shape handling from ```_update_geom_objects``` to ```_update_geom_obj```
- accordingly, ```update_ell``` and its cousins now just invoke ```_update_geom_obj```. They support undulator code currently under revision, and will be removed
- the extrusion length was wrong by a factor of 2 and the object mis-centered


mkeilman:
Ping for a look so @cchall can exercise the dipole code

moellep:
Looks good to me.

PR #4087, opened 2022-01-28
Sirepo 4079 run tests all codes

mkeilman:
This adds tests for codes where it makes sense to do so. They are the simplest tests possible, requiring code completion (in most cases) but nothing more.


robnagler (approved):
Excellent, thanks!

PR #4097, opened 2022-02-01
4014 controls screen

gurhar1133:
passed tests and works with or without instruments.

parameters.py.jinja in template/controls line 140 might not be best way of doing things, maybe you have ideas @robnagler ?

### DIFF
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -63,7 +63,18 @@
         ]
     },
     "frameIdFields": {
-        "instrumentAnimation": []
+        "instrumentAnimation": [
+            "x",
+            "y1",
+            "y2",
+            "y3",
+            "histogramBins",
+            "plotRangeType",
+            "horizontalSize",
+            "horizontalOffset",
+            "verticalSize",
+            "verticalOffset"
+        ]
     },
     "localRoutes": {
         "controls": {
@@ -96,6 +107,12 @@
         "optimizerInputsTab": {
             "title": "Optimization",
             "advanced": []
+        },
+        "instrumentAnimation":{
+            "x": ["Horizontal Value to Plot", "ValueList", ""],
+            "y1": ["Vertical Value to Plot", "ValueList", ""],
+            "histogramBins": ["Histogram Bins", "Integer", 200],
+            "colorMap": ["Color Map", "ColorMap", "afmhot"]
         },
         "command_twiss": {
             "alfx": ["Horizontal Alpha", "Float", 0.0],
@@ -165,10 +182,22 @@
         "completionState": ""
     },
     "view": {
-	"instrumentAnimation": {
-            "title": "fancy title",
-            "advanced": []
-	},
+        "instrumentAnimation": {
+                "title": "Placeholder Title",
+                "model": "instrumentAnimation",
+                "advanced": [
+                    [
+                        ["Horizontal", [
+                            "x"
+                        ]],
+                        ["Vertical", [
+                            "y1"
+                        ]]
+                    ],
+                    "histogramBins",
+                    "colorMap"
+                ]
+        },
         "beamline": {
             "title": "Beamline",
             "advanced": []
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -65,7 +65,8 @@
             "horizontalSize",
             "horizontalOffset",
             "verticalSize",
-            "verticalOffset"
+            "verticalOffset",
+            "refreshId"
         ]
     },
     "localRoutes": {
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -104,7 +104,7 @@
         "instrumentAnimation":{
             "aspectRatio": ["Aspect Ratio", "AspectRatio", "1"],
             "x": ["Horizontal Value to Plot", "ValueList", "x"],
-            "y1": ["Vertical Value to Plot", "ValueList", "px"],
+            "y1": ["Vertical Value to Plot", "ValueList", "y"],
             "histogramBins": ["Histogram Bins", "Integer", 200],
             "colorMap": ["Color Map", "ColorMap", "afmhot"]
         },
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -90,6 +90,10 @@
             "sigt": ["Bunch Length [m]", "Float", 1.0, "The bunch length $c\\space \\sigma_t$ (SIGT)."],
             "sige": ["Relative Energy Spread", "Float", 0.001, "The relative energy spread $\\sigma_E / E$ (SIGE)."]
         },
+        "optimizerInputsTab": {
+            "title": "Optimization",
+            "advanced": []
+        },
         "command_twiss": {
             "alfx": ["Horizontal Alpha", "Float", 0.0],
             "alfy": ["Vertical Alpha", "Float", 0.0],
@@ -109,6 +113,7 @@
         },
         "initialMonitorPositionsReport": {},
         "optimizerSettings": {
+            "inputs": ["", "OptimizerInputIds"],
             "method": ["Optimization Method", "OptimizationMethod", "nmead"]
         },
         "optimizerTarget": {
@@ -160,6 +165,11 @@
             "title": "Beamline",
             "advanced": []
         },
+        "optimizerInputsTab": {
+            "title": "Optimization",
+            "advanced": []
+         
+        }, 
         "command_beam": {
             "title": "Beam Settings",
             "advanced": [],
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -101,8 +101,9 @@
             "advanced": []
         },
         "instrumentAnimation":{
-            "x": ["Horizontal Value to Plot", "ValueList", ""],
-            "y1": ["Vertical Value to Plot", "ValueList", ""],
+            "aspectRatio": ["Aspect Ratio", "AspectRatio", "1"],
+            "x": ["Horizontal Value to Plot", "ValueList", "x"],
+            "y1": ["Vertical Value to Plot", "ValueList", "px"],
             "histogramBins": ["Histogram Bins", "Integer", 200],
             "colorMap": ["Color Map", "ColorMap", "afmhot"]
         },
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -192,7 +192,7 @@
     },
     "view": {
         "instrumentAnimation": {
-                "title": "Placeholder Title",
+                "title": "Cross Section",
                 "model": "instrumentAnimation",
                 "advanced": [
                     [
--- a/sirepo/package_data/static/json/controls-schema.json
+++ b/sirepo/package_data/static/json/controls-schema.json
@@ -62,6 +62,9 @@
             ["other", "Other"]
         ]
     },
+    "frameIdFields": {
+        "instrumentAnimation": []
+    },
     "localRoutes": {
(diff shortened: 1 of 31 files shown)


sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -127,7 +132,17 @@ def run_simulation(correctors):
...
             lattice = re.sub('{' + k + '}', str(values[k]), lattice)
         f.write(lattice)
-    return os.system('madx in.madx > madx.log 2> madx.err') == 0
+   
+    success = os.system('madx in.madx > madx.log 2> madx.err') == 0
+    if not success:
+        return False
+
+    cmd = 'grep "is lost" madx.log  | wc'
+    res = subprocess.check_output(cmd, shell=True)
```
moellep, line 141: Here is place where subprocess() is called. Mad-x doesn't fail if all particles are lost, but the data files will be missing and the simulation should be marked as an optimization failure. The only indicator we could find was to sum up the "is lost" lines in the madx.log file and compare against the original particle count. There is also an os.system() call above to invoke madx. @robnagler would there be a better way to do these call-outs?

robnagler: I'd recommend something like this, but definitely check the conditionals: 

```py
with open('{}') as f:
    c = 0
    for l in f.readline():
        if 'is lost' in l:
            c += 1
            if c > {{ particleCount }}:
                return False
    return True
```

I don't necessarily trust statements like "lost == particleCount". [Here's why.](https://github.com/radiasoft/pykern/wiki/DesignHints#ensure-loops-exit)

If you use wc, you can pass `-l`, if you want a count, but you could also pass `-c` to grep. However, as you see, you don't need a subprocess at all if it is simple parsing. This is done in shell scripts, but Python has reasonable enough parsing to make the code simpler and more readable to just use open/read.

I also am concerned about os.system. It's a bad example. It's always better to avoid shell parsing. There have been numerous security issues related to this. Rather, use something parsed:

```py
p = subprocess.run(('madx', 'in.madx'), text=True)
if p.returncode != 0:
    return False
```

Note that you avoid the open() and read() because you have p.stdout to decode and parse.

gurhar1133:
```python
p = subprocess.run(('madx', 'in.madx'), stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
pkio.write_text('madx.log', p.stdout)
pkio.write_text('madx.err', p.stderr)
if p.returncode != 0:
    return False

lost = pkio.read_text('madx.log').count('is lost')
return lost != {{ particleCount }}
```
How about ^
Seems to work well

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -132,15 +132,15 @@ def run_simulation(correctors):
...
+    pkio.write_text('madx.log', p.stdout)
+    pkio.write_text('madx.err', p.stderr)
+    if p.returncode != 0:
         return False
 
-    cmd = 'grep "is lost" madx.log  | wc'
-    res = subprocess.check_output(cmd, shell=True)
-    lost = int(res.decode().strip().split(' ')[0])
-    return not lost == {{ particleCount }}
+    lost = pkio.read_text('madx.log').count('is lost')
```
robnagler, line 142: Great. I didn't know about `count`! 

You have the madx.log in p.stdout so you don't need to read anything, just decode to a string with pkcompat.from_bytes.

robnagler (changes requested):

sirepo/package_data/template/controls/parameters.py.jinja
gurhar1133: good point 👍

gurhar1133:
```python
lost = pkcompat.locale_str(p.stdout).count('is lost')
return lost != {{ particleCount }}
```

like this ^ ?
didn't see pkcompat.from_bytes in docs, but pkcompat.locale_str seems to do the necessary conversion

robnagler: [from_bytes](https://github.com/radiasoft/pykern/blob/bf5946e20d77d13ceab387c33ac75c59b6579de6/pykern/pkcompat.py#L74) is there but it's hidden well. :(

from_bytes is the correct conversion, because that's what's happening (bytes to str). locale_str converts it to the locale which is not necessarily utf8. Not that this matters exactly, but I think it's better documentation than locale_str which implies there's a localization issue, which there isn't.

I'll create an issue to get rid of py2isms.

sirepo/template/controls.py
```diff
@@ -5,12 +5,15 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
+from dis import Instruction
```
e-carlin, line 8: not used

sirepo/template/controls.py
```diff
@@ -5,12 +5,15 @@
...
 from pykern import pkio
+from pykern import pkjson
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp
 from sirepo.sim_data.controls import AmpConverter
 from sirepo.template import template_common
-from sirepo.template.lattice import LatticeUtil
+from sirepo.template import madx_parser
+from sirepo.template.lattice import LatticeIterator, LatticeUtil
```
e-carlin, line 16: latticeiterator not used

sirepo/template/controls.py
```diff
@@ -22,28 +25,68 @@
...
 
 
 def background_percent_complete(report, run_dir, is_running):
     if is_running:
         return PKDict(
             percentComplete=0,
             frameCount=0,
-            elementValues=_read_summary_line(run_dir)
+            elementValues=_read_summary_line(run_dir),
+            ptcTrackColumns=get_ptc_track_columns(run_dir),
```
e-carlin, line 38: `get_ptc_track_columns` can be private

sirepo/template/controls.py
```diff
@@ -22,28 +25,68 @@
...
+    )
+
+
+def get_ptc_track_columns(run_dir):
+    if run_dir.join(_PTC_TRACK_COLUMNS_FILE).exists():
+        return pkio.read_text(_PTC_TRACK_COLUMNS_FILE).split(',')
+    return []
+
+
+def get_target_info(info_all, target):
```
e-carlin, line 57: private

sirepo/template/controls.py
```diff
@@ -22,28 +25,68 @@
...
+    for i in info_all:
+        if i.name == target:
+            return i, info_all.index(i)
```
e-carlin, line 60: can use `enumerate` instead of `.index(i)`

sirepo/template/controls.py
```diff
@@ -170,7 +213,11 @@ def _generate_parameters_file(data):
     if data.models.controlSettings.operationMode == 'DeviceServer':
         _validate_process_variables(v, data)
     v.optimizerTargets = data.models.optimizerSettings.targets
+    n = data.models.externalLattice.models.bunch.numberOfParticles
```
e-carlin, line 216: unnecessary temp var

sirepo/template/controls.py
```diff
@@ -191,20 +238,74 @@ def _set_opt(el, field, all_correctors):
         all_correctors.header.append(_format_header(el._id, _SIM_DATA.current_field(field)))
         v.ampTableNames.append(el.ampTable if 'ampTable' in el else None)
 
+    def _insert_ptc_commands(data, ptc_commands):
+        if ptc_commands:
```
e-carlin: test no instruments

sirepo/template/controls.py
```diff
@@ -22,28 +25,68 @@
...
+            return i, info_all.index(i)
+
+    raise AssertionError(f'no target={target} in info_all={info_all}')
+
+
+def sim_frame(frame_args):
+    return _extract_report_elementAnimation(frame_args, frame_args.run_dir, _PTC_TRACK_FILE)
+
+
+# TODO(e-carlin): this was copied from madx, but has been modified; need to abstract and share
```
e-carlin, line 69: remove todo

sirepo/template/madx.py
```diff
@@ -193,6 +193,13 @@ def background_percent_complete(report, run_dir, is_running):
     return PKDict(
         percentComplete=100,
         frameCount=1,
+        # TODO(e-carlin): madx produces n number of output files. We return what they are here.
```
e-carlin, line 196: remove all comment

sirepo/template/madx_parser.py
```diff
@@ -8,6 +8,7 @@
 from pykern import pkio
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdc, pkdlog, pkdp
+from sympy import EX
```
e-carlin, line 11: remove

sirepo/template/madx_parser.py
```diff
@@ -202,8 +203,12 @@ def parse_tfs_file(tfs_file, header_only=False, want_page=-1):
     for i in range(len(col_names)):
         name = col_names[i].lower()
         if name:
+
```
e-carlin, line 206: remove debugging code

sirepo/package_data/static/json/controls-schema.json
```diff
@@ -54,6 +54,21 @@
...
+            "y1",
+            "y2",
+            "y3",
+            "histogramBins",
+            "plotRangeType",
+            "horizontalSize",
+            "horizontalOffset",
+            "verticalSize",
+            "verticalOffset",
+            "refreshId"
```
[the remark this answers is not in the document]
gurhar1133: It looks like refreshId is important for making the heatmap api calls. Without it the heatmaps don't update at all during the optimization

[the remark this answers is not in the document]
e-carlin: Ah thanks for looking into it! Thinking about it more it makes sense. It is a bit of randomness added to the url for the frame which causes a full request to the server because the cached request in the browser has a different url.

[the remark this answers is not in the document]
gurhar1133: should we use something other than
```javascript
appState.models[m].refreshId = Math.random();
```
?

like a hash or a random integer rather than the float generated by Math.random()?
Is there any potential issue with the api call parser seeing the '.' in the float that we use as a refreshId?

gurhar1133:
Other than the refreshId comment, all other review changes have been made

sirepo/package_data/static/json/controls-schema.json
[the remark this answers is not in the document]
moellep: The '.' won't break anything, but really it could be removed and set frameCount to a simulation run-count or similar to make the query unique. That would require additional changes so the playback buttons would remain hidden on these "single-frame" reports. For now Math.random() is good enough.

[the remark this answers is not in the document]
gurhar1133: 👍

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -146,6 +158,15 @@ def update_and_run_simulation(correctors):
     if is_success:
         with open('{{ summaryCSV }}', 'a') as f:
             f.write('{}\n'.format(','.join([str(x) for x in correctors.tolist() + readings])))
+            # POSIT: one table option is enabled.
```
robnagler, line 161: Since this is parameters.py.jinja, wouldn't it be better to just generate the code if one table option is enabled and if not, do not generate?

robnagler (commented):
This is very impressive. Quite a lot of things going on.

sirepo/package_data/template/controls/parameters.py.jinja
gurhar1133: [onetable is always true](https://github.com/radiasoft/sirepo/blob/33ac659d1515e9e3311b8d9589cbe9acd61beb00/sirepo/template/madx.py#L874 )

a lot of our code expects that. Eg. all the report generation code that reads from the onetable file. Posit seems sufficient. Do you agree?

PR #4109, opened 2022-02-03
Raydata 10 sortcolumns

rorour:


sirepo/package_data/static/js/raydata.js
```diff
@@ -575,14 +575,17 @@ SIREPO.app.directive('scanSelector', function() {
...
                         {{ column }}
+                        <input type="checkbox" data-ng-checked="selectAllColumns" data-ng-show="showSelectAllButton($index)" data-ng-click="toggleSelectAll()"/>
                         <button type="submit" class="btn btn-primary btn-xs"  id="del-column" data-ng-show="showDeleteButton($index)" data-ng-click="deleteCol(column)"><span class="glyphicon glyphicon-remove"></span></button>
                       </th>
                   </tr>
                   </thead>
-                  <tbody ng-repeat="s in scans">
-                    <tr>
+                  <tbody>
+                    <tr ng-repeat="s in scans|orderBy:orderByColumn:reverseSortScans">
```
e-carlin, line 588: Nice! One spacing nit `s in scans | orderBy:orderByColumn:reverseSortScans`

sirepo/package_data/static/js/raydata.js
```diff
@@ -716,6 +722,14 @@ SIREPO.app.directive('scanSelector', function() {
                 });
             };
 
+            $scope.showArrowDown = (column) => {
```
e-carlin, line 725: These two function contain almost the exact same logic (except for the `!`). So, I think using one function (ex `showArrowUp()` and then passing in a true/false based on whether or not to show the up arrow. If it is false then the down arrow will show.

sirepo/package_data/static/js/raydata.js
```diff
@@ -724,6 +738,36 @@ SIREPO.app.directive('scanSelector', function() {
...
+
+            $scope.sortCol = (column) => {
+                if (column === 'selected') {
+                    return;
+                }
+                $scope.orderByColumn = column;
+                $scope.reverseSortScans = ! $scope.reverseSortScans;
+            };
+
+            $scope.toggleSelectAll = () => {
```
e-carlin, line 757: The code between the if and else is repeated. I think it could be simplified to
```js
$scope.scans.forEach((s) => {
        $scope.toggleScanSelection(
            s, 
            ! $scope.selectAllColumns,
        );
})
$scope.selectAllColumns = ! $scope.selectAllColumns;
```

sirepo/template/raydata.py
```diff
@@ -4,6 +4,7 @@
 :copyright: Copyright (c) 2021 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+from glob import glob
```
e-carlin, line 7: We try to follow the google style guide it [says this](https://google.github.io/styleguide/pyguide.html#22-imports) about imports
`Use import statements for packages and modules only, not for individual classes or functions`. [glob](https://github.com/python/cpython/blob/ae36cd1e792db9d6db4c6847ec2a7d50a71f2b68/Lib/glob.py#L13) is a function.

So just `import glob` and then `glob.glob()` when you call it.

It seems pretty redundant in this case because the module and function have the same name. But for cases where they don't it helps reduce confusion. For example, we have `sirepo.tornado.Event` which is a subclass of `tornado.locks.Event`. If we did `from tornado.locks import Event` it would be confusing which "Event" we were talking about.

sirepo/template/raydata.py
```diff
@@ -48,7 +49,8 @@ def _filename_and_image(path):
 
     def _paths():
         d = _dir_for_scan_uuid(_parse_scan_uuid(data))
-        for f in pkio.sorted_glob(d.join('*.png'), key='mtime'):
+
+        for f in glob(str(d.join('/**/*.png')), recursive = True):
```
e-carlin, line 53: nit: no spaces around = in  `recursive = True`

e-carlin (commented):
Looks great! Just a few minor comments

sirepo/package_data/static/js/raydata.js
rorour: eb41875ad13fdf2a9ff2f5e261bcf5432c048ace

sirepo/package_data/static/js/raydata.js
rorour: eb41875ad13fdf2a9ff2f5e261bcf5432c048ace

sirepo/package_data/static/js/raydata.js
rorour: eb41875ad13fdf2a9ff2f5e261bcf5432c048ace

sirepo/template/raydata.py
rorour: eb41875ad13fdf2a9ff2f5e261bcf5432c048ace

e-carlin (approved):

e-carlin:
@rorour looks good! Going to merge.

I made one change https://github.com/radiasoft/sirepo/pull/4109/commits/da56f63243354455e55663e113a55625995afb21#diff-dbbe342fcfc2c11e78a337d90abdb3234fa5b0428699a82ec9d128ac8fd1842eR629 . I couldn't find a way to describe what was going on here when we talked but I think applying the glyphicon class dynamically allows removal of repeated code. Lmk what you think.

PR #4110, opened 2022-02-03
4094 fixed aperture plots

gurhar1133:
Particle plots and monitors now have fixed axes. The editor panel at the very top 'Beamlines' has a panel that lets you change the size of the fixed axes.


moellep:
Looks good.

PR #4125, opened 2022-02-09
Fix radiasoft/raydata#11: user select databroker catalog

e-carlin:
Also, moved to using appFieldEditors with custom type to handle the
scan selection table.


e-carlin:
@moellep I added instructions to the [raydata repo](https://github.com/radiasoft/raydata#minimal-dev-install) for how to setup dev. It is a few steps but as simple as I could make it for getting a real working example.

e-carlin:
@moellep raydata ph1 is done. I'd like to merge this this week if you have time.

e-carlin:
I realized there are more dependencies that are needed and getting it to run is going to be complicated. I'm updating the raydata installer so if we win the phase II it will be easier for everyone to work on this code. Merging as is for now.

PR #4126, opened 2022-02-09
fix #4086 add logging in _send_simulation_compute

### LINKED ISSUE #4086: Supervisor log method and report
In #4068 it seem `_Op(analysis, 0FhG)` was sent to the agent. The agent received the op. And then we don't have any more information.

When the supervisor receives a stateless/stateful compute it should log the method name to give some more context for us to troubleshoot what method possibly takes time to run.

rorour:


robnagler (approved):
Looks good to me.

e-carlin (approved):

PR #4131, opened 2022-02-09
4124 optimizer error message

gurhar1133:
Optimizer error message shows if all run_sim calls fail and update_and_run fixed to attempt to get readings when is_success.

### DIFF
--- a/sirepo/package_data/template/controls/parameters.py.jinja
+++ b/sirepo/package_data/template/controls/parameters.py.jinja
@@ -7,7 +7,6 @@ from sirepo.sim_data.controls import AmpConverter
 from sirepo.template import madx_parser
 import os
 import numpy as np
-import random
 import re
 import requests
 import scipy.optimize
--- a/sirepo/package_data/template/controls/parameters.py.jinja
+++ b/sirepo/package_data/template/controls/parameters.py.jinja
@@ -53,7 +53,7 @@ targets = np.array([
     {% endfor %}
 ])
 
-finished_sims = []
+finished_sims = False
 
 
 {% if optimizerSettings_method == 'nmead' %}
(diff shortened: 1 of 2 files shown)


sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -154,26 +156,28 @@ def read_cols_and_save_report(default_file, save_file, cols_file):
...
-        keyword = columns.keyword[idx].replace('"', '')
-        if keyword == 'MONITOR':
-            readings += [float(columns.x[idx]), float(columns.y[idx])]
-        elif keyword == 'HMONITOR':
-            readings += [float(columns.x[idx])]
-        elif keyword == 'VMONITOR':
-            readings += [float(columns.y[idx])]
-    res = np.array(readings)
-    if is_success:
+    global finished_sims
```
robnagler, line 159: Unnecessary. You only need `global` for variables you are assigning. Modifications to globals don't need `global`.

That being said, why isn't finished_sims just a flag? It seems like the only time it is interesting is in the append(1) case. If it ever hits, then set it to True, and that's success. In this case, `global` is required.

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -154,26 +156,28 @@ def read_cols_and_save_report(default_file, save_file, cols_file):
...
+    if run_simulation(correctors):
+        readings = []
+        finished_sims.append(1)
+        columns = madx_parser.parse_tfs_file('twiss.file.tfs')
+        for idx in range(len(columns.name)):
```
robnagler, line 164: I think this should be columns.keyword, since that's the first thing referenced. Indeed, you might just say:

```py
for i, k in enumerate(columns.keyword):
```

I don't see a lot of value in variable names like `idx` vs `i`. I don't like the culture of removing vowels to short names, in general.

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -154,26 +156,28 @@ def read_cols_and_save_report(default_file, save_file, cols_file):
...
+            keyword = columns.keyword[idx].replace('"', '')
+            if keyword == 'MONITOR':
+                readings += [float(columns.x[idx]), float(columns.y[idx])]
+            elif keyword == 'HMONITOR':
+                readings += [float(columns.x[idx])]
+            elif keyword == 'VMONITOR':
+                readings += [float(columns.y[idx])]
```
robnagler, line 171: I would add `else` with a `pass` if that's the intent. Otherwise, `else` should be a `raise`.

robnagler (changes requested):

sirepo/package_data/template/controls/parameters.py.jinja
gurhar1133: good point. Switched it to flag with global

robnagler (approved):
Looks good, thanks.

PR #4132, opened 2022-02-10
Issue/4129 Speed up animation and report tests

moellep:
This makes the animation test about a minute faster, but it still takes 2 minutes to run - about 8 seconds per server call. All sims are now expected to complete successfully. Almost all time is spent in non simulation runtime. I also removed some report tests which seemed redundant.


robnagler:
I'll let @mkeilman review.

mkeilman (commented):
Looks fine, only one clarification around dict/PKDict

PR #4136, opened 2022-02-11
4116 one panel cross sections

gurhar1133:
Instruments are now in one panel with one editor that adjust axes and colormap.

### DIFF
--- a/sirepo/package_data/static/js/controls.js
+++ b/sirepo/package_data/static/js/controls.js
@@ -438,6 +438,7 @@ SIREPO.app.controller('ControlsController', function(appState, controlsService, 
         if (!self.instrumentAnimations){
             return;
         }
+        const m = [];
         self.instrumentAnimations.forEach((e, i) => {
             if (e.modelKey == 'instrumentAnimationAll' || e.modelKey == 'instrumentAnimationTwiss'){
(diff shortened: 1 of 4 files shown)


sirepo/package_data/static/js/controls.js
```diff
@@ -419,14 +427,29 @@ SIREPO.app.controller('ControlsController', function(appState, controlsService,
...
+        self.instrumentAnimations.forEach((e, i) => {
+            if (e.modelKey == 'instrumentAnimationAll' || e.modelKey == 'instrumentAnimationTwiss'){
+                return;
+            }
+            for (const key in appState.models[e.modelKey]) {
+                if (key != 'id'){
+                    appState.models[e.modelKey][key] = appState.models.instrumentAnimationAll[key];
+                }
+            }
+            appState.saveChanges(e.modelKey);
```
moellep, line 450: Here I would collect up all the model names that changed in an array m and then call saveChanges(m). Each call to saveChanges() does a lot of work (and maybe a server call), so in the forEach loop, add m.push(e.modelKey) and after the loop add if (m.length) { appState.saveChanges(m) }

gurhar1133:
made the change. anything else?

PR #4143, opened 2022-02-14
By being in the package.json the binaries are found

e-carlin:
So the full path to the binary nor npx are needed.

Not exactly sure what was wrong with npx but it seemed to not work
when ~/src was shared over NFS between the host and vm

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "karma-jasmine": "latest"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


e-carlin:
@rour this should fix your issue of `framework jasmine not found`. Test by running `bash test.sh` and the fronted tests should pass. If it works on your machine approve the PR and I'll merge.

robnagler (approved):

PR #4144, opened 2022-02-14
3762 markers as monitors

gurhar1133:
markers are converted to monitors. Four cases handled; markers & monitors, neither markers nor monitors and cases with one but not the other.

Also fixed bug where cross section panel renders even if no instruments

### DIFF
--- a/sirepo/template/controls.py
+++ b/sirepo/template/controls.py
@@ -5,7 +5,6 @@ u"""Controls execution template.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
-from aenum import enum
 from pykern import pkio
 from pykern import pkjson
 from pykern.pkcollections import PKDict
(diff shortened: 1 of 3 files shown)


sirepo/template/controls.py
```diff
@@ -5,6 +5,7 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
+from aenum import enum
```
moellep, line 8: Looks like an unused import

sirepo/template/controls.py
```diff
@@ -165,17 +166,29 @@ def write_parameters(data, run_dir, is_parallel):
...
+            _id=LatticeUtil.max_id(data) + 1,
+            name='M1',
+            type='MONITOR',
+        )
+        data.models.elements.append(m)
+        assert len(data.models.beamlines) == 1, \
+            f'expecting 1 beamline={data.models.beamlines}'
+        data.models.beamlines[0]['items'].append(m._id)
+        return
+    for i, e in enumerate(data.models.elements):
```
moellep, line 184: I think you could do this part as the first step in the method and then you don't need to count markers, only monitors.

sirepo/template/controls.py
```diff
@@ -165,17 +166,29 @@ def write_parameters(data, run_dir, is_parallel):
...
+        if e.type == 'MARKER':
+            m = PKDict(
```
moellep, line 186: You can remove the temporary variable m and assign directly to data.models.elements[i]

gurhar1133:
Made changes

PR #4145, opened 2022-02-14
Fix #2431: reload supervisor and flask on file changes

### LINKED ISSUE #2431: reload supervisor, job_agent and job_cmd
If files change, job_cmd (at least) should get reloaded dynamically. It would be nice if supervisor and job_agent reloaded as well.

e-carlin:
**This PR depends on radiasoft/pykern#124. Merge that before this**

Tornado's debug mode causes the server to reload whenever an imported
module changes. But, modules that aren't imported in the supervisor
(ex templates) should cause the system to reload so we pick up the
changes. When the supervisor reloads agent's are dropped and killed on
restart which forces a reload for them and any downstream processes
(ex job_cmd's).

Flask's reloader works like Tornado (reloads when imported python
module changes). But, for extra files it is only relevant to reload
flask when json files changes (ex a code's schema.json).

@mkeilman I'd like it if you gave this a try. It adds watching a lot of files. If NFS sharing between your host and vm is the culprit behind the slowdown starting the server this is probably going to make the problem worse.

### DIFF
--- a/sirepo/pkcli/job_supervisor.py
+++ b/sirepo/pkcli/job_supervisor.py
@@ -4,7 +4,6 @@
 :copyright: Copyright (c) 2019 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
-from __future__ import absolute_import, division, print_function
 from pykern import pkconfig
 from pykern import pkio
 from pykern import pkjson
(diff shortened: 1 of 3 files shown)


mkeilman:
The delay between the reload and processing a request seems unchanged...

sirepo/pkcli/job_supervisor.py
```diff
@@ -59,6 +62,10 @@ def default_command():
         websocket_ping_interval=sirepo.job.cfg.ping_interval_secs,
         websocket_ping_timeout=sirepo.job.cfg.ping_timeout_secs,
     )
+    if cfg.debug:
+        for f in sirepo.util.files_to_watch_for_reload('jinja', 'json', 'py'):
+            tornado.autoreload.watch(f)
```
mkeilman, line 67: Watching for jinja changes may not be necessary, as they influence output but no logic. I don't think reloads are necessary.

mkeilman (approved):
Looks good to me

sirepo/pkcli/job_supervisor.py
e-carlin: 58c3a5ccd

robnagler (approved):

PR #4150, opened 2022-02-15
3763 twiss from particles

gurhar1133:
Twiss is now computed from particles and twiss report is only displayed if there is tracking/instruments to generate particles to compute twiss from

### DIFF
--- a/sirepo/package_data/template/controls/parameters.py.jinja
+++ b/sirepo/package_data/template/controls/parameters.py.jinja
@@ -122,11 +122,9 @@ def check_twiss_success():
         particle_beam.read_ptc_data('ptc_track.file.tfsone')[0],
         mc2=s['constants']['particleMassAndCharge']['proton'][0],
     )
-    pkdp('\n\n\n\n res: {}', res)
     for col in ('beta_x', 'beta_y', 'alpha_x', 'alpha_y'):
+        #TODO (gurhar1133): single letter vars here
         nans = list(filter(lambda x: x != x, res[col]))
-        pkdp('\n\n\n res[col] : {}', res[col])
-        pkdp('\n\n\n nans: {}', nans)
         if nans:
             return False
     return True
--- a/sirepo/package_data/template/controls/parameters.py.jinja
+++ b/sirepo/package_data/template/controls/parameters.py.jinja
@@ -142,8 +142,7 @@ def run_simulation(correctors):
     if p.returncode != 0:
         return False
     lost = p.stdout.count('is lost')
-    #TODO (gurhar1133): is particleCount != 1 sufficient to prevent twiss report failure?
-    return lost != {{ particleCount }} and {{ particleCount }} != 1
+    return lost != {{ particleCount }}
 
 
 def read_cols_and_save_report(default_file, save_file, cols_file):
--- a/sirepo/package_data/template/controls/parameters.py.jinja
+++ b/sirepo/package_data/template/controls/parameters.py.jinja
@@ -5,6 +5,7 @@ from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp, pkdlog
 from sirepo.sim_data.controls import AmpConverter
 from sirepo.template import madx_parser
+from sirepo.template import particle_beam
 import os
 import numpy as np
 import re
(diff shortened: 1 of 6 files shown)


gurhar1133:
Just realized I need to re run tests. I'll comment when they all pass

gurhar1133:
OK, all tests passed

sirepo/template/controls.py
```diff
@@ -328,8 +318,10 @@ def _add_ptc(instruments, data):
     v.initialCorrectors = '[{}]'.format(','.join([str(x) for x in c.corrector]))
     v.correctorCount = len(c.corrector)
     v.monitorCount = len(header)
+    v.schema = SCHEMA
     if i:
         _add_ptc(i, data.models.externalLattice)
+    v.isTrackingSim = True if LatticeUtil.find_first_command(data.models.externalLattice, 'ptc_create_universe') else False
```
e-carlin, line 324: `bool(LatticeUtil.find_first_command(data.models.externalLattice, 'ptc_create_universe'))`. `If True else False` is almost always a code smell.

sirepo/template/madx.py
```diff
@@ -721,13 +721,16 @@ def _extract_report_twissFromParticlesAnimation(data, run_dir, filename):
...
-    assert set(res.keys()) == set(_PTC_OBSERVE_TWISS_COLS), \
-        f'unknown ptc twiss columns={set(res.keys())} expected={_PTC_OBSERVE_TWISS_COLS}'
+    assert set(res.keys()) == set(PTC_OBSERVE_TWISS_COLS), \
+        f'unknown ptc twiss columns={set(res.keys())} expected={PTC_OBSERVE_TWISS_COLS}'
     return extract_parameter_report(
         data,
         results=PKDict(res),
     )
 
+def _extract_report_twissFromParticlesAnimation(data, run_dir, filename):
```
e-carlin, line 731: I think this can be removed.

e-carlin (commented):
Looks good. Didn't run the code but left a few comments

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -115,6 +116,20 @@ def _optimize_polyfit(variable_count):
...
+def check_twiss_success():
+    if not {{ isTrackingSim }}:
+        return True
+    s = {{ schema }}
+    res = particle_beam.analyze_ptc_beam(
+        particle_beam.read_ptc_data('ptc_track.file.tfsone')[0],
+        mc2=s['constants']['particleMassAndCharge']['proton'][0],
+    )
+    for c in ('beta_x', 'beta_y', 'alpha_x', 'alpha_y'):
+        if list(filter(lambda x: x != x, res[c])):
```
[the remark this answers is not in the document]
gurhar1133: The x != x check is to see if it is nan or not. But I think there is an np.isNan(x) type function or something. Would something like that be better?

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -115,6 +116,20 @@ def _optimize_polyfit(variable_count):
 {% endif %}
 
 {% if controlSettings_operationMode == 'madx' %}
+def check_twiss_success():
+    if not {{ isTrackingSim }}:
+        return True
+    s = {{ schema }}
```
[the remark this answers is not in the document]
gurhar1133: Good point

sirepo/template/madx.py
gurhar1133: the private method is needed for madx since madx visualization does code generation

sirepo/package_data/template/controls/parameters.py.jinja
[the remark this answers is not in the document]
gurhar1133: Switched to 
```python
np.isnan(x)
```
 Makes it clearer what's going on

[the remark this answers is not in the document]
e-carlin: Much better, thanks

sirepo/package_data/template/controls/parameters.py.jinja
```diff
@@ -115,6 +116,19 @@ def _optimize_polyfit(variable_count):
 {% endif %}
 
 {% if controlSettings_operationMode == 'madx' %}
+def check_twiss_success():
+    if not {{ isTrackingSim }}:
+        return True
```
moellep, line 121: Rather than have the generated code look like this:
```python
def check_twiss_success():
    if not False:
        return True
    ...
```
I'd change the generated code so it never calls check_twiss_success() unless isTrackingSim is true - move the python "if" to a jinja "if".

gurhar1133: changed it

gurhar1133:
Any other changes needed?

PR #4164, opened 2022-02-17
Fix #3583: import .madx or .seq files

### LINKED ISSUE #3583: madx: include .seq files in import list
Currently you need to rename mad-x sequence files (.seq) to a .madx extension in order to import the file. The import should allow .madx or .seq directly.

gurhar1133:


sirepo/template/madx.py
```diff
@@ -328,12 +328,12 @@ def get_data_file(run_dir, model, frame, options=None, **kwargs):
 
 def import_file(req, **kwargs):
     text = pkcompat.from_bytes(req.file_stream.read())
-    assert re.search(r'\.madx$', req.filename, re.IGNORECASE), \
+    assert re.search(r'\.madx$|\.seq$', req.filename, re.IGNORECASE), \
```
moellep, line 331: We've later learned that "assert" is ignored when python is run in optimized mode, so whenever you modify an assert() line, change it to a "raise AssertionError()"

gurhar1133:
Made changes

Issue #4157, opened 2022-02-16
asynchronous analysis jobs


robnagler:
An analysis job should be two parts: start and status, like sequential and parallel. We could perhaps hold the request for a short time (100ms) and then return, but we are locking up a thread unnecessarily. Some analysis jobs will take a long time to run. In that sense, they are just sequential jobs followed by a download data file.

We could build a shim for some codes so they don't have to change.

e-carlin:
https://github.com/radiasoft/sirepo/issues/2435 is related. We need to right the code for both in such a way that we don't wait on resources while holding the flask request open. 

I imagine error handling (raising of Exceptions) will need to be modified for both

Issue #4135, opened 2022-02-10
Let user know when pending due to resources they control


e-carlin:
When a user is running a simulation and then starts a second one the second simulation will be blocked until the first one completes. In this case we should let them know that they are waiting on resources that they control. 

In the case where they are waiting on resources out of their control (ex all compute nodes are busy with other users) then just keep the same "pending" message. 

See #2435

PR #4184, opened 2022-02-23
Fix #4156 better errors

### LINKED ISSUE #4156: Better error message when analysis times out
"Report not generated" is not very friendly.

gurhar1133:
Here is a first crack at some better error handling (mostly geared towards some of the srw stuff we have discussed)
* Timeouts are now being displayed to the user if response is 'canceled'. Other ways of handling this might be better.
* Error handling for messages too large has been implemented, but for some reason the response needs to have 
 ```python
     state=job.COMPLETED
 ```
 rather than 
 ```python
     state=job.ERROR
 ```
 I'm going to look into why this is the case, but if anyone knows why this is necessary I would like to know.
* RE: 'out of memory on the server while processing the datafile'; Paul mentioned this in an email to software team. We can discuss this more.

### DIFF
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -60,7 +60,7 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             if (frameCache.getCurrentFrame(scope.modelName) == scope.prevFrameIndex) {
                 return;
             }
-            panelState.setError(scope.modelName, '')
+            panelState.setError(scope.modelName, '');
             scope.prevFrameIndex = index;
             frameCache.getFrame(scope.modelName, index, scope.isPlaying, function(index, data) {
                 if (scope.element) {
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -64,13 +64,11 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             scope.prevFrameIndex = index;
             frameCache.getFrame(scope.modelName, index, scope.isPlaying, function(index, data) {
                 if (scope.element) {
-                    // srdbg('scope.modelName', scope.modelName);
                     if (data.state == 'canceled') {
                         data.error = 'Request canceled due to timeout';
                     }
                     if (data.error) {
                         panelState.setError(scope.modelName, data.error);
-                        srdbg('data.error:', data.error);
                         return;
                     }
                     panelState.setError(scope.modelName, null);
@@ -201,11 +199,9 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
                     return;
                 }
                 panelState.requestData(scope.modelName, function(data) {
-                    srdbg('model:', appState.models[scope.modelName])
                     if (! scope.element) {
                         return;
                     }
-                    srdbg('DATA:', data);
                     forceRunCount = forceRunCount || 0;
                     if (data.x_range) {
                         scope.clearData();
@@ -2782,8 +2778,6 @@ SIREPO.app.directive('plot3d', function(appState, focusPointService, layoutServi
                 prevDomain = null;
                 $scope.dataCleared = false;
                 aspectRatio = plotting.getAspectRatio($scope.modelName, json);
-                srdbg('json.z_matrix: ', json.z_matrix)
-                srdbg('appState.models: ', appState.models);
                 heatmap = plotting.safeHeatmap(appState.clone(json.z_matrix).reverse());
                 var newFullDomain = [
                     [json.x_range[0], json.x_range[1]],
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -63,8 +63,13 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             scope.prevFrameIndex = index;
             frameCache.getFrame(scope.modelName, index, scope.isPlaying, function(index, data) {
                 if (scope.element) {
+                    // srdbg('scope.modelName', scope.modelName);
+                    if (data.state == 'canceled') {
+                        data.error = 'Request canceled due to timeout';
+                    }
                     if (data.error) {
                         panelState.setError(scope.modelName, data.error);
+                        srdbg('data.error:', data.error);
                         return;
                     }
                     panelState.setError(scope.modelName, null);
@@ -2775,6 +2780,7 @@ SIREPO.app.directive('plot3d', function(appState, focusPointService, layoutServi
                 $scope.dataCleared = false;
                 aspectRatio = plotting.getAspectRatio($scope.modelName, json);
                 srdbg('json.z_matrix: ', json.z_matrix)
+                srdbg('appState.models: ', appState.models);
                 heatmap = plotting.safeHeatmap(appState.clone(json.z_matrix).reverse());
                 var newFullDomain = [
                     [json.x_range[0], json.x_range[1]],
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -204,6 +204,7 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
                     if (! scope.element) {
                         return;
                     }
+                    srdbg('DATA:', data);
                     forceRunCount = forceRunCount || 0;
                     if (data.x_range) {
                         scope.clearData();
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -201,6 +201,7 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
                     return;
                 }
                 panelState.requestData(scope.modelName, function(data) {
+                    srdbg('model:', appState.models[scope.modelName])
                     if (! scope.element) {
                         return;
                     }
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -2774,6 +2774,7 @@ SIREPO.app.directive('plot3d', function(appState, focusPointService, layoutServi
                 prevDomain = null;
                 $scope.dataCleared = false;
                 aspectRatio = plotting.getAspectRatio($scope.modelName, json);
+                srdbg('json.z_matrix: ', json.z_matrix)
                 heatmap = plotting.safeHeatmap(appState.clone(json.z_matrix).reverse());
                 var newFullDomain = [
                     [json.x_range[0], json.x_range[1]],
--- a/sirepo/package_data/static/js/sirepo-plotting.js
+++ b/sirepo/package_data/static/js/sirepo-plotting.js
@@ -60,6 +60,7 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             if (frameCache.getCurrentFrame(scope.modelName) == scope.prevFrameIndex) {
                 return;
             }
+            panelState.setError(scope.modelName, '')
             scope.prevFrameIndex = index;
             frameCache.getFrame(scope.modelName, index, scope.isPlaying, function(index, data) {
                 if (scope.element) {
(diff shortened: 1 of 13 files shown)


gurhar1133:
Regarding 
```job.COMPLETED``` vs ```job.ERROR```

In job_supervisor.py, _receive_api_runStatus() will reset the request status to missing if state=job.ERROR:
```python
if r.state == job.ERROR:
            return self._init_db_missing_response(req)
``` 

For handling errors to make them more user friendly, do we want to change it so that errors aren't sent to the UI with state='missing' or do we want to handle certain specific cases and set state=job.COMPLETED (the latter is what I am currently doing)

robnagler:
That's a good question. The job is completed, and one or more reports can't be displayed, and the raw files can be downloaded.

The [case you showed](https://github.com/radiasoft/sirepo/blob/1a728a589dba461d16908510a2dfe711740ed106/sirepo/job_supervisor.py#L743) is only for sequential results. We are about to change that code a bit in #4135. The problem is the GUI is coupling the runSimulation with the result. If that changed (possibly with #4135) then the job would tell the job_supervisor it was completed, and the runStatus call would be simplified and wouldn't have this completed/error problem on the report.

@e-carlin thoughts?

gurhar1133:
For the out of memory error I am working with:
https://github.com/radiasoft/sirepo/blob/f922e02c7fe98a7d05610e50c322b019360ec243/sirepo/pkcli/job_cmd.py#L104-L110

Other ideas?

e-carlin:
> The [case you showed](https://github.com/radiasoft/sirepo/blob/1a728a589dba461d16908510a2dfe711740ed106/sirepo/job_supervisor.py#L743) is only for sequential results. We are about to change that code a bit in https://github.com/radiasoft/sirepo/issues/4135. The problem is the GUI is coupling the runSimulation with the result. If that changed (possibly with https://github.com/radiasoft/sirepo/issues/4135) then the job would tell the job_supervisor it was completed, and the runStatus call would be simplified and wouldn't have this completed/error problem on the report.

As far as I can see that bit of code is unlikely to change in #4135. In #4135 the job won't be completed so we will [return a status reply](https://github.com/radiasoft/sirepo/blob/1a728a589dba461d16908510a2dfe711740ed106/sirepo/job_supervisor.py#L913) and then return [before the code in question](https://github.com/radiasoft/sirepo/blob/1a728a589dba461d16908510a2dfe711740ed106/sirepo/job_supervisor.py#L736).

But that code could change in #4157. In #4157 we will be disconnecting status from analysis. 

With all that said, there is [some history](https://github.com/radiasoft/sirepo/issues/2697) to why we set the job to missing when the sequential result is in error. I think we'll need more granular job statuses (or info in the status) if we want to be able to handle all cases.

robnagler:
Good info. Lets discuss.

e-carlin:
https://github.com/radiasoft/sirepo/pull/4184#issuecomment-1049256847

We are going to separate out runStatus from sequential_result. The gui will request status, if completed, request sequential results.

gurhar1133:
Signal is now being caught:
```python
if os.WIFSIGNALED(r) and abs(r) == signal.SIGKILL:
    return PKDict(state=job.ERROR, error='Terminated Process. Possibly ran out of memory')
```
Checking abs value of subprocess exit code tells us if signal 9
From python docs:
'A negative value -N indicates that the child was terminated by signal N'

Also, canceling sim while in polling loop does not trigger the kill signal and result in memory error message being sent to UI

sirepo/pkcli/job_cmd.py
```diff
@@ -104,8 +106,7 @@ def _do_compute(msg, template):
             r = p.poll()
             i = r is None
             if not i:
-                if os.WIFSIGNALED(r):
-                    pkdp('\n\n\n\n --------- \n RET STATUS INFO: {} \n info:{}\n ------- \n\n\n\n', r, os.WTERMSIG(r))
+                if os.WIFSIGNALED(r) and abs(r) == signal.SIGKILL:
```
robnagler, line 109: `subprocess` behaves differently than `os`. `os.WIFSIGNALED` has [already been called](https://github.com/python/cpython/blob/606e496dd6e2ace298532da200169124c26ae0f2/Lib/subprocess.py#L1851). Therefore, the test should be:

```py
if r == -signal.SIGKILL:
```

We know at this point `r` is `None`. By converting the constant, the test is exactly what we want: Was it KILLed?

The documentation in subprocess could be clearer. Something bothered me about the way it was written. It is technically correct, but a bit of help for the user (note: this is not an exit code).

sirepo/pkcli/job_cmd.py
```diff
@@ -217,6 +218,8 @@ def _change_message_too_big(msg):
     if type(msg) != bytes:
         msg = pkjson.dump_bytes(msg)
     if len(msg + b'\n') > job.cfg.max_message_bytes:
+        # TODO (gurhar1133): Do we need to change things so that we don't have to send COMPLETED for
+        # this error case?
         msg = pkjson.dump_bytes(PKDict(
```
robnagler, line 223: Formatting:

```py
msg = pkjson.dump_bytes(
    PKDict(
    [snip],
    ),
)
```

Avoid multiple calls on a single line when broken out. I think it is acceptable to

```py
msg = pkjson.dump_bytes(PKDict(data=job.COMPLETED, error='Data was too large'))
```

I perhaps set a bad example with resetting a parameter, which was discussed with @e-carlin. Here, `msg` goes through several transformation so I would code this as:

```py
def _validate_and_jsonl(msg):
    m = pkjson.dump_bytes(msg)
    if len(m) >=  job.cfg.max_message_bytes:
        m = pkjson.dump_bytes(PKDict(data=job.COMPLETED, error='Response is too large to send'))
    return m + b'\n'
```

I don't think `_change_message_too_big` is right, because it implies it is going to change. Rather validate implies it is conditional. The conversion to jsonl should happen here as well. This simplifies the code in the caller and documents that it is being converted to jsonl. There isn't a need for the conditional on the bytes, because it always comes in as a PKDict. If it is exactly max_message_bytes, the `>=` fixes that, which is an implicit +1 which I don't like so you could add a -1 if you feel it is clearer.

`b` is unnecessary afaict.

sirepo/pkcli/job_cmd.py
```diff
@@ -217,6 +218,8 @@ def _change_message_too_big(msg):
...
                 state=job.COMPLETED,
                 error='Data was too large',
```
robnagler, line 225: This message should be more explanatory: Response is too large to send.  

The data is too large, but what type of data is it? It's a response. It could be too large to be stored. This indicates a problem in the communication channel, not a problem with the simulation producing too much data.

robnagler (changes requested):

sirepo/pkcli/job_cmd.py
gurhar1133: b is for the case that we call _do_download_data_file
https://github.com/radiasoft/sirepo/blob/f922e02c7fe98a7d05610e50c322b019360ec243/sirepo/pkcli/job_cmd.py#L150-L155

But maybe something like:
```python
c = _validate_and_jsonl(c)
if 'error' in c:
    return c
```
would be better. I may be more involved with c is in bytes. I'll work on this.

robnagler: I looked for this, and somehow I didn't see it, sorry.

I think two functions are warranted: _validate_msg and _validate_msg_and_json.

gurhar1133: Is two functions better than
```python
c = _validate_and_jsonl(c)
    if 'error' in pkjson.load_any(c):
        return pkjson.load_any(c)
```
for _do_download_data_file?

sirepo/pkcli/job_cmd.py
```diff
@@ -148,9 +148,9 @@ def _do_download_data_file(msg, template):
             c = pkcompat.to_bytes(pkio.read_text(r.filename)) \
                 if u.endswith(('.py', '.txt', '.csv')) \
                 else r.filename.read_binary()
-            c, b = _change_message_too_big(c)
-            if b:
-                return c
+            c = _validate_and_jsonl(c)
+            if 'error' in pkjson.load_any(c):
```
robnagler, line 152: This is not going to work, and this is going to be very expensive (load_any)

Two functions are better, because they document what's going on. Private functions are "cheap" from a cognitive load and don't incur a cost by being a public function.

robnagler (changes requested):

sirepo/pkcli/job_cmd.py
gurhar1133: Cool. I'll do it with two

robnagler: I wasn't clear about what "going to work" means. Looking for `error` in a string is a fine thing to do if one is parsing, say, the output of a code. However, in this case, there is an arbitrary result coming back, and we are looking for a specific case. You could write it as:

```py
if 'error' == pkjson.load_any(c).get('state'):
```

However, this makes an assumption, the value of `c` is a `dict` (or PKDict). That's the type of assumption that gets us into trouble in software, and is what I call an "implicit coupling". A safer test would be:

```py
x = pkjson.load_any(c)
if isinstance(x, dict) and 'error' == x.get('state'):
```

However, this is increasing the cognitive load of the whole problem, and it would be clearer to say:

```py
if len(c) >= job.cfg.max_message_bytes:
    return json....PKDict(state='error'...))
```

The problem with this is that it is duplicating code unnecessarily. We already have the concept named, and it can be "explicitly coupled" via a separate function. That's what refactoring is all about: improving the state of existing code. 

That's why two functions are better than other solutions. You factor out a concept from a clearly named function without changing what it does.

sirepo/pkcli/job_cmd.py
```diff
@@ -317,6 +325,19 @@ def _parse_python_errors(text):
...
 
+def _validate_msg(msg):
+    if len(msg) >=  job.cfg.max_message_bytes:
+        return PKDict(state=job.COMPLETED, error='Response is too large to send')
+    return None
+
+
+def _validate_msg_and_json(msg):
+    msg = pkjson.dump_bytes(msg)
+    if len(msg) >=  job.cfg.max_message_bytes:
```
robnagler, line 336: Reuse:

```py
r = _validate_msg(msg)
if r:
     msg = pkjson.dump_bytes(r)
```

sirepo/pkcli/job_cmd.py
```diff
@@ -317,6 +325,19 @@ def _parse_python_errors(text):
     return ''
 
 
+def _validate_msg(msg):
+    if len(msg) >=  job.cfg.max_message_bytes:
+        return PKDict(state=job.COMPLETED, error='Response is too large to send')
+    return None
+
+
+def _validate_msg_and_json(msg):
```
robnagler, line 334: _validate_msg_and_json**l** -- the "l" is important.

robnagler (changes requested):

sirepo/pkcli/job_cmd.py
gurhar1133: Oh. Clever!

gurhar1133:
LMK if more changes :)

sirepo/pkcli/job_cmd.py
```diff
@@ -326,16 +326,18 @@ def _parse_python_errors(text):
...
     return None
 
 
-def _validate_msg_and_json(msg):
-    m = pkjson.dump_bytes(msg)
-    if len(m) >=  job.cfg.max_message_bytes:
-        m = pkjson.dump_bytes(PKDict(state=job.COMPLETED, error='Response is too large to send'))
-    return m + b'\n'
+def _validate_msg_and_jsonl(msg):
+    r = _validate_msg(msg)
```
robnagler, line 337: Too many changes. Here's what I meant:

```py
def _validate_msg_and_json(msg):
    m = pkjson.dump_bytes(msg)
    r = _validate_msg(m)
    if r:
        m = pkjson.dump_bytes(r)
    return m + b'\n'
```

robnagler (changes requested):

gurhar1133:
Anything else?

robnagler (approved):

robnagler:
@e-carlin i'll let you direct the merges.

e-carlin (commented):
just a few nits

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -60,9 +60,13 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             if (frameCache.getCurrentFrame(scope.modelName) == scope.prevFrameIndex) {
                 return;
             }
+            panelState.setError(scope.modelName, '');
```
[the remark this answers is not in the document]
e-carlin: ok, thanks.

gurhar1133:
Made changes

PR #4194, opened 2022-02-28
4062 multiple run setup UI fix

gurhar1133:
Now multiple run_setup visualizations will have their corresponding lattices.

### DIFF
--- a/sirepo/template/elegant.py
+++ b/sirepo/template/elegant.py
@@ -1096,12 +1096,6 @@ def _output_info(run_dir):
                 return 0
             return v
 
-        def _get_lattice_id(data):
-            for c in data.models.commands:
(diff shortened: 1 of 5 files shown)


sirepo/template/elegant.py
```diff
@@ -1147,6 +1148,7 @@ def _fix(v):
                         field_range[col][1] = max(_fix(max(values)), field_range[col][1])
                     else:
                         field_range[col] = [_fix(min(values)), _fix(max(values))]
+            lattice_id = LatticeUtil.get_lattice_id_from_file_id(data, file_id)
```
e-carlin, line 1151: No need for the local variable lattice_id. Just call the function when building the PKDict

gurhar1133:
Made change

PR #4211, opened 2022-03-04
fix #3654: jupyterhub moderation requests

### LINKED ISSUE #3654: Moderate JupyterHub user creations
Create user now gives the jupyterhub_login role, but this needs to be moderated to avoid issues with cryptominers. 

- User is prompted with a box for why they want to use jupyter if they don't already have the role
- An email is sent from sirepo with IP, user-agent, email, display_name, etc. (simply key: value form "display_name: Jane Doe") with a link to approve
- The link to approve goes to the approvals list, and each user in the list can be approved or rejected
- Once approved or rejected, an email goes out to the user with the decision
- create pkcli stays the same: if we create a jupyterhub user they get the role automatically

related:
- https://github.com/radiasoft/sirepo/issues/960

rorour:


sirepo/feature_config.py
```diff
@@ -54,7 +55,7 @@ def auth_controlled_sim_types():
       frozenset:  enabled sim types that require role
     """
     return frozenset(
-        cfg().proprietary_sim_types.union(cfg().default_proprietary_sim_types),
+        cfg().proprietary_sim_types.union(cfg().proprietary_with_moderation_prompt_sim_types),
```
robnagler, line 58: I think this is `moderated_sim_types`. They may or may not be proprietary.

sirepo/feature_config.py
```diff
@@ -115,7 +116,7 @@ def b(msg, dev=False):
     _cfg = pkconfig.init(
         # No secrets should be stored here (see sirepo.job.agent_env)
         api_modules=((), set, 'optional api modules, e.g. status'),
-        default_proprietary_sim_types=(set(), set, 'codes where all users are authorized by default but that authorization can be revoked'),
+        proprietary_with_moderation_prompt_sim_types=(set(), set, 'codes where all users are authorized by default but that authorization can be revoked'),
```
robnagler, line 119: The documentation string is incorrect: codes where all users must be authorized via moderation.

sirepo/feature_config.py
```diff
@@ -143,15 +144,15 @@ def b(msg, dev=False):
             display_test_boxes=b('Display test boxes to visualize 3D -> 2D projections'),
         ),
     )
-    i = _cfg.proprietary_sim_types.intersection(_cfg.default_proprietary_sim_types)
+    i = _cfg.proprietary_sim_types.intersection(_cfg.proprietary_with_moderation_prompt_sim_types)
     assert not i, \
-        f'{i}: cannot be in proprietary_sim_types and default_proprietary_sim_types'
+        f'{i}: cannot be in proprietary_sim_types and proprietary_with_moderation_prompt_sim_types'
```
robnagler, line 149: I think these are now orthogonal. proprietary_sym_types control where the code resides. moderated_sym_types control the authorization. You might not want to have proprietary_sym_types that are moderated (ITAR codes, for example), and vice versa (jupyter).

sirepo/http_request.py
```diff
@@ -91,7 +91,7 @@ def parse_post(**kwargs):
     if kwargs.pkdel('fixup_old_data'):
         r = simulation_db.fixup_old_data(r)[0]
     res.pkupdate(req_data=r)
-    kwargs.pksetdefault(type=True)
+    kwargs.pksetdefault(type=True, require_sim_type=True)
```
robnagler, line 94: I think I need this explained to me.

sirepo/package_data/approved_role_email.jinja
```diff
@@ -0,0 +1,2 @@
+Your request for {{ role }} access has been approved.
+{{ link }}
```
robnagler, line 2: missing end of line

sirepo/package_data/denied_role_email.jinja
```diff
@@ -0,0 +1 @@
+Your request for {{ role }} access has been denied.
```
robnagler, line 1: missing end of line

sirepo/package_data/approved_role_email.jinja
```diff
@@ -0,0 +1,2 @@
+Your request for {{ role }} access has been approved.
```
robnagler, line 1: I would name this file to be something auth_role_moderation/approve_email.jinja. It's a mistake to have too many top level templates.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,105 @@
...
+def api_approveOrDenyRole():
+    sirepo.auth.check_user_is_adm()
+    def _send_approve_or_deny_email(uid, approved, info):
+        sirepo.smtp.send(
+            recipient=sirepo.auth.get_module('email').AuthEmailUser.search_by(uid=uid).unverified_email,
+            subject=f'{info.role} Access Request' + ('Approved' if approved else 'Denied'),
+            body=pkjinja.render_resource('approved_role_email' if approved else 'denied_role_email', info),
+        )
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
+    uid = req.req_data.uid
```
robnagler, line 87: if you save everything in a PKDict, e.g. `p`, and also add `mode='approve'` or `mode='deny'`, then you refer to things like `approved` as `p.mode == 'approve'` and `uid` as `p.uid`. This keeps with the single letter naming, and makes it easier to pass around parameters.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,105 @@
...
+    sirepo.auth.check_user_is_adm()
+    rows = [row.__dict__ for row in sirepo.auth_db.UserRoleInvite.get_pending_rows()]
+    headers = [key for key in rows[0].keys() if key != '_sa_instance_state'] if rows else []
+    return sirepo.http_reply.gen_json({
+                'rows': rows,
+                'headers': headers,
+            })
+
+
+@sirepo.api_perm.require_user
```
robnagler, line 77: I would expect `@sirepo.api_perm.require_adm`.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,105 @@
...
+    role = req.req_data.role
+    approved = req.req_data.approved
+    role_name = sirepo.simulation_db.SCHEMA_COMMON.appInfo[sirepo.auth_role.sim_type(role)].shortName
+    link = flask.url_for('_dispatch_empty', _external=True) + sirepo.auth_role.sim_type(role)
+    if approved:
+        sirepo.auth_db.UserRole.add_role(uid, role)
+        sirepo.auth_db.UserRoleInvite.search_by(uid=uid, role=role).delete()
+        _send_approve_or_deny_email(uid, True, PKDict(link=link, role=role_name))
+        return sirepo.http_reply.gen_json_ok()
+    sirepo.auth_db.UserRoleInvite.set_status(uid, role, 'denied')
```
robnagler, line 97: I would try to reuse the word `approve` or `deny` everywhere so we don't have to remember even if it is not conjugated correctly.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,105 @@
...
+    sirepo.auth.check_user_is_adm()
+    def _send_approve_or_deny_email(uid, approved, info):
+        sirepo.smtp.send(
+            recipient=sirepo.auth.get_module('email').AuthEmailUser.search_by(uid=uid).unverified_email,
+            subject=f'{info.role} Access Request' + ('Approved' if approved else 'Denied'),
+            body=pkjinja.render_resource('approved_role_email' if approved else 'denied_role_email', info),
+        )
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
+    uid = req.req_data.uid
+    role = req.req_data.role
```
robnagler, line 88: I'm a little concerned about taking user data here without some type of validation. Rather, I would like to ensure that the role is the expected role, that is, is validated against UserRoleInvite first. Perhaps there could be a key in UserRoleInvite that's added to UserRoleInvite when it is created, and that gets passed as a token to ensure the administrator is authorized to add that particular role for that particular user. Without that token, the admin can't add an arbitrary role to an arbitrary user.

robnagler (changes requested):
First set of comments. I would like to discuss parse_post.require_sim_type.

sirepo/http_request.py
e-carlin: Certain api's need to be accessed for a sim type before the role for that sim type has been granted. For example, if a user visits `/jupyter` and they don't have the role they will be redirected to `/jupyterhublogin#/moderation-info`. In order for that route to be loaded we need to call [`api_simulationSchema`](https://github.com/radiasoft/sirepo/blob/cf35e963cd3086be50f00f32e2e9b77a004df3ce/sirepo/server.py#L518). But, if require_sim_type was True then that api request would fail because they don't have the role for jupyterhublogin. So we'd be in a loop (can't get to jupyter because they don't have the role and can't get to the page to request the role because they don't have the role).

In addition, [admins shouldn't be required](https://github.com/radiasoft/sirepo/blob/cf35e963cd3086be50f00f32e2e9b77a004df3ce/sirepo/auth_role_moderation.py#L86) to have the role for the sim_type they are granting a user to use.

sirepo/feature_config.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/feature_config.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/feature_config.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/package_data/approved_role_email.jinja
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/package_data/denied_role_email.jinja
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/package_data/approved_role_email.jinja
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/auth_role_moderation.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/auth_role_moderation.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/auth_role_moderation.py
rorour: 84a51a4c4d292b403b922c73411c9bad58de76bf

sirepo/http_request.py
robnagler: This is a slippery slope. These type of operations should not depend on accessing the sim_type. That's an ngui problem. Too much tied up in app vs sim_type.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+            recipient=_cfg.moderator_email,
+            subject=f'{info.sim_type} Access Request',
+            body=pkjinja.render_resource('auth_role_moderation/moderation_email', info),
+        )
+
+    req = sirepo.http_request.parse_post(require_sim_type=False)
+    d = req.req_data
+    uid = sirepo.auth.logged_in_user()
+
+    sirepo.auth_db.UserRoleInvite(
```
e-carlin, line 44: If I user enters this information twice this may fail with an IntegrityError. I think we should catch that error, log it, and then return a UserAlert to the user letting them know that they've already supplied info and we are processing it.

e-carlin:
Reminding myself - we need to write a migration for existing users

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+def api_approveOrDenyRole():
+    def _send_approve_or_deny_email(info):
+        a = 'Approved' if info.approved else 'Denied'
+        if sirepo.auth.dev_and_no_method('email'):
+            pkdlog('Email not enabled. Access is {}', a)
+            return
+        sirepo.smtp.send(
+            recipient=sirepo.auth.get_module('email').AuthEmailUser.search_by(uid=info.uid).unverified_email,
+            subject=f'{info.role} Access Request' + a,
+            body=pkjinja.render_resource('auth_role_moderation/approve_email' if info.approved else 'auth_role_moderation/deny_email', PKDict(
```
robnagler, line 89: The important part about breaking up long lines is separating arguments. There's too much cognitive load when a long line can contain more than one argument. Also in this case `auth_role_moderation/xxx_email` is redundant. I think I noted elsewhere that info should contain `approve` or `deny` and then: `f'auth_role_moderation/{info.status}_email` is clearer and shorter

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+                link=flask.url_for('_dispatch_empty', _external=True) + sirepo.auth_role.sim_type(info.role),
```
robnagler, line 90: We should always construct links with sirepo.uri.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+                app_name=sirepo.simulation_db.SCHEMA_COMMON.appInfo[sirepo.auth_role.sim_type(info.role)].shortName,
```
robnagler, line 91: Needs to be abstracted. auth_role should be able to go both directions.

robnagler (changes requested):
More comments. I didn't finish. LMK when you are ready for another review.

sirepo/api_auth.py
```diff
@@ -29,11 +29,13 @@ def assert_api_def(func):
 def check_api_call(func):
     expect = getattr(func, api_perm.ATTR)
     a = api_perm.APIPerm
-    if expect in (a.REQUIRE_COOKIE_SENTINEL, a.REQUIRE_USER):
+    if expect in (a.REQUIRE_COOKIE_SENTINEL, a.REQUIRE_USER, a.REQUIRE_ADM):
         if not cookie.has_sentinel():
             raise sirepo.util.SRException('missingCookies', None)
-        if expect == a.REQUIRE_USER:
+        if expect in (a.REQUIRE_USER, a.REQUIRE_ADM):
```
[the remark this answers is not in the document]
rorour: Can you please elaborate on this?

sirepo/auth_role_moderation.py
robnagler: > app_name=sirepo.simulation_db.SCHEMA_COMMON.appInfo[sirepo.auth_role.sim_type(info.role)].shortName

I just think the abstraction of looking up appInfo to get the shortName should be somewhere and not inline in auth_role_moderation. It will be needed in the list of moderation requests as well.

> link=flask.url_for('_dispatch_empty', _external=True) + sirepo.auth_role.sim_type(info.role),

Similarly this should be encapsulated in sirepo.uri_router. This is a bit complicated, because sirepo.uri.app_root already defines the concept of mapping from a sim_type to uri, but it doesn't support absolute position. Attributes that begin with an underscore (_dispatch_empty) should not be referenced outside the module in which they are defined. This is called a leaky abstraction. I think the best place to do this would be in sirepo.uri_router so this would look like (putting it all together:

```py
t = sirepo.auth_role.sim_type(info.role)
m = 'approve' if info.approved else 'deny'
body=pkjinja.render_resource(
    f'auth_role_moderation/{m}_email',
    PKDict(
        link=sirepo.uri_router.absolute_uri(sirepo.uri.app_root(t)),
        app_name=sirepo.simulation_db.app_info(t).shortName,
```

At least that's how I see. It could be that app_info is in the wrong place.

sirepo/api_auth.py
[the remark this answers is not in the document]
robnagler: This is not something we can fix now, but it has to do with how you [design roll-based access control](https://en.wikipedia.org/wiki/Role-based_access_control#Design). Fundamentally there's a [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) issue: who has what permissions and what do those permissions allow you to do. The "what" is best managed by permissions: read, write, admin, etc. The "who" is best handled by roles: admin, user, guest, etc. In the above code the "what" is tied to the "who", and that doesn't allow enough flexibility. That being said, this was just an FYI, not something I expected you do tod.

sirepo/auth_role_moderation.py
rorour: I had already made some changes to this section at @e-carlin's recommendation, including moving app_name out to the main function body and adding sirepo.uri.server_root() to be used in the link. So now I'm a bit confused on how to implement these changes- can you please let me know if/how they still apply?

etc/run-moderation.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth/__init__.py
```diff
@@ -620,6 +637,7 @@ def _auth_state():
     return v
 
 
+
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth/email.py
```diff
@@ -156,6 +156,11 @@ def delete_changed_email(cls, user):
                 ).delete()
                 cls._session().commit()
 
+        @classmethod
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_db.py
```diff
@@ -5,6 +5,9 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_db.py
```diff
@@ -288,6 +318,35 @@ def uids_of_paid_users(cls):
...
                 ).distinct().all()
             ]
+
+    class UserRoleInvite(UserDbBase):
+        __tablename__ = 'user_role_invite_t'
+        uid = sqlalchemy.Column(UserDbBase.STRING_ID, primary_key=True)
+        role = sqlalchemy.Column(UserDbBase.STRING_NAME, primary_key=True)
+        status = sqlalchemy.Column(UserDbBase.STRING_NAME, nullable=False)
+        token = sqlalchemy.Column(UserDbBase.STRING_NAME, nullable=False, unique=True)
+        last_updated = sqlalchemy.Column(sqlalchemy.DateTime(), server_default=sqlalchemy.sql.func.now(), onupdate=sqlalchemy.sql.func.now(), nullable=False)
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_db.py
```diff
@@ -5,6 +5,9 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
+
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role.py
```diff
@@ -11,14 +11,14 @@
 ROLE_ADM = 'adm'
 ROLE_PAYMENT_PLAN_ENTERPRISE = 'enterprise'
 ROLE_PAYMENT_PLAN_PREMIUM = 'premium'
-
+_SIM_TYPE_ROLE_PREFIX = 'sim_type_'
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/package_data/static/html/moderation-info.html
```diff
@@ -0,0 +1,8 @@
+<div class="container-fluid">
+  <div class="row">
+    <div class="col-sm-6 col-sm-offset-2">
+      <h1>Moderation Info</h1>
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/package_data/auth_role_moderation/moderation_email.jinja
```diff
@@ -0,0 +1,9 @@
+Name: {{ display_name }}
+Simulation Type: {{ sim_type }}
+Reason: {{ reason }}
+Email: {{ email_addr }}
+DNS: {{ domain_name }}
+IP: {{ ip_addr }}
+
+To accept or reject:
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@sirepo.api_perm.require_adm
+def api_approveOrDenyRole():
+    def _send_approve_or_deny_email(info):
+        a = 'Approved' if info.approved else 'Denied'
+        if sirepo.auth.dev_and_no_method('email'):
+            pkdlog('Email not enabled. Access is {}', a)
+            return
+        sirepo.smtp.send(
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+        )
+    return sirepo.http_reply.gen_json_ok()
+
+
+@sirepo.api_perm.require_adm
+def api_getPendingUserRoleInvites():
+    return sirepo.http_reply.gen_json(PKDict(
+        rows=[r.as_pkdict() for r in sirepo.auth_db.UserRoleInvite.get_pending_rows()],
+        headers=[c for c in sirepo.auth_db.UserRoleInvite.column_names() if c != 'token'],
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+import sirepo.http_request
+import sirepo.simulation_db
+
+_cfg = None
+
+@sirepo.api_perm.require_user
+def api_saveModerationReason():
+    def _moderation_link(sim_type):
+        return flask.url_for('_dispatch_empty', _external=True).rstrip('/') \
+               + sirepo.uri.local_route(sim_type, 'admRoles')
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+import sirepo.http_reply
+import sirepo.http_request
+import sirepo.simulation_db
+
+_cfg = None
+
+@sirepo.api_perm.require_user
+def api_saveModerationReason():
+    def _moderation_link(sim_type):
+        return flask.url_for('_dispatch_empty', _external=True).rstrip('/') \
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

etc/run-jupyterhub.sh
```diff
@@ -1,7 +1,37 @@
 #!/bin/bash
 set -eou pipefail
 
-export SIREPO_FEATURE_CONFIG_DEFAULT_PROPRIETARY_SIM_TYPES=jupyterhublogin
+if [[ ! -d ~/mail ]]; then
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

etc/run-moderation.sh
```diff
@@ -0,0 +1,33 @@
...
+        exit 1
+    fi
+    rm ~/mail/1
+fi
+export SIREPO_FROM_EMAIL='support@radiasoft.net'
+export SIREPO_FROM_NAME='RadiaSoft Support'
+export SIREPO_SMTP_SERVER='localhost'
+export SIREPO_SMTP_SEND_DIRECTLY=1
+export SIREPO_SMTP_USER='vagrant'
+export SIREPO_AUTH_METHODS='email:guest'
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/api_auth.py
```diff
@@ -29,11 +29,13 @@ def assert_api_def(func):
...
             auth.require_user()
+            if expect == a.REQUIRE_ADM:
+                auth.require_user_is_adm()
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth/__init__.py
```diff
@@ -121,9 +121,18 @@ def api_authLogout(simulation_type=None):
 def check_user_has_role(uid, role, raise_forbidden=True):
     if auth_db.UserRole.has_role(uid, role):
         return True
-    if raise_forbidden:
+    if not raise_forbidden:
+        return False
+    if role not in [sirepo.auth_role.for_sim_type(s) for s in sirepo.feature_config.cfg().moderated_sim_types]:
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth/__init__.py
```diff
@@ -121,9 +121,18 @@ def api_authLogout(simulation_type=None):
...
         sirepo.util.raise_forbidden('uid={} role={} not found'.format(uid, role))
-    return False
+    s = sirepo.auth_db.UserRoleInvite.get_status(uid, role)
+    if s == 'pending':
+        raise sirepo.util.SRException('moderationPending', None)
+    if s == 'denied':
+        sirepo.util.raise_forbidden('uid={} role={} not found'.format(uid, role))
+    assert s is None, \
+        f'Unexpected status={s} for uid={uid} and role={role}'
+    raise sirepo.util.SRException('moderationInfo', None)
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth/__init__.py
```diff
@@ -121,9 +121,18 @@ def api_authLogout(simulation_type=None):
...
+    if not raise_forbidden:
+        return False
+    if role not in [sirepo.auth_role.for_sim_type(s) for s in sirepo.feature_config.cfg().moderated_sim_types]:
         sirepo.util.raise_forbidden('uid={} role={} not found'.format(uid, role))
-    return False
+    s = sirepo.auth_db.UserRoleInvite.get_status(uid, role)
+    if s == 'pending':
+        raise sirepo.util.SRException('moderationPending', None)
+    if s == 'denied':
+        sirepo.util.raise_forbidden('uid={} role={} not found'.format(uid, role))
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_db.py
```diff
@@ -249,11 +268,23 @@ def all_roles(cls):
                     r[0] for r in cls._session().query(cls.role.distinct()).all()
                 ]
 
+        @classmethod
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role.py
```diff
@@ -32,4 +32,8 @@ def get_all():
 
 
 def for_sim_type(sim_type):
-    return 'sim_type_' + sim_type
+    return _SIM_TYPE_ROLE_PREFIX + sim_type
+
+
+def sim_type(role):
+    return role.split(_SIM_TYPE_ROLE_PREFIX)[-1]
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+               + sirepo.uri.local_route(sim_type, 'admRoles')
+
+    def _dns_reverse_lookup(ip):
+        import dns.resolver, dns.reversename
+        try:
+            return str(dns.resolver.query(dns.reversename.from_address(ip), 'PTR')[0])
+        except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN):
+            return 'No Reverse DNS Lookup'
+
+    def _send_moderation_email(info):
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+    d = req.req_data
+    uid = sirepo.auth.logged_in_user()
+
+    sirepo.auth_db.UserRoleInvite(
+        uid=uid,
+        role=sirepo.auth_role.for_sim_type(d.simulationType),
+        status='pending',
+        token=sirepo.util.random_base62(32),
+    ).save()
+    l = _moderation_link(d.simulationType)
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+    if sirepo.auth.dev_and_no_method('email'):
+        pkdlog('Email not enabled. Moderation reason={} Link to approve/deny={}', d.reason, l)
+    else:
+        _send_moderation_email(
+            PKDict(
+                display_name=sirepo.auth_db.UserRegistration.search_by(uid=uid).display_name,
+                email_addr=sirepo.auth.get_module('email').AuthEmailUser.search_by(uid=uid).unverified_email,
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+                ip_addr=flask.request.remote_addr,
+                domain_name=_dns_reverse_lookup(flask.request.remote_addr),
+                link=l,
+                reason=d.reason,
+                role=sirepo.auth_role.for_sim_type(d.simulationType),
+                sim_type=d.simulationType,
+                uid=uid,
+                user_agent=flask.request.headers.get("User-Agent"),
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,116 @@
...
+        status='pending',
+        token=sirepo.util.random_base62(32),
+    ).save()
+    l = _moderation_link(d.simulationType)
+    if sirepo.auth.dev_and_no_method('email'):
+        pkdlog('Email not enabled. Moderation reason={} Link to approve/deny={}', d.reason, l)
+    else:
+        _send_moderation_email(
+            PKDict(
+                display_name=sirepo.auth_db.UserRegistration.search_by(uid=uid).display_name,
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

etc/run-moderation.sh
```diff
@@ -0,0 +1,33 @@
...
+        echo mail delivery test failed
+        exit 1
+    fi
+    rm ~/mail/1
+fi
+export SIREPO_FROM_EMAIL='support@radiasoft.net'
+export SIREPO_FROM_NAME='RadiaSoft Support'
+export SIREPO_SMTP_SERVER='localhost'
+export SIREPO_SMTP_SEND_DIRECTLY=1
+export SIREPO_SMTP_USER='vagrant'
```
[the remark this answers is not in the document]
rorour: f77e9717949945b988991128a2701fe97b1111d6

sirepo/auth_role_moderation.py
rorour: f77e9717949945b988991128a2701fe97b1111d6

etc/run-moderation.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo 'Testing mail delivery'
+    echo hello | sendmail vagrant@localhost.localdomain
+    sleep 2
+    if ! grep -s -d hello ~/mail/1; then
+        echo mail delivery test failed
+        exit 1
+    fi
+    rm ~/mail/1
+fi
+export SIREPO_FROM_EMAIL='support@radiasoft.net'
```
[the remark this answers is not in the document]
rorour: b3e52e371342a29e2912d6dd559da8602ec27f79

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+def api_approveOrDenyRole():
+    def _send_approve_or_deny_email(info):
+        a = 'Approved' if info.approved else 'Denied'
+        if sirepo.auth.dev_and_no_method('email'):
+            pkdlog('Email not enabled. Access is {}', a)
+            return
+        sirepo.smtp.send(
+            recipient=sirepo.auth.get_module(
+                'email',
+            ).AuthEmailUser.search_by(uid=info.uid).unverified_email,
```
robnagler, line 96: This should be user_name, not unverified_email. The user must already be verified before they can get to this point (btw, this should be tested).

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+            subject=f'{info.app_name} Access Request ' + a,
+            body=pkjinja.render_resource(
+                f'auth_role_moderation/{info.status}_email',
+                PKDict(
+                    link=sirepo.uri.server_root(external=True) + sirepo.auth_role.sim_type(info.role),
```
robnagler, line 101: I think this should be shortened to: `sirepo.uri.app_root(sirepo.auth_role.sim_type(info.role), external=True)`

By creating server_root, this breaks the abstraction provided by `app_root`.

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+                    app_name=info.app_name,
+                ),
+            ),
+        )
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
+    i = sirepo.auth_db.UserRoleInvite.search_by(token=req.req_data.token)
+    assert i, f'No record for token={req.req_data.token}'
```
robnagler, line 108: If the record is not found, it should be a UserAlert. Below the record is deleted if approved.

However, in looking at this from an auditing perspective, we could retain the record of approvals, too. This would make the table larger, but no larger than the role table, which is accessed much more frequently.  Indeed, the person who took the last action should be recorded in the table, too, so expand the table to include moderator_uid.

tests/role_moderation_test.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+from __future__ import absolute_import, division, print_function
+from pykern.pkcollections import PKDict
+import os
+import pytest
+
+def setup_module(module):
+    os.environ.update(
+        SIREPO_FEATURE_CONFIG_MODERATED_SIM_TYPES='elegant',
+        SIREPO_AUTH_ROLE_MODERATION_MODERATOR_EMAIL='vagrant@localhost.localdomain',
```
robnagler, line 15: Use `os.getlogin()` instead of `vagrant`.

tests/role_moderation_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_moderation(fc):
+    from pykern import pkunit
+    from sirepo import srunit
+    import sirepo.auth_db
+    import sirepo.pkcli.roles
+
+    with srunit.auth_db_session():
+        sirepo.auth_db.UserRole.delete_all()
```
robnagler, line 26: I was skeptical about this, but makes sense. You may want to delete all UserRoleInvites, too.

robnagler (changes requested):
It might be good to walk through these changes together. I have particular opinions here that may need more explanations.

tests/role_moderation_test.py
```diff
@@ -0,0 +1,59 @@
...
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from __future__ import absolute_import, division, print_function
+from pykern.pkcollections import PKDict
+import os
+import pytest
+
+def setup_module(module):
+    os.environ.update(
+        SIREPO_FEATURE_CONFIG_MODERATED_SIM_TYPES='elegant',
```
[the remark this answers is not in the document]
rorour: 8ef331f5ed8d6d578701a4374d899b1b06f51658

tests/role_moderation_test.py
rorour: 8ef331f5ed8d6d578701a4374d899b1b06f51658

tests/role_moderation_test.py
rorour: 8ef331f5ed8d6d578701a4374d899b1b06f51658

sirepo/auth_role_moderation.py
rorour: 8ef331f5ed8d6d578701a4374d899b1b06f51658

sirepo/auth_role_moderation.py
rorour: 7813ef1ac1103a4c5aca70b302c249a7d56d910f

sirepo/auth_role_moderation.py
rorour: 7813ef1ac1103a4c5aca70b302c249a7d56d910f

sirepo/auth_role_moderation.py
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+            ).pkupdate(sirepo.http_request.user_agent_headers())
+        )
+    return sirepo.http_reply.gen_json_ok()
+
+
+@sirepo.api_perm.require_adm
+def api_getPendingAndClarifyUserRoleInvites():
+    return sirepo.http_reply.gen_json(PKDict(
+        rows=[r.as_pkdict() for r in (
+                sirepo.auth_db.UserRoleInvite.get_pending_rows()
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+                    display_name=info.display_name,
+                ),
+            ),
+        )
+
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
+    i = sirepo.auth_db.UserRoleInvite.search_by(token=req.req_data.token)
+    assert i, f'No record for token={req.req_data.token}'
+    sirepo.auth_db.UserRoleInvite.set_status(i.uid, i.role, 'clarify')
+    pkdlog('UserRoleInvite status changed to clarify for uid={} role={} token={}',
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+        _send_approve_or_deny_email(p)
+        return sirepo.http_reply.gen_json_ok()
+    sirepo.auth_db.UserRoleInvite.set_status(p.uid, p.role, 'deny')
+    _send_approve_or_deny_email(p)
+    return sirepo.http_reply.gen_json_ok()
+
+
+@sirepo.api_perm.require_adm
+def api_clarifyModerationRequest():
+    def _send_clarify_email(info):
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+        if sirepo.auth.dev_and_no_method('email'):
+            pkdlog('Email not enabled for clarify.')
+            return
+        sirepo.smtp.send(
+            recipient=sirepo.auth.get_module(
+                'email',
+            ).AuthEmailUser.search_by(uid=info.uid).unverified_email,
+            subject=f'Additional Information Requested for {info.app_name} Access',
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+            subject=f'{info.app_name} Access Request ' + a,
+            body=pkjinja.render_resource(
+                f'auth_role_moderation/{info.status}_email',
+                PKDict(
+                    link=sirepo.uri.server_root(external=True) + sirepo.auth_role.sim_type(info.role),
+                    app_name=info.app_name,
+                ),
+            ),
+        )
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+        rows=[r.as_pkdict() for r in (
+                sirepo.auth_db.UserRoleInvite.get_pending_rows()
+                + sirepo.auth_db.UserRoleInvite.get_clarify_rows()
+        )],
+    ))
+
+
+@sirepo.api_perm.require_adm
+def api_approveOrDenyRole():
+    def _send_approve_or_deny_email(info):
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,172 @@
...
+                role=sirepo.auth_role.for_sim_type(d.simulationType),
+                sim_type=d.simulationType,
+                uid=u,
+            ).pkupdate(sirepo.http_request.user_agent_headers())
+        )
+    return sirepo.http_reply.gen_json_ok()
+
+
+@sirepo.api_perm.require_adm
+def api_getPendingAndClarifyUserRoleInvites():
```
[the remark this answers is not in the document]
rorour: f912c0cd3b5b5b02720985017a7af542f5452940

sirepo/uri.py
```diff
@@ -28,17 +22,21 @@
 PATH_INFO_CHAR = '*'
 
 
-def app_root(sim_type):
+def app_root(sim_type, external=False):
     """Generate uri for application root
 
     Args:
         sim_type (str): application name
+        external (bool): external uri? [False]
```
robnagler, line 30: `if True, make the uri absolute`

Try not to repeat the name of the parameter in the description to help reader understand. The `?` is not appropriate, because the client wants to know what the function will do when supplying it True or False, that is, write it from the perspective of the function, not what the client is asking.

sirepo/uri.py
```diff
@@ -28,17 +22,21 @@
...
 
     Returns:
         str: formatted URI
     """
-    t = http_request.sim_type(sim_type)
-    return ROOT if t is None else ROOT + t
+    return uri_router.uri_for_api(
+        'root',
+        params=PKDict(path_info=http_request.sim_type(sim_type) or ''),
```
robnagler, line 37: Now that we are not constructing the uri here, this call should be:

```py
t = http_request.sim_type(sim_type)
return uri_router.uri_for_api(
        'root',
        params=PKDict(path_info=t) if t else None,
        external=external,
)
```

sirepo/uri.py
```diff
@@ -51,14 +49,16 @@ def init(**imports):
...
+def local_route(sim_type, route_name=None, params=None, query=None, external=False):
     """Generate uri for local route with params
 
     Args:
         sim_type (str): application name
         route_name (str): a local route [defaults to local default]
         params (dict): paramters to pass to route
         query (dict): query values (joined and escaped)
+        external (bool): external uri? [False]
+
```
robnagler, line 61: No blank line here.

sirepo/uri.py
```diff
@@ -125,6 +119,8 @@ def _query(query):
     return '?' + urlencode(query)
 
 
+
```
robnagler, line 122: too many blank lines

sirepo/uri_router.py
```diff
@@ -185,7 +185,7 @@ def uri_for_api(api_name, params=None, external=True):
         str: formmatted external URI
     """
     r = _api_to_route[api_name]
-    res = (sirepo.uri.server_root(external=external) + r.base_uri).rstrip('/')
+    res = (flask.url_for('_dispatch_empty', _external=external) + r.base_uri).rstrip('/')
```
robnagler, line 188: This will not work if external is true, it will strip too many slashes with `app_root(sim_type=None)`. This is unusual, but easily fixable:

```py
res = r.base_uri
```

`res` doesn't have any slashes at this point.

Then at the end:

```py
return flask.url_for('_dispatch_empty', _external=external) + res
```

The code was broken for this case before. It would be good to have a test. Add this case to uri_router_test.py test_uri_for_api:

```py
        pkeq('/', uri_router.uri_for_api('root', params={}, external=False))
```

It will have very confusing output so run it this way:

```sh
pykern test uri_router_test.py case=uri_for_api
```

This will still be confusing, because it has to run the test case in the server, but look for the PKFail:

```txt
pykern.pkunit.PKFail: uri_router_test.py:64:t expect=/ != actual=
```

tests/role_moderation_test.py
```diff
@@ -24,8 +24,9 @@ def test_moderation(fc):
 
     with srunit.auth_db_session():
         sirepo.auth_db.UserRole.delete_all()
+        sirepo.auth_db.UserRoleInvite.delete_all()
     with pkunit.pkexcept('SRException.*moderationRequest'):
-        fc.sr_sim_data(sim_type='elegant', sim_name='SPEAR3')
+        fc.sr_sim_data(sim_type='myapp', sim_name='Scooby Doo')
```
robnagler, line 29: Now you can default these values:

```py
        fc.sr_sim_data()
```

[See sr_sim_data](https://github.com/radiasoft/sirepo/blob/c5bd99d398ea3f72e1221f057fca49e9b115f25e/sirepo/srunit.py#L541) to see the defaults.

sirepo/auth_db.py
```diff
@@ -340,10 +336,11 @@ def get_status(cls, uid, role):
                 return s.status
 
         @classmethod
-        def set_status(cls, uid, role, status):
+        def set_status(cls, uid, role, status, moderator_uid=None):
             with sirepo.util.THREAD_LOCK:
```
robnagler, line 340: Assert status is one of the valid values. Normally the db would do this if it were an enum, but we don't want to get into that.

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
 
 _cfg = None
 
+
+@sirepo.api_perm.require_adm
+def api_admModerate():
+
+    def _send_moderation_status_email(info):
+        approved_denied_subject = f'{info.app_name} Access Request '
```
robnagler, line 30: I'd vote for a single character variable here as it is a temporary that is used in the next couple of lines.

robnagler, line 30: I think this could be a module constant:

```py
_STATUS_TO_SUBJECT = PKDict(
    approve={} Access Request Approved',
    clarify='Sirepo {}: Additional Info?',
    deny={} Access Request Denied',
)
```

The denormalization is fine in this case, because the text strings are so close together. There's only two uses, too. It simplifies the code, and there's no great harm if there's a spelling error or whatever.

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
...
+            approve=approved_denied_subject + 'Approved',
+            clarify=f'Sirepo {info.app_name}: Additional Info?',
+            deny=approved_denied_subject + 'Denied',
+        )
+        if sirepo.auth.dev_and_no_method('email'):
+            pkdlog('Email not enabled. Moderation status is {}', info.status)
+            return
+        sirepo.smtp.send(
+            recipient=sirepo.auth.user_name(info.uid),
+            subject=email_subject.get(info.status),
```
robnagler, line 41: Using the constant:

```py
subject=_STATUS_TO_SUBJECT[info.status].format(info.app_name)
```

sirepo/uri.py
```diff
@@ -74,21 +74,15 @@ def local_route(sim_type, route_name=None, params=None, query=None):
...
-
-
-def server_root(external):
-    return flask.url_for('_dispatch_empty', _external=external)
+    return app_root(t, external=external) + '#' + u + _query(query)
 
 
 def server_route(route_or_uri, params, query):
     """Convert name to uri found in SCHEMA_COMMON
-
```
robnagler, line 81: Leave blank line between description and Args.

tests/role_moderation_test.py
```diff
@@ -44,16 +45,16 @@ def test_moderation(fc):
...
+        'admModerate',
         PKDict(
             token=r.rows[0].token,
-            approved=True,
+            status='approve',
         ),
     )
-    r = fc.sr_sim_data(sim_type='elegant', sim_name='SPEAR3')
-    pkunit.pkeq(r.models.simulation.name, 'SPEAR3')
+    r = fc.sr_sim_data(sim_type='myapp', sim_name='Scooby Doo')
```
robnagler, line 59: use defaults.

tests/role_moderation_test.py
```diff
@@ -44,16 +45,16 @@ def test_moderation(fc):
...
+    pkunit.pkeq(r.models.simulation.name, 'Scooby Doo')
```
robnagler, line 60: You don't need to test Scooby Doo. You can just see if you got back any models:

```py
pkunit.pkok(r.get('models'), 'no models r={}', r)
```

One of the ways to make tests more robust to changes is to avoid testing too much. In this case, if you don't get models back, that's enough. By doing a `get` the test will output the error message instead of getting a KeyError, which is less informative.

robnagler (changes requested):
I backtracked on some decisions and added more feedback. I'm being pedantic, because there are a lot of interesting coding and testing cases.

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
...
+def api_admModerate():
+
+    def _send_moderation_status_email(info):
+        approved_denied_subject = f'{info.app_name} Access Request '
+        email_subject = PKDict(
+            approve=approved_denied_subject + 'Approved',
+            clarify=f'Sirepo {info.app_name}: Additional Info?',
+            deny=approved_denied_subject + 'Denied',
+        )
+        if sirepo.auth.dev_and_no_method('email'):
```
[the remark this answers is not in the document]
rorour: I tried removing it, but I ran into this error.

```
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/uri_router.py", line 77, in call_api
    r = flask.make_response(f(**kwargs) if kwargs else f())
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth_role_moderation.py", line 86, in api_admModerate
    _send_moderation_status_email(p)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth_role_moderation.py", line 44, in _send_moderation_status_email
    recipient=sirepo.auth.user_name(info.uid),
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth/__init__.py", line 519, in user_name
    'UserModel',
AttributeError: module 'sirepo.auth.guest' has no attribute 'UserModel'
```

What should we do for guest users since they have no UserModel?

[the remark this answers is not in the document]
robnagler: We should never get a moderation request from a guest user so this code should never be executed in that case. How is a guest user asking to moderate?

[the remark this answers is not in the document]
rorour: There is nothing that prevents a guest user from asking to moderate. [This test](https://github.com/radiasoft/sirepo/blob/8cb1f59751554819c468bf4cdd36601c23a8637e/tests/role_moderation_test.py#L19) shows an example of it. Another way to show it is to set a sim_type as a moderated sim_type, log in as a guest user, delete all roles for the guest user, and then go to that sim_type. The guest user will be prompted for a moderation reason. 

How should guest users be prevented from interacting with the moderation code?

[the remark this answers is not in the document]
robnagler: When a guest visits a moderated code, simply reply Forbidden like as if they don't have the role. There would be no moderation whatsoever, because there's no way to notify the user about the moderation request. The test should use [email login](https://github.com/radiasoft/sirepo/blob/8cb1f59751554819c468bf4cdd36601c23a8637e/tests/auth/email1_test.py#L157) for creating the user. Bonus points to modularize it after copying it.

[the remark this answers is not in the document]
robnagler: I think user_name should always return something. For guests, the user_name could be `guest-<uid>`. Don't worry about auth.basic or auth.bluesky. auth.github would be fine (handle) and auth.email works, obviously. If we add ldap, it will probably be for ActiveDirectory so user names will be emails, too.

During the initial moderation check, if the user_name is not an email (use pyisemail.is_email), then we have no way of contacting them, and therefore, we can't moderate, i.e. just raise forbidden as if the role doesn't exist.

sirepo/auth/__init__.py
```diff
@@ -337,6 +337,10 @@ def process_request(unit_test=None):
         yield
 
 
+def require_adm():
+    assert check_user_has_role(logged_in_user(), sirepo.auth_role.ROLE_ADM)
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_db.py
```diff
@@ -340,10 +336,11 @@ def get_status(cls, uid, role):
...
                 s = cls.search_by(uid=uid, role=role)
                 s.status = status
+                s.moderator_uid = moderator_uid
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role.py
```diff
@@ -15,12 +15,20 @@
 _SIM_TYPE_ROLE_PREFIX = 'sim_type_'
 
 
+def for_moderated_sim_types():
+    return [sirepo.auth_role.for_sim_type(s) for s in sirepo.feature_config.cfg().moderated_sim_types]
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

tests/role_moderation_test.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

tests/role_moderation_test.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_db.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

tests/role_moderation_test.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri_router.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/uri.py
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
...
+        i.role,
+        i.token,
+    )
+    _send_moderation_status_email(p)
+    return sirepo.http_reply.gen_json_ok()
+
+
+@sirepo.api_perm.require_adm
+def api_getModerationRequestRows():
+    return sirepo.http_reply.gen_json(PKDict(
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
...
+        app_name=sirepo.simulation_db.SCHEMA_COMMON.appInfo[
+            sirepo.auth_role.sim_type(i.role)
+        ].longName,
+        display_name=sirepo.auth.user_display_name(i.uid),
+        role=i.role,
+        status=req.req_data.status,
+        uid=i.uid,
+    )
+    _set_moderation_status(p)
+    pkdlog('UserRoleInvite status changed to {} for uid={} role={} token={}',
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
```diff
@@ -22,11 +22,81 @@
...
+            info.role,
+            info.status,
+            moderator_uid=sirepo.auth.logged_in_user()
+        )
+
+    req = sirepo.http_request.parse_post(require_sim_type=False, type=False)
+    i = sirepo.auth_db.UserRoleInvite.search_by(token=req.req_data.token)
+    if not i:
+        raise sirepo.util.UserAlert(
+            f'No record in UserRoleInvite for token={req.req_data.token}',
```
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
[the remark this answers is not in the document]
rorour: ad61fc813610afe4d4c57cbe0e4322c503d6ac6c

sirepo/auth_role_moderation.py
```diff
@@ -0,0 +1,146 @@
...
+        return sirepo.uri.local_route(sim_type, route_name='admRoles', external=True)
+
+    def _send_request_email(info):
+        sirepo.smtp.send(
+            recipient=_cfg.moderator_email,
+            subject=f'{info.sim_type} Access Request',
+            body=pkjinja.render_resource('auth_role_moderation/moderation_email', info),
+        )
+
+    req = sirepo.http_request.parse_post(require_sim_type=False)
```
robnagler, line 104: This too should be different.

sirepo/auth/__init__.py
```diff
@@ -149,6 +166,10 @@ def create_new_user(uid_generated_callback, module):
     return u
 
 
+def dev_and_no_method(method):
```
robnagler, line 169: This goes away.

sirepo/http_request.py
```diff
@@ -80,6 +80,8 @@ def parse_post(**kwargs):
         model (object): `parse_model`
         name (object): `parse_name`
         template (object): `sirepo.template.import_module`
+        require_sim_type (bool): `require_sim_type`
```
robnagler, line 83: This parameter should go away. The problem is in require_sim_type():

```py
def require_sim_type(sim_type):
    if sim_type not in sirepo.feature_config.auth_controlled_sim_types():
        return
    if not _is_logged_in():
        # If a user is not logged in, we allow any sim_type, because
        # the GUI has to be able to get access to certain APIs before
        # logging in.
        return
    check_user_has_role(
        logged_in_user(),
        sirepo.auth_role.for_sim_type(sim_type),
    )
```

Comments are a good indication of a problem. If the user is not logged in, we need to return the minimal schema that allows the user to login. The app will then need to refresh itself. I think that's what happens now when you login anyway. 

I think a minimal schema is too complicated for the moment. (This is what happens when you bolt on security, instead of designing it in.)  

What we should do is add a special decorator that documents the APIs similar to who api_perm works. It could be called something `@allow_unauthorized_sim_type`. 

In uri_router, we save the current api in srcontext. You can see how to do that in http_request, for example. There's already bracketing code.

Inside auth.require_sim_type, if the type is unauthorized, it would allow the API to be authorized.  

As far as check_auth_method, this would happen in require_sim_type, not where it is now. If it is a moderated sim type, then Forbidden is raised in required_sim_type if the user is logged in, and the user doesn't have a user_name that's an email.

robnagler (changes requested):
I think I should help with this. Set up a meeting when you want to work on this.

rorour:
@robnagler where should we check for guest users to keep them from accessing the moderation apis?

sirepo/auth/__init__.py
```diff
@@ -514,10 +526,13 @@ def user_name(uid=None):
     u = getattr(
```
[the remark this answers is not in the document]
rorour: 06a93024bbd0c69ef4b9cc7be8543cfae94e9bb6

sirepo/auth/__init__.py
```diff
@@ -514,10 +526,13 @@ def user_name(uid=None):
...
         _METHOD_MODULES[m],
         'UserModel',
+        None,
     )
     if u:
         with util.THREAD_LOCK:
             return u.search_by(uid=uid).user_name
+    elif m == METHOD_GUEST:
+        return 'guest-' + uid
     raise AssertionError(
```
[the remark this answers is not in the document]
rorour: 06a93024bbd0c69ef4b9cc7be8543cfae94e9bb6

tests/role_moderation_test.py
```diff
@@ -56,5 +59,23 @@ def test_moderation(fc):
...
+    r = fc.sr_sim_data()
+    pkunit.pkok(r.get('models'), 'no models r={}', r)
+
+
+def test_no_guest():
+    from pykern import pkunit
+    from pykern.pkdebug import pkdp
+    from pykern.pkcollections import PKDict
+    from sirepo import srunit
+    fc = srunit.flask_client()
```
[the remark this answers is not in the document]
rorour: 06a93024bbd0c69ef4b9cc7be8543cfae94e9bb6

sirepo/jupyterhub.py
```diff
@@ -33,49 +34,38 @@ def __init__(self, *args, **kwargs):
...
-@contextlib.contextmanager
-def _set_cookie(handler):
-    import sirepo.auth_db
-    with sirepo.auth_db.session(), \
-         sirepo.cookie.set_cookie_outside_of_flask_request(
-             handler.get_cookie(sirepo.cookie.cfg.http_name),
-         ):
-        yield
+_cfg = pkconfig.init(
+    # TODO(e-carlin): check with rjn if this is the right way to get the uri?
```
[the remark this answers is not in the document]
robnagler: This should be in the jupyterhub config, or are you expecting to cascade the uri like agent_uri?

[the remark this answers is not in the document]
e-carlin: I'm not exactly sure what you're saying. If you mean "cascade the uri like the agent_uri" to mean that rsconf will put the uri on like how we set the `supervisor_uri` in job_driver.sbatch which ends up setting the `agent_uri` then yes.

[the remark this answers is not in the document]
robnagler: More context, and a better explanation/solution. 

This is sirepo.jupyterhub.Authenticator, which is being called by the JupyterHub process in a thread in a running JupyterHub process, which is the process that needs to be properly configured, that is, the sirepo_jupyterhub service. This process is configured by conf.py file, which is generated by rsconf. I think, actually, the uri (provided by rsconf) is supplied directly as a "Traitlet" called, say, sirepo_uri in jupyterhub.SirepoAuthenticator (the class needs to be renamed to be settable in conf.py). rsconf will need to be modified to provide this, and jupyterhub_conf.py.jinja  needs to be updated and service.py needs to be updated to supply the value for the jinja.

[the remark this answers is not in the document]
e-carlin: Look right 78abea726 ?

sirepo/pkcli/service.py
```diff
@@ -135,7 +136,11 @@ def jupyterhub():
         f = d.join('conf.py')
         pkjinja.render_resource(
             'jupyterhub_conf.py',
-            PKDict(_cfg()).pkupdate(**sirepo.sim_api.jupyterhublogin.cfg),
+            PKDict(_cfg()).pkupdate(
+                # POSIT: Running with nginx and uwsgi
+                sirepo_uri=f'http://{socket.getfqdn()}:{_cfg().nginx_proxy_port}',
```
robnagler, line 141: Since this is only for development, this is ok. You might actually want to hardwire to 127.0.0.1?

sirepo/jupyterhub.py
```diff
@@ -56,16 +55,10 @@ async def refresh_user(self, user, handler=None):
 
     def _check_permissions(self, handler):
         r = requests.post(
-            _cfg.sirepo_uri + sirepo.simulation_db.SCHEMA_COMMON.route.checkAuthJupyterhub,
+            self.sirepo_uri + sirepo.simulation_db.SCHEMA_COMMON.route.checkAuthJupyterhub,
```
robnagler, line 58: Seems like `+` is a bit too loose. If there is no `/`, the URL is broken. Is there a guarantee on these routes that they begin with `/`. Also, is there a guarantee there's no params? Seems like sirepo_uri should be the complete uri, and that's configured by rsconf. Then you wouldn't need to have simulation_db afaict.

It looks like most of the import sirepo.* are unnecessary now. This would fully decouple sirepo/jupyterhub.py.

robnagler (changes requested):

sirepo/jupyterhub.py
e-carlin: Added a POSIT for no params and fixed imports 8f4e58209

SCHEMA_COMMON.route is [guaranteed to start with a slash](https://github.com/radiasoft/sirepo/blob/78abea726afdd196a064a36d87dacf5c52bc0b0a/sirepo/uri_router.py#L316). 

I don't like having the complete uri configured by rsconf. Right now if you search for the word `checkAuthJupyterhub` you can see it in the route definition in schema_common, the `def api_checkAuthJupyterhub, and here. If we move to rsconf I think we lose context about how this all works. Someone would have to know to check rsconf (maybe they should since all cfg things get configured by rsconf but I don't think that is established knowledge on the team) . If someone were to change the route (or add params to the route) then they'd also have to fix rsconf. If it is just here then they at least stand a chance of fixing all of the places the route is used.

sirepo/pkcli/service.py
e-carlin: 127.0.0.1 doesn't work because then `flask.url_for('_dispatch_empty', _external=external)` returns 127.0.0.1 which messes up the redirect to the login pages.

sirepo/jupyterhub.py
robnagler: Fair enough.

robnagler (approved):
I think this is good to go. Is there a migration for users? Do you want to push to alpha?

e-carlin:
Yes a migration needs to be written, thanks for the reminder. I'll add that then we can merge.

e-carlin:
Thinking about this more, no migration is needed. User's who are in the approved community pool already have the sim_type_jupyterhublogin role. So, things will continue to work like normal for them. User's who are in the denied list (not actually used anywhere  currently but we kept it around) don't have any records in the db so we have no way of finding a uid for them and setting them as denied in user_role_invite_t. So, I think we're good to go. Please confirm and I'll merge.

robnagler:
Did you run a check for the denied list? I think we should run this check: validate that only users in valid pools (uspas, etc.) have the role. Any users not in these pools should get role removed.

We need to migrate from the community pool to the everybody pool once moderation is released to production. We still need pools for uspas, etc.

e-carlin:
@robnagler I've written a script (hand/check-jupyter-users/check-jupyter-users.sh) to do the validation. You can run it as is (only does reads currently). I left a TODO with a question at the top. If you could answer as well as do a full review of the script (I'd really like to improve my bash) I'd appreciate it.

PR #4224, opened 2022-03-09
fix #4106: proprietary oauth

### LINKED ISSUE #4106: Authorize codes via OAuth with expiry
Allow code roles to be authorized via OAuth. Sirepo will ask the OAuth server to validate user credentials. The OAuth server will return an access token with an expiry. 

The OAuth handoff will need to describe where it is going so that needs to be configurable on a per code basis.

Roles need an expiry and a description of an authentication method. I don't now that we can reuse sirepo.auth for this, but this is so specific that I think we can just refactor sirepo.auth.github module to support this workflow.

rorour:
This branch is based on 3654-moderate-jupyterhub-user-creations. Please review and merge #4211 first.


e-carlin:
Blocked waiting on flash center to allow our non-tls secured dev endpoint to get a response. Then need to finish https://github.com/radiasoft/sirepo/pull/4224/files#diff-6872e86bbaf9783b9521a2d7bc8efbc06219ec4d6e34d64031682626e3a13ac7R24

e-carlin:
@robnagler this is working. I'd like another review because of the security impact. 

One decision we made along the way was to not refresh tokens. Instead just have users log in again once expired. For flash the tokens expire after one hour so this will lead to an annoying ui (need to log in to flash center every hour). Should we implement using refresh tokens or go with just re-login?

robnagler:
Yes implement refresh tokens. It's unfortunately that they didn't set the default to something higher. First find out how many refreshes you get. Also, a refresh only works, I believe, if the user is active. Therefore, even if we implement refreshes, if they close their window and open it again, it might require a relogin

e-carlin:
Entering invalid credentials and clicking the "deny access" button on the flash login page leads to what looks like an approved authentication response (I would expect the ui to say username/password incorrect or at the very least a response with a payload saying the request was not approved). I've reached out to the flash folks and I'm waiting on a response. This blocks the release.

e-carlin:
@robnagler I think this is ready to merge. Please give it another review.

sirepo/auth/__init__.py
```diff
@@ -135,6 +135,27 @@ def complete_registration(name=None):
...
 
 
+def control_sim_type_role(sim_type):
+    from sirepo import oauth
+    from sirepo import auth_role_moderation
+
+    t = sirepo.template.assert_sim_type(sim_type)
+    if t not in sirepo.feature_config.auth_controlled_sim_types():
+        return
+#QUESTION(robnagler) I think this is sufficient, that is, the tests can be reversed
```
robnagler, line 145: This is the only thing I'm concerned about. LMK what you think.

sirepo/auth_role_moderation.py
```diff
@@ -142,8 +141,32 @@ def _send_request_email(info):
         return self.reply_ok()
 
 
+def control_for_user(uid, role):
```
robnagler, line 144: possibly `check_user()`?

sirepo/auth/__init__.py
```diff
@@ -135,6 +135,27 @@ def complete_registration(name=None):
     cookie.set_value(_COOKIE_STATE, _STATE_LOGGED_IN)
 
 
+def control_sim_type_role(sim_type):
```
robnagler, line 138: possibly `check_sim_type()`

robnagler (changes requested):
It was easier to edit. I think the code is right. I removed some proprietary information.

sirepo/auth/__init__.py
e-carlin: That doesn't workin in the `api_checkAuthJupyterhub` case. That API is `allow_visitor` which means that this check is always True and we return when we should really be prompting to moderate. 

Thoughts ccaa8f691 ? I don't love adding another api_perm and I struggled to find a name for it...

sirepo/auth_role_moderation.py
e-carlin: 4980a3bd5 I think it is fine. I prefixed it with raise_ since the other calls in that method are also prefixed with raise_

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -66,21 +67,18 @@ def _res_for_uri(uri):
...
-        return self.call_api(
-            'authGithubLogin',
-            kwargs=PKDict(simulation_type='jupyterhublogin'),
-        )
+        sirepo.oauth.raise_authorize_redirect('jupyterhublogin', github_auth=True)
 
     @sirepo.api_perm.require_user
     def api_redirectJupyterHub(self):
-        sirepo.auth.require_sim_type('jupyterhublogin')
+        sirepo.auth.check_sim_type_role('jupyterhublogin')
```
robnagler, line 81: `self.parse_params(req_data=PKDict(simulationType='jupyterhublogin'))`

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -33,15 +34,15 @@
...
+    @sirepo.api_perm.manual_permission_check
     def api_checkAuthJupyterhub(self):
         def _res_for_uri(uri):
             return self.reply_ok(PKDict(uri=uri))
 
         u = None
         try:
             sirepo.auth.require_user()
-            sirepo.auth.require_sim_type('jupyterhublogin')
+            sirepo.auth.check_sim_type_role('jupyterhublogin')
```
robnagler, line 45: I think this is a better method to use:

`self.parse_params(type='jupyterhublogin')`

This is part of our problem. I think if this were written with a, say, sim_type parameter like other calls e.g. downloadFile, then you wouldn't need to hardwire the type, which probably should come from a constant in the module, just for cleanliness.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -33,15 +34,15 @@
 
 
 class API(sirepo.api.Base):
-    @sirepo.api_perm.allow_visitor
+    @sirepo.api_perm.manual_permission_check
```
robnagler, line 37: What I think is going on is that this function is doing too much. It should be `@sirepo.api_perm.require_user` which will throw an exception, and then the code in sirepo.jupyterhub would do the parsing of the response. It's all inside the sirepo repo. Then you wouldn't have to catch anything in this module. All the exceptions would cause a Response to go back to the sirepo.jupyterhub.authenticate. The fall through case (creating the juptyerhub user) would also do the right thing.

sirepo/auth_role_moderation.py
```diff
@@ -142,8 +141,32 @@ def _send_request_email(info):
         return self.reply_ok()
 
 
+def raise_control_for_user(uid, role):
+    s = sirepo.auth_db.UserRoleInvite.get_status(uid, role)
+    if s in _ACTIVE:
+        raise sirepo.util.SRException('moderationPending', None)
+    if s == 'denied':
```
robnagler, line 148: Use constant

sirepo/feature_config.py
```diff
@@ -87,6 +90,17 @@ def for_sim_type(sim_type):
...
+
+    Granted access can be through oauth or manual management of the role
+
+    Returns:
+      frozenset:  enabled sim types that require role
+    """
+    return frozenset(
+        cfg().proprietary_sim_types.union(cfg().proprietary_oauth_sim_types),
+    )
 def _data_dir(value):
```
robnagler, line 104: two blank lines

sirepo/feature_config.py
```diff
@@ -119,7 +133,16 @@ def b(msg, dev=False):
             tuple,
             'Names of root packages that should be checked for codes and resources. Order is important, the first package with a matching code/resource will be used. sirepo added automatically.',
         ),
-        proprietary_sim_types=(set(), set, 'codes that require authorization'),
+        proprietary_sim_types=(
+            set(),
```
robnagler, line 137: frozenset()

sirepo/feature_config.py
```diff
@@ -119,7 +133,16 @@ def b(msg, dev=False):
...
+            set,
+            'codes that contain proprietary information and authorization to use is granted manually',
+        ),
+        proprietary_oauth_sim_types=(
+            set(),
```
robnagler, line 142: frozenset()

sirepo/sim_api/jupyterhublogin.py
e-carlin: How will sirepo.jupyterhub build the urls? We have access to self.sirepo_uri but all the methods like uri_for_api and local_route access the flask object directly which we can't do because we are outside the flask context.

robnagler: The schema is available, but not quite. It's what's necessary to map URIs from APIs. For now, these mappings could be hardwired in sirepo.jupyterhub, since we need to get this PR out. 

I don't like adding the new API permission. Another approach is to use self.call_api and have a "hidden" api that api_checkAuthJupyterhub calls and wraps in a proper response. One things we could do to fix the SRException mess is to include a header in the response (X-Sirepo-SRException) that contains a serialized version of the exception so the HTML doesn't need to be parsed (which isn't terrible, and could be better encapsulated).

Anyway, I prefer the first approach and to get to work pulling apart the request dependent (external) from the static stuff (schema and args).

e-carlin: 890681e6a

sirepo/auth_role_moderation.py
e-carlin: 890681e6a

sirepo/feature_config.py
e-carlin: b824893ad

sirepo/feature_config.py
e-carlin: b824893ad

sirepo/feature_config.py
e-carlin: b824893ad

sirepo/sim_api/jupyterhublogin.py
e-carlin: Thoughts on adding sirepo_uri? Then we can re-use the code without adding another place where we handle creation of uris

sirepo/sim_api/jupyterhublogin.py
e-carlin: 890681e6a4c0784a71559f1a88c2d2665ec78e19

sirepo/sim_api/jupyterhublogin.py
e-carlin: 890681e6a4c0784a71559f1a88c2d2665ec78e19

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -66,21 +67,18 @@ def _res_for_uri(uri):
 
     @sirepo.api_perm.require_user
     def api_migrateJupyterhub(self):
-        sirepo.auth.require_sim_type('jupyterhublogin')
+        sirepo.auth.check_sim_type_role('jupyterhublogin')
```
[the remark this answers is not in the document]
e-carlin: 890681e6a4c0784a71559f1a88c2d2665ec78e19

sirepo/jupyterhub.py
```diff
@@ -32,33 +35,60 @@ def __init__(self, *args, **kwargs):
...
-            handler.set_cookie(k, v)
-        return PKDict(r.json())
+        _maybe_html_redirect(r)
+        res = PKDict(r.json())
+        _maybe_srexception_redirect(res)
+        return res
+
+    def _redirect(self, handler, uri):
+        handler.redirect(uri)
+        raise tornado.web.Finish()
```
robnagler, line 94: I hadn't seen that before. Useful for our code.

sirepo/jupyterhub.py
```diff
@@ -32,33 +35,60 @@ def __init__(self, *args, **kwargs):
...
 
     def _check_permissions(self, handler):
+        def _cookies(response):
+            for k, v in response.cookies.get_dict().items():
+                handler.set_cookie(k, v)
+
+
+        def _maybe_html_redirect(response):
+            m = re.search(r'window.location = "(.*)"', pkcompat.from_bytes(response.content))
+            m and self._redirect(handler, f'{self.sirepo_uri}{m.group(1)}')
```
robnagler, line 60: Does there need to be a sirepo_uri? Doesn't the redirect just work on `/` without http://....?

robnagler (commented):
See question about sirepo_uri.

robnagler (approved):

robnagler:
@e-carlin I'll let you merge

PR #4227, opened 2022-03-10
fix #4070: add slack link

### LINKED ISSUE #4070: Add a Slack link/button to apps
We want to encourage people to join Slack. We want to add a prominent link/button somewhere in the app.

rorour:


robnagler:
I would like this more prominent. Please use the slack logo:

Slack_Monochrome_Black

Put it in the app header and make it gray to match existing colors. Have a tooltip for when people hover over it, and also change color similar to Notes link. 

Include the uid in the link `https://slack.sirepo.com/v1/<uid>`. In the login app header, use login for uid.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2402,6 +2402,11 @@ SIREPO.app.directive('appHeaderRight', function(appDataService, authState, appSt
                       '<li data-help-link="helpUserManualURL" data-title="User Manual", data-icon="list-alt"></li>',
                       '<li data-help-link="helpUserForumURL" data-title="User Forum", data-icon="globe"></li>',
                       '<li data-help-link="helpVideoURL" data-title="Instructional Video" data-icon="film"></li>',
+                      '<li>',
+                        '<a href="{{ slackURL }}" target="_blank">',
+                          '<span data-ng-class="glyphicon"><img width="14" src="/static/img/slack.png" /></span> Sirepo Slack',
+                        '</a>',
+                      '</li>',
```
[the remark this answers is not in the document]
robnagler: @mkeilman please review https://github.com/radiasoft/sirepo/pull/4229. The perl code in #4228 can be put in a file and run this way (on the code in #4229):

```sh
perl -pi normalize.pl sirepo-components.js
```

To test a new file, e.g. sirepo.js, you run:

```sh
perl -n normalize.pl sirepo.js
```

This will print errors and the line number in sirepo.js where they occurred. You would then fix up those template errors so that the conversion can proceed automatically. Note the changes that are in #4229, e.g. blank lines and missing commas, because you'll run into those in other js files. Once the program runs without output, you can add that file to this pull request.

sirepo/auth/__init__.py
```diff
@@ -579,6 +579,9 @@ def _auth_hook_from_header(values):
 
 
 def _auth_state():
+    def get_slack_uri():
+        return sirepo.feature_config.cfg().sim_common.join_slack_uri \
+               + (logged_in_user() if _is_logged_in() else 'login')
```
robnagler, line 584: Let's just not append anything in the case of a guest. You should use `_get_user` for this:

```py
               + (_get_user() or '')
```

We want to track the user whether they are logged in or not.

sirepo/feature_config.py
```diff
@@ -130,6 +130,8 @@ def b(msg, dev=False):
         ),
         sim_common=dict(
             hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
+            join_slack_uri=('https://slack.sirepo.com/v1/', str,
```
robnagler, line 133: Let's not expose this in the source code. Rather, just have `https://slack.com/` as the value in the default configuration. We'll set the configuration on production.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -2423,6 +2450,9 @@ SIREPO.app.directive('appHeaderRight', function(appDataService, authState, appSt
         controller: function($scope, stringsService) {
             $scope.authState = authState;
 
+            $scope.getSlackLink = function() {
```
robnagler, line 2453: I don't think this is necessary. You should be able to refer to `$scope.authState.slackUri`. 

Also, let's not change the name, i.e. it should be slackUri everywhere.

robnagler (changes requested):
Much better, thanks. I think this needs to be merged with master, because there are conflicts.

sirepo/package_data/static/js/sirepo-components.js
rorour: 1e63e3e61d6b75f5c183715519625d2fdca71b42

sirepo/feature_config.py
rorour: 1e63e3e61d6b75f5c183715519625d2fdca71b42

sirepo/auth/__init__.py
rorour: 1e63e3e61d6b75f5c183715519625d2fdca71b42

robnagler (changes requested):
I tested it, and it looks great, thanks. Just some minor nits about naming and formatting.

sirepo/auth/__init__.py
```diff
@@ -579,6 +579,9 @@ def _auth_hook_from_header(values):
 
 
 def _auth_state():
+    def get_slack_uri():
```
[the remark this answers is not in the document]
rorour: e05007898d8a512623cc3c82eb984625bf68b51c

sirepo/feature_config.py
```diff
@@ -130,6 +130,8 @@ def b(msg, dev=False):
         ),
         sim_common=dict(
             hide_guest_warning=b('Hide the guest warning in the UI', dev=True),
+            join_slack_uri=('https://slack.com/', str,
```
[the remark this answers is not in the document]
rorour: e05007898d8a512623cc3c82eb984625bf68b51c

robnagler (approved):

PR #4229, opened 2022-03-10
Fix #4228 add commas and joins

### LINKED ISSUE #4228: Normalize syntax of angular templates so they can be converted to template literals
Correcting errors for translating templates automatically with:

```perl
if (s{^(\s*template:\s*)\[\s*$}{$1`\n}) {
    $i = 1;
}
elsif (!$i) {
    # pass
}
elsif (s{^(\s*)\]\.join\(''\)}{$1`}) {
    $i = 0;
}
elsif (m{^\s*//.*}) {
    $_ = '';
}
elsif (s{^(\s*)'(.*)',\s*$}{$1$2\n}) {
    # pass
}
elsif (s{^(\s*)(\w+.*),\s*$}{$1\$\{$2\}\n}) {
    # pass
}
else {
    chomp;
    die($_);
}
```

robnagler:


e-carlin:
I thought we were switching to es6 template strings instead of joins?

mkeilman:
> I thought we were switching to es6 template strings instead of joins?

The perl script does the conversion, modulo normalizing the existing joins

robnagler:
Yes, but to do that automatically, the templates need to be properly formatted, and these had bugs, which were detected by the script (default deny).

This is one of the big reasons why coding style rigidity is important (to me). We can write a few lines of regexes and convert the code to something unanticipated.

robnagler:
And @mkeilman you are correct, my "normalize.pl" was really "convert.pl" and this is "normalize templates". So much for naming. :)

mkeilman:
Note a couple of manual fixes in ```light.js``` and ```radia.js```

mkeilman (approved):
Original changes are fine - I added the files tagged by the perl script so everything should be ready for conversion

PR #4235, opened 2022-03-14
Fix #4159 download data file

### LINKED ISSUE #4159: downloadDataFile should not wait for analysis
If a report runs too long (#4156), the user may want to request the download. However, even closing the report doesn't seem to cancel the analysis of job srw beamlimAnimation. I'm not sure downloadDataFile should wait at all. It's just I/O, and it probably should be it's own job class (also see #4157).

mkeilman:
Starting from the suggestion of a new job class, this seems to be a minimal way to grab the data while a simulation is running.  I presume there are pitfalls I have overlooked


robnagler:
I forgot why we said to create a new OP (it's not a job class/category, jobs are executed by OP_RUN). Could you explain a little more of your thinking?

mkeilman:
I thought OPs were used to distinguish between types of "jobs", like analysis vs. compute/run. Since downloadDataFIle was using OP.ANALYSIS it was getting caught "waiting for another simulation to complete" in ```job_driver._slots_ready()``` (via ```prepare_send()```). A new op skips that, but we still need an agent etc.  It works in the sense that it gets the data file without waiting for or interrupting the simulation. I don't know if it works in a conceptual sense.

There's nothing inherently "IO" about it - I used that name based on the issue comment, and maybe it's right for "thing to get data but is neither computation nor analysis".

robnagler:
Analysis is a job in the sense that it has a job id, but it's not a compute job. 

Using an op to +1 a queue length is a problem. I think we should allow multiple analyses. The GUI only asks for one at a time so this should work in this case, because it could treat download differently (jump to head of analysis queue).

Let's talk about this tomorrow.

mkeilman:
note: the downloadDataFile links in the app do not use the client-side queue but just invoke the url directly. As part of this or another issue we'll need an "IO" queue that mirrors the transient and persistent queues

robnagler:
We might be able to get away with just letting the server do the queueing. There aren't a lot of simultaneous download datafile operations since I believe they are mostly user initiated.

mkeilman:
Tested against local and sbatch, ready for another look

sirepo/job_driver/__init__.py
```diff
@@ -352,6 +352,8 @@ async def _slots_ready(self, op):
             return
         await op.op_slot.alloc('Waiting for another simulation to complete')
         await op.run_dir_slot.alloc('Waiting for access to simulation state')
+        if n in (job.OP_IO,):
```
e-carlin, line 355: This is right. But I think we now have to change the code for [`destroy_op`](https://github.com/radiasoft/sirepo/blob/4159-download-data-file/sirepo/job_driver/__init__.py#L98). We now want to only free a cpu_slot for certain types of OPS_THAT_NEED_SLOTS (probably need a better name for that too since there are now ops that need an op slot but not a cpu slot).

Hypothetical scenario with a configuration of only one cpu_slot that shows the problem:
1. run starts (job.OP_RUN so it acquires a cpu_slot)
2. io starts (job.OP_IO so it doesn't acquire a cpu_slot)
3.  io finishes, destroy_op is called, it frees the cpu_slot (the one job.OP_RUN acquired), but the run is still running
4. Another run comes in, it is able to acquire a cpu_slot, the machine is now oversubscribed (2 runs happening at the same time when there should only be one). 

I think encapsulating behaviors like this in a test would be nice. Configure the system to only have one cpu_slot. In a thread start an op_run that runs forever. Wait for it to move to running. In another thread start an op_run. Wait for it to become pending. In a third thread start an op_io. Wait for the op_io to finish. The second op_run should still be pending. With the code as it currently I think it will transition to running.

While working on this code I think refactoring 'def _slots_ready()` and some of the constants could be helpful. I think using more named constants and inner functions to separate out the different cases (ops that don't need any slots, sbatch login, ops that need an op slot but no cpu slot) would make this code clearer.

e-carlin (commented):
Looks good. One gotcha I left a comment about.

mkeilman:
This needs cleanup but I thought I'd get some feedback. Probably a bit overdone. Some notes:

- I structured the test this way so I could have any number of threads
- It seems like one needs to run different simulations in order to get to a pending state - if I run multiple 'beamlineAnimation' they either all run or look like the same request, not sure which
- Freeing the IO op slot does not cause the pending job to become running, so maybe that's working OK?
- The SIREPO_JOB_DRIVER_LOCAL_SLOTS_SEQUENTIAL/PARALLEL configuration appears to have no effect on the results. I left some code commented to see if I am doing it right

Anyway, comments awaited

sirepo/job_driver/__init__.py
e-carlin: @mkeilman has this been addressed?

mkeilman: Not quite - see my last comment

mkeilman:
Pinging for another look - see [my previous comment](https://github.com/radiasoft/sirepo/pull/4235#issuecomment-1093439189) for context. I added a little naming around cpu vs non-cpu slots but it is clunky. One note about ```test_srw_io```: the file download fails with a 404. It's not really relevant to the test but I don't know whether it is expected to fail in the test environment. The whole test could use a sanity check...

robnagler:
I don't think any tests should fail. I'll have a look tomorrow.

tests/job_concurrency1_test.py
```diff
@@ -9,9 +9,16 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 
 """
-from __future__ import absolute_import, division, print_function
 import pytest
 
+import sirepo.util
+
+def setup_module(module):
```
robnagler, line 16: setup_module means this should probably be a separate test file. The other cases will be affected by this.

More importantly, I don't think the new case is hitting the asynchrony problems. We want to test whether IO gets through when an Analysis is running. That's not what the case is test.

I'm also concerned about `fc` being used in multiple threads. I don't know enough about how flask client is constructed. It may contain state related to the request. I think for asynchrony we need a full stack requester going where there are multiple threads making IO requests. I think that's beyond the scope of this change so I think this test case can be deleted.

robnagler (changes requested):

sirepo/job_driver/__init__.py
```diff
@@ -36,7 +36,11 @@
 
 cfg = None
 
-OPS_THAT_NEED_SLOTS = frozenset((job.OP_ANALYSIS, job.OP_RUN))
+OPS_THAT_NEED_CPU_SLOTS = frozenset((job.OP_ANALYSIS, job.OP_RUN,))
+
+OPS_THAT_NEED_OTHER_SLOTS = frozenset((job.OP_IO,))
+
+OPS_THAT_NEED_SLOTS = frozenset().union(*[OPS_THAT_NEED_CPU_SLOTS, OPS_THAT_NEED_OTHER_SLOTS,])
```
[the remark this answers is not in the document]
mkeilman: frozenset does not support '+' or 'add' so I had to hunt down the union thing

sirepo/job_driver/__init__.py
```diff
@@ -343,7 +347,7 @@ def _receive_error(self, msg):
     async def _slots_ready(self, op):
         """Only one op of each type allowed"""
         n = op.opName
-        if n in (job.OP_CANCEL, job.OP_KILL):
+        if n in (job.OP_CANCEL, job.OP_KILL,):
```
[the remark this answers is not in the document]
mkeilman: That will leave ```OP_SBATCH_LOGIN``` out

mkeilman:
Ping for another look. I'd also like confirmation that it's actually solving the problem, maybe once it is on alpha.

robnagler:
I'm pretty sure it solves the problem.

tests/job_concurrency1_test.py
robnagler: @e-carlin Is the setup_module necessary for all cases? If so, that would be a separate issue/PR.

> Should we make an issue for the requester you envision, since I think the idea is that we want all our requests to eventually be asynchronous?

Yes, good idea. I started coding and then I deleted it. I think we need to have a lot of requests to the same simulation going that do different things, and ensure they all complete. I also want there to be some meta way about speaking about this similar to animation_test and report_test. I found the job_concurrency1_test too noisy.

I want to get the async tester going this month. This may address the issue naturally, because it will extensively test asynchrony.

robnagler (approved):
Let's see how it goes on alpha.

PR #4250, opened 2022-03-22
Fix #4240 - use gui in cfg instead of sim_common

### LINKED ISSUE #4240: feature_config.sim_common?
Just noticed this. Is sim_common simulation only or common to everything. If it is everything, it should just be called common. If it is simulation only, then the two items in it should get deprecated and moved to common.

Also I don't think we need `sim_common()` as `cfg()`.

mkeilman:
Also eliminates the superfluous ```sim_common()``` method


sirepo/feature_config.py
```diff
@@ -82,17 +82,7 @@ def for_sim_type(sim_type):
...
-
-def sim_common():
-    """Get cfg to use across all simulations. Separate from global cfg
-
-    Returns:
-        dict: common sim config
-    """
-    import pykern.pkcollections
-    return pykern.pkcollections.PKDict(cfg()).sim_common
+    ).pkupdate(c.gui)
```
robnagler, line 85: This is an interface change. Is this in use anywhere else?

sirepo/feature_config.py
```diff
@@ -116,6 +106,10 @@ def b(msg, dev=False):
         # No secrets should be stored here (see sirepo.job.agent_env)
         api_modules=((), set, 'optional api modules, e.g. status'),
         default_proprietary_sim_types=(set(), set, 'codes where all users are authorized by default but that authorization can be revoked'),
+        gui=dict(
```
robnagler, line 109: slack_uri does not belong in gui, because it is only used server-side (like api_modules, etc.). 

Now that I see what this looks like, I would call this schema_common just to be clear that this gets added to the schema.

You need to ReplacedBy sim_common.hide_guest_warning look in pkconfig for how to do that.

robnagler (changes requested):

sirepo/feature_config.py
mkeilman: Yes - should simulation_db go back to using a separate call then?

robnagler: No, it looks to me that raydata is not using it right. I'm not sure why it isn't using schema. The raydata use doesn't conflict so should be fine to leave as is.

robnagler (approved):
That was easy. ;-)

PR #4259, opened 2022-03-25
Fix #4254 eval rpns prior to particle gen

### LINKED ISSUE #4254: create_particle_file() needs to evaluate RPN Values
In madx, the bunch distribution fields may be RPN Values rather than floats. For example, If you run a simulation which has an expression for SIGT, the simulation fails with "unsupported operand type(s) for ** or pow(): 'str' and 'int'".

gurhar1133:
Codevar now evals the fields that UI was letting user enter as expression but was not being evaluated on the backend


gurhar1133:
Made the change, also all tests pass

PR #4270, opened 2022-03-29
Fix #3790: remove hellweg wip

### LINKED ISSUE #3790: Remove hellweg app code
The hellweg app contains deprecated code which can no longer be tested against an executable. The app logic should be removed from sirepo.

gurhar1133:
Hellweg code removed. 
```grep -r 'hellweg' ```
and
```find . | grep hellweg```
confirm this.

Anything else I should do to confirm?
Shims needed? @robnagler


moellep:
Looks good to me

e-carlin:
FWIW you can use `find . -iname '*hellweg*'

PR #4272, opened 2022-03-29
Fix #4265 eval pow before sim

### LINKED ISSUE #4265: Code_var needs to evaluated pow() expressions before handing them to in.madx
Madx conversion between other codes as well as editing variables in source tab of madx handles pow() expressions (which are not native to madx). But, if a user edits the source tab using a pow() expression and then tries to run the sim it will not be evaluated properly before being sent off to in.madx.

gurhar1133:
If a user edits bunch and commands in source tab as expressions involving pow (which is not supported by madx), those expressions will now be evaluated prior to setting them in in.madx

### DIFF
--- a/tests/template/code_variable_test.py
+++ b/tests/template/code_variable_test.py
@@ -44,6 +44,9 @@ def test_cache():
                         ),
                     ],
                     commands=[],
+                    bunch=PKDict(
+                        energy='x + 2 * y',
+                    )
                 )
             ),
             # schema
@@ -52,6 +55,9 @@ def test_cache():
                     point=PKDict(
                         p1=["P1", "RPNValue", 0],
                         p2=["P2", "RPNValue", 0],
+                    ),
+                    bunch=PKDict(
+                        energy=['Energy', 'RPNValue', 0],
                     )
                 ),
             ),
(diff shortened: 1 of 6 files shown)


moellep:
I'm not sure if it is related, or a different bug, but when I export this opal simulation as madx, the Q:1 quadurpole has a length with a pow() expression, the length value is being truncated.

https://alpha.sirepo.com/opal#/lattice/nDgEioZU

You can bring it into your dev environment by exporting the zip from the simulation, and then importing the zip into your local sirepo/OPAL server.

gurhar1133:
> 

This does seem like a tangent edge case. The way I solved the conversion problem calls _format_rpn_value() on the rpn values, but a pow expression could be in bunch, commands or in a Quad definition. I will try to cover this case in this issue as well

gurhar1133:
Changes made @moellep and opal to madx conversion tests are passing.

tests/template/code_variable_test.py
```diff
@@ -44,6 +44,9 @@ def test_cache():
                         ),
                     ],
                     commands=[],
+                    bunch=PKDict(
```
moellep, line 47: We should remove the changes to this test because the lattice ModelIterator only looks inside elements and commands values, so the top-level bunch expression would never be visited.

gurhar1133:
Changes made @moellep

PR #4274, opened 2022-03-30
4241 opal dipole orientation

gurhar1133:
Now e2 is no longer an input and is determined by e2 = angle - e1 (as per [opal docs](http://amas.web.psi.ch/opal/Documentation/2021.1/#sec.elements.bend) )
This results in correct orientation of opal dipoles


moellep:
Looks good!

Issue #3896, opened 2021-10-18
sirepo.lib handling of import resource files


cchall:
Some of the current handling of how simulation resource files (field maps, distributions, etc. listed in the input file being parsed) causes problems for common workflows in rsopt. The two main issues are:
- sirepo.lib checks to make sure that resource files exist when the simulation input file is parsed. In many cases such resources are generated in the middle of a chain of simulations and so they do not exist at parsing. One possibility is to include an optional argument for `parse_file` that allows passing file names that should not be validated.
- It seems that resources are assumed to be in the same directory as the parsed input file. However, it is often useful to keep such files in a separate directory. Keeping the full file path in the parsed input file through the validation process and then discarding the path only when `write_files` is called might be the ideal behavior to have.

robnagler:
@cchall Could you give a test case?

cchall:
Sure, here are three test cases. For each one there are two resource files needed to run the simulation: wake.sdds and distribution.sdds.
[sirepo_lib_pathing.zip](https://github.com/radiasoft/sirepo/files/8372575/sirepo_lib_pathing.zip)

### 01_missing_resources: 
Resource files are not present but are assumed to be created after `parse_file` and `write_files` have been called.

### 02_resource_pathing/elegant: 

Resource files are in directory './distr' which is included directly in the file name in the input files. Currently this leads to an assertionerror: 
```
elegant$ python run.py 
Traceback (most recent call last):
  File "run.py", line 3, in <module>
    m = lib.Importer('elegant').parse_file('elegant.ele')
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/sirepo/lib.py", line 111, in parse_file
    self.__adapter.parse_file(p),
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/sirepo/template/elegant.py", line 119, in parse_file
    _verify_files(i, i.type)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/sirepo/template/elegant.py", line 106, in _verify_files
    _input_files(model_type),
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/sirepo/lib.py", line 70, in _verify_files
    f'file={f} must be a simple name'
AssertionError: file=distr/wake.sdds must be a simple name
```
Ideally the files should be validated at parsing and then have the path information stripped (along with appropriate modification in the input files) when `write_files` is called.

[...]

PR #4280, opened 2022-04-01
4275 latticeparser look into super

gurhar1133:
Now using UpdateIterator to update commands and to downcase vars
lib_test.py is the only test that doesnt pass, but that test is broken and will be addressed in [#3896](https://github.com/radiasoft/sirepo/issues/3896)


moellep:
Looks good!

PR #4288, opened 2022-04-05
Fix #4099 radia param change viz stall

### LINKED ISSUE #4099: radia: visualization intermittently stalls after parameter changes
These cases show no errors in the logs, and usually clear up after a browser refresh

mkeilman:


mkeilman:
Notes:

- The stall was caused by a race condition between ```panelState.requestData()``` (added for #4080) and saves of geometryReport. Adding a timestamp allows the report to get updated properly without extra data fetches
- Simulation model change handling moved to viewLogic
- Kick maps can be enabled for imported models
- Other minor refactoring and cleanup

mkeilman:
Added a fix for #3521 as long as I was in the neighborhood

PR #4289, opened 2022-04-05
3896 sirepo lib resources

gurhar1133:
* sirepo.lib now no longer updates_filenames
* ingore_files arr specifies files to ignore
* lib_tests added


sirepo/foo
```diff
@@ -0,0 +1 @@
+auth_db.py
```
robnagler, line 1: accidental commit

sirepo/lib.py
```diff
@@ -7,6 +7,7 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 from __future__ import absolute_import, division, print_function
+from codecs import ignore_errors
```
robnagler, line 10: not used

sirepo/lib.py
```diff
@@ -99,10 +104,10 @@ def util(self):
 
 class Importer:
 
-    def __init__(self, sim_type):
+    def __init__(self, sim_type, ignore_files=None):
```
robnagler, line 107: Add some documentation. This is a public interface.

robnagler (changes requested):
Looks good. Minor changes

gurhar1133:
Changes made @robnagler

sirepo/lib.py
```diff
@@ -103,7 +102,12 @@ def util(self):
 
 
 class Importer:
-
+    """
```
robnagler, line 105: Needs a summary string of what the function does. Make the Args: lines one line long and shorter with defaults. Here's an example: https://github.com/radiasoft/pykern/blob/master/pykern/pkconfig.py#L273

robnagler (changes requested):

sirepo/lib.py
gurhar1133: Sure, but what do you mean 'with defaults'.

gurhar1133: And is "Uses LibAdapter to parse and import sim files" a sufficient summary of the class?

robnagler: A "what" is better than a "how". Using LibAdapter is "how".

`Imports a code's native files into Sirepo representation.`

robnagler: Here's an example with defaults: https://github.com/radiasoft/pykern/blob/master/pykern/pkyaml.py#L15

gurhar1133: Ok, clear now. thank you :)

gurhar1133:
How about now?

robnagler (approved):

PR #4302, opened 2022-04-08
Fix #4300 - vtk updated

### LINKED ISSUE #4300: Update and test latest vtk.js
This will be a recurring task as vtk releases new versions quite a bit - several a day is not unusual. As of this writing they are up to 24.4.1 (I think we are using 11.x). My current plan is to update only for "x" in "x.y.z"  unless there is a feature or bug fix that specifically addresses a problem we are having.

mkeilman:
Tested against all extant apps that use vtk.  This also replaces ```getOpenGLRenderWindow``` with ```getApiSpecificRenderWindow``` (a vtk change from last year)


robnagler (approved):

e-carlin:
If only upgrading Angular was this simple!

PR #4325, opened 2022-04-13
Fix #4188: collapse and search elems

### LINKED ISSUE #4188: "collapse all" and more robust search for beamline elements
Sirepo Elegant lattice view
Beamline Elements section (bottom right)

It looks like the beamline elements section loads with every section expanded by default, so it is somewhat difficult to navigate initially for simulations with a lot of elements. Having it load collapsed by default (or with the collapsed categories of the previous state) would be nice. A "collapse all" and "expand all" would be nice as well.

Doing a browser text "find" on the page for a name in a collapsed category also doesn't open the category to the found beamline element, meaning I can't do a page search to find the element, even though it does report the text being found on the page.

I would also be all for a dynamic search box to filter elements by pattern matching above the elements

gurhar1133:
* Beamline elements are now searchable on the lattice tab.
* There is also a collapse/expand all elements option
* Categories that have been expanded and collapsed now get saved so that refreshing/revisiting the sim will not forget which element categories were expanded
* The searchbar is sticky at the top of the beamline elements panel so you don't lose it when you scroll
* Found items are scrolled into view and highlighted

### DIFF
--- a/sirepo/package_data/static/css/lattice.css
+++ b/sirepo/package_data/static/css/lattice.css
@@ -115,3 +115,10 @@ div.sr-lattice-icon-disabled span {
   position: relative;
   overflow: hidden;
 }
+
+.sr-sticky-heading {
+  position: sticky;
+  top: 0;
+  background-color: white;
+  padding: 1em;
+}
--- a/sirepo/package_data/static/css/lattice.css
+++ b/sirepo/package_data/static/css/lattice.css
@@ -120,5 +120,10 @@ div.sr-lattice-icon-disabled span {
   position: sticky;
   top: 0;
   background-color: white;
-  padding: 1em;
+  padding: 0.5em;
+  z-index: 1000;
+}
+
+.sr-searchbar-elem {
+  margin: 0.2em;
 }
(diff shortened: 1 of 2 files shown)


moellep:
I checked in a minor change to the layout to have all the input fields on one line.

Two comments:
- I wonder if supporting partial matches (find the first element which matches the prefix) rather than complete matches would be more useful. For example, when viewing the OPAL bERLinPro simulation, it might be useful to search on VM1 and have the 1st kicker selected. It would be hard to do an exact search on elements in that lattice since the names are long and cryptic.
- Rather than storing "tree" in the sirepo-data.json with all the element info, it would be better to store only the (element type name, isCollapsed) map. Otherwise the sirepo-data.json file becomes twice as large for lattices like the bERLinPro simulation above.

gurhar1133:
Made changes:
* Now the search does partial matches and will filter/collapse based on the matches
* The full tree is no longer being saved to db, now just a map of elementType: isCollapsed

LMK if more adjustments needed

moellep:
Looks good, I made a few adjustments with 91f9ddb. Take a look and make sure it still works the way you expect.

gurhar1133:
The highlighting looks much better now and the refactoring makes sense. The only thing I changed were a couple saveQuietly() calls to saveChanges(). This is because under some conditions the expanded categories/whether or not all the categories are expanded was not being saved during refresh.

LMK what you think @moellep

PR #4330, opened 2022-04-14
Fix #4313: shadow user manual in ? icon

### LINKED ISSUE #4313: Shadow: link to the user manual from within the app
The Shadow primer can be found here,
https://forge.epn-campus.eu/attachments/download/996/Shadow3Primer.pdf

Perhaps it should be linked from within the ? icon inside the app, similar to what is done for some of the other apps.

gurhar1133:
The shaddow app now has a link to the user manual in the '?' icon dropdown:

Screen Shot 2022-04-14 at 2 28 40 PM

This look right to you @bruhwiler


bruhwiler:
This is exactly what I wanted.
I'm not qualified to review the source code.

Issue #3977, opened 2021-11-30
SRW:  the 'Idealized Undulator' tab is too inconspicuous


bruhwiler:
This is what it looks like:
![image](https://user-images.githubusercontent.com/3825435/144123161-c52ea34e-1ee4-47bf-8f84-7fd64145daf0.png)

Comment from a Sirepo Premium customer:  "At first, I didn't realize one could change the undulator settings."

The feature request is to make the report more obvious to the user, perhaps by adding an undulator icon at the top, or else by listing some of the input variables within the body of the report.

robnagler:
I agree. I don't think we should have panels with no parameters. Even having  "click the pencil icon" (which is also a link) would be useful.

gurhar1133:
How about moving everything from the pencil icon into the panel:
![Screen Shot 2022-04-15 at 10 53 19 AM](https://user-images.githubusercontent.com/49498231/163598569-017f3e38-f470-4636-9989-71acf0f4335d.png)
 ?

bruhwiler:
I suggest you keep 3 parameters in the top-level report:
- period
- length
- effective deflecting parameter

gurhar1133:
> I suggest you keep 3 parameters in the top-level report:
> 
> * period
> * length
> * effective deflecting parameter

By top level, do you mean accessible through the pencil icon?

e-carlin:
Not to be a pedant but we have names for these constructs: the panel you see without clicking on anything is called "basic editor panel" and the panel you see after clicking on the pencil icon is called the "advanced editor panel".

gurhar1133:
> Not to be a pedant but we have names for these constructs: the panel you see without clicking on anything is called "basic editor panel" and the panel you see after clicking on the pencil icon is called the "advanced editor panel".

Helpful clarification of the terms @e-carlin

gurhar1133:
Like this @bruhwiler 
![Screen Shot 2022-04-15 at 11 22 09 AM](https://user-images.githubusercontent.com/49498231/163601366-e88bd95f-05d7-4e03-b2c0-f4d7b80848d6.png)
?

bruhwiler:
Yes, I believe this is the optimal choice for the basic editor panel.

PR #4331, opened 2022-04-15
3977 conspicuous undulator panel

gurhar1133:
Now the parameters:
* period
* length
* effective deflecting parameter
Are in the advanced editor panel, and all the others have been moved into the basic editor panel

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -1536,13 +1536,12 @@ SIREPO.viewLogic('trajectoryReportView', function(appState, panelState, srwServi
 });
 
 SIREPO.viewLogic('undulatorView', function(appState, panelState, srwService, $scope) {
-    if ($scope.fieldDef == 'basic') {
-        return;
-    }
-
     $scope.whenSelected = function() {
         panelState.enableField('undulator', 'effectiveDeflectingParameter', false);
     };
+    if ($scope.fieldDef == 'basic') {
+        return;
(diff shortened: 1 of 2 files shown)


moellep:
These values should be on the "basic" panel which is shown on the source page, the other fields should go to the "advanced" panel which is shown in the modal when the pencil is selected:

period
length
effective deflecting parameter

gurhar1133:
> These values should be on the "basic" panel which is shown on the source page, the other fields should go to the "advanced" panel which is shown in the modal when the pencil is selected:
> 
> period length effective deflecting parameter

So are you saying period, length and effective deflecting parameter should be in basic or advanced? Right now they are in advanced: 

Screen Shot 2022-04-15 at 11 22 09 AM

moellep:
Yes, when David said put those fields on the top level panel, he was referring to the basic panel.

gurhar1133:
> Yes, when David said put those fields on the top level panel, he was referring to the basic panel.

But when I showed him that same screenshot he said that's what he wanted: https://github.com/radiasoft/sirepo/issues/3977

moellep:
Yes, but he meant for the basic editor: "Yes, I believe this is the optimal choice for the basic editor panel."

gurhar1133:
> Yes, but he meant for the basic editor: "Yes, I believe this is the optimal choice for the basic editor panel."

@moellep Oh, that makes sense. Thanks for clarifying :) 
Made the change

PR #4340, opened 2022-04-19
Fix #4262: opal to madx lattice

### LINKED ISSUE #4262: export madx lattice from fixed position OPAL lattice is incorrect
When exporting an OPAL simulation to MAD-X, the elements are not positioned correctly if the OPAL lattice is using fixed position element. It looks like drift elements need to be inserted to take up the space between the fixed position elements. A simple example simulation is here: https://alpha.sirepo.com/opal#/lattice/nDgEioZU

In OPAL, the lattice is two 0.3m quadrupoles spaced 1m apart
Lattice-BL1

When exported to MAD-X and then imported, there is no space between the elements. Also, it looks like the quadrupole length in the OPAL simulation has a pow() value, but it isn't getting carried over to mad-x correctly.

gurhar1133:
Drifts are now inserted to fill empty spaces when converting from opal to madx

### DIFF
--- a/sirepo/template/opal.py
+++ b/sirepo/template/opal.py
@@ -235,14 +235,14 @@ class OpalMadxConverter(MadxConverter):
                 if e._id == id:
                     return e.type
 
-        def _get_drift(data, distance):
+        def _get_drift(distance):
             for e in data.models.elements:
-                if e.l == distance:
+                if e.l == distance and e.type == 'DRIFT':
                     return e
             return False
 
         def _insert_drift(distance, beam_idx, items_idx, pos, length):
-            d = _get_drift(data, distance)
+            d = _get_drift(distance)
             n = LatticeUtil.max_id(data) + 1
             m = 'D'+str(n)
             if d:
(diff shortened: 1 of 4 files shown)


sirepo/template/opal.py
```diff
@@ -222,6 +222,60 @@ def __init__(self):
         super().__init__(SIM_TYPE, self._FIELD_MAP)
 
     def to_madx(self, data):
+
+        def _get_len_by_id(data, id):
+            for e in data.models.elements:
+                if e._id == id:
+                    return e.l
```
e-carlin, line 229: Need to handle the case where id is not found. Python implicitly returns None but if that is the desired behavior then be explicit and say `return None` (implicit behavior like this is a source of bugs https://lwn.net/Articles/794944/). Otherwise, add `raise AssertionError(f'id={id} not found in elements={data.models.elements}')`. This helps future readers know that you covered (or tried to) all cases.

sirepo/template/opal.py
```diff
@@ -222,6 +222,60 @@ def __init__(self):
...
+        def _get_drift(distance):
+            for e in data.models.elements:
+                if e.l == distance and e.type == 'DRIFT':
+                    return e
+            return False
+
+        def _insert_drift(distance, beam_idx, items_idx, pos, length):
+            d = _get_drift(distance)
+            n = LatticeUtil.max_id(data) + 1
+            m = 'D'+str(n)
```
e-carlin, line 245: spacing

gurhar1133: ``` m = 'D'+str(n)``` should be ``` m = 'D' + str(n)``` ?

e-carlin: yes

e-carlin: You can also do `m = f'D{n}' and python will convert to a string.

gurhar1133:
Changes made @e-carlin 
Thoughts @moellep

moellep:
Looks good. I refactored the test cases a little.

PR #4354, opened 2022-04-20
Fix #3810: flask messages

### LINKED ISSUE #3810: ../sirepo/util.py:222:in_flask_request flask.request is False
@e-carlin should this be coming out normally?

gurhar1133:
Now logs message when server is true

### DIFF
--- a/sirepo/server.py
+++ b/sirepo/server.py
@@ -658,8 +658,7 @@ def init(uwsgi=None, use_reloader=False, is_server=False):
     uri_router.init(_app, simulation_db)
     if is_server:
         sirepo.db_upgrade.do_all()
-        # giving this info to util so it knows to show message
-        # to help debug for: https://github.com/radiasoft/sirepo/issues/3727
+        # Currently used for a special case in sirepo.util.in_flask_request. Do not use widely, because we should avoid server vs pkcli dependencies.
         sirepo.util.is_server = True
     return _app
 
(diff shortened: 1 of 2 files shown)


sirepo/util.py
```diff
@@ -29,6 +29,9 @@
 
 cfg = None
 
+#: Treat as read only. Only set by server.py
```
robnagler, line 32: `#: This should be treated as read only and used sparingly. See discussion in sirepo.server.init`

sirepo/util.py
```diff
@@ -270,7 +273,8 @@ def in_flask_request():
     if not f.request:
         if _log_not_request < 10:
             _log_not_request += 1
-            pkdlog('flask.request is False')
+            if is_server:
+                pkdlog('flask.request is False')
```
robnagler, line 277: `# This will help debug https://github.com/radiasoft/sirepo/issues/3727`

sirepo/server.py
```diff
@@ -658,6 +658,9 @@ def init(uwsgi=None, use_reloader=False, is_server=False):
     uri_router.init(_app, simulation_db)
     if is_server:
         sirepo.db_upgrade.do_all()
+        # giving this info to util so it knows to show message
```
robnagler, line 661: `# Currently used for a special case in sirepo.util.in_flask_request. Do not use widely, because we should avoid server vs pkcli dependencies.`

robnagler (changes requested):
I wrote comments about comments. Let me know if you don't understand why I made those changes.

gurhar1133:
Made changes

@robnagler, some of your comments have ':' and some don't. What is the rule pertaining to ':' again? 

Otherwise the changes to comments make sense

robnagler:
Here's the example: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

It's a way to get the string into the generated documentation.

robnagler (approved):

PR #4357, opened 2022-04-21
Fix #4356: get srw running in parallel on Cori

### LINKED ISSUE #4356: srw not running parallel at NERSC
__srwl_logs/*.log should look like this when running in parallel
```
Calculation on 1024 cores with averaging of 20 particles/iteration.
Running      400 out of 1500000 (  0.03% complete)
Running      800 out of 1500000 (  0.05% complete)
Running     1200 out of 1500000 (  0.08% complete)
```

It will look like this when run across multiple cores but each core is running independently
```
Calculation on 1 cores with averaging of 2500000 particles/iteration.
Running       25 out of 2500000 (  0.00% complete)
Running       25 out of 2500000 (  0.00% complete)
Running       25 out of 2500000 (  0.00% complete)
```

e-carlin:
SRW uses MPI when running in parallel. On Cori we run inside of
shifter. In order for the correct MPI libs to be loaded Cori swaps MPI
libs in the running shifter image
https://docs.nersc.gov/development/shifter/how-to-use/#using-mpi-in-shifter.

This is not working properly. So, we need to manually add the
libraries ourself.

@robnagler this is enough to get things working. Obviously, less than optimal to have to manually add these libs by hand. Also, not sure if other libraries for unused parts of mpi are missing...

### DIFF
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -748,9 +748,8 @@ class _SbatchRun(_SbatchCmd):
 cat > bash.stdin <<'EOF'
 {self.job_cmd_source_bashrc()}
 {self.job_cmd_env()}
-if [[ ! $LD_LIBRARY_PATH =~ /usr/lib64/mpich/lib ]]; then
-    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib64/mpich/lib
-fi
+# Inject libmpi ourselves. See: git.radiasoft.org/sirepo/issues/4356
+export LD_LIBRARY_PATH=/opt/cray/pe/mpt/7.7.19/gni/mpich-gnu-abi/8.2/lib:$LD_LIBRARY_PATH
 exec python {template_common.PARAMETERS_PYTHON_FILE}
 EOF
 exec srun {m} {s} /bin/bash bash.stdin
--- a/sirepo/pkcli/job_agent.py
+++ b/sirepo/pkcli/job_agent.py
@@ -749,7 +749,10 @@ cat > bash.stdin <<'EOF'
 {self.job_cmd_source_bashrc()}
 {self.job_cmd_env()}
 # Inject libmpi ourselves. See: git.radiasoft.org/sirepo/issues/4356
-export LD_LIBRARY_PATH=/opt/cray/pe/mpt/7.7.19/gni/mpich-gnu-abi/8.2/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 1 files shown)


sirepo/pkcli/job_agent.py
```diff
@@ -751,6 +751,19 @@ def _processor():
 if [[ ! $LD_LIBRARY_PATH =~ /usr/lib64/mpich/lib ]]; then
```
robnagler, line 751: This probably isn't needed any more?

sirepo/pkcli/job_agent.py
```diff
@@ -751,6 +751,19 @@ def _processor():
...
     export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib64/mpich/lib
 fi
+export LD_LIBRARY_PATH=/opt/cray/pe/mpt/7.7.19/gni/mpich-gnu-abi/8.2/lib:$LD_LIBRARY_PATH
```
robnagler, line 754: This one should be documented too, e.g. libmpi_..?

Make a for loop? Perhaps it's time to add something like rs_shifter to bashrc?

```sh
declare -a x=(
    # libxpmem.so.0
    xpmem/2.2.27-7.0.3.1_3.4__gada73ac.ari
...
)
for y in "${x[@]}"; do
    LD_LIBRARY_PATH=/opt/cray/$y/lib64:$LD_LIBRARY_PATH
done
export LD_LIBRARY_PATH
```

robnagler (changes requested):

sirepo/pkcli/job_agent.py
e-carlin: It isn't https://github.com/radiasoft/sirepo/pull/4357/commits/87ef691b2e109cfb81600600cc11a39568aab666

sirepo/pkcli/job_agent.py
e-carlin: Turns out that one (imports libmpi) is the only one that is needed https://github.com/radiasoft/sirepo/pull/4357/commits/87ef691b2e109cfb81600600cc11a39568aab666

robnagler: I think this should be like it was, that is, it shouldn't reinsert in the path if already there.

e-carlin: https://github.com/radiasoft/sirepo/pull/4357/commits/78958ee6dcf1e1756d101755c8437c86914edcea

robnagler (approved):

PR #4372, opened 2022-04-28
fix #4230: start agent on visit

### LINKED ISSUE #4230: start agent when user visits app
Sometimes there is a lag when a person revisits the site after an absence and visits a page with reports. There is a noticeable delay waiting for the agent to start up to provide the report output. After the agent is running, everything works well.

It would be nice if the agent was started the first moment a logged-in user visits the site, rather than waiting for the first run-status to be called.

rorour:


moellep:
I tried it on dev and it work well. I see that the /wake-agent is called from the client, right after /simulation-list. I wonder if we could move the wake-agent call into the /simulation-list logic and avoid 2 round-trips on the client on each simulation list navigation?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -3941,6 +3949,7 @@ SIREPO.app.controller('SimulationsController', function (appState, cookieService
         appState.saveQuietly('simFolder');
     });
     loadList();
+    wakeAgent();
```
robnagler, line 3952: As @moellep said, let's avoid roundtrips, and this isn't the correct place anyway.

robnagler:
I agree with @moellep, but this needs to be integrated better with the concept of a user using the system. 

I would like to see the concept of a session added, because that's what's going on, and we need to track both in flask and the job supervisor for the same reasons. This is a raydata issue so I changed the issue to the raydata milestone. You can leave the billing for flashcap since it was relevant but the scope has just expanded, and we have to start addressing the broader issues for raydata.

We will want to know all the browse windows associated with a user. We can call these sessions, since they will be connected to web sockets. For this problem, they will be associated with a single agent. When a browser connects to the server, it will get a session id which it will save in its app state somewhere. This can be added to auth_state. A session is only started once a user has been cookied. It's important not to start a session unless cookied.

Create a class Session or something like that. It will hold this state during the request. You could rename srcontext to session, because that's what it is. Additionally, there will be a class called auth_db.Session, which persists some of this state including when the last time BEGIN_SESSION was called (see pr comments). Here's what's in the auth_db.Session:

* user agent id (primary key unique random string, generated first time)
* login state
* uid (may be None)
* start time
* request time (modulo N secs, probably something like 60)
* end time
* is active (boolean)

On every request, the db is read and put in sirepo.session.Session. If the request time + modulo N secs < now, it is updated. If the last request time + begin session delta < now, a begin session request is sent if uid is not None.

One the UI side, every request gets a new header called X-Sirepo-UserAgentId or something, which is parsed in process request.

There's a lot here so a meeting is in order. :)

robnagler (changes requested):
See https://github.com/radiasoft/sirepo/pull/4372#issuecomment-1115151168

sirepo/job_supervisor.py
```diff
@@ -758,6 +758,19 @@ async def _receive_api_statefulCompute(self, req):
     async def _receive_api_statelessCompute(self, req):
         return await self._send_simulation_compute(req)
 
+    async def _receive_api_wakeAgent(self, req):
+        c = self._create_op(job.OP_WAKEUP, req)
+        try:
+            await c.prepare_send()
+        except Awaited:
+            # OPTIMIZATION: _agent_ready is the first thing that could raise Awaited.
+            # In the event that it does, the agent is still started,
```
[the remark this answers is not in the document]
e-carlin: I'm on the fence. We originally had a `POSIT` but changed to `OPTIMIZATION`. We could do the normal awaited loop around prepare_send which would eliminate the need for a comment. But, we are just expecting one awaited since the agent starting is the first part of prepare send. This saves us from using the loop and waiting for it to terminate which is an optimization.

`Agent is running so nothing more to do.` Why say this? I don't think we know if the agent is running. For example, let's say we can't start the agent (bug in agent code). We would still get to this code.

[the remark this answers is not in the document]
robnagler: If the agent fails because of the agent code, that's a different dimension, just like if we receive the message at the same time job_supervisor is started. The entire process is an optimization, because it works without doing anything.

I think there's a larger problem here: two agents need to be started. I think the job_supervisor has to start a task asynchronously on behalf of the user, not a job.

e-carlin:
@robnagler Writing down these notes so I don't forget. To impl:

1. Create the table 
2. Modify js to read the response headers to get the `user agent id`
3. For each request set the request `user agent id`
4. In sirepo if there is no `user agent id` fail fast by checking db and otherwise create a new record
5. Set the `user agent id` header in each response


@robnagler when will the `isActive` field move to False? Will we have some auditor process that is watching and after a certain period of inactivity move to False?

robnagler:
is_active is an optimization. You want to be able to query on active records only when the table gets "large". You can do without it for now. 

I'm not thrilled with using response headers because we will want this to occur in "messages" (think what happens with web sockets). Is there a way to do this on the reply to any POST? This has to do with centralizing the request/response processing. Thoughts?

e-carlin:
> I'm not thrilled with using response headers because we will want this to occur in "messages" (think what happens with web sockets). Is there a way to do this on the reply to any POST? This has to do with centralizing the request/response processing. Thoughts?

I agree. After a brief look at the code I don't see anywhere where we know that we are just in a POST but I'll look more and if one doesn't exist create one.

Where should the `user_agent_id` be stored in the browser? If we store it anywhere in Angular then it will get blown away when we change apps which is less than ideal (maybe ok for now though?). We can't store it in the cookie because it will be shared across tabs/windows. It looks like [WebStorage](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API) and [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) suffer from the same problem as cookies (shared for all origin)s..

robnagler:
> I agree. After a brief look at the code I don't see anywhere where we know that we are just in a POST but I'll look more and if one doesn't exist create one.

flask.request.method returns this value. This could now be encapsulated in sirepo.request.Base.

> Where should the user_agent_id be stored in the browser? If we store it anywhere in Angular then it will get blown away when we change apps which is less than ideal (maybe ok for now though?). 

In Javascript (SIREPO_APP or whatever). When we change apps, it's a new session as the apps are currently written.

e-carlin:
@robnagler when should the request be instantiated? As the code is currently written we try to start the session before the API is instantiated? I think this is causing me trouble because I don't have objects when I need them. Can you provide guidance?

robnagler:
The request is not the api. The request should be instantiated right at the start of [_dispatch](https://github.com/radiasoft/sirepo/blob/2b105c4a574c0cd15eb230fa87b9768ce02adfef/sirepo/uri_router.py#L203) and passed to sirepo.auth.process_request and call_api. The latter of which will attach the request to the sapi that it creates.

e-carlin:

tests/auth2_test.py
```diff
@@ -8,19 +8,24 @@
 import pytest
 
 def test_myapp_user_dir_deleted(fc):
+    from pykern import pkjson
     from pykern import pkunit, pkcompat
     from pykern.pkcollections import PKDict
     from pykern.pkdebug import pkdp
     import sirepo.srdb
 
+    p = sirepo.srdb.root().join(
```
robnagler, line 17: is `p` needed?

sirepo/job_api.py
```diff
@@ -310,6 +310,22 @@ def _validate_and_add_sbatch_fields(self, request_content, compute_model):
...
 
 
+def begin_session():
+    try:
+        u = sirepo.auth.logged_in_user()
+    except sirepo.util.Reply:
+        # If we get any Sirepo generated errors (ex. user dir not found) then just ignore and don't
+        # try to start session (ex. agent would be started in user dir which may not exist).
+        return
+    API().request(
```
robnagler, line 320: This ok for now. Put a comment about why you have to do this.

sirepo/job_api.py
```diff
@@ -310,6 +310,22 @@ def _validate_and_add_sbatch_fields(self, request_content, compute_model):
...
+        api_name='api_beginSession',
+        _request_content=PKDict(
+            uid=u,
+            userDir=str(sirepo.simulation_db.user_path(u)),
```
robnagler, line 324: Why not put this in the try/except as well?

sirepo/session.py
```diff
@@ -0,0 +1,96 @@
...
+import sirepo.srtime
+import sirepo.uri_router  #TODO(rorour) remove
+import sirepo.util
+import sqlalchemy
+
+_RENEW_SESSION_TIMEOUT_SECS = 5 * 60
+
+_SRCONTEXT_KEY = __name__
+
+_USER_AGENT_ID_LEN = 8
```
robnagler, line 21: This should be the default (32), because it is related to authentication.

sirepo/session.py
```diff
@@ -0,0 +1,96 @@
...
+
+_Session = None
+
+
+def _event_end_api_call(kwargs):
+    i = sirepo.srcontext.get(_SRCONTEXT_KEY)
+    if i:
+        kwargs.resp.headers['X-Sirepo-UserAgentId'] = i
```
robnagler, line 29: Define a constant.

robnagler (changes requested):
The job supervisor changes are hard to follow due to the reorg. Let's walk through them together. Set up a meeting.

sirepo/session.py
```diff
@@ -0,0 +1,96 @@
...
+        t = sirepo.srtime.utc_now()
+        i = sirepo.util.random_base62(_USER_AGENT_ID_LEN)
+        _Session(
+            user_agent_id=i,
+            login_state=l,
+            uid=sirepo.auth.logged_in_user(check_path=False) if l else None,
+            start_time=t,
+            request_time=t,
+        ).save()
+        l and _begin()
```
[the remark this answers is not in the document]
e-carlin: 40fdc6ae4

tests/auth2_test.py
e-carlin: 40fdc6ae4

sirepo/job_api.py
e-carlin: 40fdc6ae4

sirepo/job_api.py
e-carlin: 40fdc6ae4

sirepo/session.py
e-carlin: 40fdc6ae4

sirepo/session.py
e-carlin: 40fdc6ae4

robnagler (approved):

PR #4373, opened 2022-04-28
fix #4041: fixup every time dev restarts

### LINKED ISSUE #4041: confusing calls to fixup on dev
When running on dev, it is very confusing that fixups do not run unless a schema file change is detected. Instead, it should probably use a new version number on every dev restart, which would guarantee that fixups are always run.

rorour:


moellep:
Works well.

PR #4402, opened 2022-05-10
fix #4312: class based api

### LINKED ISSUE #4312: move to class based api
Move all api methods in flask into a class. This is part of adding a request context to all apis (https://github.com/radiasoft/sirepo/issues/4234)

- there will be a base class for APIs
- every module that contains an API will have its own class that inherits from the base
- write a program that changes our code to do this
- call_api will have to be handled by hand to generate the object

rorour:
@robnagler Currently the API class is being instantiated in uri_router.register_api_module() - not sure if that's really the correct place, but with how things are set up right now it seemed necessary, since each module has its own API class which needs to be instantiated in order to call the apis.

Diffs for the apis might look confusing. The only changes I made were to indent them inside of the API class and add `self` as a parameter.


package.json
```diff
@@ -5,7 +5,7 @@
   "devDependencies": {
     "jasmine-core": "latest",
     "jshint": "latest",
-    "karma": "latest",
+    "karma": "^6.3.19",
```
robnagler, line 8: I think we want to push this change in a separate PR. Please create an issue and move that change there.

sirepo/api.py
```diff
@@ -0,0 +1,15 @@
+class APIRequest:
```
robnagler, line 1: Add our standard file header.

Also document the class. It's holds the request context for all API calls. On reading this, I think we should rename this module to sirepo.request and the class would be called Base. I comment on how to automate these changes.

sirepo/api.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self):
```
robnagler, line 2: If there's nothing in `__init__` then don't have one so that the superclass init is automatically called.

sirepo/api.py
```diff
@@ -0,0 +1,15 @@
...
+        return
+
+    def call_api(self, name, kwargs=None, data=None):
```
robnagler, line 5: Write some documentation describing this call. It will be used to call other APIs, checks permissions, etc.

Also add some documentation to uri_router.call_api that it should not be called, and instead self.call_api should be used.

sirepo/api.py
```diff
@@ -0,0 +1,15 @@
...
+        return uri_router.call_api(name, kwargs=kwargs, data=data)
+
+
+APIBase = APIRequest
```
robnagler, line 9: This was a quick fix. Please change all sirepo.api.APIBase to sirepo.request.Base and remove this line. I tested this so it should work:

```sh
cd ~/src/radiasoft/sirepo/sirepo
perl  -pi -e 's{(?<=sirepo.)api.APIRequest}{request.Base}; s{(?<=import sirepo.)api$}{request}; s{(?<=class )API\b}{Request}' *.py */*.py
```

Run this on a committed so that you can verify/undo these changes.

sirepo/uri_router.py
```diff
@@ -190,6 +203,22 @@ def uri_for_api(api_name, params=None, external=True):
     return res
 
 
+class _Route(PKDict):
```
robnagler, line 206: Document this: holds all route information. Document the known keys: func, cls, name, func_name, etc.

sirepo/uri_router.py
```diff
@@ -157,11 +167,14 @@ def register_api_module(module=None):
     if hasattr(m, 'init_apis'):
         m.init_apis()
     # It's ok if there are no APIs
-    for n, o in inspect.getmembers(m):
-        if n.startswith(_FUNC_PREFIX) and inspect.isfunction(o):
+    if not hasattr(m, 'API'):
```
robnagler, line 170: Document why this test exists. It doesn't seem like the right place. Probably should be in `_register_sim_api_module`.

sirepo/uri_router.py
```diff
@@ -157,11 +167,14 @@ def register_api_module(module=None):
...
+        return
+    c = m.API
```
robnagler, line 172: change to `m.Request`.

sirepo/uri_router.py
```diff
@@ -157,11 +167,14 @@ def register_api_module(module=None):
...
+    for n, o in inspect.getmembers(c):
+        if n.startswith(_FUNC_PREFIX) and inspect.isfunction(o) and n in c.__dict__:
```
robnagler, line 174: move this condition to a separate function that is named something appropriate.

robnagler (changes requested):
Looks good. Needs more documentation though.

sirepo/uri_router.py
rorour: @robnagler Without this check, tests fail because some modules like sirepo.auth.basic don't have any APIs and therefore don't have the class. Could you expand on why `_register_sim_api_modules` is the correct place and how exactly to add it there?

robnagler: Ah, I didn't remember that. I was thinking it was only sim api modules that didn't have APIs. Leave as is, then, and document including an example module.

Technically, this is an abuse of uri_router to enforce initialization. We really should sirepo.init (or something) that modules register with and declare dependencies. I thought [about this 7 years ago](https://github.com/radiasoft/pykern/wiki/ImportThoughts) (almost to the day), but never implemented. The ad hoc initialization that we are doing now is not working, because it requires too much special knowledge/handling. I'll add an issue.

sirepo/uri_router.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/uri_router.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/uri_router.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/uri_router.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/api.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/api.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/api.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

sirepo/api.py
rorour: 3ccfbc6e956420f7da6fc23176cdebb7ae57cfc9

package.json
rorour: https://github.com/radiasoft/sirepo/issues/4413

sirepo/uri_router.py
```diff
@@ -204,6 +209,13 @@ def uri_for_api(api_name, params=None, external=True):
 
 
 class _Route(PKDict):
+    """ Holds all route information for an API.
+
+    Keys:
+        func (function): The function object for the API
+        cls (class): The class in the API's module that contains the API function.
+        func_name (str): The function's name (ex: 'api_admJobs').
```
robnagler, line 217: Great. Please add more doc for decl_uri, etc. Look in _init_uris to see all the fields.

robnagler (changes requested):
Looks great. Minor doc change requested.

sirepo/uri_router.py
rorour: 8f39fbdf6e35e7cd3105d1e0fbf9da2e60f893c7

robnagler (approved):
Great!

e-carlin (commented):
Looks awesome! Some minor nitpicks you can address when you have time.

sirepo/auth/__init__.py
```diff
@@ -21,6 +20,7 @@
 import contextlib
 import datetime
 import importlib
+import sirepo.request
```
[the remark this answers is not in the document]
rorour: d49c4cd97a3c14d1d445f78b3e3c5ab78269589b

sirepo/uri_router.py
```diff
@@ -70,19 +88,22 @@ def call_api(func_or_name, kwargs=None, data=None):
...
             s = sirepo.http_request.set_sim_type(kwargs.get('simulation_type'))
-        f = check_api_call(func_or_name)
+        else:
+            kwargs = PKDict()
+        f = _check_api_call(route_or_name)
         try:
             if data:
                 p = sirepo.http_request.set_post(data)
-            r = flask.make_response(f(**kwargs) if kwargs else f())
+            i = f.cls()
```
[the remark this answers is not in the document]
rorour: b562769b31804e951044ff01bebafaaeb1c17434

sirepo/uri_router.py
```diff
@@ -50,13 +51,30 @@
...
+    """Check if `name` is executable and in allowed
+
+    Args:
+        name (str): name of the api
+        allowed (tuple): names that are allowed to be called
+    Returns:
+        str: api name
+    """
+    _check_api_call(name)
+    assert allowed, 'must pass non-zero set'
```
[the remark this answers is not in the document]
rorour: e676a795bb22b1e6c52c63a3458e39e512a0258b

sirepo/uri_router.py
```diff
@@ -50,13 +51,30 @@
...
+    if name not in allowed:
+        raise AssertionError(f'api={name} not in allowed={allowed}')
+
+
+def call_api(route_or_name, kwargs=None, data=None):
+    """Should not be called outside of Base.call_api(). Use self.call_api() to call API.
```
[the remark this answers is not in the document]
rorour: ec472d9bbdb74b05d0d0cebc3f0f3b039f997c79

PR #4414, opened 2022-05-16
fix #4413: use karma ^6.3.19 instead of latest

### LINKED ISSUE #4413: use karma ^6.3.19 instead of latest

rorour:


e-carlin:
@rorour would you mind updating the issue or commit message with the reason this change is needed? I assume latest is breaking something and in the future if we need to change karma versions it would help to know what was breaking.

rorour:
@robnagler I'm thinking we don't need this anymore, should I close this?

robnagler:
I actually don't know. I don't remember the issue so I leave it up to your judgement. :)

PR #4416, opened 2022-05-17
4386 flash improvements

### LINKED ISSUE #4399: flash not running with mpi
It looks like the flash executable is always called in single-processor mode.

moellep:
- links to setup, compile and mpi logs
- new problem file manager
- added many example sims which get populated from examples
  contained within flash.tar.gz
- initial processing takes input archive files and creates a
  simulation-specific lib file
- better flash schema handling for newly added module fields
- removed Config tab, now managed only with problem file table
- added withParticles setup option
- run flash sims using mpi, fixes #4399
- fixes to flash parser to support more example cases
- improved setup, compile and runtime error parsing
- allow apps to override simulation message based on model name
- viewing syntax highlighted archive files
- delete archive file
- download archive file
- sort archive files by type then name
- download raw report data
- Improve Parameters tab. Show modules hierarchically with field count.
- Implemented problem file add and replace with confirmation.


robnagler:
I'd like @e-carlin to review anyway.

PR #4423, opened 2022-05-18
Fix #4347: srw mpi logs

### LINKED ISSUE #4347: Show additional info from srw mpi log file
Provide more information from the SRW status file when running MPI jobs in the UI. Or provide a way to download the log file via a link.

gurhar1133:
Log file links are now available in partially coherent results, spectral flux, and coherent modes tabs.

@moellep I have run into some issues with the coherent modes tab. Not sure if reproducible, but if you run a spectral flux animation (and cancel it early once the log link becomes available) then expand the coherent modes tab you may get 'report not generated' error message in coherent modes tab or other errors.

Some errors like this were present on master branch as well. 

Maybe see if you can reproduce and I could create new issue if so?

### DIFF
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -145,6 +145,8 @@ _RSOPT_PARAMS = {
 }
 _RSOPT_PARAMS_NO_ROTATION = [p for p in _RSOPT_PARAMS if p != 'rotation']
 
+_SRW_LOG_FILE = 'run.log'
+
 _TABULATED_UNDULATOR_DATA_DIR = 'tabulatedUndulator'
 
 _USER_MODEL_LIST_FILENAME = PKDict(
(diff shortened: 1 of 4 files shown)


rorour:
looks good to me!

rorour (approved):

sirepo/template/srw.py
```diff
@@ -456,6 +456,8 @@ def get_application_data(data, **kwargs):
 
 
 def get_data_file(run_dir, model, frame, **kwargs):
+    if frame < 0:
```
robnagler, line 459: This seems like an add test. Is there no other argument you could use in kwargs to make this be more explicit?

gurhar1133: What about passing in a string to indicate that it's a SRW log file: `'isSRWLogFile'`

robnagler: The call limits you at the moment to options.suffix. I think that could be a better signal than frame < 0. Look at [zgoubi.get_data_file](https://github.com/radiasoft/sirepo/blob/868fed4ad90e3c2503664528493447f89073162b/sirepo/template/zgoubi.py#L399) which abuses suffix to be the whole file name.

gurhar1133:
Made changes to use options @robnagler. Will also work on #4428 👍

robnagler (approved):

PR #4432, opened 2022-05-19
fix #4431 number list leading - or .

### LINKED ISSUE #4431: numberList not accepting characters
Fields using numberList with a Float type can't enter a leading '-' or a '.' when typing.

moellep:


moellep:
@mkeilman please check that this still works with the radia number-list types. For number-lists without a model attribute, it was always reseting scope.values to null.

PR #4433, opened 2022-05-19
Fix #4425: Move RS caplaser examples into flash.tar.gz

### LINKED ISSUE #4425: move caplaser examples into flash.tar.gz
They are distributed as separate archives right now.

e-carlin:
@moellep I've added the examples to the flash tarball. I've put the tarball in my home dir on the usual server for you to test. 

These settings get the code to compile but neither produce any output. Any idea what settings to twiddle to get output to be produced?


moellep:
Yes, there are several settings such as plot_var_X and logging frequency. I'm pretty familiar with them now - I'll take a look tomorrow afternoon.

moellep:
I've updated the Cap Laser examples, removing the "Driver_DriverMain" overrides which was causing problems. I also added a setup arg SupportAnisoCond which seems to be needed for CapLaser3D, although with that arg enabled, the simulation doesn't compile. Without SupportAnisoCond the diff_anisoCondForEle would not be editable, and the simulation would fail when the par file turns that feature on.

moellep:
I'm not sure we got this working correctly or had working simulation files for this. Can we archive this pull request and reactivate it this addition is requested?

e-carlin:
Works for me.

PR #4435, opened 2022-05-20
Fix #4428: normalized get_data_file

### LINKED ISSUE #4428: Normalize get_data_file arg use
There is a single caller of get_data_file, which is:

```py
        r = template.get_data_file(
            msg.runDir,
            msg.analysisModel,
            msg.frame,
            options=PKDict(suffix=msg.suffix),
	)
```

There are N different ways of handling args. Since `options` was added, it is unnecessary to have `**kwargs`, because `options` is for handling this case. While I normally recommend `**kwargs` on APIs like this, I would say it confuses things.

Here are the 3 cases:

- `if options and options.suffix == _ZGOUBI_COMMAND_FILE:`
- `sfx = (options.suffix or default_sfx) if options and 'suffix' in options else default_sfx`
- `if options and options.get('suffix') == 'sr_long_analysis':`
- `if not options.suffix`

All of these need to be changed except the last one. `options.suffix` is always defined. It may be None, but comparisons to None are fine. It is sufficient to use options.suffix or default_sfx, for example.

The signature for all defs should be:

```py
```

suffix is a misnomer we'll have to live with.

gurhar1133:
get_data_file handle changed

### DIFF
--- a/sirepo/template/myapp.py
+++ b/sirepo/template/myapp.py
@@ -25,8 +25,8 @@ INPUT_NAME = 'hundli.yml'
 OUTPUT_NAME = 'hundli.csv'
 
 
-def get_data_file(run_dir, model, frame, options=None, **kwargs):
-    if options and options.get('suffix') == 'sr_long_analysis':
+def get_data_file(run_dir, model, frame, options):
+    if options.get('suffix') == 'sr_long_analysis':
         time.sleep(100)
     return OUTPUT_NAME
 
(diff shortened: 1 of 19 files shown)


sirepo/template/myapp.py
```diff
@@ -25,8 +25,8 @@
 OUTPUT_NAME = 'hundli.csv'
 
 
-def get_data_file(run_dir, model, frame, options=None, **kwargs):
-    if options and options.get('suffix') == 'sr_long_analysis':
+def get_data_file(run_dir, model, frame, options):
+    if options.get('suffix') == 'sr_long_analysis':
```
robnagler, line 29: options.suffix.

robnagler (changes requested):

gurhar1133:
Made the change @robnagler

robnagler (approved):

Issue #4167, opened 2022-02-18
User input validation


robnagler:
- real Boolean types (not "0" "1" enums)
- better type checking for Float and Integer (no strings allowed)
- share validation between client and server
- form fields should always have types and be validated, possibly interactively
- forms should have cross-validation

mkeilman:
I think this can go here: #1286

robnagler:
Look at external frameworks like [pydantic](https://pydantic-docs.helpmanual.io/#rationale)

PR #4442, opened 2022-05-24
Fix #4410: srw screen flicker reports

### LINKED ISSUE #4410: Screen flickering on rendering
When rendering partially coherent simulation frames on alpha of Young's, the report was gray/flickering. [See sideo of flickering](https://user-images.githubusercontent.com/5495179/168395175-2dbfd0ab-4cbb-4a60-a0d4-6c4e72b4fd38.mp4).

gurhar1133:
Flicker no longer happening

Given that this changed is general across sirepo I have a couple questions @moellep :
1) Should there be a check on the timeout so that this only happens for SRW? 
2) This looks good in SRW, but are there other apps I should be checking to make sure these changes didn't mess with the UI?

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1667,7 +1667,6 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
     };
 
     self.getStatusText = function(name) {
-
         if (self.isRunning(name)) {
             var count = (queueItems[name] && queueItems[name].runStatusCount) || 0;
             var progressText = (appState.models[name] || {}).inProgressText  ||
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1279,20 +1279,23 @@ SIREPO.app.factory('frameCache', function(appState, panelState, requestSender, $
         }
         var isHidden = panelState.isHidden(modelName);
         var frameRequestTime = new Date().getTime();
+        var waitedTimeElapsed = false;
         var delay = isPlaying && ! isHidden
             ? 1000 / parseInt(appState.models[modelName].framesPerSecond || 2)
             : 0;
         var requestFunction = function() {
-            if (SIREPO.SLOW_ANIMATION) {
-                // some apps like SRW may take a long time to process a frame
-                panelState.setLoading(modelName, true);
-            }
+            setTimeout(() => {
+                if (!waitedTimeElapsed) {
+                    panelState.setLoading(modelName, true);
+                }
+            }, 1000);
             requestSender.sendRequest(
                 {
                     'routeName': 'simulationFrame',
                     '<frame_id>': self.frameId(modelName, index),
                 },
                 function(data) {
+                    waitedTimeElapsed = true;
                     panelState.setLoading(modelName, false);
                     if ('state' in data && data.state === 'missing') {
                         onError();
@@ -1664,6 +1667,7 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
     };
 
     self.getStatusText = function(name) {
+
         if (self.isRunning(name)) {
             var count = (queueItems[name] && queueItems[name].runStatusCount) || 0;
             var progressText = (appState.models[name] || {}).inProgressText  ||
@@ -1672,8 +1676,8 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
             return progressText + ' ' + new Array(count % 3 + 2).join('.');
         }
         return appState.isAnimationModelName(name)
-            ? 'Requesting Data'
-            : 'Waiting';
+        ? 'Requesting Florpus'
+        : 'Waiting';
     };
 
     self.isActiveField = function(model, field) {
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1676,7 +1676,7 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
             return progressText + ' ' + new Array(count % 3 + 2).join('.');
         }
         return appState.isAnimationModelName(name)
-        ? 'Requesting Florpus'
+        ? 'Requesting Data'
         : 'Waiting';
     };
 
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -7,7 +7,6 @@ SIREPO.app.config(function() {
     SIREPO.appDefaultSimulationValues.simulation.sourceType = 'u';
     SIREPO.SHOW_HELP_BUTTONS = true;
     SIREPO.INCLUDE_EXAMPLE_FOLDERS = true;
-    SIREPO.SLOW_ANIMATION = true;
     SIREPO.SINGLE_FRAME_ANIMATION = ['coherenceXAnimation', 'coherenceYAnimation', 'coherentModesAnimation', 'fluxAnimation', 'multiElectronAnimation'];
     SIREPO.PLOTTING_COLOR_MAP = 'grayscale';
     SIREPO.PLOTTING_SHOW_FWHM = true;


sirepo/package_data/static/js/sirepo.js
```diff
@@ -1279,20 +1279,23 @@ SIREPO.app.factory('frameCache', function(appState, panelState, requestSender, $
         }
         var isHidden = panelState.isHidden(modelName);
         var frameRequestTime = new Date().getTime();
+        var waitedTimeElapsed = false;
```
mkeilman, line 1282: Use let. In general almost all "vars" can be replaced by "let" or "const". I tend to replace them as I go (not in the entire file) but YMMV.

Also, it's not major but I prefer names for booleans that are yes/no questions ("hasWaitTimeElapsed"). I don't always follow that myself though :|

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1279,20 +1279,23 @@ SIREPO.app.factory('frameCache', function(appState, panelState, requestSender, $
...
         var delay = isPlaying && ! isHidden
             ? 1000 / parseInt(appState.models[modelName].framesPerSecond || 2)
             : 0;
         var requestFunction = function() {
-            if (SIREPO.SLOW_ANIMATION) {
-                // some apps like SRW may take a long time to process a frame
-                panelState.setLoading(modelName, true);
-            }
+            setTimeout(() => {
+                if (!waitedTimeElapsed) {
```
mkeilman, line 1288: Space after bang

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1672,8 +1675,8 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
             return progressText + ' ' + new Array(count % 3 + 2).join('.');
         }
         return appState.isAnimationModelName(name)
-            ? 'Requesting Data'
-            : 'Waiting';
+        ? 'Requesting Data'
+        : 'Waiting';
```
mkeilman, line 1679: Not personally a fan of this formatting (esp. since this will fit comfortably on one line), but the indent should be consistent. See above:
```
var delay = isPlaying && ! isHidden
            ? 1000 / parseInt(appState.models[modelName].framesPerSecond || 2)
            : 0;
```

mkeilman (changes requested):
Looks fine other than some nitpicks

sirepo/package_data/static/js/sirepo.js
robnagler: 1000 and 2 are magic numbers. They should come from somewhere or be documented. The parseInt should be this:

```js
appState.models[modelName].framesPerSecond ? parseInt(appState.models[modelName].framesPerSecond) : 2
```

Maybe this should be encapsulated in a function so it is properly documented. Then the 1000 and 2 could be consts in the function and it would be clearer what they mean.

gurhar1133: @mkeilman I can def change that indent.

@robnagler I am not totally clear on the meaning behind the 1000 and the 2 myself. I can wrap the ternary expression in a function if you like. The line you wrote above however would change the meaning though wouldn't it? 

Would something like this be good?
```javascript
const getDelay = () => {
    const x = 1000;
    const y = 2;
    if (isPlaying && ! isHidden) {
        return  x/(parseInt(appState.models[modelName].framesPerSecond || y)
    }
    return 0;
}
var delay = getDelay()
```
(again, I am unclear on the semantics of 1000 and 2

gurhar1133: @robnagler are you saying framePeriod would be a better name?

moellep: I think renaming the constants x and y isn't anymore helpful than having them in place, so give the function and constants a long meaningful name. Also, make sure everything fits with 100 character lines for the new coding style.

```javascript
const millisecondsUntilNextFrameRequest = () => {
    if (isPlaying && ! isHidden) {
        const secondsToMilliseconds = 1000;
        const defaultFramesPerSecond = 2;
        return secondsToMilliseconds / (
            parseInt(appState.models[modelName].framesPerSecond
                || defaultFramesPerSecond
            )
    }
    return 0;
}
var delay = getMillisecondsUntilNextFrameRequest()
```

mkeilman (approved):
Looks good!

sirepo/package_data/static/js/sirepo.js
robnagler: I agree with @moellep on x and y. @gurhar1133 I asked you to update DesignHints: https://github.com/radiasoft/pykern/issues/187

I really don't like `parseInt(thing || default)`. defaultFramesPerSecond is already an int and it should be default milliseconds. Yes, I'm being pedantic:

```js
const millisecondsUntilNextFrameRequest = () => {
    if (! isPlaying || isHidden) {
        return 0;
    }
    const milliseconds = 1000;
    var x = appState.models[modelName].framesPerSecond; 
    if (! x) {
        return  .5 * milliseconds;
    }
    return  milliseconds / parseInt(x);
}
```

I don't think we need to document the default in this case, because "! x" clear indicates it is a default.

gurhar1133:
@robnagler I refactored millisecondsUntilNextFrameRequest, anything else?

sirepo/package_data/static/js/sirepo.js
moellep: The parseInt() is due to our current limitation with enums (framesPerSecond is an enum value) - it is always a string value mapped to a string description. This also makes booleans awkward because Boolean is an enum. So you see code like this in the python:
```python
    if v.controlSettings_readOnly == '0':
	properties.write = []
```
rather than:
```python
    if not v.controlSettings_readOnly:
	properties.write = []
```

moellep: There is already an issue related to Boolean/enum #4167

robnagler: > The parseInt() is due to our current limitation with enums (framesPerSecond is an enum value)

You did not understand the point of my refactoring. parseInt for appState.models[modelName].framesPerSecond is necessary, and I reduced the code to be more readable (imiho) to just be the parseInt on that value, not on the number 2, which is already an integer so does not need to be parsed. 

@radiasoft/software please chime in on the readability issue.

robnagler: One other item related to our discussion on React yesterday: denormalized state on the stack. [This code](https://github.com/radiasoft/sirepo/blob/5d3d71b1cfb25ef30f71aad8bf1c6b1e59cee3aa/sirepo/package_data/static/js/sirepo.js#L1316) is using denormalized model state unnecessarily:

```js
                   let endTime = new Date().getTime();
                    let elapsed = endTime - frameRequestTime;
                    if (elapsed < delay) {
                        $interval(
                            function() {
                                callback(index, data);
                            },
                            delay - elapsed,
                            1
                        );
                    }
                    else {
                        callback(index, data);
                    }
```

There's no reason to cache delay. Rather, you should ask at that point of needing it, that is, `delay` should be computed at the time that the callback is received from the server, not before. 

Furthermore, millisecondsUntilNextFrameRequest is not the right name. It really is framePeriod (in milliseconds, which is irrelevant). The reason the callback is delayed is that we want a frame rate of 2 frames per second (or whatever the particular model specifies).

I would write the code this way (ignore the indentation

```js
const framePeriod = () => {
    if (! isPlaying || isHidden) {
        return 0;
    }
    const milliseconds = 1000;
    var x = appState.models[modelName].framesPerSecond;
    if (! x) {
        return .5 * milliseconds;
    }
    return milliseconds / parseInt(x);
}

let e = framePeriod() - (new Date().getTime() - frameRequestTime);
if (e <= 0) {
    callback(index, data);
    return;
}
$interval(
    function() {
        callback(index, data);
    },
    e,
    1,
);
```

Why is this important? If a user clicks stop, then isPlaying and isHidden have already been cached in delay, and they were checked at frameRequestTime instead of now (btw, new Date().getTime() should be now() or something).

PR #4446, opened 2022-05-24
Fix #4400: put mpi log in constant

### LINKED ISSUE #4400: mpi_run.out needs to be a constant
When running with sirepo.mpi, output is writted to mpi_run.out. Several apps parse that output for errors. There should be a constant sirepo.mpi.MPI_LOG or similar rather than hard-coding 'mpi_run.out' across the modules.

https://github.com/radiasoft/sirepo/blob/master/sirepo/mpi.py#L64

gurhar1133:
mpi log is in constant defined in template common

### DIFF
--- a/sirepo/mpi.py
+++ b/sirepo/mpi.py
@@ -8,6 +8,7 @@ from pykern import pkconfig
 from pykern import pkio
 from pykern import pksubprocess
 from pykern.pkdebug import pkdc, pkdexc, pkdp, pkdlog
+from sirepo.template import template_common
 import re
 import sys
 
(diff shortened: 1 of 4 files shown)


sirepo/mpi.py
```diff
@@ -8,6 +8,7 @@
 from pykern import pkio
 from pykern import pksubprocess
 from pykern.pkdebug import pkdc, pkdexc, pkdp, pkdlog
+from sirepo.template import template_common
```
robnagler, line 11: I don't think this is a good idea. There may be circular import problems. You could put it in sirepo.const, and this would avoid that issue.

robnagler (changes requested):

sirepo/mpi.py
gurhar1133: 👍

gurhar1133:
Moved the constant to sirepo.const @robnagler

robnagler (approved):

PR #4449, opened 2022-05-25
Fix #4445 - color reset etc

### LINKED ISSUE #4445: magnet viewer color change cancel button non-functional
When you control click to select an object in the Radia magnet viewer there is an option to change the object color. When you select a new color a set of save/cancel buttons pops up, but clicking cancel doesn't revert the color selection.

image

mkeilman:
Notes:

- canceling the object color change reverts the object and the color button
- now using the html5 color picker instead of the limited palette
- if an object is selected when going back to "move" mode, it gets deselected
- appFieldEditors from join to template literals


moellep:
@mkeilman could you take a look at this pull request - it has some merge conflicts in vtk-display.html and sirepo-plotting-vtk.js

mkeilman:
> @mkeilman could you take a look at this pull request - it has some merge conflicts in vtk-display.html and sirepo-plotting-vtk.js

Looking now

mkeilman:
Conflicts should be clear now

PR #4451, opened 2022-05-25
fix #4447: change http_request.parse_* to self.parse_*

### LINKED ISSUE #4447: Change http_request.parse_* self.parse_*
_Originally posted by @robnagler in https://github.com/radiasoft/sirepo/issues/4388#issuecomment-1126771252_

rorour:


sirepo/api_auth.py
```diff
@@ -6,7 +6,7 @@
 """
 from __future__ import absolute_import, division, print_function
 from pykern.pkdebug import pkdc, pkdexc, pkdlog, pkdp
-from pykern import pkcollections
+from pykern import pkcollections, pkconfig
```
robnagler, line 9: make it's own import. once we have a formatter, this won't be an issue.

sirepo/auth/email.py
```diff
@@ -59,11 +59,11 @@ def api_authEmailAuthorized(self, simulation_type, token):
         """
         if http_request.is_spider():
             sirepo.util.raise_forbidden('robots not allowed')
-        req = http_request.parse_params(type=simulation_type)
+        req = self.parse_params(type=simulation_type)
```
robnagler, line 62: remove the import of http_request here and any other place it is not needed.

sirepo/comsol_register.py
```diff
@@ -39,8 +38,8 @@ def api_comsolRegister(self):
 Name: {}
 Email: {}
 '''.format(req.name, req.email),
-    )
-    return http_reply.gen_json_ok()
+        )
```
robnagler, line 41: thanks for the attention to detail.

robnagler (changes requested):
Looks great, minor changes.

sirepo/auth/email.py
rorour: this file still uses http_request.is_spider() on line 60, did you want me to remove that too?

sirepo/api_auth.py
rorour: b7a43c1b48e8e47121aa5d2263a14432e5409dcd

sirepo/auth/email.py
robnagler: no. let's do this incrementally. only modules which no longer rely on http_request should have their imports removed.

robnagler (approved):

PR #4455, opened 2022-05-26
Fix #4334: srw to shadow popup

### LINKED ISSUE #4334: create SRW to Shadow sim avoiding popup blocker
When the SRW to Shadow sim is created, browsers will report a blocked popup. Instead, the process should finish and show a link in the panel to the newly created simulation. Clicking the link should open the Shadow sim in a new browser window.

gurhar1133:
SRW to shadow sim now avoids popup by generating a link in the modal

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -850,7 +850,7 @@ SIREPO.app.directive('appFooter', function(appState, requestSender, srwService) 
         scope: {
             nav: '=appFooter',
         },
-    template: `
+        template: `
             <div data-common-footer="nav"></div>
             <div data-import-python=""></div>
             <div data-confirmation-modal="" data-is-required="" data-id="sr-shadow-dialog" data-title="Open as a New Shadow Simulation" data-modal-closed="resetURL()" data-cancel-text="{{ displayLink() ? \'Close\' : \'Cancel\' }}" data-ok-text="{{ displayLink() ? \'\' : \'Create\' }}" data-ok-clicked="openShadowSimulation()">


moellep:
Looks good. I tried it out and have a few items

Formatting would look better like this:
- buttons aligned right
- Cancel button changes to Close after sim has been created
- x close button in dialog header

Screen Shot 2022-05-26 at 2 47 45 PM

And the results something like this with a clickable link which opens in a new window. Also, if possible, dismiss the dialog when the link is clicked.

Screen Shot 2022-05-26 at 2 51 12 PM

moellep:
Also, each time you select the "Open as a New Shadow Simulation" it should be back in the initial state which allows you to create a totally new sim. Currently it is still showing the dialog with the results link from the previous create.

gurhar1133:
> Looks good. I tried it out and have a few items
> 
> Formatting would look better like this:
> 
> * buttons aligned right
> * Cancel button changes to Close after sim has been created
> * x close button in dialog header
> 
> Screen Shot 2022-05-26 at 2 47 45 PM
> 
> And the results something like this with a clickable link which opens in a new window. Also, if possible, dismiss the dialog when the link is clicked.
> 
> Screen Shot 2022-05-26 at 2 51 12 PM

What do you mean 'dismiss the dialog when the link is clicked'?

moellep:
After you select "Create", you would see the link to the shadow simulation. Clicking the link would open the Shadow simulation in a new window and dismiss the "Open as a New Shadow Simulation" modal in SRW. That way when you return to the browser window with the SRW sim, you don't need to select the Close button.

gurhar1133:
> After you select "Create", you would see the link to the shadow simulation. Clicking the link would open the Shadow simulation in a new window and dismiss the "Open as a New Shadow Simulation" modal in SRW. That way when you return to the browser window with the SRW sim, you don't need to select the Close button.

Gottcha

gurhar1133:
@moellep adding ```data-dismiss="modal"``` attribute to the <a> for the link makes it so the modal closes but then doesn't navigate to the link. Do you have other suggestions for closing the modal when clicking the link?

gurhar1133:
@moellep modal closes now on opening the link

moellep:
Formating and closing the modal looks good.

It still has the problem that you can't invoke the feature multiple times. When I create a new Shadow sim, and then return to the menu item in SRW, it doesn't allow me to create a new Shadow simulation.

gurhar1133:
> Formating and closing the modal looks good.
> 
> It still has the problem that you can't invoke the feature multiple times. When I create a new Shadow sim, and then return to the menu item in SRW, it doesn't allow me to create a new Shadow simulation.

I think I fixed what you are talking about. Take a look and LMK @moellep

PR #4460, opened 2022-05-27
fix #4337 encode login redirect

### LINKED ISSUE #4337: Strange error when logged out and go to shared link
To reproduce:
- make sure you are logged out of sirepo.com
- go to https://www.sirepo.com/srw#/beamline/Fc7wsTmf?tab=multi
- see the error
Screen Shot 2022-04-18 at 4 52 56 PM

If you go to https://www.sirepo.com/srw#/beamline/Fc7wsTmf? (no `tab=multi`) then the link works.

garsuga:


moellep:
I'm seeing some cases when I test this that the encoded value has trailing "=" characters which is causing the sirepo cookie code to fail. This may be related to the length of the url being shared. 

```
angular.js:15567 Error: prev: Cookie name/value cannot contain delimiters |:;=
    at add (sirepo.js?20220531.162618:4095:19)
    at svc.addCookie (sirepo.js?20220531.162618:4048:9)
    at saveCookieRedirect (sirepo.js?20220531.162618:2057:23)
```

For example, attempting to share this url created this invalid encoded value:
http://v2.radia.run:8000/srw#/source/vLJ52U48 --> "srw L3NvdXJjZS92TEo1MlU0OA=="

@robnagler do you have an opinion on how to further escape the output from btoa() to avoid this?

robnagler:
Probably should just use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent). I think it does the right thing for cookie values.

garsuga:
> Probably should just use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent). I think it does the right thing for cookie values.

That is a good tip, I didn't know about those functions. `encodeUriComponent` appears to work fine. I have swapped the old calls for these.

moellep:
Works great!

PR #4476, opened 2022-06-01
Fix #4448: undulator length auto updates

### LINKED ISSUE #4448: updating idealized undulator length should update beam drift
In SRW when you edit the Length of an Idealized Undulator, it should update the Electron Beam Drift if the beam's "Drift Calculation Method" is set to Automatic.

For example, if you select the "NSLS-II CHX beamline" example, and change the Idealized Undulator Length to 3.1, the Single-Electron Spectrum report is all wrong:
Single-Electron-Spectrum-20-5m

If you edit the Electorn Beam, and then toggle the Drift Calculation Method between Automatic and Manual, the value changes from -1.54 to the correct value -1.59 and the Single-Electron Spectrum now looks correct again:

Single-Electron-Spectrum-20-5m (1)

That change should happen automatically in this case.

gurhar1133:
The undulator seems to be auto updating correctly now

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -177,7 +177,9 @@ SIREPO.app.factory('srwService', function(activeSection, appDataService, appStat
         });
     };
 
-    self.computeBeamParameters = function() {
+    self.computeBeamParameters = function(callback=null) {
+
+        srdbg('computingBeamParams');
         requestSender.sendStatelessCompute(
             appState,
             function(data) {
@@ -189,6 +191,9 @@ SIREPO.app.factory('srwService', function(activeSection, appDataService, appStat
                     ebeam[f] = data[f];
                 });
                 appState.models.electronBeamPosition.drift = data.drift;
+                if (callback) {
+                    callback();
+                }
             },
             {
                 method: 'process_beam_parameters',
@@ -820,7 +825,10 @@ SIREPO.app.controller('SourceController', function (appState, panelState, srwSer
     $scope.$on('modelChanged', function(e, name) {
         if (name == 'undulator' || name == 'tabulatedUndulator') {
             // make sure the electronBeam.drift is also updated
-            appState.saveQuietly('electronBeamPosition');
+            srwService.computeBeamParameters(() => {
+                    appState.saveChanges(['electronBeamPosition', 'electronBeam']);
+                }
+            )
         }
         else if (name == 'gaussianBeam') {
             appState.models.sourceIntensityReport.photonEnergy = appState.models.gaussianBeam.photonEnergy;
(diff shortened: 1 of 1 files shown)


sirepo/package_data/static/js/srw.js
```diff
@@ -177,7 +177,7 @@ SIREPO.app.factory('srwService', function(activeSection, appDataService, appStat
         });
     };
 
-    self.computeBeamParameters = function() {
+    self.computeBeamParameters = function(callback=null) {
```
robnagler, line 180: `callback` should be named something like `replyHandler`.

sirepo/package_data/static/js/srw.js
```diff
@@ -820,7 +823,10 @@ SIREPO.app.controller('SourceController', function (appState, panelState, srwSer
     $scope.$on('modelChanged', function(e, name) {
         if (name == 'undulator' || name == 'tabulatedUndulator') {
             // make sure the electronBeam.drift is also updated
-            appState.saveQuietly('electronBeamPosition');
+            srwService.computeBeamParameters(() => {
```
robnagler, line 826: I'm concerned about this. The changes are sitting somewhere and then there's an async call which may never return (user closes laptop). The assumption you can just save a random set of changes "some time later" is not a good one. It seems to me that the saveChanges should be happening in the sendStatelessCompute callback, but I could be wrong. However, again, there's no synchronization with the model state at the time of the request and in the response.

sirepo/package_data/static/js/srw.js
```diff
@@ -1462,10 +1468,6 @@ SIREPO.viewLogic('tabulatedUndulatorView', function(appState, panelState, reques
     $scope.whenSelected = updateUndulator;
     $scope.watchFields = [
         ['tabulatedUndulator.undulatorType'], updateUndulator,
-        [
```
robnagler, line 1465: This seems ok. I just don't know enough about the change.

robnagler (changes requested):

sirepo/package_data/static/js/srw.js
gurhar1133: We don't need that watchfield anymore since line 823 -
```javascript
 $scope.$on('modelChanged', function(e, name) {
        if (name == 'undulator' || name == 'tabulatedUndulator') {
```
listens and triggers self.computeBeamParameters

sirepo/package_data/static/js/srw.js
gurhar1133: Not sure I follow here
```
if (replyHandler) {
    replyHandler();
}
```
occurs in the `requestSender.sendStatelessCompute` call. Do you mean something else by saveChanges "should be happening in sendStatelessCompute"?

PR #4478, opened 2022-06-01
Fix #4133: is_save_images now False

### LINKED ISSUE #4133: Turn off image saving
> is_save_images=False

This needs to be off on parallel runs if not always, since I doubt we use them.

gurhar1133:
is_save_images is now False by default. examples/"sample from image" coherent results and partially coherent results both work with this change.


gurhar1133:
@robnagler Thoughts here?

robnagler:
I don't actually know, but I'll approve since it is a small change.

robnagler (approved):

PR #4487, opened 2022-06-04
fix #4482: Change http_reply.gen_* to self.reply.*

### LINKED ISSUE #4482: Change http_reply.gen_* to self.reply.*
Change http_reply.gen_* to self.reply.* (class methods)

_Originally posted by @robnagler in https://github.com/radiasoft/sirepo/issues/4388#issuecomment-1126771252_

rorour:


rorour:
@robnagler We don't seem to be setting `type` on the `sapi` object, so I didn't remove `sim_type` as a parameter from `login_success_response()` yet. Should we add that somewhere?
Also, some tests are failing because we are getting to line 289 in `auth.login_success_response() `without` sapi`.

robnagler (approved):
Looks good, thanks.

PR #4499, opened 2022-06-08
Fix #4493 - added io timeout

### LINKED ISSUE #4493: job_timeout_test failing
tests/job_timeout_test.py is failing beginning with commit 6194651695f1822319eb5501c816822b26dfd347. 

```
=================================== FAILURES ===================================
_____________________________ test_myapp_analysis ______________________________
  File "/home/vagrant/src/radiasoft/sirepo/tests/job_timeout_test.py", line 81, in test_myapp_analysis
    pkunit.pkok(r.status_code == 404, 'r={}', r)
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkunit.py", line 339, in pkok
    pkfail(fmt, *args, **kwargs)
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkunit.py", line 305, in pkfail
    raise PKFail('{} {}'.format(call, msg))
pykern.pkunit.PKFail: tests/job_timeout_test.py:81:test_myapp_analysis r=<WrapperTestResponse streamed [200 OK]>
=================== 1 failed, 1 passed in 123.30s (0:02:03) ====================
```

mkeilman:
This was falling through for ```downloadDataFile``` because there was no ```max_secs``` defined for an IO op. It seems like it deserves its own so I have added it - but the value is negotiable.


e-carlin (approved):
I don't know about the best time amount for it either but the concept looks good to me

mkeilman:
> I think the number should be computed based on the size of the object. The network speed is constant except of course NERSC, but that's very fast. Therefore, I don't think it is good to be a configuration parameter.

Previously downloadDataFile was using the "analysis" timeout (144s) then the "sequential" timeout (360s). For the purposes of this pr, perhaps using one of those is sufficient. Then we can have a new issue for a variable io timeout.

robnagler:
It can just reuse one of those or be computed (better).

moellep:
Is this good enough for Tuesday's alpha, or does it need more work?

mkeilman:
> Is this good enough for Tuesday's alpha, or does it need more work?

IMHO setting the timeout to 144 is good enough to be getting on with...getting the file size at the point we need to set ```max_secs``` is proving tricky, and a fixed value has been working all this time.

PR #4502, opened 2022-06-08
fix #4484: add user email column to adm request list

### LINKED ISSUE #4484: api_getModerationRequestRows add email
When you visit `myapp#/adm-roles` you see just the columns from user_role_invit_t
Screen Shot 2022-06-03 at 9 19 10 AM

Add "email" as the first column. This will make it easier when moderating users to know what user one is moderating.

rorour:


sirepo/auth_db.py
```diff
@@ -310,6 +310,7 @@ def uids_of_paid_users(cls):
 
     class UserRoleInvite(UserDbBase):
         __tablename__ = 'user_role_invite_t'
+        email = sqlalchemy.Column(UserDbBase.STRING_NAME, primary_key=True)
```
e-carlin, line 313: We don't need to add this field to the table since the data is already in auth_email_user_t we could join on uid with it and get it if we ever need it from there.

e-carlin (changes requested):
Looks good. One comment

sirepo/auth_db.py
rorour: @e-carlin should I do the join in api_getModerationRequestRows then?

rorour: e49bbdccbaf2d849005479713d0b0f6cc190d19b

rorour:
@e-carlin with this join, if a user was in user_role_invite_t but for some reason not in auth_email_user_t , they wouldn’t appear on the adm moderation page. I’m assuming this should never be the case, but I wanted to point it out.

sirepo/auth_db.py
```diff
@@ -324,8 +325,16 @@ class UserRoleInvite(UserDbBase):
 
         @classmethod
         def get_moderation_request_rows(cls):
+            import sirepo.auth
```
e-carlin, line 328: Don't qualify imports inside of function. When you import like this the module is added to sys.modules and then ins't removed when we exit the function which it should be (a bug of python). So import like `from sirepo import auth`

sirepo/auth_db.py
```diff
@@ -324,8 +325,16 @@ class UserRoleInvite(UserDbBase):
...
             with sirepo.util.THREAD_LOCK:
-                return cls.search_all_by(status='pending') + cls.search_all_by(status='clarify')
+                t = sirepo.auth.get_module('email').UserModel
```
e-carlin, line 330: This can be outside of the THREAD_LOCK. When using locks it is good practice to do as little as possible inside the lock to reduce time that other threads may be waiting on the lock

sirepo/auth_db.py
```diff
@@ -324,8 +325,16 @@ class UserRoleInvite(UserDbBase):
...
+                return [r[1].as_pkdict().pkupdate(
```
e-carlin, line 331: Thanks for figuring out the query. sqlalchemy never makes much sense to me...

This is a bit of code golf but I see two minor issues with your query
1. You select everything from the auth_email_user_t when all you need is the user_name. We have such few users it doesn't really matter but if we had enough users to the point of being careful about what we are pulling into memory you'd want to select only the fields that you care about. 
2. You do filter and then another filter. You can reduce those to one call to filter. sqlalchemy ends up producing the same query regardless but I think it is a little cleaner to reduce to one.

What do you think of this?
```python
                q = cls._session().query(t, cls).with_entities(
                    t.user_name.label('email'),
                    *cls.__table__.columns,
                ).filter(
                    t.uid == cls.uid,
                    sqlalchemy.sql.expression.or_(cls.status == 'pending', cls.status == 'clarify')
                ).all()
            return [dict(zip(r.keys(), r)) for r in q]
```

e-carlin (commented):
Looks good! Few minor comments.

e-carlin:
> @e-carlin with this join, if a user was in user_role_invite_t but for some reason not in auth_email_user_t , they wouldn’t appear on the adm moderation page. I’m assuming this should never be the case, but I wanted to point it out.

Yes that's ok. All of the user_role_invite code expects email.

sirepo/auth_db.py
rorour: 6a18786a1c0ea80d306b558a3c9d83ef8d1df2d8

sirepo/auth_db.py
rorour: 6a18786a1c0ea80d306b558a3c9d83ef8d1df2d8

sirepo/auth_db.py
rorour: 6a18786a1c0ea80d306b558a3c9d83ef8d1df2d8

e-carlin (approved):

PR #4527, opened 2022-06-14
fix #4524 remove vtk global key listeners

### LINKED ISSUE #4524: 3d plot changing as input typed in unrelated field
In radia, if you type the letters 's', 'v', 'or 'w' in an input field (for example in the Max Iterations field), the Magnet View plot changes:

Type 's'  and the plot disappears. Type 'v' and it becomes dots. Type 'w' and it becomes wireframe:
Screen Shot 2022-06-14 at 11 17 59 AM

This problem affects all apps with 3D plots, ex. warpvnd.

moellep:


moellep:
@mkeilman please review this one

In VTK, they are adding key listeners to document which get fired anytime something is typed in a field anywhere:
https://github.com/Kitware/vtk-js/blob/cf86a866109741d6e6bca0e39d3378d4785ff49e/Sources/Rendering/Core/RenderWindowInteractor/index.js#L208

This change removes the global key listener, but I saw there was some other key bindings going on in sirepo-plotting-vtk.js which might break.

PR #4536, opened 2022-06-16
fix container-sirepo-ci#4: Add karma globally and fix test.sh

rorour:
radiasoft/container-sirepo-ci#4

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -10,9 +10,9 @@ test_err() {
 }
 
 test_js() {
-    npm run lint -- "${jsfiles[@]}"
+    jshint --config=etc/jshint.conf "${jsfiles[@]}"
     if [[ ! ${sirepo_test_no_karma:-} ]]; then
-        npm run test
+        karma start etc/karma-conf.js
     fi
 }
 
--- a/test.sh
+++ b/test.sh
@@ -10,9 +10,7 @@ test_err() {
 }
 
 test_js() {
-    if [[ ! -x ./node_modules/karma/bin/karma || ! -x ./node_modules/jshint/bin/jshint ]]; then
-        npm install
-    fi
+    karma --version && jshint --version || npm install
     npm run lint -- "${jsfiles[@]}"
     if [[ ! ${sirepo_test_no_karma:-} ]]; then
         npm run test
--- a/test.sh
+++ b/test.sh
@@ -10,7 +10,6 @@ test_err() {
 }
 
 test_js() {
-    karma --version && jshint --version || npm install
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -10,9 +10,7 @@ test_err() {
 }
 
 test_js() {
-    if [[ ! -x ./node_modules/karma/bin/karma || ! -x ./node_modules/jshint/bin/jshint ]]; then
-        npm install
-    fi
+    karma --version && jshint --version || npm install
```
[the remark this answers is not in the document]
robnagler: npm is quite the pain. I think we may need to install globally always. There doesn't seem to be a way to ask npm if a package is installed either globally or locally. You have to query each separately, which is annoying.

If we assume a global install, then we might as well just add it to sirepo-dev. This would eliminate the need for a check here.

To answer Evan's specific question:

```bash
local p
if ! [[ $(type -p karma) && $(type -p jshint) ]]; then
     npm install
fi
```

You don't need a `>& /dev/null` this way. 

Do you both agree?

rorour:
@robnagler ready for re-review

test.sh
```diff
@@ -10,9 +10,6 @@ test_err() {
 }
 
 test_js() {
-    if [[ ! -x ./node_modules/karma/bin/karma || ! -x ./node_modules/jshint/bin/jshint ]]; then
-        npm install
-    fi
     npm run lint -- "${jsfiles[@]}"
```
robnagler, line 13: This needs to be:

```sh
jshint --config=etc/jshint.conf "${jsfiles[@]}"
``` 

The global install doesn't not (thankfully) install the scripts:

```js
  "scripts": {
    "lint": "jshint --config=etc/jshint.conf",
    "test": "karma start etc/karma-conf.js"
  },
```

These should be removed from package.json, too, but at a later date.

test.sh
```diff
@@ -10,9 +10,6 @@ test_err() {
...
     if [[ ! ${sirepo_test_no_karma:-} ]]; then
         npm run test
```
robnagler, line 15: This needs to be:

```sh
karma start etc/karma-conf.js
```

robnagler (changes requested):

test.sh
rorour: 0d0cbd9727a002b1ea964084a02b7d51b8ab1f25

test.sh
rorour: 0d0cbd9727a002b1ea964084a02b7d51b8ab1f25

PR #4550, opened 2022-06-22
Fix #4503 - moved overlay sizing to style

### LINKED ISSUE #4503: VTK loading spinner misplaced if vtk axes are not defined
The spinner appears at in the top left corner rather than the center. Also the png snapshot is blank gray.

mkeilman:
`data-attr-width` etc. doesn't actually do anything - the overlay was expanding to fit the axes div, which had the size passed in to the directive. When that div was excluded the overlay collapsed. The sizing belongs in the style attribute.


mkeilman:
pinging @garsuga

garsuga (approved):
Looks good, I see I missed a closing tag for sr-screenshot on an earlier commit which is also fixed here.

PR #4575, opened 2022-07-04
radiasoft/download#233: env now in env_keep instead of kernel.json

e-carlin:


e-carlin:
We discussed that this isn't needed [in other environments](https://github.com/radiasoft/rsconf/pull/208#issuecomment-1176191774). But, for dev singleuser with our current version it is needed.

e-carlin:
I mean [`etc/run-jupyterhub`](https://github.com/radiasoft/sirepo/blob/8bf6d7c2fdf28c9979075d305d1988c4484a786a/etc/run-jupyterhub.sh)

PR #4576, opened 2022-07-05
Fix #4532: radia 3d axes update

### LINKED ISSUE #4532: Radia 3D visual axes do not update when mouse is outside visual
The axes in the 3D 'Magnet Viewer' visuals do not update when the mouse leaves the bounds of the visual while clicking and dragging but the visual continues to update

https://user-images.githubusercontent.com/55895241/173963193-d84c8354-2d18-4247-8651-070e5b8be26d.mov

gurhar1133:
Binding the listeners to the sim view seems to have fixed to issue.

@mkeilman while working on this I noticed a potential bug (on my branch and master). If you move the 3d element around a lot and then try to quickly click on the simulations link you may run into the js error:
```
 Cannot read properties of undefined (reading '0')
```
or 
```
 Cannot read properties of undefined (reading '1')
```
I think this has to do with the coords calculations for `lCoord` still being in progress 
```javascript
viewPortPoint(worldPoint) {
        // this is required to do conversions for different displays/devices
        const pixels = window.devicePixelRatio;
        this.worldCoord.setCoordinateSystemToWorld();
        this.worldCoord.setValue(worldPoint.coords());
        const lCoord = this.worldCoord.getComputedLocalDisplayValue()
            .map(x => x / pixels);
        return new SIREPO.GEOMETRY.Point(lCoord[0], lCoord[1]);
    }
```

Should I open an issue for this?

### DIFF
--- a/sirepo/package_data/static/js/sirepo-plotting-vtk.js
+++ b/sirepo/package_data/static/js/sirepo-plotting-vtk.js
@@ -2645,7 +2645,7 @@ SIREPO.app.directive('vtkDisplay', function(appState, geometry, panelState, plot
                     }
                 });
                 Object.keys(eventHandlers).forEach(function (k) {
-                    rw[k] = function (evt) {
+                    body[k] = function (evt) {
                         eventHandlers[k](evt);
                         if (hdlrs[k]) {
                             hdlrs[k](evt);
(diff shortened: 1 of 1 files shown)


mkeilman:
Yes, go ahead and open an issue. I wonder if it's object destruction vs. debounce delay...

mkeilman:
Zoom does not update the axes. I think that handler (`onwheel`) needs to stay on the render window

mkeilman (changes requested):
See my comment about the zoom handler

gurhar1133:
@mkeilman I think it's fixed now. LMK if anything else 👍

mkeilman (approved):
Works well!

mkeilman:
> @mkeilman can we merge this one? Paul says if you approve he does too.

Yep!

PR #4584, opened 2022-07-06
Fix #4581 - set element title in loop

### LINKED ISSUE #4581: SRW beamline element name change does not propagate to ml export
The list of ml-eligible elements fails to update

mkeilman:
Also changes some lets to consts as long as I'm in there


rorour (approved):
Looks good. Should we be using const wherever possible for javascript variables then?

mkeilman:
> Looks good. Should we be using const wherever possible for javascript variables then?

I think so - you can modify consts, you just can't re-assign them

PR #4585, opened 2022-07-06
Fix #4578 - jinja replace spaces with underscores

### LINKED ISSUE #4578: rsopt script fails when element names contain spaces
If the element is named 'foo bar' with a parameter 'baz' then the jinja generates the function `def set_rsopt_params_dummy(foo bar_ baz,):`. We'll need to remove or replace spaces.

mkeilman:
This adds a jinja macro to replace spaces with underscores in the element title. It makes no such substitutions in the model itself.


rorour (approved):
looks good to me

PR #4592, opened 2022-07-08
Fix #4583: fix refresh_user loop and add default_proprietary_sim_types back in

### LINKED ISSUE #4583: Jupyterlogin default_proprietary_sim_type is not working

e-carlin:
default_proprietary_sim_types:
This was removed in https://git.radiasoft.org/sirepo/pulls/4211 . Add it back in
to keep backwards compatibility.

refresh_user:
sirepo.jupyterhub.SirepoAuthenticator.refresh_user should not redirect. Doing so causes a redirect
loop. In the event that a user is not authenticated return False which is what the jupyterhub
protocol expects.


sirepo/feature_config.py
```diff
@@ -130,6 +131,11 @@ def b(msg, dev=False):
     _cfg = pkconfig.init(
         # No secrets should be stored here (see sirepo.job.agent_env)
         api_modules=((), set, "optional api modules, e.g. status"),
+        default_proprietary_sim_types=(
+            set(),
```
robnagler, line 135: frozenset()

robnagler (approved):
I will let you merge.

sirepo/feature_config.py
e-carlin: [9f4cd42](https://github.com/radiasoft/sirepo/pull/4592/commits/9f4cd42ddac349ea4d63da51aa50109a210b521b)

PR #4595, opened 2022-07-08
fix #4580: jupyter moderation get DNS name(s) associated with the IP

### LINKED ISSUE #4580: jupyter moderation
In the moderation email we get `<dns.resolver.Answer object at 0x7f1f4c221a58>` for the `DNS` field. We should unwrap that object to get the underlying value.

rorour:


sirepo/http_request.py
```diff
@@ -174,11 +174,13 @@ def _dns_reverse_lookup(ip):
...
         try:
             if ip:
-                return ', '.join(
-                    [str(i) for i in dns.resolver.resolve(
-                        dns.reversename.from_address(ip),
-                        "PTR"
-                    ).rrset.items]
+                return ", ".join(
+                    [
+                        str(i)
```
robnagler, line 179: join these two lines `str(i) for ...(`

robnagler (changes requested):

sirepo/http_request.py
rorour: Black did the formatting

robnagler: ah i was wondering

robnagler (approved):

PR #4598, opened 2022-07-08
Fix #4559 blicon run srw with rsopt

### LINKED ISSUE #4559: Use rsopt to run srw after ml data generation
Currently the ml data generation script invokes srw directly, using inefficient assignment of tasks to workers. Instead use rsopt, which is designed around efficient parallel execution. This will also help to inform #3699

mkeilman:
The generated script now uses rsopt to both create input data and run SRW with it. This allows us to remove a lot of conditional and inefficient code from the python.


mkeilman:
Ping @gurhar1133

sirepo/package_data/template/srw/exportRsOpt_post.py.jinja
```diff
@@ -0,0 +1,38 @@
...
+{%- endmacro -%}
+
+{%- macro fixSpaces(t) -%}
+{{ t | replace(" ", "_") }}
+{%- endmacro -%}
+
+def main():
+    import h5py
+    import numpy
+    data = numpy.load("{{ fileBase }}.npy")
```
gurhar1133, line 22: `data` might not be necessary. maybe:
```python
for r in numpy.load("{{ fileBase }}.npy"):
    ...
```
?

sirepo/package_data/template/srw/parameters.py.jinja
```diff
@@ -376,94 +366,24 @@ def _read_srw_file(filename):
...
-    param_set = param_arr[start:end]
-    for vals in param_set:
-        _rsopt_run_single(vals, proc_num, task_num)
-        task_num += 1
-
-
-def _rsopt_run_single(param_vals, proc_num, task_num):
-    vp = _rsopt_set_params(*param_vals)
+def rsopt_run({{ rsOptFuctionSignature() }}):
+    vp = _rsopt_set_params({{ rsOptFuctionSignature() }})
```
gurhar1133, line 370: Unless I am mistaken, it looks like `vp` is declared and only used once. If so maybe the variable is not necessary and you could call
```python
f = _get_beamline_param(
    _rsopt_set_params({{ rsOptFuctionSignature() }}),
    'ws_fni',
    '{{ ws_fni_desc }}'
)

?

gurhar1133 (approved):
Looks good 👍 
Left a couple comments/questions about minor style things.

sirepo/package_data/template/srw/parameters.py.jinja
mkeilman: It's a little unclear, but `{{srwMain}}`, references `vp` once jinja expands it.

PR #4611, opened 2022-07-14
4395 reset examples admin

gurhar1133:
As discussed with @moellep, this issue was hand tested by moving sims out of `/package_data/template/<sim_name>/examples` and `reset_examples` successfully removed. Also tested by messing with the `.lastModified` attr on `run/user/.../sirepo-data.json` to test that sims older than 6 months are deleted and replaced

Regarding handling corrupted data, I removed/altered necessary fields from sims that `reset_examples`/`create_examples` was trying to access and try/except prints the name of the simulation file which isn't readable and continues

@moellep lmk if more needed for handling bad data file

@robnagler @moellep should prob sign off on this one when he gets back from vacation, but if you have thoughts on this one, LMK


sirepo/pkcli/admin.py
```diff
@@ -6,13 +6,15 @@
 """
 from __future__ import absolute_import, division, print_function
```
robnagler, line 7: remove

sirepo/pkcli/admin.py
```diff
@@ -44,6 +50,25 @@ def audit_proprietary_lib_files(*uid):
 
 def create_examples():
     """Adds missing app examples to all users"""
+
+    def _create():
+        for sim_type in feature_config.cfg().sim_types:
+            names = [
```
robnagler, line 56: Since this is used with `in` use a set().

robnagler, line 56: This is done twice, and really, three times. Create a function that returns `{name: example}` and then you can reuse in three places.

sirepo/pkcli/admin.py
```diff
@@ -44,6 +50,25 @@ def audit_proprietary_lib_files(*uid):
...
+                x.name
+                for x in simulation_db.iterate_simulation_datafiles(
+                    sim_type,
+                    simulation_db.process_simulation_list,
+                    {"simulation.isExample": True},
+                )
+            ]
+            for example in simulation_db.examples(sim_type):
```
robnagler, line 64: simulation_db.examples(sim_type) is a constant so cache in a dict outside the call to _access_sim_db_and_callback so you don't incur the cost of an expensive loop invariant for every user. This would be reused below, too.

sirepo/pkcli/admin.py
```diff
@@ -44,6 +50,25 @@ def audit_proprietary_lib_files(*uid):
...
+                if example.models.simulation.name not in names:
+                    _create_example(example)
+
+    _access_sim_db_and_callback(_create)
+
+
+def _access_sim_db_and_callback(callback):
```
robnagler, line 71: I would call this `_iterate_all_users` and make it a generator so you would use it:

```py
for _ in _iterate_all_users():
     the code that does the work
```

sirepo/pkcli/admin.py
```diff
@@ -55,20 +80,50 @@ def create_examples():
...
+    _access_sim_db_and_callback(_reset)
+
+
+def _build_delete_and_revert_lists(delete_list, revert_list, simulations, sim_type):
+    n = [n.models.simulation.name for n in simulation_db.examples(sim_type)]
+    for sim in simulations:
+        t = (
+            srtime.utc_now_as_milliseconds() - sim.simulation.lastModified
+        ) / _MILISECONDS_PER_MONTH
+        if sim.name not in n:
```
robnagler, line 112: Do this test first and continue.

sirepo/pkcli/admin.py
```diff
@@ -55,20 +80,50 @@ def create_examples():
...
+            delete_list.append((sim, sim_type))
+        elif t > _MAXIMUM_SIM_AGE_IN_MONTHS:
```
robnagler, line 114: Make this:
  
```py
if _example_is_too_old(sim.simulation.lastModified):
```

This will simplify the test so you don't need a temporary.
    revert_list.

sirepo/pkcli/admin.py
```diff
@@ -55,20 +80,50 @@ def create_examples():
...
-                    if example.models.simulation.name not in names:
-                        _create_example(example)
+            callback()
+
+
+def reset_examples():
+    def _reset():
+        c = []
+        r = []
+        for sim_type in feature_config.cfg().sim_types:
```
robnagler, line 90: I think the word "sim" appears too many times so labeling this `t` will make it easier to see in the code.

sirepo/pkcli/admin.py
```diff
@@ -55,20 +80,50 @@ def create_examples():
...
+                )
+            ]
+            _build_delete_and_revert_lists(c, r, sims, sim_type)
+        _revert_sims(r)
+        _delete_sims(c)
+
+    _access_sim_db_and_callback(_reset)
+
+
+def _build_delete_and_revert_lists(delete_list, revert_list, simulations, sim_type):
```
robnagler, line 106: I don't like that they are modifying values in this particular way. I would say create a `ops = PKDict(delete=[], revert=[])` and pass ops to _build_ops (_build_delete_and_revert_lists is too complicated a name). Then rename functions _delete and _revert, which are all nested functions in reset.

robnagler (changes requested):

rorour (approved):

sirepo/pkcli/admin.py
gurhar1133: I think simulation_db.examples(sim_type) is going to vary by user. 

I think we need 
```python
import sirepo.auth_db
import sirepo.server

sirepo.server.init()
for d in pkio.sorted_glob(simulation_db.user_path().join("*")):
    if _is_src_dir(d):
        continue
    uid = simulation_db.uid_from_dir_name(d)
    with sirepo.auth_db.session_and_lock(), auth.set_user_outside_of_http_request(
         uid
    ):
```
to get the relevant types to pass into `simulation_db.examples(sim_type)`

otherwise we run into: `AssertionError: no _non_threaded_context`

robnagler: [simulation_db.examples](https://github.com/radiasoft/sirepo/blob/8bf6d7c2fdf28c9979075d305d1988c4484a786a/sirepo/simulation_db.py#L144) does not vary by user. [simulation_db.iterate_simulation_data_files](https://github.com/radiasoft/sirepo/blob/8bf6d7c2fdf28c9979075d305d1988c4484a786a/sirepo/simulation_db.py#L282) does.

sirepo/pkcli/admin.py
```diff
@@ -141,9 +135,85 @@ def move_user_sims(target_uid=""):
...
+        elif _example_is_too_old(sim.simulation.lastModified):
+            ops.revert.append((sim.name, sim_type))
+            ops.delete.append((sim, sim_type))
+
+
 def _create_example(example):
-    simulation_db.save_new_example(example)
+    try:
+        simulation_db.save_new_example(example)
+    except Exception as e:
```
e-carlin, line 151: Why do we want to ignore exceptions here? Also, in the pkdlog below add `stack={}`. You can get the stack by calling `pkdexc()`

sirepo/pkcli/admin.py
```diff
@@ -141,9 +135,85 @@ def move_user_sims(target_uid=""):
...
+
+
+def _delete(ops):
+    for s, t in ops.delete:
+        simulation_db.delete_simulation(t, s.simulationId)
+
+
+def _example_is_too_old(last_modified):
+    t = (srtime.utc_now_as_milliseconds() - last_modified) / _MILLISECONDS_PER_MONTH
+    if t > _MAXIMUM_SIM_AGE_IN_MONTHS:
```
e-carlin, line 164: return `t > _MAXIMUM_SIM_AGE_IN_MONTHS`

sirepo/pkcli/admin.py
```diff
@@ -141,9 +135,85 @@ def move_user_sims(target_uid=""):
...
+
+def _get_example_by_name(name, sim_type):
+    for example in simulation_db.examples(sim_type):
+        if example.models.simulation.name == name:
+            return example
+    raise AssertionError(f"Failed to find example simulation with name={name}")
+
+
+def _get_examples_by_type():
+    e = PKDict()
```
e-carlin, line 177: I think a comprehension is simpler and more "pythonic"
`return PKDict({ t: simulation_db.examples(t) for t in feature_config.cfg().sim_types})`

sirepo/pkcli/admin.py
```diff
@@ -141,9 +135,85 @@ def move_user_sims(target_uid=""):
...
+
+
+def _get_named_sims(all_sim_types):
+    res = PKDict()
+    for t in all_sim_types:
+        res[t] = PKDict()
+        for x in simulation_db.iterate_simulation_datafiles(
+            t,
+            simulation_db.process_simulation_list,
+            {"simulation.isExample": True},
```
e-carlin, line 190: use PKDict

e-carlin (changes requested):

sirepo/pkcli/admin.py
gurhar1133: In the event of a corrupted data file. Not exactly sure on the particulars here and waiting for @moellep to get back from vacation to clarify.

robnagler: I think this maybe can be removed now. The point of this exercise is to make sure we don't have corrupt examples.

sirepo/pkcli/admin.py
gurhar1133: @robnagler I think I addressed caching of simulation_db.examples lmk what you think

sirepo/pkcli/admin.py
gurhar1133: > I think the corrupted data file is if a user has an existing example simulation that is corrupted. Not if we have an example that we are trying to create that is corrupted. We'd want to know and crash in that case.

I think that is the correct case and I think it also would make sense to take fail fast approach

sirepo/pkcli/admin.py
```diff
@@ -23,6 +24,10 @@
 import shutil
 
 
+_MILLISECONDS_PER_MONTH = 30.4167 * 24 * 60 * 60 * 1000
```
robnagler, line 27: This should not be 30.4167. It should be 30. That's sufficiently accurate for this purpose and keeps this number an int, which it should be.

sirepo/pkcli/admin.py
```diff
@@ -44,31 +49,20 @@ def audit_proprietary_lib_files(*uid):
...
-                ]
-                for example in simulation_db.examples(sim_type):
-                    if example.models.simulation.name not in names:
-                        _create_example(example)
+
+def reset_examples():
+    e = _get_examples_by_type()
+    for t, s in _iterate_sims_by_users(e.keys()):
+        ops = PKDict(delete=[], revert=[])
+        _build_ops(ops, list(s[t].values()), t, e)
```
robnagler, line 63: ops should be created in this function and returned by it.

sirepo/pkcli/admin.py
```diff
@@ -141,9 +135,87 @@ def move_user_sims(target_uid=""):
         shutil.move(lib_file, target)
 
 
+def _build_ops(ops, simulations, sim_type, examples):
+    n = set([x.models.simulation.name for x in examples[sim_type]])
+    for sim in simulations:
+        if sim.name not in n:
+            ops.delete.append((sim, sim_type))
```
robnagler, line 142: Does this delete all simulations, not just all simulation examples?

robnagler (changes requested):
Is there a test for this? If so, it should have non-example simulations to make sure they aren't deleted by this code.

sirepo/pkcli/admin.py
```diff
@@ -44,31 +49,20 @@ def audit_proprietary_lib_files(*uid):
 
 def create_examples():
     """Adds missing app examples to all users"""
-    import sirepo.auth_db
-    import sirepo.server
+    e = _get_examples_by_type()
+    for t, s in _iterate_sims_by_users(e.keys()):
+        for example in e[t]:
```
gurhar1133, line 54: * switch e and examples names

sirepo/pkcli/admin.py
gurhar1133: `simulations` name made it seem like all simulations when it was already filtered by examples

robnagler (approved):
I'll let Paul review when he gets back.

e-carlin (approved):
+1 paul can give final word

Issue #4607, opened 2022-07-11
ML functional Keras code generation


moellep:
We need to use the functional Keras calls to support skip connections. For example, the 2019 World Happiness example currently has the generated code:
```python
model = Sequential()
model.add(Dense(10, activation="relu", input_dim=6))
model.add(Dense(10, activation="relu"))
model.add(Dense(10, activation="sigmoid"))
model.add(Dense(1, activation="linear"))
```
This would need to be generated in a functional style:
```python
input_args = Input(shape=(6,))
x = Dense(10, activation="relu")(input_args)
x = Dense(10, activation="relu")(x)
x = Dense(10, activation="sigmoid")(x)
x = Dense(1, activation="linear")(x)
model = Model(input_args, x)
```
This will be more complicated when the layers branch and combine.

PR #4614, opened 2022-07-18
Fix: #4607 functional keras layers

gurhar1133:
Functional style keras layers are being generated by jinja now


mkeilman (approved):
Works and looks fine to me. There might be a way to make it "cuter" with jinja macros, but likely not worth the effort

sirepo/package_data/template/ml/build-model.py.jinja
```diff
@@ -1,26 +1,32 @@
...
-    {%- elif layer.layer == 'Dropout' -%}
-    {{ layer.dropoutRate }}
-    {%- elif layer.layer == 'Flatten' -%}
-    {%- elif layer.layer == 'GaussianDropout' -%}
-    {{ layer.gaussianDropoutRate }}
-    {%- elif layer.layer == 'GaussianNoise' -%}
-    {{ layer.gaussianNoiseStddev }}
-    {%- endif -%}
-    {%- if loop.index0 == 0 -%}, input_dim={{ inputDim }} {%- endif -%}))
+{% if loop.index0 == 0 %}
```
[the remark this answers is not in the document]
robnagler: Thanks. I think I wasn't clear. The naming is the problem. ActivationActivation should just be Activation is the "layer" is Activation so then you can control by using one name.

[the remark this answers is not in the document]
gurhar1133: Oh, ok. Is there no reason for denseActivation and activationActivation both being in the schema then?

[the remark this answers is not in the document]
robnagler: Where does layer.layer come from? In any event, this should all be in Python not Jinja.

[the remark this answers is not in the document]
gurhar1133: I'll work on the python over jinja issue. I think some substantial schema changes might be necessary for that end.

But to answer your question, `data.models. ... .neuralNetLayers` has layers. Each layer has an attribute `layer` referring to the layer type. Should it be renamed `layer.type`?

gurhar1133:
@robnagler layers now in python. LMK if any more changes

sirepo/template/ml.py
```diff
@@ -620,11 +620,57 @@ def _generate_parameters_file(data):
...
+            if i == 0:
+                c = f'({l.denseDimensionality}, activation="{l.denseActivation}")(input_args)'
+            else:
+                c = f"({_layer_args(l)})(x)"
+            res += f"x = {l.layer}{c}\n"
+        return res
+
+    return f"""
+from keras.models import Model, Sequential
+from keras.layers import Dense, Input {_import_layers(v)}
```
robnagler, line 665: remove space  after `Input`

sirepo/template/ml.py
```diff
@@ -620,11 +620,57 @@ def _generate_parameters_file(data):
...
         )
-    res += template_common.render_jinja(SIM_TYPE, v, "build-model.py")
+    res += _build_model_py(v)
     res += template_common.render_jinja(SIM_TYPE, v, "train.py")
     return res
 
 
+def _build_model_py(v):
+    def _import_layers(v):
+        i = ", ".join(n for n in v.layerImplementationNames if not n == "Dense")
```
robnagler, line 630: Simpler:

```py
return "".join(", " + n for n in v.layerImplementationNames if not n == "Dense")
```

sirepo/template/ml.py
```diff
@@ -620,11 +620,57 @@ def _generate_parameters_file(data):
...
+            return layer.dropoutRate
+        if l == "Flatten":
+            return ""
+        if l == "GaussianDropout":
+            return layer.gaussianDropoutRate
+        if l == "GaussianNoise":
+            return layer.gaussianNoiseStddev
+        raise ValueError(f"invalid layer.layer={l}")
+
+    def _build_layers(v):
```
robnagler, line 653:
```py
def _build_layers(layers):
```

sirepo/template/ml.py
```diff
@@ -620,11 +620,57 @@ def _generate_parameters_file(data):
...
+                c = f"({_layer_args(l)})(x)"
+            res += f"x = {l.layer}{c}\n"
+        return res
+
+    return f"""
+from keras.models import Model, Sequential
+from keras.layers import Dense, Input {_import_layers(v)}
+
+input_args = Input(shape=({v.inputDim},))
+{_build_layers(v)}
```
robnagler, line 668: `{_build_layers(v.neuralNetLayers)}`

sirepo/template/ml.py
```diff
@@ -620,11 +620,57 @@ def _generate_parameters_file(data):
...
+            return ""
+        if l == "GaussianDropout":
+            return layer.gaussianDropoutRate
+        if l == "GaussianNoise":
+            return layer.gaussianNoiseStddev
+        raise ValueError(f"invalid layer.layer={l}")
+
+    def _build_layers(v):
+        res = ""
+        for i, l in enumerate(v.neuralNetLayers):
```
robnagler, line 655: `for i, l in enumerate(v.neuralNetLayers):`

robnagler (approved):
That's great. Minor changes requested. I'll let somebody who knows what an NN is merge. :)

sirepo/template/ml.py
gurhar1133: You mean `for i, l in enumerate(layers):` correct? since we pass layers into `_build_layers`

sirepo/template/ml.py
```diff
@@ -620,11 +620,49 @@ def _generate_parameters_file(data):
...
+    res += _build_model_py(v)
     res += template_common.render_jinja(SIM_TYPE, v, "train.py")
     return res
 
 
+def _build_model_py(v):
+    def _import_layers(v):
+        return "".join(", " + n for n in v.layerImplementationNames if not n == "Dense")
+
+    def _layer_args(layer):
```
moellep, line 632: I think the "d = PKDict()" is actually a constant and shouldn't get recreated on each call to _layer_args(). I'd recommend putting that as a named variable within _build_model_py() but outside of _layer_args().

gurhar1133: But we pass the layer into `_layer_args` each time and that will change with each call.

moellep: Ah yes, I see. It is awkward that it constructs a full dict of values every time but really only needs one entry from it. It could be a constant PKDict of lambda functions for formatting, but the current implementation is good enough for now.

gurhar1133: I'll try out the lambda functions for the new layers issue. I was running into some problems and that might be a nice fix

PR #4620, opened 2022-07-19
Fix #4285: new particlematterinteractionType

### LINKED ISSUE #4285: update opal's particlematterinteraction command
It looks like the TYPE parameter for the PARTICLEMATTERINTERACTION command has changed in version 2021.1 and the options are now either SCATTERING or BEAMSTRIPPING.
See: http://amas.web.psi.ch/opal/Documentation/2021.1/#chp.partmatter

gurhar1133:
New types for particlematterinteraction command type; BEAMSTRIPPING, SCATTERING


sirepo/package_data/static/json/opal-schema.json
```diff
@@ -291,9 +291,8 @@
         ],
         "ParticlematterinteractionType": [
             ["", "NOT SELECTED"],
-            ["CCOLLIMATOR", "CCOLLIMATOR"],
```
e-carlin, line 294: I believe there should be a fixup made because these types have been removed. Idk if you've written a fixup before. If not I'm happy to share more info on how they're done.

e-carlin (approved):

sirepo/package_data/static/json/opal-schema.json
gurhar1133: I'll reach out to you when I fix the test failures. (maybe related to fixup) Thanks @e-carlin

gurhar1133: I removed DEGRADER files from the import test, if all tests pass, then I'll reach out about the fixup. In addition to removing references to the old particlematterinteractionTypes, will it be necessary to write tests involving SCATTERING and BEAMSTRIPPING types?

e-carlin: I don't have the best intuition for what will break so what should be tested. But, I think adding a test for one of SCATTERING OR BEAMSTRIPPING would be good. Some thoughts for writing the tests: change all instances of DEGRADER in the old test data to one of the new words. My guess is things will work since we are just parsing and not running opal. If we were running it may produce something invalid. Another idea is to go to the opal docs. Many codes have examples with their docs.

sirepo/sim_data/opal.py
```diff
@@ -34,6 +34,10 @@ def fixup_old_data(cls, data):
             if cmd._type == "filter":
                 cmd.type = cmd.type.upper()
             elif cmd._type == "particlematterinteraction":
+                if cmd.type not in [
```
gurhar1133, line 37: What do you think of this fixup @moellep ?

sirepo/sim_data/opal.py
```diff
@@ -34,6 +34,10 @@ def fixup_old_data(cls, data):
...
+                    t[1] for t in cls.schema().enum.ParticlematterinteractionType
```
e-carlin, line 38: I think this should be t[0]. t[1] is the display name and we store the value inside of models.

sirepo/sim_data/opal.py
```diff
@@ -34,6 +34,10 @@ def fixup_old_data(cls, data):
...
+                ]:
+                    cmd.type = "NOT SELECTED"
```
e-carlin, line 40: I think this should be `cmd.type = ""`. Again, we want to set it to the value not the display name

e-carlin (changes requested):

sirepo/sim_data/opal.py
gurhar1133: Good catch. Fixed

e-carlin:
lgtm but Paul should have the final say. He can also determine if a new test should be written to replace the removed one.

sirepo/sim_data/opal.py
moellep: Yes, that looks good. It doesn't look like particlematterinteraction is a very commonly used feature.

moellep:
This looks ready to merge. @e-carlin you have a change request which needs to get approved, I think.

e-carlin (approved):

e-carlin:
Sorry. lgtm. I merged.

PR #4628, opened 2022-07-20
fix #4627 do not allow override of uid

### LINKED ISSUE #4627: api_ownJobs should not allow override of args
This construct is a security hole:

```py
                uid=sirepo.auth.logged_in_user(), **self.parse_post()
```

robnagler:


rorour (approved):
I assume this means that previously if self.parse_post() contained the key 'uid' it would override the value sirepo.auth.logged_in_user() in _request_content?

robnagler:
Yes

PR #4631, opened 2022-07-20
Fix #4617 proxy react (npm server) through server.py

### LINKED ISSUE #4617: Proxy or emulate npm start via sirepo service http
Either we can emulate what Node is doing or we can proxy it. Either way, we'll need a way to edit files and have the server behave like it does when we do an npm start. We don't want to build every time.

robnagler:
@garsuga please test this branch by running `npm start` in your virtual machine, and then starting sirepo this way:

```sh
SIREPO_SERVER_REACT_SERVER=http://localhost:3000/ sirepo service http
```

I did not test updates to reactapp in real time. It's not clear how that works, because it only opens one port. However, I was able to get the basic app working just fine.

### DIFF
--- a/sirepo/api.py
+++ b/sirepo/api.py
@@ -58,6 +58,13 @@ class Base:
     def reply_html(self, path):
         return http_reply.render_html(path)
 
+    def reply_as_proxy(self, response):
+        r = http_reply.gen_response(response.content)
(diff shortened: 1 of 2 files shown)


robnagler:
@garsuga please review

garsuga (approved):
Looks good and seems to be working

PR #4644, opened 2022-07-26
Fix #4641: shadow srw conversion fix

### LINKED ISSUE #4641: shadow to srw conversion issues
If a shadow mirror has a "Mirror Dimensions" of Infinite, it shouldn't use the mirror width settings, but instead should set the mirror to a size 1e6 or similar.

Also, If a "Mirror Boundary" of Ellipse, use Width (X) and Length (Y) values, rather than the rectangular field values. The resulting SRW mirror will still be rectangular but the width/height should be set from the ellipse dimensions.

gurhar1133:
If a shadow mirror has a "Mirror Dimensions" of Infinite or if it has a "Mirror Boundary" of Ellipse, we set the srw mirror tangential and sagittal sizes to 10. 

Test data added for shadow mirrors with these properties


moellep:
For shadow mirrors with a mirror boundary of Ellipse or Ellipse with Hole, use the Width/Length fields (externalOutlineMajorAxis, externalOutlineMinorAxis) for the SRW size. That way the rough mirror size is reflected in the SRW sim.

gurhar1133:
> For shadow mirrors with a mirror boundary of Ellipse or Ellipse with Hole, use the Width/Length fields (externalOutlineMajorAxis, externalOutlineMinorAxis) for the SRW size. That way the rough mirror size is reflected in the SRW sim.

I think we discussed this already. externalOutlineMajorAxis and externalOutlineMinorAxis were too big for SRW and things broke. The limit ended up being 10:

https://github.com/radiasoft/sirepo/blob/6581a2ecdb574dd6042a3c8261216c952c0810fe/sirepo/template/srw_shadow.py#L785-L788

Do you want me to set to 10 if externalOutlineMajorAxis or externalOutlineMinorAxis are > 10? Also I only did the check for Ellipse and not with Ellipse with Hole, so I can change to cover both regardless.

gurhar1133:
@moellep made changes.

sirepo/template/srw_shadow.py
```diff
@@ -781,6 +782,13 @@ def __sim_to_srw(self, shadow, srw):
         srw.simulation.distanceFromSource = shadow.plotXYReport.distanceFromSource
         srw.initialIntensityReport.colorMap = shadow.initialIntensityReport.colorMap
 
+    def __size_srw_mirror(self, srw_mirror, shadow_mirror):
```
moellep, line 785: Still not quite right - if fhit_c is 0 the shadow mirror is infinite, so the 10m value should be used - none of the other values can be trusted. If fhit_c is 1 (finite) then use fshape to determine which fields to translate (rectangular use halfWidthX1, halfWidthX2, halfLengthY1, halfLengthY2), (ellipse or ellipse with hole use externalOutlineMajorAxis, externalOutlineMinorAxis)

Otherwise the SRW sim will fail if an infinite mirror is selected and all the shadow mirror sizes are 0.

gurhar1133:
@moellep Made changes

moellep:
Looks good!

PR #4651, opened 2022-07-28
fix radiasoft/raydata#17 Data Source tab won't display scans until analysis finishes running

rorour:


rorour:
@e-carlin This seems to have introduced an issue where polling for new scans is 'Waiting for another simulation to complete' when an analysis is already running. This wasn't an issue in f3012db38c35f2eba88c5c242412d5595329afec before changing the analysis to parallel. Any ideas why?

e-carlin:
Hmm my best guess is because the polling job is a parallel simulation too. So now it is contending for resources with the analyses. It is like we want a 3rd class of jobs or ability to set priority so the polling job always has priority over the analyses. Being able to set priority on jobs and re-order the queue is part of the work to do on raydata. So maybe the time is upon us to implement. I'll setup some time for us and @robnagler to talk.

e-carlin:
I'm a little confused now about what is going on. I ran the code on master and there is actually contention on cpu_slot not on op_slot. Contention on cpu_slot doesn't make sense since there is only one user and we have one sequential agent. Something strange is happening.

e-carlin:
After some discussion spurred on by https://github.com/radiasoft/sirepo/issues/4660 I think the best course of action is to revert back to using all non-animation (sequential) reports. We will then assign the sequential container more resources and allow more concurrent sequential jobs inside of the agent.

e-carlin:
@robnagler I need your advice on how to proceed:

Here is what I want to accomplish. x number of sequential run ops and y number of analysis ops running concurrently for a user (someday to be a group). I’d like to make sure that no matter how many sequential runs an agent is executing concurrently that it can always execute an analysis too.

No matter which option below that we pick we are going to need to do some of the same work
- Add configuration for op_slots so we can set run ops queue size to x and analysis op queue size to y.
- Add configuration so we can assign a configurable amount of cores to a sequential docker container (maybe x + y cores so everything gets it’s own)

Then we have two different options for ensuring that an agent is able to execute the analysis and run(s) concurrently:
1.  Make sure that the number of cpu_slots on the machine > x + y. This ensures that we can always run a sequential run and analysis concurrently. I think for now we can assume there will only be one user of the raydata app. Even once we deploy to BNL I think that will be the case for some time. So we can make sure that we pick values for x and y that don’t oversubscribe the machine for just that one user. And/or we could say there could be n number of users and pick values where cpu_slots * n > (x + y) * n
2. We can implement https://github.com/radiasoft/sirepo/issues/4660#issuecomment-1217079144 . active_agent_slots will be configured as a ratio between the number of cores on the machine and x + y. compute_bound_slots will be set to x + y. 

My preference would be option 2. I don’t foresee it being that much work to add the 2 new queues and it solves other issues…

Thanks for your help!

robnagler:
See https://github.com/radiasoft/sirepo/issues/4660#issuecomment-1217382022. I think we can implement active_agent_slots, and it will simplify things.

Every op_slot type could be configurable except a few (cancel, kill, etc.). The equation cpu_slots * n > (x + y) * n is not explicit. System sizing will be more complicated. We are allowing some combination of active_agent_slots and op_slots.

Note that is will reduce fairness. Consider M to be sum(agent_ops_max.values()) where agent_ops_max is {op_type: op_max}. The larger M is the less likely an agent can move from active to idle, that is, the busier an agent can get, the harder it will be to catch it in a quiescent state. We cannot shut down an agent with ops. That is not supported by the current implementation. Therefore, active users (consider multiple browser windows per user) could keep other users from resources. While this isn't a problem we have now, the larger M may cause it to be a problem, especially if we allow the supervisor to be the sole queue manager (planned for React GUI). 

I'm not sure if there is anything we can do here other than be aware to not make M too large.

We need to [allow cores to be configurable for sequential agents](https://github.com/radiasoft/sirepo/blob/9d3a7c4cafd1ac8679c5fbb3463a12421552b13e/sirepo/job_driver/docker.py#L105). This is a one line change, but I'm just noting here. We will want to be aware of making more cores available does not reduce the problem of oversubscribing a single agent. M again plays in here. If we guess that OP_IO doesn't need to be constrained as much, and we are wrong (e.g. an OP_IO consumes 4 cores), we can cause contention. 

Also note that M being large increases I/O contention on the host. Speaking of which, it's unclear if we need OP_IO if we implement this solution. It was introduced to allow a data file download while an analysis was running, but if we allow multiple analyses, then this is just normal contention within a single user's resources. Furthermore, RUN and ANALYSIS really are the "long" and "short" ops for a particular kind of agent with this new approach. Until we implement some "run time" parameter (and therefore can implement some type of priority queuing based on time), we might want to rename these ops to be "OP_LONG" and "OP_SHORT", since they have no other meaning that than this (and OP_RUN is confusing anyway, because OP_ANALYSIS [not OP_RUN] is being used for statefulCompute and statelessCompute operations).

We need to configure op_slot_q per agent type. This [will simplify sbatch](https://github.com/radiasoft/sirepo/blob/9d3a7c4cafd1ac8679c5fbb3463a12421552b13e/sirepo/job_driver/sbatch.py#L50).

We will want to be [more sophisticated with "situation" messages](https://github.com/radiasoft/sirepo/blob/9d3a7c4cafd1ac8679c5fbb3463a12421552b13e/sirepo/job_driver/__init__.py#L353) (how large is waiting queue vs running).

e-carlin:
I agree on all points. Thanks for filling in with details. We'll get going on implementing. Closing this PR for the time being to reduce noise but I'll point back to it and #4660 in the fix.

PR #4653, opened 2022-07-29
Fix #4648: more logs download

### LINKED ISSUE #4648: add `__srwl_logs__/*.log` to download for a simulation
This is is similar to the run.log. The name of the log varies by the simulation so may need a specific api for this.

Here's where should be added:

Screen Shot 2022-07-28 at 11 13 54 AM

I would could the current link "Download run log" and the new link "SRW internal log". Maybe @moellep has better suggestions.

gurhar1133:
Progress log download link added

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2456,6 +2456,8 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
         template: `
            <div data-ng-if="(simState.getFrameCount() > 0) || errorMessage()" class="well well-lg">
               <a style="position: relative;" href="{{ logFileURL() }}" target="_blank">SRW log file</a>
+              <br>
+              <a style="position: relative;" href="{{ progressLogURL() }}" target="_blank">SRW Progress log file</a>
            </div>
             <form name="form" class="form-horizontal" autocomplete="off" novalidate>
               <div data-canceled-due-to-timeout-alert="simState"></div>
@@ -2537,6 +2539,20 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
             }
 
             $scope.logFileURL = () => {
+                if (! appState.isLoaded()) {
+                    return '';
+                }
+                return  requestSender.formatUrl('downloadDataFile', {
+                    '<simulation_id>': appState.models.simulation.simulationId,
+                    '<simulation_type>': SIREPO.APP_SCHEMA.simulationType,
+                    '<model>': $scope.simState.model,
+                    '<frame>': -1,
+                    '<suffix>': 'run.log',
+                });
+            };
+
+            $scope.logFileURL = () => {
+                return 'sirepo.com';
                 if (! appState.isLoaded()) {
                     return '';
                 }
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2455,9 +2455,9 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
         },
         template: `
            <div data-ng-if="(simState.getFrameCount() > 0) || errorMessage()" class="well well-lg">
-              <a style="position: relative;" href="{{ logFileURL() }}" target="_blank">SRW log file</a>
+              <a style="position: relative;" href="{{ logFileURL() }}" target="_blank">SRW run log file</a>
               <br>
-              <a style="position: relative;" href="{{ progressLogURL() }}" target="_blank">SRW Progress log file</a>
+              <a style="position: relative;" href="{{ progressLogURL() }}" target="_blank">SRW progress log file</a>
            </div>
             <form name="form" class="form-horizontal" autocomplete="off" novalidate>
               <div data-canceled-due-to-timeout-alert="simState"></div>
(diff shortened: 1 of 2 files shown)


sirepo/template/srw.py
```diff
@@ -507,9 +509,19 @@ def get_application_data(data, **kwargs):
...
     if options.suffix == _SRW_LOG_FILE:
         return template_common.text_data_file(_SRW_LOG_FILE, run_dir)
+    if options.suffix == _PROGRESS_LOG_DIR:
+        d = run_dir.join(_PROGRESS_LOG_DIR)
+        f = _get_min_mtime_file(pkio.sorted_glob(d.join("*.log"))).path
+        return template_common.text_data_file(f.basename, d)
     return get_filename_for_model(model)
 
 
+def _get_min_mtime_file(files):
```
robnagler, line 519: Is this min_mtime or max_mtime? Instead of naming it by the algorithm, you should name it by the "what": most_recent.

sirepo/template/srw.py
```diff
@@ -507,9 +509,19 @@ def get_application_data(data, **kwargs):
 def get_data_file(run_dir, model, frame, options):
     if options.suffix == _SRW_LOG_FILE:
         return template_common.text_data_file(_SRW_LOG_FILE, run_dir)
+    if options.suffix == _PROGRESS_LOG_DIR:
+        d = run_dir.join(_PROGRESS_LOG_DIR)
+        f = _get_min_mtime_file(pkio.sorted_glob(d.join("*.log"))).path
```
robnagler, line 514: _get_min_mtime_file should just return basename.

sirepo/template/srw.py
```diff
@@ -158,6 +158,8 @@
 
 _SRW_LOG_FILE = "run.log"
```
robnagler, line 159: Drop this.

sirepo/template/srw.py
```diff
@@ -507,9 +509,19 @@ def get_application_data(data, **kwargs):
 def get_data_file(run_dir, model, frame, options):
     if options.suffix == _SRW_LOG_FILE:
```
robnagler, line 510: template_common.RUN_LOG

sirepo/template/srw.py
```diff
@@ -507,9 +509,19 @@ def get_application_data(data, **kwargs):
...
         return template_common.text_data_file(_SRW_LOG_FILE, run_dir)
```
robnagler, line 511: options.suffix

sirepo/template/srw.py
```diff
@@ -507,9 +509,19 @@ def get_application_data(data, **kwargs):
...
+    if options.suffix == _PROGRESS_LOG_DIR:
+        d = run_dir.join(_PROGRESS_LOG_DIR)
+        f = _get_min_mtime_file(pkio.sorted_glob(d.join("*.log"))).path
+        return template_common.text_data_file(f.basename, d)
     return get_filename_for_model(model)
 
 
+def _get_min_mtime_file(files):
+    m = [PKDict(path=f, time=os.path.getmtime(f)) for f in files]
```
robnagler, line 520: This would be simpler:

```py
return sorted(files, key=lambda f: f.mtime())[-1].basename
```

sirepo/package_data/static/js/srw.js
```diff
@@ -2596,6 +2596,16 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return true;
             }
 
+            function logFileRequest(suffix) {
+                return  requestSender.formatUrl('downloadDataFile', {
+                    '<simulation_id>': appState.models.simulation.simulationId,
+                    '<simulation_type>': SIREPO.APP_SCHEMA.simulationType,
+                    '<model>': $scope.simState.model,
+                    '<frame>': -1,
```
robnagler, line 2604: This really should be a constant....

sirepo/package_data/static/js/srw.js
```diff
@@ -2596,6 +2596,16 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return true;
             }
 
+            function logFileRequest(suffix) {
```
robnagler, line 2599: This is a bit funny, because it is really not a "suffix" but a logical name. You could call it log_kind, as the function, and it would get passed to suffix.

robnagler (changes requested):

sirepo/template/srw.py
gurhar1133: Oops, its supposed to be max, but I will change to most_recent

sirepo/package_data/static/js/srw.js
gurhar1133: constant declared where? if outside of `logFileRequest()` but in either `progressLogUrl()` or `logFileUrl()`, then it needs to be in both. If outside of all and declared at the top of the directive, then that wont work. If inside `logFileRequest()` then I am not sure I understand the point.

robnagler: -1 is something and should be documented as such:

```sh
$ grep -i -- frame.*.-1 *.js
opal.js:            '<frame>': -1,
radia.js:                    '<frame>': -1,
rs4pi.js:                        $scope.advanceFrame(-1);
sirepo-plotting.js:        scope.prevFrameIndex = -1;
sirepo-plotting.js:            scope.prevFrameIndex = -1;
sirepo-plotting.js:                    scope.prevFrameIndex = -1;
sirepo-plotting.js:                    scope.prevFrameIndex = -1;
sirepo-plotting.js:              <button type="button" class="btn btn-default" data-ng-disabled="isFirstFrame()" data-ng-click="advanceFrame(-1, true)"><span class="glyphicon gl\
yphicon-step-backward"></span></button>
sirepo-plotting.js:                $scope.prevFrameIndex = -1;
srw.js:                '<frame>': -1,
srw.js:                    '<frame>': -1,
zgoubi.js:                    '<frame>': appState.isAnimationModelName(modelKey) ? 1 : -1,
```

gurhar1133: oh. Thansk for clarifying 👍  I thought you meant the whole object. I'll make it a constant.

robnagler (approved):

rorour (commented):
looks good

PR #4665, opened 2022-08-03
Fix #4604: skip connections

### LINKED ISSUE #4604: ML Neural Net layers with skip connections
The current UI only support simple sequential models. It should be expanded to support branching and joining via "Add" and "Concatenate" elements.
For example, if an Add element was added, it would display two sections to define the elements in both branch:
Screen Shot 2022-07-11 at 3 40 19 PM

Nesting layers could be many layers deep - ex. one branch could include another Concatenate element.

gurhar1133:
Skip connections are now possible with Add layers which let you nest layers in UI
Backend `template/ml.py` generates functional layers which reflect the branching

@moellep lmk what you think about UI layout and any changes required there as well


gurhar1133:
Just found and edge case I am trying to work out.

mkeilman:
@gurhar1133 fyi the python-ci check failed but it looks like a back-end error (`ERROR: Failed building wheel for uwsgi`)

sirepo/package_data/static/js/ml.js
```diff
@@ -1517,7 +1565,7 @@ SIREPO.app.directive('neuralNetLayersForm', function(appState, mlService, panelS
             };
 
             $scope.deleteLayer = function(idx) {
-                appState.models.neuralNet.layers.splice(idx, 1);
+                $scope.layerLevel.splice(idx, 1);
```
mkeilman, line 1568: (comment here because there is no good place) - If you add a layer and then cancel:
1. The UI does not revert
2. If you then try to add a new layer you get this error:
```javascript
TypeError: Cannot read properties of undefined (reading 'layer')
    at $scope.fieldTrack (ml.js?20220804.150015:1575:30)
    at fn (eval at compile (angular.js:16418:15), <anonymous>:4:313)
    at z (angular.js:33785:18)
    at angular.js:33844:25
    at angular.js:19002:13
    at m.$digest (angular.js:19140:23)
    at b.$apply (angular.js:19500:13)
    at HTMLSelectElement.<anonymous> (angular.js:35709:9)
    at sg (angular.js:3824:3)
    at HTMLSelectElement.d (angular.js:3812:9)
```

sirepo/package_data/static/js/ml.js
```diff
@@ -1616,6 +1668,41 @@ SIREPO.app.directive('neuralNetLayersForm', function(appState, mlService, panelS
...
             }
 
+            function getLayerLevel() {
+                if ($scope.layerTarget){
+                    return $scope.layerTarget.layers;
+                }
+                return appState.models.neuralNet.layers;
+            }
+
+            function setLevelName(){
```
mkeilman, line 1678: Space after ()

sirepo/package_data/static/js/ml.js
```diff
@@ -1616,6 +1668,41 @@ SIREPO.app.directive('neuralNetLayersForm', function(appState, mlService, panelS
 
             }
 
+            function getLayerLevel() {
+                if ($scope.layerTarget){
```
mkeilman, line 1672: Space after ()

sirepo/package_data/static/js/ml.js
```diff
@@ -1616,6 +1668,41 @@ SIREPO.app.directive('neuralNetLayersForm', function(appState, mlService, panelS
...
+                    return $scope.layerTarget.layers;
+                }
+                return appState.models.neuralNet.layers;
+            }
+
+            function setLevelName(){
+                if ($scope.layerTarget){
```
mkeilman, line 1679: Space after ()

sirepo/package_data/static/js/ml.js
```diff
@@ -1616,6 +1668,41 @@ SIREPO.app.directive('neuralNetLayersForm', function(appState, mlService, panelS
...
+                    $scope.lName = $scope.layerTarget.name;
+                    return;
+                }
+                appState.models.neuralNet.name = "x";
```
mkeilman, line 1683: Single quotes (we may need to discuss moving to double quotes to match the new python style?)

sirepo/template/ml.py
```diff
@@ -582,6 +582,8 @@ def _fit_animation(frame_args):
 def _generate_parameters_file(data):
     report = data.get("report", "")
     dm = data.models
+    for l in dm.neuralNet.layers:
+        pkdp('\n\n\n\n layer -> {}', l)
```
mkeilman, line 586: Remove logging

mkeilman (changes requested):
Main issue is a bug with canceling added layers. The UI could use refinement but I wouldn't hold things up because of it

sirepo/package_data/static/js/ml.js
gurhar1133: Good catch. I'll work on that. @moellep  and I discussed some other changes as well that will be necessary for the upcoming network importing feature so I am converting this PR to draft until I have worked some stuff out.

sirepo/template/ml.py
```diff
@@ -626,6 +626,34 @@ def _generate_parameters_file(data):
...
+        layer_level.name = _name_layer(layer_level, first_level, parent_level_name)
+        layer_level.parent_name = parent_level_name
+        for i, l in enumerate(layer_level.layers):
+            if _branching(l):
+                for c in l.children:
+                    l.parent_name = layer_level.name
+                    if i == 0:
+                        _name_layers(c, parent_level_name)
+                    else:
+                        _name_layers(c, layer_level.name)
```
mkeilman, line 655: Or `_name_layers(c, parent_level_name if i == 0 else layer_level.name)`

mkeilman (changes requested):
I see a couple of UI bugs: canceling changes to the nn layers still does not revert the changes. Also "add" layers on the right do not get reordered when clicking the arrows.

Also, when I start new training, it fails with `name 'Dense' is not defined`. Indeed it is not imported in `parameters.py`. You might be missing the output layer when building the import list.

mkeilman:
> * Re the cancelling not working, Paul and I discussed and it looks like a deeper angular issue, maybe to be addressed on another issue or @moellep will look into it

OK

> * I'm not exactly sure what you mean about add layers not reordering. If I click the arrows I see add layers being moved up or down.

Oh,  I was expecting the two child panels to reorder, not their contents, but I guess that is consistent behavior. nvm

moellep:
Layout-wise, I have the following comments marked with red circles on the screenshot.
Screen Shot 2022-08-10 at 4 07 36 PM

A) Don't overlap the sub-panel on the cancel button - possibly layout slightly below the button.
B) Add a small top margin on the "Add another child" button.
C) Don't extend the sub-panels outside the main "Neural Network Layers" panel. If you want it to pop-out, expand the parent panel's right margin.
D) The "Add Layer" striped row's color should extend across the entire row, similar to the other striped rows above.

gurhar1133:
@moellep made the UI changes

mkeilman (approved):
Looks cool

moellep:
Here's some hacked HTML were I removed some columns and made the first "add" a colspan=5.

Screen Shot 2022-08-11 at 2 37 42 PM

PR #4671, opened 2022-08-04
Fix #4602 nuline url data source

### LINKED ISSUE #4602: ML import dataset from URLs
A new Data Source field should be added, “Data Origin” which has a dropdown for “From File” or “From URL”. If “From File” is selected, the application should proceed normally. If “From URL” is selected an URL field should be displayed for input, rather than the Data File input field. When OK is selected, the datafile will be downloaded and then installed in the user’s lib directory. 

For example, the Oxford pet data input could look like this:

Input and Output Format:	Image to Image
Input Data Origin:		From URL
Input Data URL: 		www.robots.ox.ac.uk/~vgg/data/pets/data/images.tar.gz
Input Data Path: 		images/*.jpg
…

The lib directory should be a cache of the URL datafile. So if a simulation is run, it will first check the simulation lib directory for the file, and if it is not it will download the data and store it as a lib file first.

mkeilman:
Notes:
- n.b. this is "file from somewhere to server", which differs from both downloadDataFile ("file from server to client") and uploadFile ("file from client to server")
- The interesting bit is using `urllib` in a stateless compute to get the file into the `lib` directory. I think this is the correct way but lmk if not. Anyway using a "simulation" fails because it cannot access `lib`
- The data is written in chunks so that we can query the file size along the way
- On the client side, we invoke stateless compute twice: once to get the file size (actually all the headers) and again to trigger the download. What we need is something similar to a simulation (but again not an actual "simulation"), I think. (as an aside, should this process be an `OP_IO`? Seems suited to it)
- `getBytesLoaded` is intended to be run in a loop so we keep the user informed of the progress. However it is not easy to use the run queue for arbitrary purposes. A hand-crafted loop fails because the stateless compute has to wait on the download to finish, defeating its purpose (see comment re `OP_IO` above)
- The UI includes a data format setting (image/text) whose current purpose is just to keep a bunch of "error" reports from showing up. It will tie into loading image data
- tldr; the code works but needs further attention


sirepo/package_data/static/js/ml.js
```diff
@@ -30,6 +30,14 @@ SIREPO.app.config(function() {
...
           <div data-trim-button="" data-model-name="modelName" data-model="model" data-field="field"></div>
         </div>
+        <div data-ng-switch-when="URL" class="col-sm-5" data-field-class="fieldClass">
+          <input type="text" data-ng-model="model[field]" class="form-control" style="text-align: right" data-lpignore="true" />          
+          <span data-ng-show="model.dataOrigin === 'url'" style="font-style: italic; font-size: 80%;">{{ model.bytesLoaded || 0 }} of {{ model.contentLength || 0 }} bytes</span>
+          <div class="sr-input-warning"></div>
+          <div data-ng-show="model.contentLength && ! model.bytesLoaded" class="progress">
+            <div class="progress-bar progress-bar-striped active" role="progressbar" aria-valuenow="100" aria-valuemin="0" aria-valuemax="100" style="width:100%"></div>
+          </div>
+        </div>
```
moellep, line 40: You can change the URL type handler class from "col-sm-5" to "col-sm-7" and it will show in a larger field. You can then remove the style="text-align: right" which was probably a work-around for the small field.

moellep:
When I create a new simulation, select Regression, URL, and set a URL to a csv file, with data format=text and select Save Changes, it doesn't load the data into the plots.

If we want the "Data Origin" to be only for Image data, we should make "Data Format" the 2nd field on the panel and hide "Data Origin" and "Data URL" if Text is selected.

sirepo/package_data/static/js/ml.js
```diff
@@ -1943,20 +1898,188 @@ SIREPO.viewLogic('partitionView', function(appState, panelState, $scope) {
...
+        const partition = appState.models.partition;
+        if (dataFile.appMode === 'regression'
+            && partition.training + partition.testing >= 100) {
+            ['training', 'testing', 'validation'].forEach(function(f) {
+                delete partition[f];
+            });
+            appState.setModelDefaults(partition, 'partition');
+        }
+        else if (dataFile.appMode === 'classification') {
+            if (partition.training + partition.testing < 100) {
```
robnagler, line 1956: This deserves a comment. Why the fixup?

sirepo/template/ml.py
```diff
@@ -287,13 +287,83 @@ def stateful_compute_compute_column_info(data):
...
+    strides={layer.strides},
+    padding="{layer.padding}"
+    """
+
+    def _pooling_args(layer):
+        return f'''pool_size=({layer.size}, {layer.size}),
+    strides={layer.strides},
+    padding="{layer.padding}"'''
+
+    args_map = PKDict(
```
robnagler, line 324: Should be defined before other functions.

sirepo/template/ml.py
```diff
@@ -287,13 +287,83 @@ def stateful_compute_compute_column_info(data):
...
+        GlobalAveragePooling2D=lambda layer: "",
+        MaxPooling2D=lambda layer: _pooling_args(layer),
+        SeparableConv2D=lambda layer: _conv_args(layer),
+        Conv2DTranspose=lambda layer: _conv_args(layer),
+        UpSampling2D=lambda layer: f'size={layer.size}, interpolation="{layer.interpolation}"',
+        ZeroPadding2D=lambda layer: f"padding=({layer.padding}, {layer.padding})",
+    )
+
+    def _layer_args(layer):
+        assert layer.layer in args_map, ValueError(f"invalid layer.layer={layer.layer}")
```
robnagler, line 344: I don't think _layer_args is necessary. Returning ValueError on assert doesn't make sense. Just let it naturally fail fast.

sirepo/template/ml.py
```diff
@@ -287,13 +287,83 @@ def stateful_compute_compute_column_info(data):
...
+        return args_map[layer.layer](layer)
+
+    def _build_layers(layers):
+        res = ""
+        for i, l in enumerate(layers):
+            if i == 0:
```
robnagler, line 350: Share code:

```py
c = "input_args" if i == 0 else "x"
res += f"x = {l.layer}({args_map[l.layer](l)})({c})"
```

sirepo/template/ml.py
```diff
@@ -326,6 +396,20 @@ def _get_matrix():
...
 
+def _cols_with_non_unique_values(filename, has_header_row, header):
+    # TODO(e-carlin): support npy
+    assert not re.search(
+        r"\.npy$", str(filename)
+    ), f"numpy files are not supported path={filename}"
+    v = sirepo.numpy.ndarray_from_csv(_filepath(filename), has_header_row)
+    res = PKDict()
+    for i, c in enumerate(np.all(v == v[0, :], axis=0)):
+        if not c:
```
robnagler, line 407: unnecessary use of continue. Flip the condition.

sirepo/template/ml.py
```diff
@@ -736,6 +744,43 @@ def _get_fit_report(report, x_vals, y_vals):
...
+    filename = os.path.basename(urllib.parse.urlparse(url).path)
+    try:
+        with urllib.request.urlopen(url) as r:
+            if headers_only:
+                return PKDict(headers=_header_str_to_dict(r.headers))
+            chunk_size = 1024 * 1024
+            with open(
+                _SIM_DATA.lib_file_write_path(
+                    _SIM_DATA.lib_file_name_with_model_field(
+                        "dataFile", "file", filename
```
robnagler, line 762: trailing comma

sirepo/template/ml.py
```diff
@@ -736,6 +744,43 @@ def _get_fit_report(report, x_vals, y_vals):
...
+    import urllib
+    from urllib import parse
+    from urllib import request
+
+    filename = os.path.basename(urllib.parse.urlparse(url).path)
+    try:
+        with urllib.request.urlopen(url) as r:
+            if headers_only:
+                return PKDict(headers=_header_str_to_dict(r.headers))
+            chunk_size = 1024 * 1024
```
robnagler, line 758: This is a constant. Move out off the try and possibly global scope.

sirepo/template/ml.py
```diff
@@ -736,6 +744,43 @@ def _get_fit_report(report, x_vals, y_vals):
...
+                    )
+                ),
+                "wb",
+            ) as f:
+                while True:
+                    c = r.read(chunk_size)
+                    if not c:
+                        break
+                    f.write(c)
+    except urllib.error.HTTPError as e:
```
robnagler, line 772: Why are we return this error and not other errors?

sirepo/template/ml.py
```diff
@@ -736,6 +744,43 @@ def _get_fit_report(report, x_vals, y_vals):
...
+        return PKDict(error=e.code)
+    return PKDict(filename=filename)
+
+
+# if this conversion is not done, the header gets returned as a newline-delimited string
+# EmailMessage headers pseduo-dicts and can have duplicated keys, which we ignore
+def _header_str_to_dict(h):
+    d = {k: v for k, v in h.items()}
```
robnagler, line 780: temporary unnecessary. Why isn't this just h.copy?

sirepo/template/ml.py
```diff
@@ -808,6 +853,23 @@ def _read_file_with_history(run_dir, filename, report=None):
     return res
 
 
+def _remote_data_bytes_loaded(filename):
+    import os
```
robnagler, line 857: This can be up top

sirepo/template/ml.py
```diff
@@ -736,6 +744,43 @@ def _get_fit_report(report, x_vals, y_vals):
     return param_vals, param_sigmas, plots
 
 
+def _get_remote_data(url, headers_only):
+    import os
```
robnagler, line 748: Move these imports up top and fully qualify them.

sirepo/package_data/static/js/ml.js
```diff
@@ -1943,20 +1898,188 @@ SIREPO.viewLogic('partitionView', function(appState, panelState, $scope) {
...
+            }
+        );
+    }
+
+    function computeDefaultPartition() {
+        const size = appState.models.columnInfo.rowCount;
+        const partition = appState.models.partition;
+        if (! partition.cutoff0 || ! partition.cutoff1
+            || partition.cutoff0 > size
+            || partition.cutoff1 > size) {
```
robnagler, line 1938: bring over "){" so conditional separates from block.

robnagler (changes requested):

sirepo/package_data/static/js/ml.js
mkeilman: [2683a56](https://github.com/radiasoft/sirepo/pull/4671/commits/2683a56139bbe105b349d17c92a723693e839d80)

I think the text align was from a copy-paste job

mkeilman:
> When I create a new simulation, select Regression, URL, and set a URL to a csv file, with data format=text and select Save Changes, it doesn't load the data into the plots.
> 
> If we want the "Data Origin" to be only for Image data, we should make "Data Format" the 2nd field on the panel and hide "Data Origin" and "Data URL" if Text is selected.

We should support urls for both, including reading them from archives. I'll get something in place

sirepo/package_data/static/js/ml.js
mkeilman: [f90e3ef](https://github.com/radiasoft/sirepo/pull/4671/commits/f90e3efda45cc21eb1b8a354ef981d7b5e4c3358)

sirepo/package_data/static/js/ml.js
mkeilman: One for @moellep. I just moved these functions bodily from controller to viewLogic.

sirepo/template/ml.py
mkeilman: [0ca67b9](https://github.com/radiasoft/sirepo/pull/4671/commits/0ca67b9604d6873b11bb2249fbd9a6565dca7226)

sirepo/template/ml.py
mkeilman: [bd53c46](https://github.com/radiasoft/sirepo/pull/4671/commits/bd53c468e02d539b2eb871f87914ee0c2a6c2f98)

sirepo/template/ml.py
mkeilman: [362e593](https://github.com/radiasoft/sirepo/pull/4671/commits/362e593e0ff9368fb1921fc602e5fdc57c40b06d) [60b7fd0](https://github.com/radiasoft/sirepo/pull/4671/commits/60b7fd0a92120093b77dc6a32d4ba2835f0c1757)

sirepo/template/ml.py
mkeilman: [9567d25](https://github.com/radiasoft/sirepo/pull/4671/commits/9567d2543e869bedeebbd4ed87255d8e96e64a96)

(is a mibibyte reasonable?)

sirepo/template/ml.py
mkeilman: [3000344](https://github.com/radiasoft/sirepo/pull/4671/commits/3000344484fe9bb5b46259d9515b146e2c99a068)

sirepo/template/ml.py
mkeilman: [4cfdb3d](https://github.com/radiasoft/sirepo/pull/4671/commits/4cfdb3d512aaeeaddd29148bdb714879f9b062bc)

There is no `copy()` defined on `h` for whatever reason

sirepo/template/ml.py
mkeilman: [57d671f](https://github.com/radiasoft/sirepo/pull/4671/commits/57d671f1c5317c57d30d5462e48538d24a099224)

sirepo/template/ml.py
mkeilman: [62cb8c7](https://github.com/radiasoft/sirepo/pull/4671/commits/62cb8c7be3a8ee06ce5e89ed8e3ce2ea1e7e708c)

sirepo/template/ml.py
mkeilman: I'm not sure which one you mean. args_map?

sirepo/template/ml.py
mkeilman: [6b900a3](https://github.com/radiasoft/sirepo/pull/4671/commits/6b900a3f9696f740a6319dfed724a0efbe5d1173)

mkeilman:
> When I create a new simulation, select Regression, URL, and set a URL to a csv file, with data format=text and select Save Changes, it doesn't load the data into the plots.

[9a9dd2b](https://github.com/radiasoft/sirepo/pull/4671/commits/9a9dd2bcc2e4b7bf5821eecb32e2ababca234be2)

mkeilman:
Ping

sirepo/package_data/static/js/ml.js
moellep: The fix ups are to ensure that training percentage + testing percentage == 100 for classification sims. For regression (training + testing + validation) == 100. If you switch a sim from classification to/from regression, the percentages need to get adjusted or invalidated.

sirepo/template/ml.py
moellep: I think he means the args_map var should be first in the _build_model_py() method.

Alternatively, the _conv_args(), _pooling_args(), and args_map could be moved to a separate method:
```python
def _build_args_map():
    def _conv_args(layer):
        #...
    def _pooling_args(layer):
        #...
    return PKDict(
        #...
        AveragePooling2D=lambda layer: _pooling_args(layer),
        #...
    )
```

moellep:
I've verified the loading a csv from an URL and running full sim works OK now.

sirepo/template/ml.py
robnagler: > I think he means the args_map var should be first in the _build_model_py() method.

The args_map could be a constant, possibly declared at the global scope. That's kind of the point of having maps, otherwise a bunch of if's is probably more efficient and just as clear. 

`_conv_args` does not need to be defined in order to use it in a lambda definition. The name is bound at the time of the call, not at compile time of the lambda, just like for normal function definitions.

Defining it in a function just makes it more confusing.

sirepo/package_data/static/js/ml.js
robnagler: Still think it needs a comment, e.g.


```py
            if (validation > 0) {
                partition.validation = parseFloat(validation.toFixed(2));
            }
```

Why is partition.validation not set to zero when validation is 0?

PR #4674, opened 2022-08-05
Fix #4649: effective deflecting param fix

### LINKED ISSUE #4649: recompute effective deflecting parameter
When converting a shadow undulator simulation to SRW, the undulator Effective Deflecting Parameter needs to be recomputed from the values. For example, if an example has only a Vertical Deflecting Parameter, then the Effective Deflecting Parameter should match that value. Below is an example where the values were out-of-sync after converting:
Screen Shot 2022-07-28 at 10 04 18 AM

This example was create by converting the NSLS-II FMX beamline from SRW to shadow and then back to SRW.

gurhar1133:
Fixed the deflection parameter setting with `__set_effective_deflecting_parameter()`
Added a test case


rorour (approved):
Looks good to me

PR #4675, opened 2022-08-05
fix radiasoft/raydata#18: Analysis status panel not showing all running analysis jobs

rorour:
Analysis panel shows everything that is selected, pending or running. Scans do not get unselected automatically even if they are outside of search window. Added Unselect all scans button to make it easy to unselect scans that are not visible in search results.

~~Based off https://github.com/radiasoft/sirepo/pull/4651. Maybe want to finish and merge that to master first.~~

### DIFF
--- a/sirepo/package_data/static/js/raydata.js
+++ b/sirepo/package_data/static/js/raydata.js
@@ -598,6 +598,9 @@ SIREPO.app.directive('scanSelector', function() {
                     </tr>
                   </tbody>
                 </table>
+                <div data-ng-if="awaitingScans">
+                  <p>Loading scans...</p>
+                </div>
               </div>
             </div>
             <div data-column-picker="" data-title="Add Column" data-id="sr-columnPicker-editor" data-available-columns="availableColumns" data-save-column-changes="saveColumnChanges"></div>
@@ -610,6 +613,7 @@ SIREPO.app.directive('scanSelector', function() {
 
             $scope.appState = appState;
             $scope.availableColumns = [];
+            $scope.awaitingScans = false;
             // POSIT: same columns as sirepo.template.raydata._DEFAULT_COLUMNS
             $scope.defaultColumns = ['start', 'stop', 'suid'];
             $scope.orderByColumn = 'start';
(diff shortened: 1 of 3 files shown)


sirepo/package_data/static/js/raydata.js
```diff
@@ -141,8 +141,13 @@ SIREPO.app.factory('raydataService', function(appState, panelState, requestSende
     };
 
     self.updateScanInfoTableColsInCache = function(cols) {
-        simulationDataCache.scanInfoTableCols = cols;
-        return cols;
+        if (cols.length) {
```
rorour, line 144: Allowing simulationDataCache.scanInfoTableCols to be empty/undefined was causing the analysisStatus panel to be empty when the search query returned no scans while some scans were still checked. You can see what I mean if you check out 4b1993d72ae35b2ac185fae6990eee18b822c621, search for scans, select some, then make a search query that has no results (for example, start and end time the same), then visit the analysis tab.

e-carlin:
Thanks, @rorour. Any chance you can base it off master and not [raydata17-data-source-loading](https://github.com/radiasoft/sirepo/tree/raydata17-data-source-loading)? It is going to be a little while before that PR is merged because we need to do some more thinking and I don't want it to hold this up.

rorour:
@e-carlin done. I added back in some changes from the other branch that helped this work.

I just realized that analysisStatusPanel still doesn't actually show all running/pending jobs, just selected scans. This current fix at least solves the problem of scans from polling disappearing. Let me know if you think it's still necessary to show all running/pending jobs - it seems like that is something we would need to get from the job supervisor. (Also, in that case it seems like we would also want to include completed analysis jobs, but then would there be any way to remove completed ones from the panel or are they stuck there forever? Maybe adding a link to ownjobs could be a simpler fix)

sirepo/package_data/static/js/raydata.js
```diff
@@ -741,6 +746,16 @@ SIREPO.app.directive('scanSelector', function() {
                 $scope.selectAllColumns = ! $scope.selectAllColumns;
             };
 
+            $scope.unselectAllScans = () => {
+                Object.keys(appState.models.selectedScans.uids).forEach((u) => {
```
e-carlin, line 750: I think we can just say `appState.models.selectedScans.uids = []` instead of deleting each one individually.

sirepo/package_data/static/js/raydata.js
```diff
@@ -741,6 +746,16 @@ SIREPO.app.directive('scanSelector', function() {
...
+                    delete appState.models.selectedScans.uids[u];
+                });
+                for (let i = 0; i < $scope.scans.length; i++) {
```
e-carlin, line 753: This is up for debate but I prefer functional style. So, that would be

`$scope.scans.map(s => $scope.toggleScanSelection(s, false))`

e-carlin (commented):
Looks good. A couple bugs and some minor style changes.

Issue #4688, opened 2022-08-08
lib ignore_files should default to None


moellep:
LibAdapterBase has a new ignore_files arg which should default to None for backward compatibility (madx LibAdapter).

PR #4689, opened 2022-08-08
LibAdapterBase ignore_files defaults to None

moellep:


moellep:
Fixes #4688 lib ignore_files defaults to None

sirepo/lib.py
```diff
@@ -22,11 +22,11 @@
 class LibAdapterBase:
     """Common functionality between code specific LibAdapter implementations."""
 
-    def __init__(self, ignore_files):
+    def __init__(self, ignore_files=None):
         m = inspect.getmodule(self)
         self._sim_data, _, self._schema = sirepo.sim_data.template_globals(m.SIM_TYPE)
         self._code_var = m.code_var
-        self._ignore_files = ignore_files
+        self._ignore_files = ignore_files if ignore_files else []
```
gurhar1133, line 29: simpler to do
```python
def __init__(self, ignore_files=[]):
    ...
    self.ignore_files = ignore_files
```
?

gurhar1133 (commented):
Just one comment

sirepo/lib.py
moellep: Yes, that is a python gotcha - you should never have an mutable as a default arg:
https://docs.python-guide.org/writing/gotchas/

gurhar1133: 👍

gurhar1133 (approved):

sirepo/lib.py
e-carlin: You'll get surprising behavior using mutable default values for kwargs (lists are mutable). So, we don't use them.
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

PR #4692, opened 2022-08-08
Fix #4685 radia update field on path

### LINKED ISSUE #4685: field on path plot does not update
When the start or end points for a field path are changed the plot refreshes but does not actually change the plotted bounds. The bounds only seem to update correctly if you first switch to a different path and then switch back to the edited path.
For example see: https://www.sirepo.com/radia#/visualization/W8qxqvxD

mkeilman:
Previously the plot was being updated only if the user selected a new path, not on changes to the current path.

This also amends the current density units


moellep:
Works well!

PR #4699, opened 2022-08-09
Fix #4630 radia fillets

### LINKED ISSUE #4630: Allow arbitrary planar cuts to objects
We already use `Radia.ObjCutMag` to slice bevels into objects. We have a need to bisect objects, which the bevel interface is not suited for, but which we should generalize.

mkeilman:
Notes:
- I repurposed this issue since I realized I could accomplish the desired fillets with existing bevel cuts without introducing arbitrary planar cuts
- Fillets are defined similarly to bevels: click the shape button of an object and then the Modifications tab to see a table of fillets to add to or edit
- Pulled some of the cloudier bevel code into functions


sirepo/package_data/static/js/radia.js
```diff
@@ -1221,13 +1224,15 @@ SIREPO.app.directive('bevelTable', function(appState, panelState, radiaService)
                 <col style="width: 20ex">
                 <col style="width: 20ex">
                 <col style="width: 20ex">
+                <col style="width: 20ex">
```
rorour, line 1227: Could do this in fewer lines with `<col span="4" style="width: 20ex">`. Also, just wondering why there are 4 of these but 5 columns (not including edit button). I guess the last one would be redundant since it can just size to whatever is left?

sirepo/package_data/static/js/radia.js
```diff
@@ -1343,6 +1349,144 @@ SIREPO.app.directive('bevelTable', function(appState, panelState, radiaService)
...
+                    <td>{{ bevelEdge(item.edge) }}</td>
+                    <td>{{ item.radius }}mm</td>
+                    <td>{{ item.numSides }}</td>
+                  <td style="text-align: right">
+                    <div class="sr-button-bar-parent">
+                        <div class="sr-button-bar" data-ng-class="sr-button-bar-active" >
+                             <button data-ng-click="editItem(item)" class="btn btn-info btn-xs sr-hover-button">Edit</button>
+                             <button data-ng-click="deleteItem(item, $index)" class="btn btn-danger btn-xs"><span class="glyphicon glyphicon-remove"></span></button>
+                        </div>
+                    <div>
```
rorour, line 1394: Should be `</div>`, also applies to L1254.

sirepo/package_data/static/js/radia.js
```diff
@@ -1343,6 +1349,144 @@ SIREPO.app.directive('bevelTable', function(appState, panelState, radiaService)
...
+                <tr>
+                  <th>Axis</th>
+                  <th>Edge</th>
+                  <th>Radius</th>
+                  <th>Resolution</th>
+                  <th></th>
+                </tr>
+              </thead>
+             <tbody>
+            <tr>
```
rorour, line 1381: Indentation seems off. Also applies to bevelTable

sirepo/package_data/static/js/radia.js
```diff
@@ -1343,6 +1349,144 @@ SIREPO.app.directive('bevelTable', function(appState, panelState, radiaService)
...
+                             <button data-ng-click="deleteItem(item, $index)" class="btn btn-danger btn-xs"><span class="glyphicon glyphicon-remove"></span></button>
+                        </div>
+                    <div>
+                  </td>
+                </tr>
+            </tbody>
+            </table>
+            <button data-ng-click="addItem()" id="sr-new-fillet" class="btn btn-info btn-xs pull-right">New Fillet <span class="glyphicon glyphicon-plus"></span></button>
+        `,
+        controller: function($scope, $element) {
```
rorour, line 1401: A lot of the controller contents seem duplicated/similar to bevelTable, could be an opportunity to abstract this out?

sirepo/package_data/static/js/radia.js
mkeilman: [752caf2](https://github.com/radiasoft/sirepo/pull/4699/commits/752caf200761300ff36e71c8dd89b4553dd3782e)
(4 instead of 5 was a mistake)

sirepo/package_data/static/js/radia.js
mkeilman: [681e384](https://github.com/radiasoft/sirepo/pull/4699/commits/681e384af5430814034d4778780c0f6a0ff5ab39)

sirepo/package_data/static/js/radia.js
mkeilman: [681e384](https://github.com/radiasoft/sirepo/pull/4699/commits/681e384af5430814034d4778780c0f6a0ff5ab39)

sirepo/package_data/static/js/radia.js
mkeilman: Absolutely should be abstracted, but I think it belongs in a separate issue so as not to complicate this one

rorour (approved):

PR #4702, opened 2022-08-10
fix #4701 show beam position report for single-update use cases

### LINKED ISSUE #4701: show beam position report  in single-update mode
In the controls app, the Beam Position at Monitors report should be shown when individual correctors are updated using the UI. Currently the Beam Position at Monitors report is only displayed when an optimization is running.

moellep:
- added controlSettings.simMode to track if app is optimizing or running single updates

### DIFF
--- a/sirepo/package_data/static/js/controls.js
+++ b/sirepo/package_data/static/js/controls.js
@@ -227,37 +227,6 @@ SIREPO.app.controller('ControlsController', function(appState, controlsService, 
         return res;
     }
 
-    function getInitialMonitorPositions() {
-        if (self.simState && self.simState.isProcessing()) {
-            // optimization is running
-            return;
-        }
-        if (! appState.applicationState().externalLattice) {
-            return;
-        }
-        if (controlsService.isDeviceServerReadOnly()) {
-            return;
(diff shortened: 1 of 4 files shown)


mkeilman:
Can you explain how to test this?

moellep:
It is related to running with DeviceServer. Here is the step to reproduce the bug before the fix:

- open the "FODO with instruments" example
- from the cogwheel, select Control Settings
- change the operation mode to "DeviceServer" and enter some values in user, procName, procId and machine (on dev, the values can be any string)
- copy the simulationId from the URL
- In a bash terminal, go to ~/src/radiasoft/sirepo/etc and start the madx_device_server with the simulationId:
```bash
python madx_device_server.py -simulationId-
```
- Run an optimization using the Linear Response Matrix
- When the optimization finishes note the values on the Beam Position at Monitors plot.
- Edit HV_KICKER_1 and change the horizontal current 17.253333027e-5
- Notice the Monitors plots update, but the Beam Position at Monitors plot doesn't change when it should

The expected plot would be something like this:

Screen Shot 2022-08-11 at 9 48 46 AM

PR #4706, opened 2022-08-10
Fix #4703 radia und example axis path

### LINKED ISSUE #4703: undulator example field path shows no initial axis
When I edit the "Y-Axis" field path on the Undulator example, it doesn't show any Axis selected. I think it should show "y" selected.
Screen Shot 2022-08-10 at 2 28 42 PM

mkeilman:
Discarding changes to example will show the correct axis now. The wiggler example had the same problem. IMHO this is not worth adding fixup code, but I will do so if needed.


mkeilman:
@moellep lmk if you think I should add fixup code

PR #4718, opened 2022-08-12
Fix #4712 ml to activait

### LINKED ISSUE #4712: rename ml to activait
Pulling this out of #4406 - most of it is done but it should not wait on the image stuff.

mkeilman:
Just a bunch of renaming, but a couple of questions:

- old sims in the "ml" directory would be unreachable (not deleted though). Is it worth adding a mechanism to copy them into the new "activait" directory? I doubt there are any "real" sims out there and I guess we could hand-copy them if anyone complains...
- should we redirect the ml url? That ties in to the previous question - a link to a sim would not be found (/ml#/data/abcd -> /activait#/data/abcd) unless we copy it.


sirepo/sim_data/activait.py
```diff
@@ -13,6 +13,8 @@
 class SimData(sirepo.sim_data.SimDataBase):
     @classmethod
     def fixup_old_data(cls, data):
```
e-carlin, line 15: If we aren't carrying over old simulations then I think all fixup code can be deleted.

sirepo/simulation_db.py
```diff
@@ -232,6 +232,8 @@ def _last_modified(data, path):
             data.simulationType = "warppba"
         elif data.simulationType == "fete":
             data.simulationType = "warpvnd"
+        elif data.simulationType == "ml":
```
e-carlin, line 235: If we aren't carrying over old simulations then I don't think we will ever have a case where data.simulationType == "ml"

e-carlin (commented):
default-data.json still mentions ml. Some docstrings (ex pkcli/activait) still mention ml. 

My preference would be to put in a redirect for ml to activait and write a db_upgrade to move all existing ml simulations to activait. If not, then I think we should write a db_upgrade to delete all ml simulations.

mkeilman:
> My preference would be to put in a redirect for ml to activait and write a db_upgrade to move all existing ml simulations to activait. If not, then I think we should write a db_upgrade to delete all ml simulations.

What is the procedure for db_upgrades? From my playing around I see they only get applied once, so how does one test them, and clear out previous attempts?

e-carlin:
Yes, they are applied once [on startup](https://github.com/radiasoft/sirepo/blob/9d3a7c4cafd1ac8679c5fbb3463a12421552b13e/sirepo/server.py#L727).  We [store a record of them in sqlite](https://github.com/radiasoft/sirepo/blob/9d3a7c4cafd1ac8679c5fbb3463a12421552b13e/sirepo/db_upgrade.py#L33) and only run the functions in db_upgrade.py that don't have a corresponding record in db_upgrade_t. 

There are a few ways to get around this for testing. I think for your use case I would do the following:
```bash
$ git checkout master
$ rm -rf run/
$ sirepo service http
$ # go to /ml and create some simulations
$ # stop the server
$ cp -r run/ run.bak/ # Make a backup of the db so we can apply upgrades to it over and over
$ git checkout 4712-ml-to-activait
$ sirepo service http
$ # db upgrade is applied
$ # stop server
$ # edit the upgrade to fix issues with last application
$ cp -r run.bak/ run/ # copy over the old db so we can apply our fixed upgrade to it
$ sirepo service http
$ # db upgrade is applied
$ # repeat from cp -r run.bak/ run/
```

mkeilman:
I'm going to convert this to draft until we get to a calm point in other nuline development, to minimize disruption

mkeilman:
The latest commits clean things up and include code for migrating from old to new simulation types.  I have some questions on the details:
- Where should `migrate_sim_type()` live, or should we consider it a one-shot and put it all in db_upgrade? (a nice thing about a separate function is that I can test without constantly messing with the database)
- Does the lib dir need to be copied recursively?
- This skips the examples since `simulation_dir()` will create them for the new sim type. Do we want to preserve the user's changes to the examples?

moellep:
- I think migrate_sim_type() should live in db_upgrade, possibly parameterized so it could be used again if another sim needs migration.
- lib dir should be a flat list with no subdirectories (at least for this app and most others)
- it is OK to skip the examples, we can run "sirepo admin create-examples" after the migration

mkeilman:
Last commit moved migrate_sim_type to db_upgrade - I think everything else is covered

PR #4735, opened 2022-08-18
Fix #4734: handle case when there is no stdout (only stderr)

### LINKED ISSUE #4734: _SbatchPrepareSimulationCmd self.stdout may not exist
There may just be stderr and in that case we should still be able to send back the error message.

e-carlin:


gurhar1133 (approved):
looks good 👍

Issue #4406, opened 2022-05-12
Add image processing to activait


mkeilman:
Blicon will use this to analyze srw output. A couple of notes about the initial work:

- use regression only
- use convolution layers instead of neural net layers

moellep:
A roadmap for the new features related to image processing
[Sirepo_activait Improvements.pdf](https://github.com/radiasoft/sirepo/files/9075557/Sirepo_activait.Improvements.pdf)

mkeilman:
I believe @gurhar1133's efforts took care of this, correct?

moellep:
Yes, this item is completed.

PR #4736, opened 2022-08-19
Fix #4723 activait archive file list

### LINKED ISSUE #4723: Present paths to files in archives and allow selection
If the user chose a file or url that is an archive (.zip, .tar.gz, .h5), show a list of the files within to select for analysis.

mkeilman:
Notes:
- The main feature here is the ArchiveManager class, which uses a context manager to handle regular files as well as zips and tarballs (h5 is on deck). It reads data within the archive with no need to extract.
- For text data, the user can select from a list of files in the archive to analyze. For image data, the column selector is replaced by a list of directories to use for input or output. Actual image processing is to come (#4406)
- There is now a cache for urls so data is not re-fetched. We need a reload button for which I'll make an issue


moellep:
When I select "new simulation", it shows the error message: Cannot read properties of undefined (reading 'endsWith') from the isArchiveFile(filename) method in ml.js.

sirepo/package_data/static/js/sirepo-utils.js
```diff
@@ -11,6 +11,32 @@ class SirepoUtils {
         return s.charAt(0).toUpperCase() + s.slice(1);
     }
 
+    static orderOfMagnitude(val, binary=false) {
+        const MAGS = ['', 'k', 'M', 'G', 'T', 'P', 'E'];
+        let i = 0;
+        let v = 0;
+        const base = binary ? 1024 : 1000;
+        for (i = 0; i < MAGS.length; ++i) {
+            v = val / Math.pow(base, i);
```
robnagler, line 20: Doesn't this break if val is negative?

sirepo/package_data/static/js/sirepo-utils.js
```diff
@@ -11,6 +11,32 @@ class SirepoUtils {
...
+            if (v < base) {
+                break;
+            }
+        }
+        return {
+            order: i,
+            suffix: MAGS[i] + (binary ? 'i' : ''),
```
robnagler, line 27: This breaks if it is larger than "E". It's maybe fine, but the error message will be confusing.

sirepo/package_data/static/js/sirepo-utils.js
```diff
@@ -11,6 +11,32 @@ class SirepoUtils {
         return s.charAt(0).toUpperCase() + s.slice(1);
     }
 
+    static orderOfMagnitude(val, binary=false) {
+        const MAGS = ['', 'k', 'M', 'G', 'T', 'P', 'E'];
```
robnagler, line 15: Since this supports binary, it should use Ki, Mi, Gi, Ti, Pi, Ei.

sirepo/package_data/static/js/sirepo-utils.js
```diff
@@ -11,6 +11,32 @@ class SirepoUtils {
...
+        for (i = 0; i < MAGS.length; ++i) {
+            v = val / Math.pow(base, i);
+            if (v < base) {
+                break;
+            }
+        }
+        return {
+            order: i,
+            suffix: MAGS[i] + (binary ? 'i' : ''),
+            val: v,
```
robnagler, line 28: val is confusing. It should be magnitude or exponent.

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+        ".zip",
+    )
+
+    def __init__(self, file_path):
+        import zipfile
+        import tarfile
+
+        self.file_path = file_path
+        self.filename = file_path if isinstance(file_path, str) else file_path.basename
+        if self._is_archive_type(".zip"):
```
robnagler, line 82: self.pkupdate would work if ArchiveManager was a subclass of PKDict.  Why not use _ARCHIVE_EXTENSIONS possibly in a loop checking the archives that are supported?

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+            self.dir_check = "is_dir"
+            self.extractor = "open"
+            self.file_ctx = zipfile.ZipFile
+            self.item_name = "filename"
+            self.lister = "infolist"
+        if self._is_archive_type(".tar.gz"):
```
robnagler, line 88: elif

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+            self.dir_check = "isdir"
+            self.extractor = "extractfile"
+            self.file_ctx = tarfile.open
+            self.item_name = "name"
+            self.lister = "getmembers"
+
```
robnagler, line 94: else? what happens if it is not a file that can be handled?

sirepo/template/ml.py
```diff
@@ -301,13 +301,26 @@ def stateless_compute_remote_data_bytes_loaded(data):
...
     pkio.write_text(
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
         _generate_parameters_file(data),
     )
 
 
+def _archive_file_list(filename, data_type):
+    def _filter(item):
+        is_dir = getattr(item, a.dir_check)()
+        return is_dir if data_type == "image" else not is_dir
```
robnagler, line 318: `is_dir == (data_type =="image")`?

sirepo/template/ml.py
```diff
@@ -301,13 +301,26 @@ def stateless_compute_remote_data_bytes_loaded(data):
...
 def write_parameters(data, run_dir, is_parallel):
     pkio.write_text(
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
         _generate_parameters_file(data),
     )
 
 
+def _archive_file_list(filename, data_type):
+    def _filter(item):
+        is_dir = getattr(item, a.dir_check)()
```
robnagler, line 317: Single letter variables aren't used outside scope. Pass it into _filter or name it and define it above _filter.

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
         if name:
             return [cls.lib_file_name_with_model_field("dataFile", "file", name)]
         return []
+
+
+class ArchiveManager:
+    import contextlib
```
robnagler, line 69: This can be a global import.

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+
+    _ARCHIVE_EXTENSIONS = (
+        ".tar.gz",
+        ".zip",
+    )
+
+    def __init__(self, file_path):
+        import zipfile
```
robnagler, line 77: These can be global.

sirepo/package_data/template/ml/scale.py.jinja
```diff
@@ -64,21 +65,22 @@ def scale_columns(values, column_types, col_type, scaler):
...
     out_idx = scale_columns(v, column_types, 'output', outputs_scaler)
-    os.remove(path)
+    os.remove(arch_mgr.file_path)
     np.save('{{ scaledFile }}', v)
     return v, in_idx, out_idx
 
 
 scaled, in_idx, out_idx = scale_file(
-    '{{ dataFile }}',
+    sirepo.sim_data.ml.ArchiveManager('{{ dataFile }}'),
```
robnagler, line 82: The name DataReader seems more appropriate.

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+        import tarfile
+
+        self.file_path = file_path
+        self.filename = file_path if isinstance(file_path, str) else file_path.basename
```
robnagler, line 81: `self.path = pkio.py_path(file_path)`

Then you can say self.path.filename where you need it. And, self.path.ext gives you a lot of what you want. If it is a gz file, then it is an archive, and you can further qualify.

sirepo/package_data/template/ml/scale.py.jinja
```diff
@@ -9,16 +10,16 @@ from sklearn.preprocessing import {{ dataFile_inputsScaler }}
 from sklearn.preprocessing import {{ dataFile_outputsScaler }}
 {% endif %}
 
-def read_data(path, **kwargs):
-    return sirepo.numpy.ndarray_from_csv(
-        path,
+def read_data(arch_mgr, data_path, **kwargs):
```
robnagler, line 13: archive_manager Avoid abbreviations unless they are necessary. In this case, arch_mgr is used once.

sirepo/sim_data/ml.py
```diff
@@ -62,3 +63,57 @@ def _lib_file_basenames(cls, data):
...
+        return any(
+            [self._is_archive_type(s) for s in ArchiveManager._ARCHIVE_EXTENSIONS]
+        )
+
+    @contextlib.contextmanager
+    def data_ctx(self, data_path):
+        if not self.is_archive():
+            yield open(self.file_path)
+        else:
+            with self.file_ctx(self.file_path, mode="r") as f:
```
robnagler, line 108: I don't like the name data_ctx. It doesn't tell me anything. It's return an open file that can be read. 

However, I see how it is used, and I think what you really want is something that matches your problem better, eg. return a numpy array or csv depending on what the data is.

robnagler (changes requested):
I think there's an impedance mismatch between the problem and the ArchiveManager. It's better to say "what do I need here", and define a function that gives you what you need, exactly, e.g.  res.colsWithNonUniqueValues = _cols_with_non_unique_values is very clear. However, to implement _cols_with_non_unique_values, don't add parameters, remove them. I think this is a general trend I see in your code: you add parameters instead of removing them. Try to flip that around. _compute_column_info could be solved by using a class. It's polymorphic. Imbue the context of the question (compute column info) into that class so it doesn't have to be generalized.

mkeilman:
> When I select "new simulation", it shows the error message: Cannot read properties of undefined (reading 'endsWith') from the isArchiveFile(filename) method in ml.js.

[2eb1305](https://github.com/radiasoft/sirepo/pull/4736/commits/2eb13051a54a3478cc227049639b69ca6b3356fc)

sirepo/package_data/static/js/ml.js
```diff
@@ -2113,6 +2190,51 @@ SIREPO.viewLogic('dataFileView', function(appState, panelState, persistentSimula
...
+        requestSender.sendStatelessCompute(
+            appState,
+            d => {
+                setDataList(d.datalist);
+            },
+            {
+                method: 'get_archive_file_list',
+                filename: dataFile.file,
+                data_type: dataFile.dataFormat,
+            }
```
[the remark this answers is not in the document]
mkeilman: [b3ee94f](https://github.com/radiasoft/sirepo/pull/4736/commits/b3ee94f3970ac0893c3e239fe52b10a40a84f694)

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -786,6 +786,7 @@ SIREPO.app.directive('fieldEditor', function(appState, keypressService, panelSta
         controller: function($scope, $element) {
             $scope.appState = appState;
             $scope.utilities = utilities;
+            $scope.UTILS = SIREPO.UTILS;
```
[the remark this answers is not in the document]
mkeilman: When I need some utility function that I think is of general interest, I (now) put it in the `SirepoUtils` class instead of the `utilities` angular service. It needs to be here because this is the scope referenced in the URL field definition. It's named that way as an analog to `$scope.utilities`, but it can be `SIREPO_UTILS` or whatever works best

sirepo/package_data/static/js/sirepo-utils.js
mkeilman: [adbb0c2](https://github.com/radiasoft/sirepo/pull/4736/commits/adbb0c2fb917f892c1960a6c2cdea3bdcdeffe6c)

sirepo/package_data/static/js/sirepo-utils.js
mkeilman: [885738e](https://github.com/radiasoft/sirepo/pull/4736/commits/885738ee5b12c100e3da7b191ff4bbf9ab36ae0c)

sirepo/package_data/static/js/sirepo-utils.js
mkeilman: The "i" is added on returning, but also [d8a178b](https://github.com/radiasoft/sirepo/pull/4736/commits/d8a178b58ca0e460e6e7c512b169e0dba74f0c53)

sirepo/package_data/static/js/sirepo-utils.js
```diff
@@ -11,6 +11,32 @@ class SirepoUtils {
         return s.charAt(0).toUpperCase() + s.slice(1);
     }
 
+    static orderOfMagnitude(val, binary=false) {
+        const MAGS = ['', 'k', 'M', 'G', 'T', 'P', 'E'];
+        let i = 0;
+        let v = 0;
+        const base = binary ? 1024 : 1000;
+        for (i = 0; i < MAGS.length; ++i) {
```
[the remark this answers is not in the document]
mkeilman: Yup! [a5702ff](https://github.com/radiasoft/sirepo/pull/4736/commits/a5702ff466de34a620658a869b164a959c9c3b73)

sirepo/sim_data/ml.py
mkeilman: [87b24e3](https://github.com/radiasoft/sirepo/pull/4736/commits/87b24e3f31c05fa6bb938a16525863872465d1b9)

sirepo/sim_data/ml.py
mkeilman: [3c866a1](https://github.com/radiasoft/sirepo/pull/4736/commits/3c866a1fe074693237b4be774648761a1e03b649)

sirepo/package_data/template/ml/scale.py.jinja
mkeilman: agreed [0227019](https://github.com/radiasoft/sirepo/pull/4736/commits/02270195eb3d6ebd58565db4e4c6e1a45eafc004)

sirepo/sim_data/ml.py
mkeilman: Any file is supported, but if it's not a supported archive those properties are not used and the file is accessed directly

sirepo/sim_data/ml.py
mkeilman: [6e561ce](https://github.com/radiasoft/sirepo/pull/4736/commits/6e561ced430aef5c3176897462da7003d538df7e)

I think this better...???

sirepo/sim_data/ml.py
mkeilman: [6e561ce](https://github.com/radiasoft/sirepo/pull/4736/commits/6e561ced430aef5c3176897462da7003d538df7e)

sirepo/sim_data/ml.py
mkeilman: [6e561ce](https://github.com/radiasoft/sirepo/pull/4736/commits/6e561ced430aef5c3176897462da7003d538df7e)

sirepo/template/ml.py
mkeilman: [4871694](https://github.com/radiasoft/sirepo/pull/4736/commits/4871694b7b14e7a7e255c3872bdfae5afe91af43)

sirepo/package_data/template/ml/scale.py.jinja
mkeilman: [115e90b](https://github.com/radiasoft/sirepo/pull/4736/commits/115e90b0f6255ac567774bd06a36493f01dafe09)

sirepo/sim_data/ml.py
mkeilman: [5895f25](https://github.com/radiasoft/sirepo/pull/4736/commits/5895f2509fbb49bd496a21aaa025dbd3c8768977)

sirepo/template/ml.py
mkeilman: I think that passes all files for images and none for text...???

sirepo/sim_data/ml.py
mkeilman: Aside: [.gz is not an archive](https://stackoverflow.com/questions/33592099/how-do-i-list-contents-of-a-gz-file-without-extracting-it-in-python), so we really do need to use .tar.gz explicitly. I played with `pathlib` a bit, but it got too complicated.

mkeilman: [b5a6fd8](https://github.com/radiasoft/sirepo/pull/4736/commits/b5a6fd8c12d7927066d3b9bc60ad131a8daed3f6)

sirepo/sim_data/ml.py
```diff
@@ -104,30 +103,27 @@ def __init__(self, file_path):
     def _get_archive_extension(self):
         x = list(
             filter(
-                lambda s: self._is_archive_type(s),
+                lambda s: self.path.basename.endswith(s),
```
e-carlin, line 104: You can say `self.path.ext == f".{s}"` which I think is more explicit about what you're after.

sirepo/sim_data/ml.py
```diff
@@ -104,30 +103,27 @@ def __init__(self, file_path):
...
                 DataReader._SUPPORTED_ARCHIVE_EXTENSIONS,
             )
         )
         return x[0] if len(x) else None
```
e-carlin, line 108: An empty list in python is False so you can omit the len().

sirepo/sim_data/ml.py
mkeilman: That won't work for ".tar.gz"; its `.ext` is ".gz", which by itself is not an archive (just compression). We must explicitly test for .tar.gz.

sirepo/sim_data/ml.py
mkeilman: [3b9b562](https://github.com/radiasoft/sirepo/pull/4736/commits/3b9b562469fab32bc1a6dbd85b7b1c715dcb140a)

mkeilman:
ping just to see if the changes so far are ok, modulo fixing the "impedance mismatch"

robnagler (approved):
I don't have time to work on this right now. Something to circle back to.

mkeilman:
@moellep could you test to make sure the new simulation bug is fixed?

Issue #4693, opened 2022-08-08
more modular tallies


moellep:
We need to build up tallies similar to openmc sources - There are many filter to mix and match which can combine into 1 talley. For example, a tally could have an EnergyFilter and a MaterialFilter applied. Also, tally scores should support the full list of openmc tally scores.

PR #4743, opened 2022-08-22
Fix #4720 tally colormap

### LINKED ISSUE #4720: Add colormap parameter to tallies
Remove the hard-coded jet colormap and let the user select it

mkeilman:
Also tally display now responds to opacity control


mkeilman:
I'm going to convert this to draft since it will have to be woven into the extensive changes in #4693

mkeilman:
ping - should be cool with the tally changes

PR #4771, opened 2022-08-25
Fix radiasoft/prop22_cloudmc#13: export statepoints

e-carlin:
Add a "Statepoints" link in the DAGMC Gemoetry report panel that downloads
a zip of all statepoint files output by the simulation.

In addition, create sirepo.util.write_zip which encapsulates all of the
args commonly supplied when creating a zipfile.


mkeilman:
FYI I tested the export in radia and it was broken for other reasons. The fix is in place in master so the export will work once this is merged.

PR #4778, opened 2022-08-29
4693 modular tallies

moellep:
- Tallies can be created and managed on the Settings / Tallies tab.
- Mesh and material filters can be used together
- Multiple tally scores can be computed at once
- All tally data extraction is moved to sim_frame() rather than in parameters.py
- Openmc errors are parsed and returned to the app


mkeilman:
I'm getting an error - the generated python looks like

```python
def create_tallies():
    ]
    t1.scores = ['heating']

    tallies = openmc.Tallies([
        t1
    ])
    tallies.export_to_xml()
```
with that stray closing bracket at the top

moellep:
Yes, I see that error if there is no mesh defined. I'll fix that. To work around it for now, create a mesh like this for the fuel pin example.
Screen Shot 2022-08-29 at 4 14 45 PM

sirepo/template/cloudmc.py
```diff
@@ -279,3 +309,160 @@ def _generate_space(space):
...
+                _generate_tally(t, data.models.volumes)
+                for t in data.models.settings.tallies
+            ]
+        )
+        + f"""
+tallies = openmc.Tallies([
+    {','.join(['t' + str(tally._index + 1) for tally in data.models.settings.tallies])}
+])
+tallies.export_to_xml()
+"""
```
mkeilman, line 329: No problem, but I'm curious when to use this construction instead of just putting everything in the jinja file

mkeilman:
FYI the voxel changes are ready but will have to be modified for modular tallies...

PR #4785, opened 2022-08-31
Fix #4782 kickmap error

### LINKED ISSUE #4782: kickmap sdds report broken
The functions refer to outdated model schema, similar to #4780

mkeilman:
This also cleans up `_save_kick_map_sdds()` and removes the download as text option, which was never implemented and is of unclear importance.

### DIFF
--- a/sirepo/template/radia.py
+++ b/sirepo/template/radia.py
@@ -10,6 +10,7 @@ Radia "instance" goes away and references no longer have any meaning.
 from pykern import pkcompat
 from pykern import pkinspect
 from pykern import pkio
+from pykern import pkjson
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdc, pkdp, pkdlog
 from scipy.spatial.transform import Rotation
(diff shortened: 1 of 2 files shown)


mkeilman:
Kickmaps are for undulator magnets, so this can be tested with the one under Examples.

sirepo/template/radia.py
```diff
@@ -1233,29 +1231,15 @@ def _save_fm_sdds(name, vectors, scipy_rotation, path):
...
-    for i in range(len(x_vals)):
-        for j in range(len(y_vals)):
-            h.append(h_vals[i][j])
-            v.append(v_vals[i][j])
-    col_data.append([x])
-    col_data.append([y])
-    col_data.append([h])
-    col_data.append([v])
+    col_data = [
+        [numpy.tile(0.001 * numpy.array(km_data.x), len(km_data.x)).tolist()],
```
gurhar1133, line 1238: maybe put 0.001 in a constant with a name that clarifies it's role?

gurhar1133 (approved):
I approved with one suggestion, see comments.

But looks good

sirepo/template/radia.py
mkeilman: Good call, [86bc328](https://github.com/radiasoft/sirepo/pull/4785/commits/86bc3289d0fd30f6855bef62135ad38a493cf383)

PR #4790, opened 2022-09-05
Fix #4303 radia electron trajectories

### LINKED ISSUE #4303: Add electron trajectories
Add the trajectories calculated by Radia's ```RadFldPtcTrj``` function in both 2d and 3d

mkeilman:
Notes:
- Radia's trajectory function expects the coordinates in a certain order, so we have to rotate both the fields and the input coords, then rotate the output coords back.
- Tested against various dipoles and undulators
- Examples updated


gurhar1133 (approved):
Looks good. 

One observation: `extract_report_data()` in template/radia.py is getting really big. Maybe should be broken up into sub calls? eg.
```python
def _geometry_report():
    ...

def _electron_report():
    ...

def extract_report_data(...):
    ...
    PKDict(
        ...
        geometryReport=_geometry_report,
        ...
        electronTrajectoryReport=_electron_report,
        ...
    )[sim_in.report]()
```

moellep:
The new report looks good. I did find that the "Save as SDDS/CSV/SRW" no longer works on this branch for the Field on Path or the other reports on the Visualization page.

mkeilman:
> One observation: `extract_report_data()` in template/radia.py is getting really big. Maybe should be broken up into sub calls? eg.
> 
> ```python
> def _geometry_report():
>     ...
> 
> def _electron_report():
>     ...
> 
> def extract_report_data(...):
>     ...
>     PKDict(
>         ...
>         geometryReport=_geometry_report,
>         ...
>         electronTrajectoryReport=_electron_report,
>         ...
>     )[sim_in.report]()
> ```

Agreed

mkeilman:
> The new report looks good. I did find that the "Save as SDDS/CSV/SRW" no longer works on this branch for the Field on Path or the other reports on the Visualization page.

[966bf43](https://github.com/radiasoft/sirepo/pull/4790/commits/966bf43e9db1fc616aa1bcee29e7501c622637c6)

I'm building the trajectory report download now

mkeilman:
> Looks good - the download works on the reports, except for the Kick Map report for the Undulator example is having this error when download as SDDS (or CSV) is selected:
> 
> ```
> File "sdds.py", line 255, in setColumnValueLists
>     raise Exception(msg)
> Exception: invalid column name xpFactor
> ```

I can't seem to reproduce this error - I tried with a fresh example and it was fine. ???

moellep:
I can't reproduce it with a reset example. Looks good - I'll merge this.

PR #4797, opened 2022-09-06
4795 angular distribution

### LINKED ISSUE #4795: angular distribution error
Running the "reactor" simulation with these angular distribution settings causes the simulation to fail with an error: '_generate_angleribution' is not defined

Screen Shot 2022-09-06 at 10 42 57 AM

moellep:
fix #4795 angular distribution corrected
- also implement EnergyFilter, EnergyOutFilter, ParticleFilter code generation

### DIFF
--- a/sirepo/package_data/static/json/cloudmc-schema.json
+++ b/sirepo/package_data/static/json/cloudmc-schema.json
@@ -239,7 +239,7 @@
             "r": ["R distribution", "Univariate", {"_type": "None"}],
             "phi": ["Phi distribution", "Univariate", {"_type": "None"}, "Distribution of phi-coordinates (azimuthal angle) in a reference frame specified by the origin parameter"],
             "z": ["Z distribution", "Univariate", {"_type": "None"}],
-            "origin": ["Origin", "Point3D [cm]", [0, 0, 0], "Coordinates (x0, y0, z0) of the center of the cylindrical reference frame for the source."]
+            "origin": ["Origin [cm]", "Point3D", [0, 0, 0], "Coordinates (x0, y0, z0) of the center of the cylindrical reference frame for the source."]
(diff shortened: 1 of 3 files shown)


mkeilman:
I get an error running with energy or energyout filters:
```
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/job_cmd.py", line 223, in _do_get_simulation_frame
    msg.data.copy().pkupdate(run_dir=msg.runDir),
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/template_common.py", line 639, in sim_frame_dispatch
    res = o(frame_args)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/cloudmc.py", line 105, in sim_frame
    :, :, t.get_score_index(frame_args.score)
  File "/home/vagrant/src/radiasoft/openmc/openmc/mesh.py", line 736, in write_data_to_vtk
    volume_normalization=volume_normalization
  File "/home/vagrant/src/radiasoft/openmc/openmc/mesh.py", line 236, in write_data_to_vtk
    raise RuntimeError(errmsg.format(label))
RuntimeError: The size of the dataset mean should be equal to the number of cells
```

mkeilman (changes requested):
Minor comment, but the `write_data_to_vtk()` call fails for energy and energyOut filters

moellep:
The "The size of the dataset mean should be equal to the number of cells" error will happen if you have an EnergyFilter and a MeshFilter on the same tally - the results creates a mesh for every energy step and then it can't convert it correctly. I'll add a better error message in this case.

mkeilman (approved):
All good - I think disabling or hiding incompatible filters would be a good idea for a future issue

PR #4802, opened 2022-09-07
20220923 Fix #4635 implement Jspec in React

### LINKED ISSUE #4635: Jspec in React
Jspec will be the first React app. It's less used, and it's a simple code. React will use the existing Sirepo API. 

This is a "bundled" issue. The first step is to come up with a plan of action of how this will happen. 

This will be the third deliverable for Afcloud at the end of Sep.

Plan:
- [x] Animations
- [x] Writing back data to server with forms (already a WIP)
- [x] Simulation folder view like all apps have
- [x] Some editors have tabs within the panels
- [x] Simulation start panel, the one with the 'start simulation' button and simulation status
- [x] Panels need to dynamically show/enter the view when a simulation completes

garsuga:
This is partially complete but @robnagler wanted to be able to review the code. Jspec simulations cannot be displayed yet but it is almost at the point where that is possible.


garsuga:
Here is what is left right now

- [ ] Fix graph y axis direction
- [ ] Add File upload type for forms
- [ ] Add report button next to File type like in Jspec
- [ ] Add legend to graphs
- [ ] Add download button to reports
- [ ] Add simulation start button and create / refresh report visuals on data receive
- [ ] Receive partial data from run-status
- [ ] Handle refreshing reports that automatically run
- [ ] Express report data dependency in schema

react/src/data/dependency.js
```diff
@@ -0,0 +1,69 @@
+import { useSelector, useDispatch, useStore } from 'react-redux';
```
robnagler, line 1: I rewrote this to show a few things that we do here differently. 

We name objects by what they are, not how they are used. An SPath is the thing. A Dependency is how it is used. It could be used for other things. 

Fail fast is a deep philosophy. Unless there is a good everything should be fail fast. This surfaces defects instead of burying them.

I'm not sure about the names, but I tired to express what I thought they were. The point is intent, not prescription. 

We don't put blank lines in a function. That usually indicates another function is needed that describes the block of code by its name.

Temporary variables are avoided. They introduce cognitive load. "Why does this name mean?" "Where is it referenced?" "What are the implicit couplings and dependencies?" And so on.

The single letter variable naming system is to make that apparent. More than three single letter variables indicates modularization is required.

It took me quite a while to figure out this code. The long names hindered my ability to understand, because I needed to understand what the "hook" part of it. However, that's not what it does. It's about mapping fields to values. The algorithm is important for Redux to reduce dependencies, but that should be described by the name where it is used. HookFieldMap gives enough of a clue that this is about hooks. FieldMap is about what this code does.

To that end, I turned it into a map so filtering didn't need to happen on lookups. That code was confusing:

```js
    getHookedDependency = (dependency) => {
        return this.hookedDependencies.find(hookedDependency => {
            return (dependency.modelName === hookedDependency.modelName &&
                dependency.fieldName === hookedDependency.fieldName);
        })
    }
```

By making SPath (below) responsible for storing the fully qualified name (spath) and the decomposed name (modelName and fieldName), the method was trivialized, and also surfaced the problem that it wasn't fail fast. If the model is not found, the above code will return None. 

I did make the assumption that fields wanted to be reduced, too. This could be wrong. If so, this should be called out in the code clearly: preserving duplicates.

I don't understand modern javascript well enough, but I think the `...` and `{}` "expansions" simplified the code. Again, this eliminates many unnecessary temporary variables.

There are other [DesignHints](https://github.com/radiasoft/pykern/wiki/DesignHints) here. Please read through the code and discuss with me or in this PR.

```js
export class SPath {
    constructor(spath) {
        {this.modelName, this.fieldName} = spath.match(/^(\w[^\.]*))\.(\w[^\.]*)$/);
        if (!this.modelName ){
            throw new Error(`spath=${spath} must be in model.field form`);
        }
        this.spath = spath;
    };
};

export class HookFieldMap {
    constructor({ dependencies, models, schemaModels }) {
        let cache = {};
        let modelInfo = (name) => {
            if (! (name in cache)) {
                //TODO: is this a fail fast object?
                let m = models.forModel(name);
                cache[name] = {
                    ...m,
                    modelSchema: schemaModels[name],
                    modelValue: m.hookModel(),
                };
            }
            return cache[name];
        };
        this.fields = Object.fromEntries(
            dependencies.map(
                (spath) => {
                    let m = modelInfo(spath.modelName);
                    let s = m.modelSchema[spath.fieldName];
                    if(! s) {
                        throw new Error(`field=${path.field} not in model=${spath.model} schema=${m.modelSchema}`)
                    }
                    return [
                        spath,
                        {
                            ...spath,
                            ...s,
                            model: m,
                            value: m.value[spath.fieldName],
                        },
                    ],
                },
            ),
        );
    };

    find = (spath) => {
        //TODO: needs to be a fail fast object
        if (! (spath in this.fields) {
            throw new Error(`field=${spath} not found in HookFieldMap`);
        }
        return this.fields[spath];
    };
}
```

garsuga:
This is a new set of tasks that should be completed as part of JSPEC's implementation

- [ ] Handle manual report panels somehow (hide/show, loading indicator, message?)
- [ ] Particle 3d visual
- [ ] Hide/show tabs instead of create/destroy to preserve state
- [ ] Investigate caching issue with report execution
- [ ] Scientific notation where applicable
- [ ] Fix graph axes long numbers being cut off

garsuga:
This will be ready to complete after merging https://github.com/radiasoft/sirepo/tree/4927-react-heatplot with https://github.com/radiasoft/sirepo/pull/4944

garsuga:
This PR will fix https://github.com/radiasoft/sirepo/issues/4964 as well

garsuga:
Accidentally committed to the wrong branch above. Reverted it, haha

garsuga:
@robnagler all tasks are complete now. I have opened some issues for other low priority fixes but this should be enough to start any future applications on react.

garsuga:
This PR now fixes https://github.com/radiasoft/sirepo/issues/4980

react/src/layout/navbar.js
```diff
@@ -0,0 +1,204 @@
...
+        let isDirty = formController.isFormStateDirty();
+        let isValid = formController.isFormStateValid();
+        let actionButtons = <ViewPanelActionButtons canSave={isValid} onSave={_submit} onCancel={_cancel}></ViewPanelActionButtons>
+
+        let { Portal: NavbarPortal, portalRef } = usePortal({
+            bindTo: document && document.getElementById(NavbarContainerId)
+        })
+
+        if(portalRef && portalRef.current) {
+            portalRef.current.classList.add("float-left");
```
robnagler, line 71: Seems like modularization is important here.

react/src/layout/navbar.js
```diff
@@ -0,0 +1,204 @@
...
+            iconElement = <FontAwesomeIcon fixedWidth icon={Icon[icon]}></FontAwesomeIcon>;
+        }
+
+        // TODO fix button cursor on hover
+        return (
+            <>
+                <NavbarPortal>
+                    <Col>
+                        <div onClick={() => updateModalShown(true)} variant="secondary">
+                            <span>{title}<a className="ms-2">{iconElement}</a></span>
```
robnagler, line 87: ms-2 is used elsewhere.

react/src/layout/navbar.js
```diff
@@ -0,0 +1,204 @@
...
+                        </div>
+                    </Col>
+                </NavbarPortal>
+
+                <Modal show={modalShown} onHide={() => _cancel()} size="lg">
+                    <Modal.Header className="lead bg-info bg-opacity-25">
```
robnagler, line 93: I think classNames should come from somewhere to avoid errors where the names aren't found. bg-opacity-25 is something that should be defined elsewhere. `lead` is also something about formatting and should be defined in the style.

react/src/layout/panel.js
```diff
@@ -0,0 +1,81 @@
...
+            showButtons: formController.isFormStateDirty(),
+            formValid: formController.isFormStateValid(),
+            mainChildren,
+            modalChildren,
+            title: title || config.name,
+            id: config.name
+        }
+
+        return (
+            <Col md={6} xl={4} className="mb-3">
```
robnagler, line 75: Modularize this style. It should have a name.

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+        frameIndex,
+        reportName,
+        simulationId,
+        appName,
+        computeJobHash,
+        computeJobSerial,
+        ...((hookedDependendies || []).map(v => v.value))
+    ]
+
+    return frameIdElements.join('*');
```
robnagler, line 98: constant.

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+
+export class ManualRunReportLayout extends View {
+    getFormDependencies = (config) => {
+        let { reportLayout } = config;
+        let layoutElement = this.layoutsWrapper.getLayoutForConfig(reportLayout);
+        return layoutElement.getFormDependencies();
+    }
+
+    component = (props) => {
+        let { config } = props;
```
robnagler, line 109: The blank lines are arbitrary and confusing. Better off not having them.

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+                }
+
+                updateReportData(undefined);
+
+                let { state } = simulationData;
+                if(state == "completed") {
+                    simulationInfoPromise.then(({simulationId}) => {
+                        let { frameCount, computeJobHash, computeJobSerial } = simulationData;
+                        let frameId = getFrameId({
+                            frameIndex: frameCount - 1, // pick last frame for now
```
robnagler, line 154: "on the side" comments" are hard to manage. Rather, name a function "lastFrame(frameCount)"

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+        useEffect(() => {
+            panelController.setShown(shown && !!reportData);
+        }, [shown, !!reportData])
+
+        let VisualComponent = reportData ? layoutElement.component : undefined;
+
+        // set the key as the key for the latest request sent to make a brand new report component for each new request data
+        return (
+            <>
+                {VisualComponent && shown && <VisualComponent key={reportEventsVersionRef.current} config={reportLayout} simulationData={reportData}></VisualComponent>}
```
robnagler, line 183: simulationData -> simData or just reportData.

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+            </>
+        )
+    }
+}
+
+export class SimulationStartLayout extends View {
```
robnagler, line 189: Let's ban "Simulation" and just use "Sim".

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+        let endSimulationButton = <Button variant="primary" onClick={endSimulation}>End Simulation</Button>;
+        let startSimulationButton = <Button variant="primary" onClick={startSimulation}>Start Simulation</Button>
+
+        if(lastSimulationData) {
+            let { state } = lastSimulationData;
+
+            let getStateBasedElement = (state) => {
+                let elapsedTimeSeconds = Math.ceil(((new Date()) - lastSimulationStartTime) / 1000)
+                switch(state) {
+                    case 'pending':
```
robnagler, line 256: I saw these used elsewhere. Modularize.

react/src/layout/report.js
```diff
@@ -0,0 +1,297 @@
...
+                        return (
+                            <Stack gap={2}>
```
robnagler, line 258: Modularize.

robnagler (changes requested):
The main comment is modularize styling and move out of from business logic. The coding style needs to change to be more aligned how we do things here. There are good reasons for no naked prints, consistency in naming, etc. Read the python style guide to understand those reasons and apply them to Javascript.

garsuga:
This PR will also fix https://github.com/radiasoft/sirepo/issues/4982

react/src/component/app.js
```diff
@@ -0,0 +1,94 @@
...
+import { formStatesSlice } from "../store/formState";
+import { useSetup } from "../hook/setup";
+import { compileSchemaFromJson } from "../utility/schema";
+import { Provider } from "react-redux";
+import { SimulationBrowserRoot } from "../component/simbrowser";
+import { Layouts } from "../layout/layouts";
+import "./app.scss";
+
+function SimulationListInitializer(props) {
+    let stateFn = useState;
```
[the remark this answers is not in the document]
garsuga: These aliases are just to make React's linter happy. They don't like it when we access hooks from 'uncertain' code paths. We base things off of the schema which is assumed to be constant so even some conditionals can be considered to be consistent across all renders if they are just asking about the schema. The aliases aren't needed in all the places they are present. I need to go around and check them again. I will also look into ignoring the linter checks but they may help our other developers get used to React

react/src/utility/compute.js
```diff
@@ -0,0 +1,108 @@
+export function pollCompute({ doFetch, pollInterval, callback }) {
+    console.log("started polling compute");
+    let doIteration = () => {
+        doFetch().then(async (resp) => {
+            let respObj = await resp.json();
+            let { state } = respObj;
+            console.log("polled compute: " + state);
```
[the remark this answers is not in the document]
garsuga: Sure, most of the debug here is not designed for long term use anyways and I should go around and remove it now that the functionality is mostly stable. Should we have that debug function ready for this PR or is that a separate issue?

react/src/component/app.js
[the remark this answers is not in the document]
robnagler: I think it is ok to use something like "linterUseState" or something like that. Could be something shorter that's unique. I assume it can't be a suffix e.g. userStateRLH (react linter happy).

react/src/utility/compute.js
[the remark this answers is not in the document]
robnagler: Genesis is the priority. Make these issues including linter aliases. We need to wrap errors, prints, etc. Once Genesis is done, aligning the code bases will be a priority.

react/src/layout/report.js
garsuga: How about using 'report' universally. I already refer to the concept of a 'manually started simulation' as a 'report group' within the schema and code elsewhere. Would the name ReportGroupStatusLayout be a good replacement here or do you prefer 'sim'?

robnagler: There are two concepts: reports (subclass animations) and computes. When you press "Simulation start" you are staring a compute, not a report. This language needs to be made clear, since a compute maps to many reports. I don't like ReportGroup, because we already try to distinguish this on the server with compute and report and ReportGroup is not necessarily the only way computes are accessed. Sometimes you want to see if a compute is running. I think Jspec looks at a group of Elegant computes for twiss parameters to be used as input to Jspec.

robnagler (approved):
@garsuga please make sure any comments in this PR are issues.

robnagler:
@garsuga please resolve conflicts

garsuga:
@robnagler done

garsuga:
I am getting errors after the merge saying "no module found pykern.quest". Is there a dependency I need to upgrade?

robnagler:
I don't think the merge worked properly. See failing tests: NameError: name 'cfg' is not defined in server.py

You need to:

cd ~/src/radiasoft/pykern; git pull

PR #4805, opened 2022-09-07
fix #4804 ML app improvements

### LINKED ISSUE #4804: Improvements for ML app
Several improvements for the ML app:
- change padding and interpolation from type String to enum values
- default padding should be "valid"
- no longer filter available layers based on dev status
- show layer columns in API order, esp for Dimensionality to be the first arg, define column order in schema

moellep:
- change padding and interpolation from type String to enum values
- default padding should be "valid"
- no longer filter available layers based on dev status
- show layer columns in API order, esp. for Dimensionality to be the first arg, define column order in schema


gurhar1133 (approved):
Looks good!

PR #4809, opened 2022-09-08
Fix #4719 complete tally voxels

### LINKED ISSUE #4719: Complete tally voxels
The current tally visualization displays a single side of the cell of interest. Fill in the rest.

mkeilman:
Notes:
- Builds a voxel per mesh cell by creating vtkCubeSources and adding their data to nested vtkAppendPolyData objects. Nesting is done because vtkAppendPolyData seems to give up adding data at some point (it would be nice to find the cause and contribute to the vtk.js repo if there's a bug)
- Assumes a single mesh filter
- Moves the colorMap and inset percentage to a separate model


mkeilman:
I can't figure out why the check failed before but it passed with [4912137](https://github.com/radiasoft/sirepo/pull/4809/commits/49121370e725fe22b76ac3b9fc8859c8283bcfcd)...

robnagler:
> I can't figure out why the check failed before but it passed with [4912137](https://github.com/radiasoft/sirepo/pull/4809/commits/49121370e725fe22b76ac3b9fc8859c8283bcfcd)...

if you need help, supply links to the failed checks and passing checks. this will make it easier to trace.

moellep:
This example from alpha uses one mesh for t1 and a different mesh for t2. With the new software, the t2 plot doesn't show, but it works on alpha:

https://alpha.sirepo.com/cloudmc#/visualization/xxCWI7bf

mkeilman:
> This example from alpha uses one mesh for t1 and a different mesh for t2. With the new software, the t2 plot doesn't show, but it works on alpha:
> 
> https://alpha.sirepo.com/cloudmc#/visualization/xxCWI7bf

[41eae9e](https://github.com/radiasoft/sirepo/pull/4809/commits/41eae9eac4cb51cc0290294aa43ba07a5c03754a)

moellep:
That's fixed and it looks good. It does take way too long to build the voxels for a 100x100x100 grid. I think vtkCubeSource or something similar must have too much overhead. I'm going to see if it is easy to just calculate one polydata with all the facets.

mkeilman:
> That's fixed and it looks good. It does take way too long to build the voxels for a 100x100x100 grid. I think vtkCubeSource or something similar must have too much overhead. I'm going to see if it is easy to just calculate one polydata with all the facets.

One can build the points and cells by hand; that was my original approach. I don't know if that counts as "easy" :) but it may well be faster.

moellep:
I've checked in a change to manually calculate the voxels. It doesn't use voxelInsetPct, but I think that could be easily added to the calculation. It renders 100x100x100 in 9 seconds or so, and 3 seconds of that is calculating the colors which I think can also be sped up.

I also removed the POINTS and POLYS from the data returned from the server - I think the client only needs the FIELD data.

I couldn't figure out how to get the polydata working with the ActorBundle, so I left some TODOs in the code for that.

mkeilman:
ping - cleaned up a few things. I think performance improvements re: colors can move to a fresh issue

mkeilman:
> ping - cleaned up a few things. I think performance improvements re: colors can move to a fresh issue

or you could just do it now :) [5c6e1e9](https://github.com/radiasoft/sirepo/pull/4809/commits/5c6e1e92f68dd1faf28300235488ac428ef7b471) 

Is this ready then?

PR #4842, opened 2022-09-14
4549 browser log viewer

### LINKED ISSUE #4549: 20220915 log file viewer in browser
It would be nice if the log file links across various applications (elegant, srw, flash, etc) would show a modal window with the file contents, with an option to download, rather than the current behavior which is to immediately download the log file.

An example could be found in the flash Problem Files panel which allows viewing the individual files in the problem set. This could be extended to add a "download" icon in the header to allow the user to download the file contents.

Screen Shot 2022-06-22 at 12 04 35 PM

Benbenbenin0:
Fix #4549: Replaced "View log" hyperlink with new directive that views in browser

### DIFF
--- a/sirepo/package_data/static/html/elegant-visualization.html
+++ b/sirepo/package_data/static/html/elegant-visualization.html
@@ -34,7 +34,6 @@
                 <div data-view-log-iframe></div>
               </div>
             </div>
-            <!--<div data-run-simulation-fields=""></div>-->
             <div class="col-sm-6 pull-right" data-ng-show="visualization.simState.isStopped()">
               <button class="btn btn-default" data-ng-click="visualization.startSimulation()">{{ visualization.startButtonLabel() }}</button>
             </div>
--- a/sirepo/package_data/static/html/elegant-visualization.html
+++ b/sirepo/package_data/static/html/elegant-visualization.html
@@ -33,7 +33,6 @@
                 </div>
                 <div data-view-log-iframe></div>
               </div>
-
             </div>
             <!--<div data-run-simulation-fields=""></div>-->
             <div class="col-sm-6 pull-right" data-ng-show="visualization.simState.isStopped()">
--- a/sirepo/package_data/static/html/elegant-visualization.html
+++ b/sirepo/package_data/static/html/elegant-visualization.html
@@ -33,6 +33,9 @@
                 </div>
                 <a data-ng-href="{{ visualization.logFileURL() }}" target="_blank">View elegant log</a>
               </div>
+
+              <div data-view-log-iframe></div>
+
             </div>
             <!--<div data-run-simulation-fields=""></div>-->
             <div class="col-sm-6 pull-right" data-ng-show="visualization.simState.isStopped()">
--- a/sirepo/package_data/static/html/elegant-visualization.html
+++ b/sirepo/package_data/static/html/elegant-visualization.html
@@ -31,10 +31,8 @@
                   <p class="elegant-error-box">{{ visualization.simulationAlerts }}</p>
                   <br>
                 </div>
-                <a data-ng-href="{{ visualization.logFileURL() }}" target="_blank">View elegant log</a>
+                <div data-view-log-iframe></div>
               </div>
-
-              <div data-view-log-iframe></div>
 
             </div>
             <!--<div data-run-simulation-fields=""></div>-->
--- a/sirepo/package_data/static/js/elegant.js
+++ b/sirepo/package_data/static/js/elegant.js
@@ -285,6 +285,26 @@ SIREPO.app.factory('elegantService', function(appState, commandService, requestS
 
     };
 
+    self.dataFilePath = function(model, filename) {
+        if (! appState.isLoaded()) {
+            return '';
+        }
+        
+        requestSender.sendStatelessCompute(
+            appState,
+            (data) => {
+                srdbg("dataFilePath: ", data.html);
+                appState.models.elegantLogHTML = data.html
+            },
+            {
+                method: 'log_to_html',
+                model: model,
+                filename: filename,
+                simulationId: appState.models.simulation.simulationId
+            });
+        
+    };
+
     self.findFirstCommand = function(types, commands) {
         if (! commands) {
             if (! appState.isLoaded()) {
@@ -620,6 +640,13 @@ SIREPO.app.controller('VisualizationController', function(appState, elegantServi
 
     self.logFileURL = function() {
         return elegantService.dataFileURL(self.simState.model, -1);
+    };
+
+    self.logFilePath = function() {
+        elegantService.dataFilePath(self.simState.model, 'elegant.log');
+        srdbg("appstate", appState.models.simulationStatus.animation.computeModel);
+        srdbg("simState", self.simState.model);
+        return appState.models.elegantLogHTML;
     };
 
     self.runningStatusText = function() {
@@ -1440,6 +1467,73 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
                     ['bunch.emit_x', 'bunch.emit_y', 'bunch.emit_nx', 'bunch.emit_ny'],
                     updateEmittance);
             });
+        },
+    };
+});
+
+
+SIREPO.app.directive('viewLogIframe', function(appState, requestSender) {
+    return {
+        restrict: 'A',
+        scope: {
+
+        },
+        template: `
+            <div class="well well-lg">
+                <div data-ng-if="visualization.simulationAlerts">
+                  <div class="text-warning"><strong>{{ visualization.errorHeader() }}</strong></div>
+                  <p class="elegant-error-box">{{ visualization.simulationAlerts }}</p>
+                  <br>
+                </div>
+                <a href data-ng-click="viewLog()">Look At Log</a>
+            </div>
+
+            <div class="modal fade" id="sr-iframe-text-view" tabindex="-1" role="dialog">
+              <div class="modal-dialog modal-lg">
+                <div class="modal-content">
+                  <div class="modal-header bg-warning">
+                    <button type="button" class="close" data-dismiss="modal">
+                      <span>&times;</span>
+                    </button>
+                    <span class="lead modal-title text-info">Title</span>
+                  </div>
+                  <div class="modal-body" style="padding: 0">
+                    <iframe id="sr-text-iframe"
+                      style="border: 0; width: 100%; height: 80vh" src=""></iframe>
+                  </div>
+                </div>
+              </div>
+            </div>
+        `,
+        controller: function(appState, requestSender, $scope) {
+
+            //function setIFrameHTML(html) {
+                //$('#sr-text-iframe').contents().find('html').html(html);
+            //}
+
+            $scope.viewLog = function(model, filename) {
+                if (! appState.isLoaded()) {
+                    return '';
+                }
+
+                //$('#sr-iframe-text-view').modal('show');
+                //setIFrameHTML('<div style="text-align: center; padding: 1em">Loading Log</div>');
+
+                requestSender.sendStatelessCompute(
+                    appState,
+                    (data) => {
+                        srdbg("dataFilePath: ", data.html);
+                        appState.models.modelLog = data.html;
+                        srdbg("form app state: ", appState.models.modelLog);
+                    },
+                    {
+                        model: appState.models.simulationStatus.animation.computeModel,
+                        filename: 'elegant.log',
+                        simulationId: appState.models.simulation.simulationId
+                    });
+                
+                //setIFrameHTML(appState.models.modelLog)
+            };
         },
     };
 });
--- a/sirepo/package_data/static/js/elegant.js
+++ b/sirepo/package_data/static/js/elegant.js
@@ -1453,18 +1453,21 @@ SIREPO.app.directive('viewLogIframe', function(appState, requestSender) {
         },
         // TODO: At error case for when log cannot be found, edit case below
         template: `
-            <div class="well well-lg">
-                <a href data-ng-click="viewLog()">Look At Log</a>
-            </div>
-
+            <a href data-ng-click="viewLog()">View Log</a>
             <div class="modal fade" id="sr-iframe-text-view" tabindex="-1" role="dialog">
               <div class="modal-dialog modal-lg">
                 <div class="modal-content">
                   <div class="modal-header bg-warning">
-                    <button type="button" class="close" data-dismiss="modal">
-                      <span>&times;</span>
-                    </button>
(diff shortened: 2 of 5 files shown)


sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,81 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                <div class="modal-content">
+                  <div class="modal-header bg-warning">
+                    <span class="lead modal-title text-info">Log</span>
+                    <div class="sr-panel-options pull-right">
+                      <a data-ng-href="{{ downloadLog() }}" target="_blank">
+                        <span class="sr-panel-heading glyphicon glyphicon-cloud-download" style="margin-bottom: 0"></span>
+                      </a>
+                      &nbsp&nbsp
+                      <button type="button" class="close" data-dismiss="modal">
+                        <span>&times;</span>
```
moellep, line 1468: Rather than the nbsp nbsp, you could add a style="margin-left: 1ex" to the close button.

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,81 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+            <div class="modal fade" id="sr-iframe-text-view" tabindex="-1" role="dialog">
+              <div class="modal-dialog modal-lg">
+                <div class="modal-content">
+                  <div class="modal-header bg-warning">
+                    <span class="lead modal-title text-info">Log</span>
+                    <div class="sr-panel-options pull-right">
+                      <a data-ng-href="{{ downloadLog() }}" target="_blank">
+                        <span class="sr-panel-heading glyphicon glyphicon-cloud-download" style="margin-bottom: 0"></span>
+                      </a>
+                      &nbsp&nbsp
```
moellep, line 1466: If you remove the whitespace before the end anchor, the download icon won't show that odd blip when you hover over it.

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,81 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                if (! appState.isLoaded()) {
+                    return '';
+                }
+                
+                requestSender.sendStatelessCompute(
+                    appState,
+                    (data) => {
+                        appState.models.modelLog = data.html;
+                        setIFrameHTML(appState.models.modelLog);
+                    },
```
moellep, line 1508: You don't need to set anything on appState. You can call setIFrameHTML(data.html) directly.

sirepo/template/elegant.py
```diff
@@ -1453,3 +1456,23 @@ def _sdds_init():
     _x = getattr(_s, "SDDS_LONGDOUBLE", None)
     _SDDS_DOUBLE_TYPES = [_s.SDDS_DOUBLE, _s.SDDS_FLOAT] + ([_x] if _x else [])
     _SDDS_STRING_TYPE = _s.SDDS_STRING
+
+
+def stateless_compute_log_to_html(data):
```
robnagler, line 1461: This is a stateful_compute, because it refers to a specific run dir which must have "state" (the log file). 

Here's where it is explained: https://github.com/radiasoft/sirepo/issues/2201#issuecomment-831591025

@e-carlin is there documentation?

sirepo/template/elegant.py
```diff
@@ -1453,3 +1456,23 @@ def _sdds_init():
...
+    text = pkio.read_text(r.join(data.filename))
+    t = pygments.highlight(
+        text,
+        pygments.lexers.get_lexer_by_name("text"),
+        pygments.formatters.HtmlFormatter(
+            noclasses=False,
+            linenos=False,
+        ),
+    )
+    return PKDict(html=t)
```
robnagler, line 1478: missing newline at eof. Maybe set up your editor to always put in a newline at eof.

sirepo/template/elegant.py
```diff
@@ -1453,3 +1456,23 @@ def _sdds_init():
...
+def stateless_compute_log_to_html(data):
+    r = simulation_db.simulation_run_dir(
+        PKDict(
+            simulationType=data.simulationType,
+            simulationId=data.simulationId,
+            report="animation",
+        )
+    )
+    text = pkio.read_text(r.join(data.filename))
+    t = pygments.highlight(
```
robnagler, line 1470: text and t temporaries unnecessary. They don't shorten the code, and it's not too deep nesting.

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,81 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                    {
+                        method: 'log_to_html',
+                        model: appState.models.simulationStatus.animation.computeModel,
+                        filename: 'elegant.log',
+                        simulationId: appState.models.simulation.simulationId
+                    });
```
moellep, line 1514: Don't pass filename to the server - that would allow this API to access any file on the server. Instead, the server side should use the ELEGANT_LOG_FILE constant for the filename.

sirepo/package_data/static/html/elegant-visualization.html
```diff
@@ -31,8 +31,9 @@
                   <p class="elegant-error-box">{{ visualization.simulationAlerts }}</p>
                   <br>
                 </div>
-                <a data-ng-href="{{ visualization.logFileURL() }}" target="_blank">View elegant log</a>
+                <div data-view-log-iframe></div>
               </div>
+
```
robnagler, line 36: extra space?

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,81 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
         },
     };
 });
+
+
+SIREPO.app.directive('viewLogIframe', function(appState, requestSender) {
+    return {
+        restrict: 'A',
+        scope: {
+
```
robnagler, line 1452: unnecessary space? `scope: {}`?

robnagler (changes requested):
minor comments

sirepo/template/elegant.py
moellep, line 1478: You can remove the temporary variables "text" and "t" and just place them inline.

sirepo/template/elegant.py
```diff
@@ -1453,3 +1456,23 @@ def _sdds_init():
...
     _SDDS_DOUBLE_TYPES = [_s.SDDS_DOUBLE, _s.SDDS_FLOAT] + ([_x] if _x else [])
     _SDDS_STRING_TYPE = _s.SDDS_STRING
+
+
+def stateless_compute_log_to_html(data):
+    r = simulation_db.simulation_run_dir(
+        PKDict(
+            simulationType=data.simulationType,
+            simulationId=data.simulationId,
+            report="animation",
```
moellep, line 1466: Use "report=data.model" rather than "animation".

moellep (commented):
This is a good start - a few minor things to review.

sirepo/template/elegant.py
e-carlin: I think that is an analysis job. statefulCompute is just sirepo-data.json (doesn't go to NERSC for example)

https://github.com/radiasoft/sirepo/wiki/Interacting-with-Simulations-Architecture

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,75 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                return requestSender.formatUrl('downloadDataFile', {
+                    '<simulation_id>': appState.models.simulation.simulationId,
+                    '<simulation_type>': SIREPO.APP_SCHEMA.simulationType,
+                    '<model>': model,
+                    '<frame>': index,
+                });
+            }
+
+            $scope.viewLog = function(model, filename) {
+                if (! appState.isLoaded()) {
```
e-carlin, line 1495: I don't believe this check is necessary. I think that appState is guaranteed to be loaded by the time this directive is rendered. Is there a case this is covering that I'm missing?

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,75 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                }
+                
+                requestSender.sendAnalysisJob(
+                    appState,
+                    (data) => {
+                        setIFrameHTML(data.html);
+                    },
+                    {
+                        method: 'log_to_html',
+                        computeModel: appState.models.simulationStatus.animation.computeModel,
```
e-carlin, line 1506: I think the more "correct" way to get this is to specify `elegantService` in the arguments to the controller function above. Then here to say `elegantService.computeModel()`. 

Every app has a `<app-name>Service` in Sirepo. And every one of those (or almost every one) has an abstraction for getting the computeModel.

sirepo/template/elegant.py
```diff
@@ -28,6 +28,9 @@
 import sirepo.lib
 import sirepo.sim_data
 import stat
+import pygments
```
e-carlin, line 31: Sort imports alphabetically.

sirepo/template/elegant.py
```diff
@@ -1453,3 +1456,24 @@ def _sdds_init():
     _x = getattr(_s, "SDDS_LONGDOUBLE", None)
     _SDDS_DOUBLE_TYPES = [_s.SDDS_DOUBLE, _s.SDDS_FLOAT] + ([_x] if _x else [])
     _SDDS_STRING_TYPE = _s.SDDS_STRING
+
+
+def analysis_job_log_to_html(data):
+    r = simulation_db.simulation_run_dir(
```
e-carlin, line 1462: No need to fix here but I'll fix at some point https://github.com/radiasoft/sirepo/issues/4850

e-carlin (commented):
Looks good. Just a few nits.

sirepo/package_data/static/html/elegant-visualization.html
```diff
@@ -31,7 +31,7 @@
                   <p class="elegant-error-box">{{ visualization.simulationAlerts }}</p>
                   <br>
                 </div>
-                <a data-ng-href="{{ visualization.logFileURL() }}" target="_blank">View elegant log</a>
+                <div data-view-log-iframe></div>
               </div>
             </div>
             <!--<div data-run-simulation-fields=""></div>-->
```
gurhar1133, line 37: Can remove this comment

gurhar1133 (approved):
Looks good to me. Good job!

sirepo/package_data/static/js/elegant.js
Benbenbenin0: Good point think its unneccesary

sirepo/package_data/static/js/elegant.js
Benbenbenin0: Got it good to know.

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,72 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+            </div>
+        `,
+        controller: function(appState, elegantService, requestSender, $scope) {
+
+            function setIFrameHTML(html) {
+                $('#sr-text-iframe').contents().find('html').html(html);
+            }
+
+            function getDownloadURL(model, index) {
+                if (! appState.isLoaded() || ! model) {
```
e-carlin, line 1483: I don't believe this appState or model check are needed

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,72 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                      style="border: 0; width: 100%; height: 80vh" src=""></iframe>
+                  </div>
+                </div>
+              </div>
+            </div>
+        `,
+        controller: function(appState, elegantService, requestSender, $scope) {
+
+            function setIFrameHTML(html) {
+                $('#sr-text-iframe').contents().find('html').html(html);
```
e-carlin, line 1479: I think this can be simplified to `$('#sr-text-iframe').attr("srcdoc", html);`. srcdoc is the attribute for setting an html string on an iframe https://www.w3schools.com/tags/att_iframe_srcdoc.asp

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,72 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                    '<model>': model,
+                    '<frame>': index,
+                });
+            }
+
+            $scope.viewLog = function(model, filename) {
+                requestSender.sendAnalysisJob(
+                    appState,
+                    (data) => {
+                        srdbg(elegantService)
```
e-carlin, line 1498: Remove debug statement

sirepo/package_data/static/js/elegant.js
```diff
@@ -1443,3 +1443,72 @@ SIREPO.app.directive('srBunchEditor', function(appState, panelState) {
...
+                    {
+                        method: 'log_to_html',
+                        computeModel: elegantService.computeModel(),
+                        simulationId: appState.models.simulation.simulationId
+                    });
+                $('#sr-iframe-text-view').modal('show');
+            };
+
+            $scope.downloadLog = function() {
+                return getDownloadURL(appState.models.simulationStatus.animation.computeModel, -1);
```
e-carlin, line 1510: Since `getDownloadUrl` is only ever called here and there is no other code in this function I think the contents of `getDownloadUrl` can be moved here and that `getDownloadUrl` function can be removed

e-carlin (approved):
Just a couple nits but looks good. I'll let Paul approve and merge when he reviews.

sirepo/package_data/static/js/elegant.js
moellep: appState.isLoaded() will always be true, but model is "appState.models.simulationStatus.animation.computeModel" which gets set with the sim status response, so it may be null initially.

sirepo/package_data/static/js/elegant.js
moellep: Also, setIFrameHTML() is a single line method which is called once, so the method could be removed and the code called directly in the sendAnalysisJob data handler.

sirepo/package_data/static/js/elegant.js
moellep: You can catch these cases by running "sh test.sh" in the ~/sirepo directory before checking in - it will run jslint on the code, some karma tests, and then the full test suite. For UI changes like this, I usually kill the "sh test.sh" after it gets past the jshint and karma tests because the full test suite takes a long time to run.

PR #4846, opened 2022-09-15
Fix #4836 conftest port

### LINKED ISSUE #4836: 20220924 conftest needs a configured port
This is for devbox environments. conftest has a hardwired port. I suggest adding SIREPO_CONFTEST_PORT and have conftest create all ports based off of this (+1, +2, etc.).

mkeilman:
(initial pr to see if tests work - they often fail locally due to resource constraints)

This may be "overcooked", but it uses a local dict to keep track of the ports. It looks like everything happens in one method so I put everything there.

While I was at it, I also made constants for the ports in `service.py`


mkeilman:
ping for first impressions

tests/conftest.py
```diff
@@ -303,26 +303,41 @@ def _slurm_not_installed():
...
+    import sirepo.pkcli.service
     from pykern.pkcollections import PKDict
 
+    ports = PKDict()
+
+    # different port than default so can run tests when supervisor running
+    def _next_conftest_port(port_key, port_pkg):
+        p = ports.get(port_key)
+        if p:
+            ports[port_key] = p + 1
```
robnagler, line 316: This is a bit wonky. Let's define all cases in PORT_DEFAULTS so we don't have this type of thing and instead assert ports in all cases. 

I think the code in _job_supervisor_check could be generalized to test for the socket so we can ask the question here if the socket is in use. If it is in use, then we just increment by the default amount. This would simplify the problem of collisions. Perhaps that's the best approach. Thoughts?

tests/conftest.py
```diff
@@ -303,26 +303,41 @@ def _slurm_not_installed():
...
     """setup the supervisor"""
     import os
+    import sirepo.job
+    import sirepo.pkcli.service
     from pykern.pkcollections import PKDict
 
+    ports = PKDict()
+
+    # different port than default so can run tests when supervisor running
+    def _next_conftest_port(port_key, port_pkg):
```
robnagler, line 313: `_port()`. We are in conftest so we don't need to add this name to a function. This is nested function so the space/use is limited. I do like "port_key" though. :)

tests/conftest.py
```diff
@@ -303,26 +303,41 @@ def _slurm_not_installed():
 def _subprocess_setup(request, cfg=None, uwsgi=False):
     """setup the supervisor"""
     import os
+    import sirepo.job
+    import sirepo.pkcli.service
     from pykern.pkcollections import PKDict
 
+    ports = PKDict()
```
robnagler, line 310: It could be argued that ports should be an object so we don't have the unrelated values. I'm not sure the complexity is worth it...

tests/conftest.py
```diff
@@ -303,26 +303,41 @@ def _slurm_not_installed():
...
     cfg.pkupdate(
         PYKERN_PKDEBUG_WANT_PID_TIME="1",
         SIREPO_PKCLI_JOB_SUPERVISOR_IP=_LOCALHOST,
         SIREPO_PKCLI_JOB_SUPERVISOR_PORT=p,
         SIREPO_SRDB_ROOT=str(pkio.mkdir_parent(pkunit.work_dir().join("db"))),
     )
     if uwsgi:
-        cfg.SIREPO_PKCLI_SERVICE_PORT = "8102"
-        cfg.SIREPO_PKCLI_SERVICE_NGINX_PROXY_PORT = "8180"
+        cfg.SIREPO_PKCLI_SERVICE_PORT = _next_conftest_port("DEFAULT_PORT", sirepo.job)
```
robnagler, line 336: It could be argued that this name is incorrect. It should be SIREPO_PKCLI_SERVICE_HTTP_PORT and separately *_UWSGI_PORT.  Having explicit values would simplify things. The PORT_DEFAULTS would be the same, which would argue against this...

robnagler (changes requested):
Big question is whether we should test for ports. The problem we are solving is really general: is there going to be a port collision? I think if we put in a port in use check, we can avoid many issues. There are always race conditions with this type of test, but given that this is just for testing, I think we can live with it. Thoughts?

tests/conftest.py
mkeilman: Would there be a loop to keep checking the next port (to some limit) or one and done?

robnagler: Correct. You would loop until you found a free port.

robnagler: You need to close the port once if is opened. You can continue to bind the same socket, btw, so the try/except only needs to be around the bind.

mkeilman:
Notes on latest push:
- Sockets now use a context manager so we can use `with as`.
- Sockets cannot be rebound once they are closed.  Doing so raises `[Errno 9] Bad file descriptor`.
- We still need to keep track of the "busy" ports but an array is sufficient. Alternatively we could make sockets without `socket.SO_REUSEADDR `and leave them unclosed - then the next socket's bind will fail. I think that means avoiding with...as and requires recursion. I don't know if that's better.

mkeilman:
ping for comments on latest approach

sirepo/pkcli/service.py
```diff
@@ -306,13 +323,9 @@ def _cfg_ip(value):
...
 
 def _cfg_port(value):
-    return _cfg_int(3000, 32767)(value)
-
-
-def _cfg_react_port(value):
     if not value:
         return None
-    return _cfg_int(3000, 32767)(value)
+    return _cfg_int(sirepo.const.PORT_DEFAULTS.react, sirepo.const.PORT_MAX)(value)
```
robnagler, line 328: Please use PORT_MIN 1025, not react.

tests/conftest.py
```diff
@@ -240,26 +241,11 @@ def _config_sbatch_supervisor_env(env):
...
-    except Exception:
-        raise AssertionError(
-            "job_supervisor still running on ip={} port={}".format(
-                env.SIREPO_PKCLI_JOB_SUPERVISOR_IP,
-                env.SIREPO_PKCLI_JOB_SUPERVISOR_PORT,
-            ),
-        )
-    finally:
-        s.close()
+    _ = _port(
```
robnagler, line 244: The assignment to _ is unnecessary and confusing.

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
...
+def _port(port, busy_ports, ip=_LOCALHOST):
+    import socket
+    from pykern.pkdebug import pkdlog
+
+    p = int(port)
+    while sirepo.const.PORT_MIN <= p <= sirepo.const.PORT_MAX:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            try:
+                assert p not in busy_ports
```
robnagler, line 279: why is this an assertion and not simply a continue?

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
...
+                s.bind(
+                    (
+                        ip,
+                        p,
+                    )
+                )
+                busy_ports.append(p)
+                return str(p)
+            except Exception:
+                pkdlog(f"ip={ip} port={p} in use")
```
robnagler, line 289: don't log. it's a normal part of the algorithm

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
...
+            p += 1
+    raise AssertionError(f"ip={ip} port={p} invalid address")
```
robnagler, line 291: error is "unable to allocate port".

tests/conftest.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import requests
 import subprocess
+import sirepo.const
```
robnagler, line 6: don't import at this level in conftest, only at the place you need it.

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
     return fc_module
 
 
+def _port(port, busy_ports, ip=_LOCALHOST):
+    import socket
+    from pykern.pkdebug import pkdlog
+
+    p = int(port)
+    while sirepo.const.PORT_MIN <= p <= sirepo.const.PORT_MAX:
```
robnagler, line 275: This should be in in a smaller range that's outside the normal range for manually started servers, e.g. 10000-11000.  This can be called "TEST_START". If we are testing 30K+ ports, there's a bug. Narrowing the range avoids too long of a wait.

Start at a random point within the range so that we don't need to keep "busy_ports" state.

tests/conftest.py
```diff
@@ -312,17 +323,20 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
     from pykern import pkunit
     from pykern import pkio
 
-    # different port than default so can run tests when supervisor running
-    p = "8101"
+    ports = []
+    delta = sirepo.const.PORT_DELTA_FOR_TEST
```
robnagler, line 327: I think we should just have a range of ports for tests. sirepo.const defines this range.

tests/conftest.py
```diff
@@ -312,17 +323,20 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
     from pykern import pkunit
     from pykern import pkio
 
-    # different port than default so can run tests when supervisor running
-    p = "8101"
+    ports = []
```
robnagler, line 326: "ports" doesn't need to be saved.

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
...
 
 
+def _port(port, busy_ports, ip=_LOCALHOST):
+    import socket
+    from pykern.pkdebug import pkdlog
+
+    p = int(port)
+    while sirepo.const.PORT_MIN <= p <= sirepo.const.PORT_MAX:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
robnagler, line 277: We may not need to REUSEADDR if we pick a random starting point. This would be better.

tests/conftest.py
```diff
@@ -281,6 +267,30 @@ def _fc(request, fc_module, new_user=False):
...
+            try:
+                assert p not in busy_ports
+                s.bind(
```
robnagler, line 280: This should be on line line.

tests/conftest.py
```diff
@@ -303,6 +313,7 @@ def _slurm_not_installed():
 def _subprocess_setup(request, cfg=None, uwsgi=False):
     """setup the supervisor"""
     import os
+    import sirepo.const
```
robnagler, line 316: from sirepo import const when it is inside a function

sirepo/const.py
```diff
@@ -4,9 +4,23 @@
 :copyright: Copyright (c) 2021 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
-from __future__ import absolute_import, division, print_function
-
+from pykern.pkcollections import PKDict
```
robnagler, line 7: Do not import at this level. Put a comment to that effect.

robnagler (changes requested):

sirepo/pkcli/service.py
mkeilman: [2294f37](https://github.com/radiasoft/sirepo/pull/4846/commits/2294f370d4dfcddf29ae5982bdbeef19caea77c7)

tests/conftest.py
mkeilman: [a373bd2](https://github.com/radiasoft/sirepo/pull/4846/commits/a373bd279a8441720bda7dbf8a71bb3fe53e54b2)

tests/conftest.py
mkeilman: [52e56a4](https://github.com/radiasoft/sirepo/pull/4846/commits/52e56a438b57f19b8e689580bc4ada1cdc76541f)

tests/conftest.py
mkeilman: [632d1ba](https://github.com/radiasoft/sirepo/pull/4846/commits/632d1ba1de5b3a29582cb87843fdc777033b59ee)

tests/conftest.py
mkeilman: [308777d](https://github.com/radiasoft/sirepo/pull/4846/commits/308777d2447203787dc3df0950124714dd1d90f7)

tests/conftest.py
mkeilman: [308777d](https://github.com/radiasoft/sirepo/pull/4846/commits/308777d2447203787dc3df0950124714dd1d90f7)

tests/conftest.py
mkeilman: To use the same logging as bind failure - about to be obviated

mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

tests/conftest.py
mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

tests/conftest.py
mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

tests/conftest.py
mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

tests/conftest.py
mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

tests/conftest.py
mkeilman: [739348c](https://github.com/radiasoft/sirepo/pull/4846/commits/739348cec891a109dbd3ad4390a8c9390d92ba7a)

sirepo/const.py
mkeilman: I'm not sure what "this level" means? No imports in const.py? Does that mean `PORT_DEFAULTS` has to live in a function, or be a regular dict, or broken up into individual constants?

mkeilman:
Latest push:
- `_port()` now returns a random port from a range defined in `sirepo/const`
- ports already selected are not tracked

tests/conftest.py
```diff
@@ -397,3 +401,6 @@ def _subprocess(cmd):
         for x in p:
             x.terminate()
             x.wait()
+
```
robnagler, line 404: Global is unnecessary

tests/conftest.py
```diff
@@ -312,20 +318,18 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
     from pykern import pkunit
     from pykern import pkio
 
-    # different port than default so can run tests when supervisor running
-    p = "8101"
```
robnagler, line 316: p = _port()

tests/conftest.py
```diff
@@ -312,20 +318,18 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
...
     cfg.pkupdate(
         PYKERN_PKDEBUG_WANT_PID_TIME="1",
         SIREPO_PKCLI_JOB_SUPERVISOR_IP=_LOCALHOST,
-        SIREPO_PKCLI_JOB_SUPERVISOR_PORT=p,
+        SIREPO_PKCLI_JOB_SUPERVISOR_PORT=_PORT,
```
robnagler, line 324: use p here

tests/conftest.py
```diff
@@ -312,20 +318,18 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
...
     )
     if uwsgi:
-        cfg.SIREPO_PKCLI_SERVICE_PORT = "8102"
-        cfg.SIREPO_PKCLI_SERVICE_NGINX_PROXY_PORT = "8180"
+        cfg.SIREPO_PKCLI_SERVICE_PORT = _port()
+        cfg.SIREPO_PKCLI_SERVICE_NGINX_PROXY_PORT = _port()
     for x in "DRIVER_LOCAL", "DRIVER_DOCKER", "API", "DRIVER_SBATCH":
         cfg["SIREPO_JOB_{}_SUPERVISOR_URI".format(x)] = "http://{}:{}".format(
-            _LOCALHOST, p
+            _LOCALHOST, _PORT
```
robnagler, line 332: p

tests/conftest.py
```diff
@@ -312,20 +318,18 @@ def _subprocess_setup(request, cfg=None, uwsgi=False):
     from pykern import pkunit
     from pykern import pkio
 
-    # different port than default so can run tests when supervisor running
-    p = "8101"
     cfg.pkupdate(
         PYKERN_PKDEBUG_WANT_PID_TIME="1",
         SIREPO_PKCLI_JOB_SUPERVISOR_IP=_LOCALHOST,
```
robnagler, line 323: All services should use _LOCALHOST so set  SIREPO_PKCLI_SERVICE_IP=_LOCALHOST.

tests/conftest.py
```diff
@@ -281,6 +271,22 @@ def _fc(request, fc_module, new_user=False):
     return fc_module
 
 
+def _port(ip=_LOCALHOST):
```
robnagler, line 274: Don't make ip a param.

tests/conftest.py
```diff
@@ -281,6 +271,22 @@ def _fc(request, fc_module, new_user=False):
...
+    import random
+    from sirepo import const
+
+    r = range(const.TEST_PORT_START, const.TEST_PORT_END)
+    port = random.choice(r)
+    for _ in r:
+        try:
+            _check_port(ip, port)
```
robnagler, line 282: Remove "ip" from _check_port(). It's always _LOCALHOST in tests.

tests/conftest.py
```diff
@@ -281,6 +271,22 @@ def _fc(request, fc_module, new_user=False):
     return fc_module
 
 
+def _port(ip=_LOCALHOST):
+    import random
+    from sirepo import const
+
+    r = range(const.TEST_PORT_START, const.TEST_PORT_END)
+    port = random.choice(r)
+    for _ in r:
```
robnagler, line 280: Simplify `for p in random.sample(r, 100):`

Then you don't need port calc below, and 100 is better than 1000 when searching for ports. If we can't find a port after 100, something is wrong.

robnagler (changes requested):

tests/conftest.py
mkeilman: That was my original approach, but a global does appear necessary. When `supervisor_uri` gets set - first appearing here:
https://github.com/radiasoft/sirepo/blob/e80ddebeaed4493e940a0f493f61352a311354fe/sirepo/job_api.py#L353-L355
and used here:
https://github.com/radiasoft/sirepo/blob/e80ddebeaed4493e940a0f493f61352a311354fe/sirepo/job_api.py#L310-L311
it does not change until the next test file runs, while `_subprocess_setup` runs for every test case. That causes connection errors like this:
```
Sep 27 00:48:51 285570     0 ../../sirepo/pkcli/job_supervisor.py:78:default_command ip=127.0.0.1 port=10368
sirepo/job_api.py:231:request api=api_beginSession runDir=None
sirepo/session.py:81:_begin error=HTTPConnectionPool(host='127.0.0.1', port=10891): Max retries exceeded with url: /job-api-request (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fdf113c1fd0>: Failed to establish a new connection: [Errno 111] Connection refused')) trying job_api.begin_session stack=Traceback (most recent call last):
```
You can see the two different ports (which matched for the 1st test). Is the global `_PORT` working around a "bug" then?

tests/conftest.py
mkeilman: [2f06d99](https://github.com/radiasoft/sirepo/pull/4846/commits/2f06d994f5be7e9228fe8afef8319ca0979397c3)

tests/conftest.py
mkeilman: [2f06d99](https://github.com/radiasoft/sirepo/pull/4846/commits/2f06d994f5be7e9228fe8afef8319ca0979397c3)

tests/conftest.py
mkeilman: [61c6ce8](https://github.com/radiasoft/sirepo/pull/4846/commits/61c6ce81f04d1b0e1421ea52ca338c45c93bd8f3)

tests/conftest.py
mkeilman: Cool, that's better
[c490615](https://github.com/radiasoft/sirepo/pull/4846/commits/c4906152e17b741ed638943314bf35aa7222c15d)

I also moved the range into `const` since we only use the start and end ports to make a range anyway.

tests/conftest.py
mkeilman: ping for feedback on https://github.com/radiasoft/sirepo/pull/4846#discussion_r982451782

robnagler: It is a bug. It might be that _auth_fc_module needs a `fixture(scope="module")`. I don't know which tests are failing, but once the supervisor starts, it shouldn't be started again. The config should be fixed for the whole module.

robnagler (changes requested):

tests/conftest.py
mkeilman: [6b64346](https://github.com/radiasoft/sirepo/pull/4846/commits/6b64346d02c7b5877d8917e09313c94cf007366c)

tests/conftest.py
mkeilman: [6b64346](https://github.com/radiasoft/sirepo/pull/4846/commits/6b64346d02c7b5877d8917e09313c94cf007366c)

tests/conftest.py
mkeilman: [6b64346](https://github.com/radiasoft/sirepo/pull/4846/commits/6b64346d02c7b5877d8917e09313c94cf007366c)

tests/conftest.py
mkeilman: Adding the fixture scope did the trick here, but now supervisor_purge_free_sims_test.py fails on the 2nd test case (the status is "completed" instead of "job_run_purged") and I haven't figured that out yet. The same happens if I add the scope to a fresh master checkout with no other changes. I'll keep at it...

mkeilman: Update: the module-scoped _`auth_fc_module` version sees the purge loop collect and purge the json file for the 2nd test while it is still running, while the unscoped version skips it.

mkeilman: Update: the problem was the the "adjustTime" call - previously, when a new supervisor was created for every test case, the time shift was freshly applied. With a single supervisor, the time shift gets applied twice, pushing subsequent sim files outside the purge threshold.

tests/supervisor_purge_free_sims_test.py
```diff
@@ -67,16 +67,14 @@ def _status_eq(next_req, status):
...
-        params=PKDict(days=_PURGE_FREE_AFTER_DAYS + 1),
-    )
+    _adjust_time(fc)
     time.sleep(_CACHE_AND_SIM_PURGE_PERIOD + 1)
     _status_eq(next_req_free, "job_run_purged")
     _check_run_dir(should_exist=0)
     fc.sr_email_login(user_premium)
     _status_eq(next_req_premium, "completed")
     _check_run_dir(should_exist=7)
+    _adjust_time(fc, forward=False)
```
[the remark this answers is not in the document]
mkeilman: Yes, that's right. I commented to that effect but it's not in the commit message. Is it OK to make a "meaningless" commit and include the reasoning in that message?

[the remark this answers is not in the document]
robnagler: I see https://github.com/radiasoft/sirepo/pull/4846#discussion_r997667401 now. No need for a commit message. It's part of the logic now. I would change "forward" to "restore" and invert the logic, since that's what's happening. I didn't understand why it had to be moved back.

robnagler (approved):

PR #4858, opened 2022-09-19
fix raydata#24: run analysis notebook and display queued scans

rorour:
depends on https://github.com/radiasoft/raydata/pull/25


e-carlin (approved):
I committed my pr review directly to keep things moving but take a look and lmk if you have any questions.

PR #4867, opened 2022-09-21
Fix #4792 tally colorbar

### LINKED ISSUE #4792: Colorbar in visualization window
Feedback from a potential customer:
Need to have some scale information for the visualized output.

A colorbar next to the plot would be the easiest thing, next would be voxel value information upon mouseover.

I understand this mouse-over option would be more difficult for 3d tallies, so let's do a colorbar now and push the difficult thing to Phase II.

mkeilman:
Notes:
- Adds a typical color bar as used in other apps
- Points to the proper location on mouseover
- Displays the field value and coordinates in the toolbar area
- Defines constants for field units that differ from "numbers" (which have no units) and allows katex handling


moellep:
Very cool! This may be a VTK limitation, but I do find it odd when I'm mousing over a 200x1x200 tally with Y between (0, 1)cm that the Y values isn't always a constant. For example, the Utah Teapot shows values between (0 and 0.171) on the 0 side and (0.337 and 1) on the 1 side.

PR #4869, opened 2022-09-21
Fix #4811: dat vs bkp slurm

### LINKED ISSUE #4811: 20220923 use *.dat.bkp or *.dat whichever is newer
Customer writes:
> On a few occasions, I was getting no result at the end of multi-hour MPI calculation because of this at NERSC. I hope you can change it back to: “['wm_fbk', '', '1', …”?
> I note that this “backup” file is not created by copying from original *.dat file, rather, the data is saved directly to the *dat.bkp file periodically; so when showing the data in the Intensity (or Degree of Coherence) report, one would need to depict the content of either *.dat or *.dat.bkp, whatever is the most recent (previously, you were probably depicting always only the content or *.dat, even if *.dat.bkp was generated ?)

This will be tricky, because *.dat.bkp is not written atomically so it may not be complete so it's whichever file is newer and not corrupt.

`wm_fbk` (above) controls this. This is only `1` (true) [in sbatch slurm](https://github.com/radiasoft/sirepo/blob/1bfd2d5f5f17800c316e9c0d7da63997ab1c1b91/sirepo/template/srw.py#L2315). Therefore, the bkp file will not always exist.

gurhar1133:
.bkp will never be selected if does not exist
if does exist, and is more recent and has sufficient line count, then .bkp is selected

### DIFF
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -2324,8 +2324,6 @@ def _set_magnetic_measurement_parameters(run_dir, v):
 
 
 def _set_parameters(v, data, plot_reports, run_dir):
-    # TODO (gurhar1133): undo this later
-    # v.sbatchBackup = "1"
     report = data.report
     is_for_rsopt = _is_for_rsopt(report)
     dm = data.models
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -41,6 +41,7 @@ _CANVAS_MAX_SIZE = 65535
 
 _OUTPUT_FOR_MODEL = PKDict(
     coherenceXAnimation=PKDict(
+        check_backup=True,
         title="",
         filename="res_int_pr_me_dcx.dat",
         dimensions=3,
@@ -48,6 +49,7 @@ _OUTPUT_FOR_MODEL = PKDict(
         units=["m", "m", ""],
     ),
     coherenceYAnimation=PKDict(
+        check_backup=True,
         title="",
         filename="res_int_pr_me_dcy.dat",
         dimensions=3,
@@ -94,6 +96,7 @@ _OUTPUT_FOR_MODEL = PKDict(
         units=["m", "m", "m"],
     ),
     multiElectronAnimation=PKDict(
+        check_backup=True,
         title="E={photonEnergy} eV",
         filename="res_int_pr_me.dat",
         dimensions=3,
@@ -366,6 +369,10 @@ def _extract_coherent_modes(model, out_info):
     )
     return out_file
 
+def _check_backup(animation_meta_data):
+    # get most recent unless most recent has less lines
+    # this would mean it is partially written
+    pass
 
 def extract_report_data(sim_in):
     r = sim_in.report
@@ -379,6 +386,8 @@ def extract_report_data(sim_in):
         return _extract_trajectory_report(dm.trajectoryReport, out.filename)
     if r == _SIM_DATA.EXPORT_RSOPT:
         return out
+    if out.get("check_backup"):
+        out.filename = _check_backup(out)
     # TODO(pjm): remove fixup after dcx/dcy files can be read by uti_plot_com
     if r in ("coherenceXAnimation", "coherenceYAnimation"):
         _fix_file_header(out.filename)
@@ -584,6 +593,7 @@ def sim_frame(frame_args):
         # some reports may be written at the same time as the reader
         # if the file is invalid, wait a bit and try again
         for i in (1, 2, 3):
+            # TODO (gurhar1133): maybe test here
             try:
                 return extract_report_data(frame_args.sim_in)
             except Exception:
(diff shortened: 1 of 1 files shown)


sirepo/template/srw.py
```diff
@@ -367,6 +370,36 @@ def _extract_coherent_modes(model, out_info):
...
+
+def _count_lines(text):
+    return len(pkio.read_text(text).splitlines())
+
+
+def _recent_has_sufficient_lines(recent, old):
+    return _count_lines(recent) >= _count_lines(old)
+
+
+def _check_backup(animation_meta_data):
```
robnagler, line 393: sort the functions, and animation_meta_data is too long and not relevant to the function. Pass just the file name. `_check_backup` takes a `primary` (or base or original).

sirepo/template/srw.py
```diff
@@ -367,6 +370,36 @@ def _extract_coherent_modes(model, out_info):
...
+    b = animation_meta_data.filename + ".bkp"
+    if not pkio.py_path(b).check():
+        return animation_meta_data.filename
+    i = _get_recent(animation_meta_data.filename, b)
+    if _recent_has_sufficient_lines(i.recent, i.old):
```
robnagler, line 398: nice check

sirepo/template/srw.py
```diff
@@ -367,6 +370,36 @@ def _extract_coherent_modes(model, out_info):
...
+            recent=backup_data_filename,
+            old=data_filename,
+        )
+    return PKDict(
+        recent=data_filename,
+        old=backup_data_filename,
+    )
+
+
+def _count_lines(text):
```
robnagler, line 385: It will be more efficient (and sufficient) to check for the size of the files. py.path has a size check so should be easy.

sirepo/template/srw.py
```diff
@@ -41,13 +41,15 @@
 
 _OUTPUT_FOR_MODEL = PKDict(
     coherenceXAnimation=PKDict(
+        check_backup=True,
```
robnagler, line 44: This will require a db upgrade. I think it is better if we deal with this programmatically inside srw.py, that is, in the code below (which already makes assumptions about the report name anyway) it adds code based on the report name.

sirepo/template/srw.py
```diff
@@ -367,6 +370,36 @@ def _extract_coherent_modes(model, out_info):
     return out_file
 
 
+def _get_recent(data_filename, backup_data_filename):
+    if os.path.getmtime(backup_data_filename) > os.path.getmtime(data_filename):
```
robnagler, line 374: These names are long. You can call them "secondary" and "primary". To make the code simpler, simply check the size, not the time, since we know the file continues to grow, and the count lines check overrides the date check.

sirepo/template/srw.py
```diff
@@ -379,6 +412,8 @@ def extract_report_data(sim_in):
         return _extract_trajectory_report(dm.trajectoryReport, out.filename)
     if r == _SIM_DATA.EXPORT_RSOPT:
         return out
+    if out.get("check_backup"):
+        out.filename = _check_backup(out)
```
robnagler, line 416: Use `_best_data_file(out.filename)`. This is not really about check_backup, but about picking between a primary and second file.

sirepo/template/srw.py
```diff
@@ -367,6 +370,36 @@ def _extract_coherent_modes(model, out_info):
...
+def _count_lines(text):
+    return len(pkio.read_text(text).splitlines())
+
+
+def _recent_has_sufficient_lines(recent, old):
+    return _count_lines(recent) >= _count_lines(old)
+
+
+def _check_backup(animation_meta_data):
+    b = animation_meta_data.filename + ".bkp"
```
robnagler, line 394: To simplify, have another nested function that checks existence and lines (both files may not exist) so the code looks like:

```py
def _best_data_file(primary):
    def _size(path):
        return path.size if path.exists() else 0

    p = pykern.pkio.py_path(primary)
    s = p.new(ext='dat.bkp')
    if _size(s) > _size(primary):
        p = s
    return p.basename
```

sirepo/template/srw.py
```diff
@@ -379,6 +412,8 @@ def extract_report_data(sim_in):
         return _extract_trajectory_report(dm.trajectoryReport, out.filename)
     if r == _SIM_DATA.EXPORT_RSOPT:
         return out
+    if out.get("check_backup"):
```
robnagler, line 415: See discussion above `if r in ("coherenceXAnimation", "coherenceYAnimation", "coherentModesAnimation")`

robnagler (changes requested):

sirepo/template/srw.py
moellep: You need to check time because you want the most recent. Current and backup files will contain the same number of points but may be of different file sizes due to number to string in the data file.

robnagler: Thanks.

gurhar1133: Ok, so check time and use line count rather than file size?

robnagler: I updated the [_best_data_file code](https://github.com/radiasoft/sirepo/pull/4869#discussion_r977871466).

robnagler (approved):

PR #4871, opened 2022-09-22
Fix #4596: coherent modes on nersc

### LINKED ISSUE #4596: 20221031 Cross spectral density doesn't run on NERSC
Selecting NERSC from the execution mode for the SRW Coherent Modes report (on the Source tab) runs locally and does not prompt for NERSC credentials.

gurhar1133:
Now runs on sbatch

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2453,7 +2453,7 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
             model: '@simulationStatusPanel',
             title: '@',
         },
-        template: `******
+        template: `
            <div data-ng-if="(simState.getFrameCount() > 0) || errorMessage()" class="well well-lg">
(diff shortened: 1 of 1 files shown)


sirepo/package_data/static/js/srw.js
```diff
@@ -2656,8 +2656,8 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                     j.jobRunMode = 'parallel';
                 }
                 frameCache.setFrameCount(0, $scope.model);
+                appState.saveQuietly($scope.simState.model);
```
robnagler, line 2659: Why does this work, and how does it affect other models. What does the quite mode modification do?

moellep: Gurhar and I discussed this. The bug was the changes to the simState.model were not being saved except for the multiElectronAnimation case. The saveQuietly saves the data with the expectation that another save (or autosave) will push it to the server. There is already a saveAndRun() call in this method, so no need to call saveChanges() followed by saveAndRunSimulation() because it will post to the server twice.

PR #4878, opened 2022-09-24
raydata: replay panel working, add fixup

rorour:


sirepo/package_data/static/js/raydata.js
```diff
@@ -218,8 +218,34 @@ SIREPO.app.directive('replayPanel', function() {
             modelName: '=',
         },
         template: `
-          <button type="submit" class="btn btn-primary" data-ng-click="">Start Replay</button>
+          <button type="submit" class="btn btn-primary" data-ng-click="startReplay()">Start Replay</button>
```
e-carlin, line 221: This button should be not visible until the user has supplied the catalogs and numScans. It should also be grayed out while the replay is running (maybe add some spinner indicating things are running too).

sirepo/package_data/static/js/raydata.js
```diff
@@ -218,8 +218,34 @@ SIREPO.app.directive('replayPanel', function() {
...
         `,
+        controller: function(appState, errorService, panelState, raydataService, requestSender, $scope) {
+            $scope.startReplay = () => {
+                if (! (appState.models.replay.sourceCatalogName && appState.models.replay.destinationCatalogName && appState.models.replay.numScans)) {
```
e-carlin, line 225: This can be removed once we don't show the button until all fields are valid.

sirepo/template/raydata.py
```diff
@@ -90,6 +90,10 @@ def stateless_compute_catalog_names(data):
     )
 
 
+def stateless_compute_replay_scans(data):
+    return PKDict(data=_request_scan_monitor(PKDict(method="begin_replay", data=data)))
```
e-carlin, line 94: Make this method name "begin_replay" the same as the stateless compute method "replay_scans" so it is easier to follow the code. I don't have a preference for which name is chosen.

e-carlin (changes requested):
Few comments

sirepo/package_data/static/js/raydata.js
rorour: f69cb0460a842c98b329b6e5cbe841843fbbb762

sirepo/template/raydata.py
rorour: f69cb0460a842c98b329b6e5cbe841843fbbb762

sirepo/package_data/static/js/raydata.js
rorour: f69cb0460a842c98b329b6e5cbe841843fbbb762

sirepo/package_data/static/js/raydata.js
```diff
@@ -197,6 +197,29 @@ SIREPO.app.directive('dateTimePicker', function() {
     };
 });
 
+SIREPO.app.directive('dotsAnimation', function() {
```
rorour, line 200: The existing uses of dots used [persistentSimulation](https://github.com/radiasoft/sirepo/blob/master/sirepo/package_data/static/js/sirepo.js#L2755) and relied on simulation status and other things that I don't think apply to this case, so I made a directive.

e-carlin: Looks good!

e-carlin (approved):
perfect

PR #4885, opened 2022-09-26
fix #4832: srw points to /en/xray-beamlines.html in dev

### LINKED ISSUE #4832: 20220930 Synchrotron Radiation Workshop points to /en/xray-beamlines.html in dev
While it's not a real problem, it does cause confusing and will cause confusion in local downloads, because that file doesn't exist in dev.

rorour:


sirepo/package_data/static/js/light.js
```diff
@@ -253,7 +253,7 @@ app.directive('pageHeading', function() {
         template: `
             <nav class="navbar navbar-default navbar-static-top">
               <div class="container-fluid">
-                <a class="navbar-brand" href="/en/xray-beamlines.html">
+                <a class="navbar-brand" href="">
```
rorour, line 256: @robnagler  When is this code used? I don't see code from light.js executing when I visit SRW.

rorour:
@robnagler When I was figuring out this issue I noticed that elegant points to "/en/particle-accelerators.html" which doesn't seem to exist either. Should that also be removed?

sirepo/package_data/static/js/light.js
robnagler: light.js is used at BNL. does href="" do the "right" thing?

sirepo/feature_config.py
```diff
@@ -181,7 +181,6 @@ def b(msg, dev=False):
             "Link to Sirepo Slack workspace; uid will be appended",
         ),
         srw=dict(
-            app_url=("/en/xray-beamlines.html", str, "URL for SRW link"),
```
robnagler, line 184: The "app_url" defaults as /srw  in dev, but in prod it's /en/xray...

sirepo/package_data/static/js/srw.js
```diff
@@ -1595,18 +1595,14 @@ SIREPO.app.directive('appHeader', function(appState, panelState, srwService) {
               ${rightNav}
             </div>
             <div data-ng-if="srwService.isApplicationMode(\'default\')">
-              <div data-app-header-brand="" data-app-url="{{ ::appURL() }}"></div>
```
robnagler, line 1598: This needs to stay

sirepo/package_data/static/js/srw.js
```diff
@@ -1595,18 +1595,14 @@ SIREPO.app.directive('appHeader', function(appState, panelState, srwService) {
...
+              <div data-app-header-brand=""></div>
               <div class="navbar-left" data-app-header-left="nav"></div>
               ${rightNav}
             </div>
         `,
         controller: function($scope) {
             $scope.srwService = srwService;
 
-            $scope.appURL = function() {
```
robnagler, line 1606: Stays.

robnagler (changes requested):

sirepo/feature_config.py
rorour: Should I just change app_url to "/srw"?

sirepo/package_data/static/js/light.js
rorour: 9847975a7cb50761b59da0f3bbff9d40064e7526

sirepo/feature_config.py
rorour: df44df69b0cf6074618593a4c872ea9c2213cc9a

sirepo/package_data/static/js/srw.js
rorour: df44df69b0cf6074618593a4c872ea9c2213cc9a

sirepo/package_data/static/js/light.js
```diff
@@ -253,9 +253,9 @@ app.directive('pageHeading', function() {
         template: `
             <nav class="navbar navbar-default navbar-static-top">
               <div class="container-fluid">
-                <a class="navbar-brand" href="/en/xray-beamlines.html">
```
robnagler, line 256: This should stay for production. It can go to app_url for dev.

sirepo/feature_config.py
```diff
@@ -181,7 +181,7 @@ def b(msg, dev=False):
...
+            app_url=("/srw", str, "URL for SRW link"),
```
robnagler, line 184: This needs to be conditional on channel_in("dev").

robnagler (changes requested):

robnagler:
This is a larger issue with all apps, and needs to be addressed more generally in React.

PR #4888, opened 2022-09-27
Fix #4820 radia export script error

### LINKED ISSUE #4820: 20220930 Radia export script produces error when run
Setting up a simple Radia simulation with a cube (https://www.sirepo.com/radia#/visualization/Nggh4m4O) and then exporting it to Python produces an error when the Python script is run:

```
  File "export_test.py", line 596, in <module>
    o_id = _add_object({'_super': 'geomObject', 'bevels': [], 'center': '1.2970168612191981,11.251621271076525,0', 'color': '#cccccc', 'fillets': [], 'groupId': '', 'id': 'J6WN5PpYnsB15kmel0iouqyvguEbRA0V', 'layoutShape': 'rect', 'magnetization': '0.0, 0.0, 0.0', 'material': 'Xc06', 'materialFile': '', 'name': 'cuboid 1', 'remanentMag': 0.5, 'segments': '1, 1, 1', 'shapeButton': {}, 'size': '10.0, 10.0, 10.0', 'transforms': [], 'type': 'cuboid', 'super_classes': ['_', 'model', 'geomObject', 'radiaObject', 'layoutObject'], 'h_m_curve': None})
  File "export_test.py", line 555, in _add_object
    if o['model'] == _MODEL_BOX:
KeyError: 'model'
```

I think I should have the latest version of sirepo installed in my environment. I also saw this when I tried to export an existing dipole simulation I had.

mkeilman:
This brings the python export up to date with the sirepo code. The two should be consolidated so they do not get out of sync later, but that is beyond the scope of this issue. The exported code creates the geometry, solves the relaxation, and dumps a file. It also calculates various fields, with the understanding that the user can modify the script to do something with them (plots etc.)

Parallel is not supported but will be added in a new issue.


mkeilman:
Select "Python Source" from the gear menu and run `python <exported file>` to test the fix

PR #4898, opened 2022-09-29
Fix #4850: analysis jobs now have run_dir arg

### LINKED ISSUE #4850: 20220930 template_common.analysis_job_dispatch should include run_dir
So we can remove code like https://github.com/radiasoft/sirepo/pull/4842/files#diff-24496517cc6cc5369fc69a13a07158c876115ef9eec8ceb098a25b374b083f7fR1462

similar methods like sim_frame_dispatch do this.

gurhar1133:
Calls to simulation_db.simulation_run_dir have been removed and now analysis jobs take a run_dir arg.
Analysis jobs now take **kwargs.

### DIFF
--- a/sirepo/template/template_common.py
+++ b/sirepo/template/template_common.py
@@ -9,6 +9,7 @@ from pykern import pkio
 from pykern import pkjinja
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdc, pkdlog, pkdp, pkdexc
+from sirepo import simulation_db
 from sirepo.template import code_variable
 import math
 import os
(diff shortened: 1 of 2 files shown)


sirepo/template/template_common.py
```diff
@@ -9,6 +9,7 @@
 from pykern import pkjinja
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdc, pkdlog, pkdp, pkdexc
+from sirepo import simulation_db
```
robnagler, line 12: This import should be localized in analysis_job_dispatch. There are circular import issues here.

sirepo/template/template_common.py
```diff
@@ -243,10 +244,17 @@ def __set_from_pc(cls, particle, energy):
 
 def analysis_job_dispatch(data):
     t = sirepo.template.import_module(data.simulationType)
+    d = simulation_db.simulation_run_dir(
```
robnagler, line 247: Passing data should work. parse_model looks for computeModel so you don't need this extra pkdict.

sirepo/template/template_common.py
```diff
@@ -243,10 +244,17 @@ def __set_from_pc(cls, particle, energy):
...
+            simulationType=data.simulationType,
+            simulationId=data.simulationId,
+            report=data.computeModel,
+        )
+    )
     return getattr(
         t,
         f"analysis_job_{_validate_method(t, data)}",
-    )(data)
+    )(data, d)
```
robnagler, line 257: temporary variable is unnecessary since it is only used once.

robnagler (changes requested):

gurhar1133:
@robnagler should be fixed now

robnagler (approved):

PR #4933, opened 2022-10-12
Fix #4932: move all databroker calls out of sirepo.

### LINKED ISSUE #4932: raydata: remove all databroker calls
All databroker calls should live in the scan monitor. This means we only need to supply sirepo with configuration for interacting with the scan monitor instead of config for the scan monitor and for interacting with databroker (mongo)

e-carlin:
They were moved into the raydata scan_monitor. There will be a companion PR there.

### DIFF
--- a/sirepo/pkcli/raydata.py
+++ b/sirepo/pkcli/raydata.py
@@ -9,35 +9,6 @@ from pykern.pkdebug import pkdp, pkdlog
 from sirepo.template import template_common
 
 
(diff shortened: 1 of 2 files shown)


e-carlin:
@gurhar1133 since @rorour is out would you mind reviewing? I don't think it is worth the hassle to get your machine setup so you can run the code. But, please just look over it. There is also a companion raydata pr coming shortly.

gurhar1133 (approved):
I dont know too much about how this works, but it looks good to me

PR #4938, opened 2022-10-13
Fix #4922: cifar image samples

### LINKED ISSUE #4922: Activait CIFAR example Data Source page image examples
* show 10 or so random data sample image/labels on the Data Source page

gurhar1133:
Now the data tab has a image sample viewer with 5 sets of 25 images to scroll through

### DIFF
--- a/sirepo/package_data/static/js/activait.js
+++ b/sirepo/package_data/static/js/activait.js
@@ -1120,6 +1120,19 @@ SIREPO.app.directive('columnSelector', function(appState, mlService, panelState,
     };
 });
 
+
+SIREPO.app.directive('imagePreviewPanel', function(appState) {
+    return {
+        restrict: 'A',
+        scope: {},
+        template: `hello world`,
+        controller: function($scope, appState) {
+
+        }
+    }
+})
+
+
 SIREPO.app.directive('dataPathSelector', function(appState, mlService, panelState, utilities) {
     return {
         restrict: 'A',
--- a/sirepo/package_data/static/js/activait.js
+++ b/sirepo/package_data/static/js/activait.js
@@ -1128,13 +1128,21 @@ SIREPO.app.directive('imagePreviewPanel', function(appState, requestSender) {
         template: `
         <div>
           <img class="img-responsive srw-processed-image" />
-          <button data-ng-click="prev()"> prev </button>
-          <button data-ng-click="next()"> next </button>
+          <div data-ng-if="isLoading()" class="progress">
+            <div class="progress-bar progress-bar-striped active" role="progressbar" aria-valuenow="{{ simState.getPercentComplete() }}" aria-valuemin="0" aria-valuemax="100" data-ng-attr-style="width: {{ simState.getPercentComplete() || 100 }}%"></div>
+          </div>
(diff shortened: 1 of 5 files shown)


sirepo/template/activait.py
```diff
@@ -299,6 +299,35 @@ def stateful_compute_compute_column_info(data):
...
+    from pykern.pkcompat import from_bytes
+
+    def _data_url(filename):
+        f = open(filename, "rb")
+        u = "data:image/jpeg;base64," + from_bytes(b64encode(f.read()))
+        f.close()
+        return u
+
+    with h5py.File(_filepath("CIFAR-4.h5"), "r") as f:
+        x = f["images"]
```
moellep, line 314: Looks like a hard-coded datafile - this should probably come from a call to lb_file_name_with_model_field() similar to other calls in template/activait.py

sirepo/package_data/static/js/activait.js
```diff
@@ -1120,6 +1120,74 @@ SIREPO.app.directive('columnSelector', function(appState, mlService, panelState,
...
+                        }
+                    }
+                );
+            };
+
+            loadImageFile();
+        }
+    };
+});
+
```
moellep, line 1189: Creating a new ML simulation in a fresh sirepo install is causing this error:
angular.js:15567 TypeError: Cannot read properties of undefined (reading '0')
    at setImageFromUriIndex (activait.js?20221017.155522:1164:60)

I think the imagePreviewPanel needs to be created only if the Data Source is images and an appropriate file has been added to the simulation.

sirepo/template/activait.py
mkeilman: I'll add that the `DataReaderFactory` is designed to simplify getting the data from archive files so you can avoid opening them directly - I think it can be used here.

sirepo/template/activait.py
```diff
@@ -299,6 +299,38 @@ def stateful_compute_compute_column_info(data):
...
+        y = f["metadata/image_types"]
+        u = []
+        for i in range(0, 125, 25):
+            plt.figure(figsize=[10, 10])
+            for j in range(25):
+                plt.subplot(5, 5, j + 1)
+                plt.xticks([])
+                plt.yticks([])
+                plt.imshow(x[i + j])
+                plt.xlabel(str(f["metadata/labels"][y[i + j]]).replace("b'", "'"))
```
mkeilman, line 324: `pkcompat.from_bytes()` will take care of those pesky "b'"s

sirepo/template/activait.py
```diff
@@ -299,6 +299,38 @@ def stateful_compute_compute_column_info(data):
     return _compute_column_info(data.args.dataFile)
 
 
+def stateful_compute_sample_images(data):
+    import matplotlib.pyplot as plt
+    from base64 import b64encode
+    from pykern.pkcompat import from_bytes
```
mkeilman, line 305: Use `from pykern import pkcompat` and then `pkcompat.from_bytes()`. You might also want to do the import at the top of the file, but I'm not sure of the protocol there.

mkeilman (commented):
Just a few style things I noticed

gurhar1133:
addressed your comments @mkeilman

PR #4944, opened 2022-10-17
Fix #4927 react heatplot

### LINKED ISSUE #4927: 20221014 react heatmap plot
Implement a heatmap plot in react.

moellep:
fix #4927 added react heatplot
supports axes, zoom and pan
still needs a colormap and colorbar


react/src/component/util.js
```diff
@@ -0,0 +1,81 @@
...
+export function graphMetrics(dim) {
+    return {
+        graphHeight: dim.height - heightAdjustment(),
+        graphWidth: dim.width - widthAdjustment(),
+        graphX: yAxisSize + margin,
+        graphY: margin,
+    };
+}
+
+export function useRefSize(ref, aspectRatio) {
```
garsuga, line 54:
```suggestion
export function useRefSize(ref) {
```

Unless it is not possible to do this later, this function should not do aspect ratio scaling and other size adjustments. It is called 'useRefSize' so it should only return the size of a ref. This is more reusable and the other scaling should be made into other functions

react/src/component/util.js
```diff
@@ -0,0 +1,81 @@
...
+    useLayoutEffect(() => {
+        if (! ref || ! ref.current) {
+            return;
+        }
+        const handleResize = debounce(() => {
+            if (! ref || ! ref.current) {
+                return;
+            }
+            const w = Number.parseInt(ref.current.offsetWidth);
+            const h = heightAdjustment() + Number.parseInt((w - widthAdjustment()) * aspectRatio);
```
garsuga, line 68: This function should not use the width or height adjustment. We need another function that can do this separately.

react/src/component/util.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export function graphMetrics(dim) {
+    return {
+        graphHeight: dim.height - heightAdjustment(),
+        graphWidth: dim.width - widthAdjustment(),
+        graphX: yAxisSize + margin,
+        graphY: margin,
+    };
+}
```
garsuga, line 52:
```suggestion
export function graphContentBounds(dim) {
    return {
        height: dim.height - heightAdjustment(),
        width: dim.width - widthAdjustment(),
        offsetX: yAxisSize + margin,
        offsetY: margin,
    };
}
```

Let's use the word 'content' to differentiate between the size including the margin and axes and the size without those. I think we should avoid prefixes like 'graph' when we name fields within objects. This will help us unify types later. For instance, this is describing a 'rect' and anywhere else in Sirepo where we describe rectangles will likely have the same values.

robnagler:
Very nice comments @garsuga!

react/src/component/util.js
moellep: I've made useRefSize() reusable - it now only returns the dimensions of the component ref. I added useGraphContentBounds() which uses useRefSize() and sets the component height based on the graph's aspect ratio, and returns the graph position info.

I also updated the method names to match your suggestions, but take a new look at utility/component.js to make sure it looks OK.

garsuga:
I may need to merge this manually. I renamed and moved the schema as part of another issue and this branch makes changes to the schema that no longer exists.

garsuga (approved):

PR #4975, opened 2022-10-19
Fix #4973 - removed default fieldPath, do not run if no paths

### LINKED ISSUE #4973: New radia simulation creates empty field path causing fieldLineoutReport to fail
The suspect path looks like `{'id': '1jfR3peOnf64oc7R', 'name': ''}`, which has no counterpart in the model. Two issues: first, this object should not be created. Second, fieldLineoutReport should not run at all if no (real) field paths have been created.

mkeilman:
The default value for a fieldPath model is an id and an empty name, with no data besides, which causes problems if the app tries to use it. This removes that from the schema, and also prevents fieldLineoutReport from running if there is no path defined.

### DIFF
--- a/sirepo/package_data/static/js/radia.js
+++ b/sirepo/package_data/static/js/radia.js
@@ -1840,7 +1840,7 @@ SIREPO.app.directive('fieldLineoutAnimation', function(appState, persistentSimul
                 $scope.dataCleared = false;
             });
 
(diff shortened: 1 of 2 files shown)


mkeilman:
Latest commit was to run the simulation after the data is saved, not just changed. Otherwise the "frame data" gets out of sync with the actual state (with the result here that the plotAxis is incorrect)

PR #4976, opened 2022-10-19
Fix #4974: move Raydata examples into Sirepo

### LINKED ISSUE #4974: raydata: move examples into sirepo
If the raydata examples live in Sirepo then we don't need to do anything special (ex have a special raydata sirepo docker image) to get sirepo to work with raydata enabled. They currently live in raydata but the examples themselves don't contain any proprietary information.

e-carlin:
They previously lived in the private raydata repo. By moving them into Sirepo then Sirepo can create the raydata examples w/o being taught about a special place to look for them (ex package_path).

### DIFF
--- a/sirepo/package_data/template/raydata/examples/chx.json
+++ b/sirepo/package_data/template/raydata/examples/chx.json
@@ -10,7 +10,7 @@
         "simulation": {
             "folder": "/Examples",
             "isExample": true,
-            "name": "XPCS SAXS Analysis"
+            "name": "CHX Analysis"
         },
         "catalog": {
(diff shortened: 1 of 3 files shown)


rorour:
Do we still need two examples?

e-carlin:
I think so. One for chx and one for csx. But that reminded me to change the name of chx so it matches better with everything https://github.com/radiasoft/sirepo/pull/4976/commits/4cf3d0b961f7fe9bcadb65b36b0f7188719bacb4

rorour (approved):
Looks good to me

PR #4977, opened 2022-10-20
fix radiasoft/raydata#30: bug fixes

rorour:
I've tried and haven't been able to recreate the super long loading time that happened during the demo. The longest loading time I'm seeing right now is about 30 seconds.

see comment on radiasoft/raydata#30

### DIFF
--- a/sirepo/package_data/static/js/raydata.js
+++ b/sirepo/package_data/static/js/raydata.js
@@ -107,15 +107,18 @@ SIREPO.app.directive('catalogPicker', function() {
             field: '=',
         },
         template: `
-            <select class="form-control" data-ng-model="model[field]" data-ng-options="name as name for name in catalogNames"></select>
+            <div data-ng-hide="!awaitingCatalogNames">Loading...</div>
+            <select class="form-control" data-ng-hide="awaitingCatalogNames" data-ng-model="model[field]" data-ng-options="name as name for name in catalogNames"></select>
         `,
         controller: function($scope, appState, errorService, requestSender) {
             $scope.catalogNames = [];
+            $scope.awaitingCatalogNames = true;
 
             requestSender.sendStatelessCompute(
                 appState,
                 json => {
                     $scope.catalogNames = json.data.catalogs;
+                    $scope.awaitingCatalogNames = false;
                 },
                 {
                     method: 'catalog_names',
@@ -316,6 +319,7 @@ SIREPO.app.directive('scansTable', function() {
                     </tr>
                   </tbody>
                 </table>
+                <div ng-if="awaitingScans">Loading scans...</div>
               </div>
             </div>
             <div data-column-picker="" data-title="Add Column" data-id="sr-columnPicker-editor" data-available-columns="availableColumns" data-save-column-changes="saveColumnChanges"></div>
@@ -356,6 +360,7 @@ SIREPO.app.directive('scansTable', function() {
         `,
         controller: function(appState, errorService, panelState, raydataService, requestSender, $scope, $interval) {
             $scope.availableColumns = [];
+            $scope.awaitingScans = false;
             // POSIT: status + sirepo.template.raydata._DEFAULT_COLUMNS
             $scope.defaultColumns = ['status', 'start', 'stop', 'suid'];
             $scope.images = null;
@@ -407,23 +412,25 @@ SIREPO.app.directive('scansTable', function() {
                         return;
                 }
                 function doRequest() {
-                        requestSender.sendStatelessCompute(
-                            appState,
-                            (json) => {
-                                $scope.scans = json.data.scans.slice();
-                            },
-                            {
-                                method: 'scans',
-                                args: {
-                                    analysisStatus: $scope.analysisStatus,
-                                    catalogName: appState.models.catalog.catalogName,
-                                    searchStartTime: m.searchStartTime,
-                                    searchStopTime: m.searchStopTime,
-                                    selectedColumns: appState.models.metadataColumns.selected,
-                                }
-                            },
-                            errorOptions
-                        );
+                    $scope.awaitingScans = true;
+                    requestSender.sendStatelessCompute(
+                        appState,
+                        (json) => {
+                            $scope.awaitingScans = false;
+                            $scope.scans = json.data.scans.slice();
+                        },
+                        {
+                            method: 'scans',
+                            args: {
+                                analysisStatus: $scope.analysisStatus,
+                                catalogName: appState.models.catalog.catalogName,
+                                searchStartTime: m.searchStartTime,
+                                searchStopTime: m.searchStopTime,
+                                selectedColumns: appState.models.metadataColumns.selected,
+                            }
+                        },
+                        errorOptions
+                    );
                 }
                 if (scanRequestInterval) {
                     $interval.cancel(scanRequestInterval);
--- a/sirepo/package_data/static/js/raydata.js
+++ b/sirepo/package_data/static/js/raydata.js
@@ -107,18 +107,18 @@ SIREPO.app.directive('catalogPicker', function() {
             field: '=',
         },
         template: `
-            <div data-ng-hide="!awaitingCatalogNames">Loading...</div>
-            <select class="form-control" data-ng-hide="awaitingCatalogNames" data-ng-model="model[field]" data-ng-options="name as name for name in catalogNames"></select>
+            <div data-ng-hide="haveCatalogNames">Loading...</div>
+            <select class="form-control" data-ng-show="haveCatalogNames" data-ng-model="model[field]" data-ng-options="name as name for name in catalogNames"></select>
         `,
         controller: function($scope, appState, errorService, requestSender) {
             $scope.catalogNames = [];
-            $scope.awaitingCatalogNames = true;
+            $scope.haveCatalogNames = false;
 
             requestSender.sendStatelessCompute(
                 appState,
                 json => {
                     $scope.catalogNames = json.data.catalogs;
-                    $scope.awaitingCatalogNames = false;
+                    $scope.haveCatalogNames = true;
                 },
                 {
                     method: 'catalog_names',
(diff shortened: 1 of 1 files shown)


e-carlin:
In what situation are you seeing a 30 second loading time? That seems abnormally long.

sirepo/package_data/static/js/raydata.js
```diff
@@ -107,15 +107,18 @@ SIREPO.app.directive('catalogPicker', function() {
             field: '=',
         },
         template: `
-            <select class="form-control" data-ng-model="model[field]" data-ng-options="name as name for name in catalogNames"></select>
+            <div data-ng-hide="!awaitingCatalogNames">Loading...</div>
```
e-carlin, line 110: Make the variable something like "haveCatalogNames". Then you can get rid of the `!`. `!` can be confusing (I fall in the group of people it confuses) and should be avoided when possible. It is especially confusing in this case where true means we do want to hide something (I ascribe hide to a "negative"). Hard for me to keep it all straight... 

I can't find a good quote but Steve McConnell in his frequently praised book Code Complete was the first person I saw mention this. Here is a blog that sums it up https://schneide.blog/2014/08/03/dont-ever-not-avoid-negative-logic/

sirepo/package_data/static/js/raydata.js
```diff
@@ -316,6 +319,7 @@ SIREPO.app.directive('scansTable', function() {
                     </tr>
                   </tbody>
                 </table>
+                <div ng-if="awaitingScans">Loading scans...</div>
```
e-carlin, line 322: I don't love this popping up and going away. What about a spinner that is just running as long as the we have the interval running?

e-carlin (commented):
Couple comments.

One bug I think that should be fixed is if you get an error response from the request for scans the interval just keeps on running. This causes the same error to happen repeatedly. What about canceling the interval if we get an error response back?

rorour:
> In what situation are you seeing a 30 second loading time? That seems abnormally long.

It is abnormal, but I was trying to stress it to recreate the loading issue.

Replay 10 scans chx -> chx, click on queued tab while replay still in progress. 22 seconds to load
Then go to Executed tab, change stop time while Catalog Name still loading. 41 seconds to load

sirepo/package_data/static/js/raydata.js
rorour: I found this comment a little confusing. The line after this used `awaitingCatalogNames` without the `!`, so renaming the variable wouldn't get rid of the need for `!`. But I changed one of the `data-ng-hide` to `data-ng-show` to get rid of the `!`. Is that what you had in mind?

9834229108193197e2be278d92997ffbf798781f

sirepo/package_data/static/js/raydata.js
rorour: Do you mean regardless of whether it's actively awaiting a request for scans? I think there will be no way to tell whether there are no scans in the result or if it's just loading indefinitely.

Do you want to use the animation from loadingSpinner directive? Its style has its position set as absolute, so I will need to modify it or make a new directive if that's the one you would like to use.
Though to be honest I think the text looks cleaner and makes it clear what it's loading.

sirepo/package_data/static/js/raydata.js
e-carlin: Agh I'm sorry I totally misread the code! I didn't see the use below. I think using ng-hide and ng-show is actually more confusing. Better to use the same one and have the variable negated in one case. I don't have a preference on haveCatalogNames vs awaitingCatalogNames. Sorry to make you do this unnecessary work.

sirepo/package_data/static/js/raydata.js
e-carlin: > Do you mean regardless of whether it's actively awaiting a request for scans?

Yes. In my eyes (please push back if you disagree) the user doesn't care that a request is happening _right now_. What we're trying to let the user now is that when there are new scans to display we will show them. The GUI is "working behind the scenes" and they don't need to refresh the page to see something new. We happen to be searching on an interval every few seconds. But, when we switch to websockets maybe we won't poll. Updates will be pushed to the GUI whenever there are new scans. So, showing "loading scans..." isn't so important. What is important is that the user knows when there are new scans they will be displayed.  That process of "searching for new scans", as far as the user cares, is happening constantly. 

This may be worth opening a discuss issue because I could be in the minority but I also find things that show/hide on an interval to be distracting. For example, whenever someone has a GIF in a presentation slide I find it distracting that the GIF is just on a loop over and over while they talk.

e-carlin:
Hmm, that seems like a long time for the requests to happen. Can you take a look at the code and see where the slowdown is? You can do this by putting prints with timestamps throughout critical sections. I want to make sure that we don't have a bug in our code. If it is just waiting on databroker then that is fine. We aren't going to improve that speed (right now). But, I may have written a bug in the raydata_scan_monitor where the cooperative multi-tasking isn't very cooperative and one task is waiting on some blocking operation preventing other tasks from executing.

e-carlin:

rorour:
I'm still trying to figure out the long loading times - it seems like the requests are not really being handled asynchronously. A request doesn't enter `_incoming` until the previous request has been responded to. Any ideas? 

I have a fix for the modal issue, I'll push what I have.

sirepo/package_data/static/js/raydata.js
rorour: [da3290d](https://github.com/radiasoft/sirepo/pull/4977/commits/da3290dd1c15209bcdbe06064ce9ba0668b5b172)

rorour:
There seemed to be 2 reasons for the modal issue. Firstly, `$scope.showAnalysisOutputModal()` was being called from inside a `$scope.$watch` on `'selectedScan'`, but we weren't setting `selectedScan` to null after closing the modal, so the $watch wouldn't be triggered if the same scan was clicked again. I added `$scope.setSelectedScan(null);` inside `el.on('hidden.bs.modal'`, but that still didn't fix the problem. 

I noticed that there was sometimes a significant delay between `el.on('hidden.bs.modal'` and `$scope.$watch('selectedScan'` after the modal was actually closed. Since `$scope.showAnalysisOutputModal()` was inside `$scope.$watch('selectedScan'`, I guess the modal could only be shown once the digest cycle caught up with the new value of `selectedScan`.

Since `selectedScan` is only changed through `$scope.setSelectedScan`, I just moved `$scope.showAnalysisOutputModal();` into there so we don't need to wait on the digest.

Thanks for encouraging me to look deeper into that one.

e-carlin:
> A request doesn't enter _incoming until the previous request has been responded to

There are 2 things that could be going on here:

1. Flask is running single threaded (in dev). So it is only able to handle one request at a time and send one request to the scan_monitor at a time
2. The scan_monitor is blocking. No where in the databroker code is there any asynchrony. I just made an issue to solve this https://github.com/radiasoft/raydata/issues/37 . 

Thanks for looking into it.

> Since selectedScan is only changed through $scope.setSelectedScan, I just moved $scope.showAnalysisOutputModal(); into there so we don't need to wait on the digest.

Nice! That's a great find.

I'm going to test and assuming all works well merge this PR. We can fix https://github.com/radiasoft/raydata/issues/37 in a separate PR.

sirepo/package_data/static/js/raydata.js
rorour: There are 2 pieces of information here: whether or not we are currently awaiting a request, and the fact that new scans matching the query will be displayed automatically.

Without the first piece, there is no way for a user to know for certain if there are no scans matching their query or if they're still awaiting a response. I think that's a frustrating experience for a user, and that's the benefit of the current "Loading...". An alternative solution would be to display something when the scan monitor returns an empty list, like "No scans to display". I still like the the loading indicator better.

The second piece is also useful information. I don't think a spinner is the best way to convey it though. 

I thought of a potential way to combine them though. Instead of popping up and going away, what if we had something that alternated between "Loading" and "Last updated at \<timestamp\>"? Here's a mockup I made with figma - text could change, just wanted to show the general idea. I'm not married to this but I think it's a possible solution.
gifmock

rorour:
Sounds good. I still wanted to respond to the loading spinner discussion - left a comment above.

sirepo/package_data/static/js/raydata.js
e-carlin: You bring up some very good points. Thanks for creating that mockup. @moellep would you mind weighing in on the design?

To get you up to speed here is the problem we are working on: We have a table that we fill with data about analyses. Each row is one analysis and the columns are metadata about it (status, start time, stop time, etc). The query to get the data to populate the table takes long enough that we need some indication that it is happening. Likewise, we are polling on an interval for new data so the user doesn't have to refresh. On response there may be no data returned or there may be data. Do you have any thoughts on how to indicate all of this information to the user?

rorour: @moellep not sure if you saw this

moellep: Thanks - I'll take a look

moellep: I agree that a spinner is distracting and plain text is fine. It is a bit confusing that it says "Loading scans..." or "Fetching scans..." but then does nothing when there are no scans, almost like the user is expecting something to get loaded.

I'm also not a fan of the text appearing and disappearing because it causes the panel border to constantly be changing height. This could be fixed by putting a place-holder black space in when the "loading scans.." is not shown. Something to keep the border from flipping back and forth.

Just a thought - how about show "Checking for new scans..." where the "..." changes every second, . . ...
And when it finds a new one, it could change to "Fetching scan" for a second, or until the new scan is available, and then immediately change back to "Checking for new scans..."?

e-carlin: @rorour is there an issue to track this work?

rorour: @e-carlin yes, https://github.com/radiasoft/raydata/issues/38

e-carlin: Thanks!

Issue #4049, opened 2022-01-18
allow multiple execution of single-core background tasks


moellep:
When the SRW beamline page was changed from a series of individual reports into one background task which produces multiple plots, it had the side-effect of blocking if any other background (heavy, multi-core tasks) were already running.

I think single core background tasks shouldn't have to be in the same single-run-per-user queue as the other background tasks which can take many hours to complete. Maybe we could add some extra config which lets the app choose which queue a particular task should run in?

robnagler:
@e-carlin would runMulti help with this?

e-carlin:
Closing the loop after discussion in SW meeting last week:

runMulti won't fix. The problem is the jobs are now a parallel job which we only allow one of at a time.

Planned fix:
Fix the GUI so we can have a "start new simulation" button but send as a sequential job.

Issue #3886, opened 2021-10-12
selecting serial execution mode still runs in parallel


moellep:
When running an SRW simulation with "Execution Mode" Serial, it is still running in parallel. On alpha, running this sim: 

https://alpha.sirepo.com/srw#/beamline/6rWDTEuO?tab=multi 

shows 8 mpi processes running and the srwl log shows "Calculation on 8 cores with averaging of 5 particles/iteration." which is the same message when running normally with 8 cores.

e-carlin:
[This is the code](https://github.com/radiasoft/sirepo/blob/ec3a0ccc76d4be9819cd99ec1291fcee0712ac07/sirepo/sim_data/__init__.py#L251) we use to determine if we use `run` vs `run-background`. So anything with animation in the name will use run-background even if we set the jobRunMode to serial.  I think we may have other places in the code where we assume animation is parallel. Not sure off of the top of my head.

When deciding to send to a parallel or sequential agent we also use `sim_data.is_parallel` (not jobRunMode). 

jobRunMode should probably override inferring from the model name. Need to think more about how to implement and how this could impact sbatch run vs analysis.

robnagler:
dup  #1874

PR #4987, opened 2022-10-21
Fix #4942: coherent modes errors

### LINKED ISSUE #4942: 20221025 file error in coherent modes on Young's Double Slit Experiment
I looked at srwlib.py, and it is possible that `_file_path=fp_cm` is None. I was running on my machine with one core so maybe it's related. Here's the stack from the run.log:

```txt
Coherent Mode Decomposition done in: 45.415 s
Common Quadratic Terms were added to CMs in: 0.015 s
  File "parameters.py", line 231, in <module>
    main()
  File "parameters.py", line 229, in main
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwl_bl.py", line 3173, in calc_all
    _cm_wfr = lstWfrCM) #OC02072021
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwl_bl.py", line 1951, in calc_wfr_emit_prop_me
    _file_bkp = _fbk, _rand_opt = _op_rnd, _file_form = _fform, _n_mpi=_nmm, _n_cm=_ncm) #OC27062021
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwlib.py", line 10386, in srwl_wfr_emit_prop_multi_e
    srwl_uti_save_wfr_cm_hdf5(cohModes, None, _awfr=wfrA, _file_path=fp_cm) #OC28062021
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/srwlib.py", line 6828, in srwl_uti_save_wfr_cm_hdf5
    with h5.File(_file_path, 'w') as hf:
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/h5py/_hl/files.py", line 395, in __init__
    name = filename_encode(name)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.7/site-packages/h5py/_hl/compat.py", line 111, in filename_encode
    filename = fspath(filename)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

gurhar1133:
Error is thrown when coherent modes animation has too few cores

### DIFF
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2534,23 +2534,26 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return $scope.model == 'coherentModesAnimation';
             }
 
+            function coherentModesCoresError(cores, minCores) {
+                if (cores < minCores) {
+                    throw new Error(`${cores} cores found, should be > 3 for coherent modes`);
+                }
+            }
 
             function coherentModesEnoughCores() {
-                var a = appState.models.coherentModesAnimation;
-                srdbg(a);
-                if (a.jobRunMode == 'sequential') {
+                var c = appState.models.coherentModesAnimation;
+                if (c.jobRunMode == 'sequential') {
                     throw new Error('Coherent modes should not be run sequentially');
                 }
-                else if (a.jobRunMode == 'parallel') {
-                    // TODO (gurhar1133): leave comment explanation
-                    const n = Number(appState.enumDescription('JobRunMode', a.jobRunMode).split(" ")[0])
-                    if (n < coreMinimum) {
-                        throw new Error(`${n} cores found, should be > 3 for coherent modes`);
-                    }
+                else if (c.jobRunMode == 'parallel') {
+                    coherentModesCoresError(
+                        // Using enumDescription to obtain cores count from UI
+                        Number(appState.enumDescription('JobRunMode', c.jobRunMode).split(" ")[0]),
+                        coreMinimum,
+                    );
+                    return;
                 }
-                else if (a.sbatchCores < coreMinimum) {
-                    throw new Error(`${a.sbatchCores} cores found, should be > 3 for coherent modes`);
-                }
+                coherentModesCoresError(c.sbatchCores, coreMinimum);
             }
 
             function setActiveAnimation() {
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2541,10 +2541,10 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 var a = appState.models.coherentModesAnimation;
                 if (checkRunMode(a.jobRunMode)) {
                     return true;
-                };
+                }
                 if (a.sbatchCores < 3) {
                     throw new Error(`${a.sbatchCores} cores found, should be > 3 for coherent modes`);
-                };
+                }
             }
 
             function setActiveAnimation() {
--- a/sirepo/package_data/static/js/srw.js
+++ b/sirepo/package_data/static/js/srw.js
@@ -2533,6 +2533,20 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return $scope.model == 'coherentModesAnimation';
             }
 
+            function checkRunMode(runMode) {
+                return (runMode == 'sequential') || (runMode == 'parallel');
+            }
+
+            function coherentModesEnoughCores() {
+                var a = appState.models.coherentModesAnimation;
+                if (checkRunMode(a.jobRunMode)) {
+                    return true;
+                };
+                if (a.sbatchCores < 3) {
+                    throw new Error(`${a.sbatchCores} cores found, should be > 3 for coherent modes`);
+                };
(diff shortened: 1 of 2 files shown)


sirepo/package_data/static/js/srw.js
```diff
@@ -2533,6 +2534,20 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return $scope.model == 'coherentModesAnimation';
             }
 
+            function checkRunMode(runMode) {
+                return (runMode == 'sequential') || (runMode == 'parallel');
+            }
+
+            function coherentModesEnoughCores() {
+                var a = appState.models.coherentModesAnimation;
+                if (checkRunMode(a.jobRunMode)) {
```
robnagler, line 2543: The failure can occur in parallel, too. I don't think it should ever be run sequentially.

sirepo/package_data/static/js/srw.js
```diff
@@ -2533,6 +2534,20 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
...
+                    return true;
+                }
+                if (a.sbatchCores < coreMinimum) {
+                    throw new Error(`${a.sbatchCores} cores found, should be > 3 for coherent modes`);
```
robnagler, line 2547: Does this show nicely to the user?

robnagler (dismissed):

sirepo/package_data/static/js/srw.js
gurhar1133: Do you want the error coming through in the sim status panel?

robnagler: Yes, because the user should know. On production, they won't see an error if parallel is set, but will see an error if sequential is set. @moellep please correct me if I'm wrong about sequential.

sirepo/package_data/static/js/srw.js
robnagler: Yes. I believe you should get the number of cores correctly in dev.

sirepo/package_data/static/js/srw.js
moellep: I think it is OK if the error shows up in the page header - it won't be common except on dev. Single-core execution isn't supported for srw - see #4049

moellep: Wrong issue number - the correct is #3886 SRW can't run the coherent modes in serial mode on prod anyway.

robnagler: I agree. Let's make this server side and throw a UserAlert.

sirepo/package_data/static/js/srw.js
gurhar1133: So '2 cores (SMP)'  should be a failure case?

robnagler: Yes.

sirepo/package_data/static/js/srw.js
```diff
@@ -2533,6 +2534,28 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 return $scope.model == 'coherentModesAnimation';
             }
 
+            function coherentModesCoresError(cores, minCores) {
+                if (cores < minCores) {
+                    throw new Error(`${cores} cores found, should be > 3 for coherent modes`);
```
robnagler, line 2539: Where does this show up?

gurhar1133: Do you mean where does the error message show up? In the page header.

robnagler: I think this is still the wrong way about it. Since this is an error in configuration, it really should originate from the server and come back as a UserAlert. This way the server protects itself. If there are other users of the API (e.g. tests or bluesky), the server will not produce a reasonable error message.

moellep: Yes, I agree it should be on the server and it should throw an exception which would automatically be displayed in the report window. The problem is we don't want it to throw the exception when the code is generated, only when it is run and we can't find the appropriate hook for this on the server because pkcli.srw doesn't use run_background(). Where is the hook in the server to do this check?

robnagler:
```diff
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -1052,6 +1052,8 @@ def write_parameters(data, run_dir, is_parallel):
         p = ""
         _export_rsopt_config(data)
     else:
+        if data.report == "coherentModesAnimation":
+            raise sirepo.util.UserAlert("ruh roh!")
         p = _trim(_generate_parameters_file(data, run_dir=run_dir))
     pkio.write_text(
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
```

gurhar1133: The error from UserAlert doesn't show when in_slurm == True.
eg.
```
Simulation Error: no reply
Elapsed time: 00:00:05
```

 It will only show when not in vagrant/slurm.

robnagler: That's a bug. Create an issue. UserAlerts should cascade.

moellep:
Looks great!

PR #4990, opened 2022-10-21
Fix #4918: loss function options

### LINKED ISSUE #4918: Activait UI needs update to reflect more options for loss functions
We currently have a list of loss functions the user can select, but some loss functions have important input parameters that change their meaning.
eg.

```python
model.compile(
    ...
    loss='sparse_categorical_crossentropy',
   ...
)
```

```python
model.compile(
    ...
    loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
   ...
)
```

The UI needs to give the user the ability to specify inputs to the loss function selection, and the template code needs to be able to generate the loss function as specified by the user in the UI

gurhar1133:
loss functions that should have from_logits=True now get the option


sirepo/package_data/template/activait/train.py.jinja
```diff
@@ -18,7 +21,14 @@ history = model.fit(
 
 {% else %}
 
-model.compile(loss='{{ neuralNet_losses }}', optimizer='{{ neuralNet_optimizer }}')
+model.compile(
+{% if "keras" in neuralNet_losses %}
+    loss={{ neuralNet_losses }},
+{% else %}
+    loss='{{ neuralNet_losses }}',
+{% endif %}
```
moellep, line 29: This part is a little awkward. How about have the generated code like:
```
   loss=keras.losses.{{ neuralNet_losses }}({{ lossArguments }}),
```
And then set lossArguments to '' by default in template.activait, and to 'from_logits=True' in the case where you need it. That way you don't need to look for 'keras' in the loss value which is confusing.

PR #4997, opened 2022-10-24
Fix #4966 radia point table error

### LINKED ISSUE #4966: 20221024 Imported point table causes TypeError
Specifically, `o.referencePoints is not defined`.

mkeilman:
Notes:
- If the final point loaded from the file is the same as the first, it seems to confuse Radia. This removes it
- Also fixes a couple of random bugs
- Attaching a couple of files to test with
[Fillet Points.csv](https://github.com/radiasoft/sirepo/files/9855676/Fillet.Points.csv)
[quad_pole2.csv](https://github.com/radiasoft/sirepo/files/9855677/quad_pole2.csv)


mkeilman:
Fixes #5013 as well

PR #5009, opened 2022-10-25
fix #4986: implement genesis in react

### LINKED ISSUE #4986: 20221115 convert genesis to react

garsuga:
This implements Genesis in React. The schema has been converted, animations can now be viewed frame-by-frame like in Sirepo, and everything appears to be working correctly. I also decided to 'shown' the Scan and Time Dependence panels which were defined in the schema as views but not displayed in Genesis. I could not determine why they were being hidden and they appear to function correctly. Let me know if those should remain hidden.


garsuga:
This builds on the jspec branch so I am not sure what order you would like things merged in.

garsuga:
Should this be opened as a PR to the jspec branch instead?

robnagler:
Can we merge the jspec branch? Is CI passing?

garsuga:
I believe it is passing on JSPEC. It is unusual with the timing but maybe it would be better to handle your comments on the JSPEC PR after Genesis is merged and we begin work on typescript. I think being able to set up types while we make the abstractions we are missing would be quicker than separating those steps

moellep:
We want to continue hiding the Scan and Time Dependence panels - we don't support those features of Genesis yet - there is no handling for it on the server and visualization side.

robnagler:
I think we should merge Jspec. Then create a Genesis PR off master. There are too many changes here to figure out which is which. I can compare the branches, but that makes it hard to put comments in this PR.

moellep:
The genesis app works well - thanks for adding the plot paging feature.

PR #5064, opened 2022-11-01
Fix #5018 remove global uid in flask

### LINKED ISSUE #5018: 20221031 resolve logged_in_user in non-auth environments
sirepo.job needs a uid to cascade. This causes job_agent to need to init sirepo_auth so that it is visible so it has to call modules.init_and_import to get auth. Removing sirepo.auth from the loop of logged_in_user in non-auth environments would eliminate this.

robnagler:
I haven't don't extensive click around testing. Have not run jupyter. I'll do more testing. Just wanted to start the PR.


moellep:
Calls to SRW api_listFiles is failing with an assertion ```uid not supplied and no logged_in_user config```. You can reproduce by visiting the beamline page for the CHX simulation. The listFiles is called for the mirror element's profile data file.

robnagler:
warpvnd issues was related to #5118. Catching all places that require qcall is tricky when errors are suppressed. python_soruce_for_model is tricky. I'll check a few.

robnagler:
Tested all apps with a few examples each and exports from all. If tests pass, I think it's ready for a merge.

PR #5065, opened 2022-11-01
fix radiasoft/pykern#305: add eof newlines to pass new ci check

rorour:
tests had to be updated because they depend on sirepo/package_data/template/common-header.py.jinja


rorour:
Thanks @moellep. I'll get the corresponding pykern pr finished up today

PR #5076, opened 2022-11-03
Fix #5015 radia built in shapes dont stay where dragged

### LINKED ISSUE #5015: 20221026 Built-in extruded shapes do not stay where dragged in workspace
The shapes snap back to their original position. A few notes:
- Setting the center in the popup form positions the object correctly
- In any case the object is in the correct position in the visualization tab
- #5013 must be fixed first

mkeilman:
Notes:
- There was a race condition because getting the points from the server via stateful (now stateless) compute is asynchronous. The saves now live in callbacks so they get called only when the new points are returned
- This should also help strange occasional misplacement of shapes built from points in files


moellep:
@mkeilman - could you take a quick look at this PR - I think a prior radia PR has a merger conflict with it

mkeilman:
> @mkeilman - could you take a quick look at this PR - I think a prior radia PR has a merger conflict with it

Should be good now

PR #5078, opened 2022-11-03
Fix #4968 Import binary stl's (with test)

### LINKED ISSUE #4968: 20221130 Allow importing binary STL files to Radia
Currently only ASCII stl files can be imported, does not save file if binary.

Likely can be fixed with trimesh.exchange.stl [library](https://trimsh.org/trimesh.exchange.stl.html)

Benbenbenin0:
Can now import binary and ascii stl files.

Successfully visualizes a single binary, single ascii, and both binary and ascii in the same simulation.


robnagler:
Congrats!!!

PR #5104, opened 2022-11-08
Fix #4908: export h5 models

### LINKED ISSUE #4908: 20221130 Export h5 model
The Activait visualization tab should provide a download link to access the neural net layers export as a h5 file. The lattice builder should provide a un-weighted model, and the visualization output should provide a model with the calculated weights.

gurhar1133:
Both weighted and unweighted models are now saved and links for download provided in UI

### DIFF
--- a/sirepo/package_data/static/html/activait-regression.html
+++ b/sirepo/package_data/static/html/activait-regression.html
@@ -26,7 +26,9 @@
     <div class="clearfix"></div>
     <div data-ng-repeat="report in regression.reports track by report.columnNumber">
       <div class="col-sm-3">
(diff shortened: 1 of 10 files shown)


moellep:
I don't think mlService needs the hasFrames entry - instead the
modelDownloadLink could import frameCache and ask directly.

Don't pass the filename in the downloadDataFile suffix. That would
allow any file to be downloaded via that call and could be
abused. Instead, set the "model" field to one of the models, ex
"neuralNet" and "mlModel" and on the server, look for those values and
return the appropriate weighted or unweighted file.

I'd recommend putting the "Download weighted model" link on the
Training Error panel (epochAnimation model), because there can be
multiple Fit panels depending on the output dimension and it wouldn't
make sense to show the link on all the Fit panels if it is the same
file for each.

PR #5125, opened 2022-11-14
fix #5073 ml refactor

### LINKED ISSUE #4072: Add Slack link to JupyterHub
On JupyterHub pages, we want a prominent link to Slack.

### LINKED ISSUE #5072: zip file with csv workflow is awkward
This workflow doesn't seem to work without page refreshes:

Create new sim.
Select a zip file which contains a csv for Data File.
The Data field appears, but is empty (bad).
Select OK.
The Data field populates and autocompletes to the csv contained in the zip.
The Columns panel appears, but no columns are shown (bad).
Refresh the browser window.
The columns appear and the column reports are populated.

### LINKED ISSUE #5073: simplify data source selection
The Data Source panel for Regression is too complicated with what we currently support.

When Regression is initially selected for Application Mode only "Data Origin" and "Data File"/"Data URL" should be shown.

If "Data File" is selected and it is a CSV file, show "Inputs Scalar" and "Outputs Scalar". 

If "Data File" is selected and an H5 file, don't show any other fields. We'll add special scalar functions specific to image and/or intensity data.

If we later add image support from zip/gz files, we can expand the UI, but for now we should only allow CSV or H5 files.

moellep:
- add 2 analysis examples
- fix #5072 removes zip option for now
- refactor javascript


gurhar1133 (approved):
Looks good. Clicked around with the example you sent and that worked.

PR #5133, opened 2022-11-15
Fix #5046 - updated css

### LINKED ISSUE #5046: 20221230 radia.css affecting simulations list
The styling in radia.css is affecting the simulations list page which causes it to render differently than other apps. This could probably be prefixed with a radia-specific css tag.

x

Also, it would be nice if mousing over the table rows didn't cause the row height to change which causes the lower part of the page to shift slightly while moving the mouse. Keeping the border width (2px) the same on both normal and hover would work, and the color could be changed to hover to indicate it is active.

mkeilman:
This uses outline instead of border, the difference being that the outline does not contribute to the size of elements so the layout doesn't get recalculated.

I also added alternating row colors but not sure about it...


moellep:
Looks great - thanks Mike!

PR #5141, opened 2022-11-15
fix radiasoft/raydata#38: loading indicator for scans table

rorour:
> Just a thought - how about show "Checking for new scans..." where the "..." changes every second, . . ...
And when it finds a new one, it could change to "Fetching scan" for a second, or until the new scan is available, and then immediately change back to "Checking for new scans..."?

Had to make a change from @moellep's [comment](https://github.com/radiasoft/sirepo/pull/4977#discussion_r1014258261), the time interval for "Fetching scan" was so short that it wouldn't show at all. So for now "Checking for new scans..." is still popping up and disappearing. I'm open to more suggestions.

### DIFF
--- a/sirepo/package_data/static/js/raydata.js
+++ b/sirepo/package_data/static/js/raydata.js
@@ -320,7 +320,10 @@ SIREPO.app.directive('scansTable', function() {
                     </tr>
                   </tbody>
                 </table>
-                <div ng-if="awaitingScans">Loading scans...</div>
+                <div style="height: 20px;">
+                  <div ng-if="awaitingScans" data-dots-animation="" data-text="Checking for new scans"></div>
+                  <div ng-if="noScansReturned">No matching scans returned.</div>
+                </div>
               </div>
             </div>
             <div data-column-picker="" data-title="Add Column" data-id="sr-columnPicker-editor" data-available-columns="availableColumns" data-save-column-changes="saveColumnChanges"></div>
@@ -365,6 +368,7 @@ SIREPO.app.directive('scansTable', function() {
             // POSIT: status + sirepo.template.raydata._DEFAULT_COLUMNS
             $scope.defaultColumns = ['status', 'start', 'stop', 'suid'];
             $scope.images = null;
+            $scope.noScansReturned = false;
             $scope.orderByColumn = 'start';
             $scope.reverseSortScans = false;
             $scope.scans = [];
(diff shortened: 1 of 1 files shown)


sirepo/package_data/static/js/raydata.js
```diff
@@ -320,7 +320,10 @@ SIREPO.app.directive('scansTable', function() {
                     </tr>
                   </tbody>
                 </table>
-                <div ng-if="awaitingScans">Loading scans...</div>
+                <div style="height: 20px;">
+                  <div ng-if="awaitingScans" data-dots-animation="" data-text="Checking for new scans"></div>
+                  <div ng-if="noScansReturned">No matching scans returned.</div>
```
e-carlin, line 325: What about `No scans found`? It is simpler. In addition matching implies the user can change their query which they can't really. And when I see "returned" it makes me think that maybe the data just didn't come back which isn't what's going on here.

e-carlin: Also,  is this available on the "Executed" tab? I couldn't get it to appear.

e-carlin (changes requested):

sirepo/package_data/static/js/raydata.js
rorour: That's weird, it's appearing for me on the executed tab.

And sure, my initial thinking was that I wanted to make it clear that there were no scans matching the query, rather than there being no scans in the catalog. I think you're right that `No scans found` is clear enough though.

What do you mean by this though?

> In addition matching implies the user can change their query which they can't really.

There are no scans matching the date range and catalog, but they could change those which changes their query. Or am I misunderstanding?

rorour: [07aefb1](https://github.com/radiasoft/sirepo/pull/5141/commits/07aefb105646ca785918a47cad0460fddb2d4854)

e-carlin: > That's weird, it's appearing for me on the executed tab.

Ah my bad. I hadn't selected a date range so it wasn't querying. I see it now.

> There are no scans matching the date range and catalog, but they could change those which changes their query. Or am I misunderstanding?

Since I was only viewing on the Queued scans tab I wasn't thinking about the date range. The only thing the user could change was the catalog name which isn't much. This is 100% my own personal feeling so if you don't agree we can go with your text.

e-carlin (approved):
LGTM

e-carlin:
I rebased and force pushed to update. GH was complaining that the branch was out of date. I'll merge once the checks finish.

Issue #4971, opened 2022-10-19
20221130 Visualizing Radia objects with the same STL file only displays one object


Benbenbenin0:
When you create multiple objects with the same STL file in Radia, only one of those object is displayed in the visualization.

Benbenbenin0:
Seems to be additional consequence of issue #4972, will likely merge issues

Benbenbenin0:
Merging issue into #4972, multiple objects display when using correct center values for shapes (both objects were displaying on top of each other).

PR #5143, opened 2022-11-16
Fix #4972 Shape center and size not reflective of actual

### LINKED ISSUE #4972: 20221130 Center and sizes for STL objects not reflected accurately in Radia
Object center is dependent on how how the STL is exported, sometimes isn't aligned with the shape. Can likely be fixed by using trimesh.center_mass (or .centroid) to determine appropriate center.

Object size is currently just default values, can likely use trimesh bounding box to determine actual size.

Benbenbenin0:
Merged #4971 into this PR along with #4972

Shapes are now built at center provided by dragging in the design tab, also now displays shapes using the same STL file.

Shape size is now updated to the actual after building from the imported file, however the design tab still leaves default values in the "size [mm]" field that do nothing when changed.

@mkeilman think we need a new issue for that? Not sure whether to let the user scale the STL's size in the designer, remove the field for STL's or do something else.


mkeilman:
> Shape size is now updated to the actual after building from the imported file, however the design tab still leaves default values in the "size [mm]" field that do nothing when changed.
> 
> @mkeilman think we need a new issue for that? Not sure whether to let the user scale the STL's size in the designer, remove the field for STL's or do something else.

We should still display the size but disable the input fields. See https://github.com/radiasoft/sirepo/blob/dcb249c0625782ee3cf8ed4877244a6b437f1b27/sirepo/package_data/static/js/radia.js#L3841
It's a one-line fix if you want to add it to this PR, but a new issue is fine too

sirepo/template/radia_util.py
```diff
@@ -304,6 +304,7 @@ def build_stl(**kwargs):
     g_id = radia.ObjPolyhdr(
         d.vertices, (numpy.array(d.faces) + 1).tolist(), d.magnetization
     )
+    g_id = radia.TrfOrnt(g_id, radia.TrfTrsl([d.center[0], d.center[1], d.center[2]]))
```
mkeilman, line 307: This doesn't quite work on its own. We want to use the center defined by the user, irrespective of the points in the stl file. I think the easiest way is to subtract off the stl's center first:
```python
o.stlCentroid = mesh.centroid.tolist()
...
g_id = radia_util.build_stl(
...
    center=[x - o.stlCentroid[i] for i, x in enumerate(o.center)],
)
```

mkeilman (changes requested):
See comment on centering

sirepo/template/radia_util.py
Benbenbenin0: Ah, I was testing radia.TrfOrnt() without realizing I was using (0,0,0) as my initial center, misunderstood the function.

sirepo/package_data/static/js/radia.js
```diff
@@ -3870,9 +3870,16 @@ SIREPO.viewLogic('geomObjectView', function(appState, panelState, radiaService,
         panelState.showField('geomObject', 'materialFile', o.material === 'custom');
 
         panelState.enableField('geomObject', 'size', true);
+
+        if (o.type == 'stl') {
+            panelState.enableField('geomObject', 'size', false);
+            //TODO(BG): Only disables 'size' field, need to build shape to get sizes to update values (likely will need to send request since python)
```
mkeilman, line 3876: I'll make a separate issue for this

mkeilman (commented):
One minor comment

mkeilman:
ping @Benbenbenin0 - a couple of changes still in queue here. If you need to focus on ldap that's ok - I'll fix this up; lmk one way or the other

mkeilman (approved):

Benbenbenin0:
@mkeilman Thanks for the fixes and merge

PR #5145, opened 2022-11-16
fix #5144 improve regression tab layout for xxl width

### LINKED ISSUE #5144: 20221116 regression tab layout on xxl monitor
The layout should show both the Neural Net Layers and simulation settings side by side when the browser is xxl wide.

moellep:


gurhar1133 (approved):
Looks good to me. Pulled the branch and clicked around and seems all good 👍

PR #5155, opened 2022-11-16
Fix #5086 radia rotate about object center

### LINKED ISSUE #4700: 20221026 Clicking (not dragging) transform items in freehand editor throws error
```javascript
TypeError: Cannot read properties of null (reading 'center')
    at self.shapeForObject (radia.js?20220810.192321:604:56)
    at $scope.dropSuccess (sirepo-plotting-vtk.js?20220810.192321:2252:47)
    at fn (eval at compile (angular.js:16418:15), <anonymous>:4:306)
    at ngDraggable.js?20220810.192321:342:37
    at angular.js:21593:28
    at Tg.completeTask (angular.js:21231:7)
    at angular.js:6812:7
```
The code seems to be treating transforms as "shapes", which they are not

### LINKED ISSUE #5086: 20221118 Add "rotate around axis" transform
This is similar to the existing "rotate" transform except there is not input for the center point. The center of the rotation is just always kept at the center of the shape, and automatically updated if the user moves the shape's center. This is really a QoL feature but it's come up a lot trying to test a quadrupole build. 

I'm not sure if it's better to have an entirely new transform listed for this or if it should just be a switch on the existing rotate transform.

### LINKED ISSUE #5101: 20221118 Remove top-level translation transform
Since users can drag objects around the workspace, adding a translation is redundant and confusing. It's still meaningful for clones however. @cchall @j-edelen is this a sensible change?

mkeilman:
Notes:
- 54375ca7c10abb19b5c8a3c6075af25c2d8f14f8 fixes #5101 
- 3dfa3c729d049901a560967b5b979cc578dea28c fixes #4700
- Rotations default to using the object center
- Added a bit of code cleanup

Regarding dragging objects that are rotated around a specified point: it's a bit tricky because the object being dragged has the transform applied to it, but we really want to move the un-transformed object. I'll try to think of a natural way to present that to the user.


moellep:
@mkeilman could you take a look at the merge conflicts with this pull request (sirepo-plotting-vtk.js)

Issue #5174, opened 2022-11-20
20221130 split db_upgrade into different parts and as pkcli


robnagler:
db_upgrade is doing too much. There are different databases: auth_db, job db, and user db. auth_db has to run as the server starts. job db has to run when the supervisor starts (and in a separate process since they may import templates). user db should run in the server (in separate process). These should be separate modules, and they can share db_upgrade_t, just need to make sure the methods don't collide.

Coordinating all this while multiple servers are starting is problematic. We need a command line utility that gets run when the server stops on alpha/beta/prod that knows how to start and stop servers so that we have a quiescent system during the upgrade. In dev and single user (sirepo.run) docker images, we can stick with automatic upgrades for convenience.

Issue #3933, opened 2021-11-02
20221230 Refine ModuleNotFound in import_submodule


robnagler:
The actual error is:

`ModuleNotFoundError: No module named 'radia'`

The message was:

```txt
../sirepo/http_reply.py:285:_gen_exception_error unsupported exception=<class 'AssertionError'> msg=cannot find submodule=template.radia in package_path=('sirepo',)
```

It was confusing, of course, because there is such a module (sirepo.template.radia).

The fix was `radia_run beamsim-codes`, but I couldn't have known that.

PR #5175, opened 2022-11-20
Fix #4872 auth_db package holds models

### LINKED ISSUE #4872: 20221123 auth_db classes and auth models at outer scope
Possibly be able to get rid init_model

robnagler:
- auth_db is a package with all models re-exported on sirepo.auth_db
- all auth_db model modules loaded dynamically (with package path)
- Renamed session to spa_session
- auth_db.create_or_upgrade is called by server and pkcli.admin_db_upgrade
- Created pkcli.admin.db_upgrade for #5174


robnagler:
Requires https://github.com/radiasoft/pykern/pull/322 so checks won't pass until that's merged.

sirepo/auth_db/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+        r = cls._execute_raw_sql(f"PRAGMA table_info({t})")
+        for c in r.all():
+            if not c[1] == column:
+                continue
+            assert c[2] == column_type, (
+                f"unexpected column={c} when adding column={column} of",
+                f" type={column_type} to table={table}",
+            )
+            return
+        cls._execute_raw_sql(f"ALTER TABLE {t} ADD {column} {column_type}")
```
e-carlin, line 59: I think a comment here that this is never used with user defined data could be good and maybe moving _execute_raw_sql into this function so it is never used elsewhere.

sirepo/auth_db/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+
+def db_filename():
+    return sirepo.srdb.root().join(_SQLITE3_BASENAME)
+
+
+def init_module():
+    def _classes():
+        res = PKDict()
+        for r in sirepo.feature_config.cfg().package_path:
+            p = pykern.pkinspect.module_name_join((r, "auth_db"))
```
e-carlin, line 229: This isn't backwards compatible. Packages on package path may not provide an auth_db sub-package.

e-carlin (approved):
Couple nits. LGTM. I'll let you merge since it is large and depends on pykern change.

sirepo/auth_db/__init__.py
robnagler: I was wondering whether to use package_path or not. I'm happy to remove. The failure is immediate so not a huge problem. Should I remove?

e-carlin: What about adding a check? If the package has it then use it. That is what the other areas of the code do in regards to package path.

sirepo/auth_db/__init__.py
robnagler: Thanks. I read about this, and we should just use placeholders, which seem to work fine for DDL. 57dd3717a

sirepo/auth_db/__init__.py
robnagler: This is a trickier case, because the other cases are looking for a module along package_path. This is loading all modules along package_path. #3933 is key here. If there's a syntax error in `auth_db/__init__.py`, it gets a not found, and there won't be an sirepo.auth_db module. The other issue is if the package path is being used to improve auth_db model or creating a new one. We don't handle the former case, and I don't think we can support the latter case. Therefore, I think this is a YAGNI, and I'll take it out.

e-carlin: Sounds good.

PR #5204, opened 2022-11-28
Fix #4995: batch at a time training with image shuffle

### LINKED ISSUE #4995: 20221230 support shuffle for 2d image processing
Currently the "Shuffle Before Epoch" option is ignored for image processing because the hdf5 reader does not support this. We want to avoid reading the full dataset into memory, but there may be keras data access utilities which allow loading data only for the required batch when training, which would allow for shuffling to be allowed.

gurhar1133:
Image data loader loads and shuffles batches at a time. Scaling is done with partial fitting the batches. An error message is passed to user when training fails


moellep:
I think the SirepoHDF5ImageGenerator should be moved into the generated code - define the class in loadImages.py.jinja

Otherwise, template.activate imports h5imagegenerator which brings in all the tensorflow libraries (including trying to load libcudart.so) anytime the template is accessed.

moellep:
I'm having problems running the ML on this branch - for example, an image dataset is not able to run and create the Training Error plot. I'm seeing "nan" in the fit.csv file:
```
epoch,accuracy,loss,val_accuracy,val_loss
0,0.2507777810096741,nan,0.24899999797344208,nan
1,0.2511666715145111,nan,0.24899999797344208,nan
```

gurhar1133:
> I'm having problems running the ML on this branch - for example, an image dataset is not able to run and create the Training Error plot. I'm seeing "nan" in the fit.csv file:
> 
> ```
> epoch,accuracy,loss,val_accuracy,val_loss
> 0,0.2507777810096741,nan,0.24899999797344208,nan
> 1,0.2511666715145111,nan,0.24899999797344208,nan
> ```

I'll look into it

e-carlin:
@gurhar1133 fyi I updated the sirepo-ci image and tests are now passing. As far as having h5imagegenerator is concerned this can be merged.

gurhar1133:
> @gurhar1133 fyi I updated the sirepo-ci image and tests are now passing. As far as having h5imagegenerator is concerned this can be merged.

Awesome, thanks @e-carlin !

moellep:
@gurhar1133 please take a look at the changes I merged. It now scales the Y values per column, and allows setting the SirepoSequence to return only x or y values which is needed to create the fit plot data.

gurhar1133:
> @gurhar1133 please take a look at the changes I merged. It now scales the Y values per column, and allows setting the SirepoSequence to return only x or y values which is needed to create the fit plot data.

The changes look and work well. I think having a dedicated SirepoScaler class is a good move

PR #5214, opened 2022-11-29
fix #5032: refactor react and convert to typescript

### LINKED ISSUE #5032: 2022/11/31 react needs more abstraction, less code in UI components
This issue includes taking care of the comments raised in https://github.com/radiasoft/sirepo/pull/4802 and completing that work in Typescript if possible

### LINKED ISSUE #5210: 20221201 show old animation report output when returning to a simulation
The react app doesn't show previous run results when you return to a simulation. The runStatus should be called to see if there is existing data on the server.

garsuga:
This also includes Paul's changes to fix #5210.

This PR should probably come before the next app is implemented into React. The next apps should be built off of this branch.


garsuga:
I do not know why the old commits are shown on this PR but the diff is correct.

Issue #1972, opened 2019-10-24
Discard Changes should be an API


robnagler:
Right now discard changes, deletes the example, and relies on findByName to copy in the example. That's decoupled, and difficult to know the path through the code.

moellep:
This seems broken - the SRW fixups are never getting run. template.srw.fixup_old_data() never is called - was this supposed to go in sim_data.fixup_old_data() instead?

moellep:
See commit 8d1ef2356630e30c88ebbc40ea4881ebdccf2309 with the change which broke this.

moellep:
I think this is the problem:
```
        if 'horizontalDeflectingParameter' not in dm.undulator:
            cls.__fixup_old_data_by_template(data)
```
I'll try it with running the fixup every time.

robnagler:
Discard Changes is still not an API.

robnagler:
See https://github.com/radiasoft/sirepo/issues/5342#issuecomment-1377353881

api_revertExample should keep the same simulationId and increment simulationSerial. All other data should be replaced with the example data found by name. If the simulation is not an example, raise a UserAlert. This might happen if an example gets deleted (rare but possible). The api should return the data to the client.

PR #5216, opened 2022-11-30
Fix #3699 integrate rsopt with srw

### LINKED ISSUE #3699: 20221231 integrate rsopt with srw
Give the user the option to generate the data and run the srw simualtions directly from sirepo

mkeilman:
Notes:
- Adds a machine learning tab with the rsopt panel and an animation to both generate the data and run srw
- Removes the rsopt panel from the gear menu
- Allows direct download of the results file or opening in activait
- rsopt does not work on NERSC yet, so we'll have to see what that looks like when this is on alpha

To open in activait, this copies the results file to the activiat lib directory with `sim_data. sim_file_to_other_sim_lib()`. I'd like to know if that function is "legal" in the sense of not having any "cd .." lurking underneath. There are a few other niceties to consider as we go - like using a symlink instead of a copy, or returning a url, or allowing an upload to a site of the user's choice - but this is good enough for testing datasets of moderate size.


sirepo/template/srw.py
```diff
@@ -1432,8 +1448,16 @@ def _enum_text(name, model, field):
...
-    return _write_rsopt_zip(data, _rsopt_jinja_context(data))
+def _export_rsopt_config(data, run_dir):
+    ctx = _rsopt_jinja_context(data)
+    f = _write_rsopt_zip(data, ctx)
+    if _SIM_DATA.is_for_ml(data.report):
+        s = f"{_SIM_DATA.EXPORT_RSOPT}_run.sh"
+        pkio.write_text(s, template_common.render_jinja(SIM_TYPE, ctx, s))
+        template_common.subprocess_output(
+            ["bash", s],
+        )
```
moellep, line 1459: You definitely don't want to run rsopt during write_parameters(). This is also why the background_percent_complete() isn't working. Instead, write out a parameters.py file which calls this. 

You can add "machineLearningAnimation" to the "not mpi" part of sim_data.srw.is_run_mpi() similar to the beamlineAnimation exception there. That way it won't run the parameters.py with mpi.

moellep:
A few minor UI issues:

The "Start New ML Data Generation" button extends off the panel, and the panel isn't taking up the full page width:
img1

The "Random Seed" editor is offset from the other editors. I think if you wrap the input and 2 buttons in a div with class "row" it will show correctly. When you nest a bootstrap "col" within a "col" you will see odd offsets like this - instead have "col" "row" "col" nested.
img2

I think the fixed height on the element selector can be removed now that it isn't showing in a modal dialog. Otherwise it is a bit odd when it cuts off parts of the input fields.
img3

moellep:
@robnagler Could you comment on the "this copies the results file to the activiat lib directory with sim_data. sim_file_to_other_sim_lib()" part of this PR?

robnagler (changes requested):
More thought needed about how to simply and not generalize.

sirepo/template/srw.py
```diff
@@ -1996,6 +2021,23 @@ def _load_user_model_list(model_name, qcall=None):
     return _load_user_model_list(model_name, qcall=qcall)
 
 
+def _machine_learning_percent_complete(run_dir, res):
+    dm = simulation_db.read_json(run_dir.join(template_common.INPUT_BASE_NAME)).models
+    count = len(
+        pkio.sorted_glob(
+            run_dir.join("ensemble").join("worker*").join("sim*").join("values.npy")
```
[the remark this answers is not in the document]
mkeilman: > I think the fixed height on the element selector can be removed now that it isn't showing in a modal dialog. Otherwise it is a bit odd when it cuts off parts of the input fields. img3

I think it's still useful once you get more than a couple of elements. [78e56ad](https://github.com/radiasoft/sirepo/pull/5216/commits/78e56ad71e64ce99c96485f35e4534d2b6f62247) adds a border - does that help?

mkeilman:
> The "Start New ML Data Generation" button extends off the panel, and the panel isn't taking up the full page width

[575bdc6](https://github.com/radiasoft/sirepo/pull/5216/commits/575bdc623e1bac4c79136d1f9afa73f457c52a43) - the button styling is universal so I just shortened the label a bit.

> The "Random Seed" editor is offset from the other editors. I think if you wrap the input and 2 buttons in a div with class "row" it will show correctly. When you nest a bootstrap "col" within a "col" you will see odd offsets like this - instead have 
> 
[2969ca2](https://github.com/radiasoft/sirepo/pull/5216/commits/2969ca25a40b7f41d72f501efde9fa8861225132)

sirepo/template/srw.py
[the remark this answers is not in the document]
moellep: I like the border, but I really think you don't need to fix the height to 360px because it won't always display well in different browser sizes. You could have the height be variable and set to fill to the remaining browser height, but that would probably be too much work (elegant does this on the Lattice panel - the Beamline Editor panel fills the full browser window).

For example - here there is space available to show more values:
pic1

And here there isn't and you get 2 scrollbars which is bad:
pic2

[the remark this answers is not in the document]
mkeilman: > join takes multiple args. It seems to me you just want to count values.npy with a walk tree, but I could be wrong. Hardwiring the directory structure here is going to make this rigid/fragile.

[e6ba5ef](https://github.com/radiasoft/sirepo/pull/5216/commits/e6ba5efeda1abec0d3acee9cb34b4a59c5b8d084)

sirepo/sim_data/__init__.py
```diff
@@ -552,6 +552,32 @@ def schema(cls):
     def sim_file_basenames(cls, data):
         return cls._sim_file_basenames(data)
 
+    @classmethod
+    def sim_file_to_other_sim_lib(
```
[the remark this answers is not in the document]
mkeilman: [7a6fbb4](https://github.com/radiasoft/sirepo/pull/5216/commits/7a6fbb4fe80e1402e24544c941b02ec06d1b1fa3) - made private and removed the docstring. Also made #5226 to address our discussion of managing sim_db files

sirepo/sim_data/__init__.py
```diff
@@ -552,6 +552,32 @@ def schema(cls):
...
+        from sirepo import simulation_db
+
+        f = (
+            cls.lib_file_name_with_model_field(model_name, field, basename)
+            if model_name and field
+            else basename
+        )
+        t = simulation_db.simulation_lib_dir(other_sim_type, qcall=qcall).join(f)
+        r = _request(
+            "GET", _cfg.supervisor_sim_db_file_uri + cls._sim_file_uri(sim_id, basename)
```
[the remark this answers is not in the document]
mkeilman: [7a6fbb4](https://github.com/radiasoft/sirepo/pull/5216/commits/7a6fbb4fe80e1402e24544c941b02ec06d1b1fa3)

sirepo/template/srw.py
```diff
@@ -1996,6 +2021,23 @@ def _load_user_model_list(model_name, qcall=None):
...
+        )
+    )
+    res.frameCount = count
+    res.percentComplete = 100 * count / dm.exportRsOpt.totalSamples
+    if res.percentComplete >= 100:
+        res.resultsFile = _SIM_DATA.ML_OUTPUT
+        _SIM_DATA.put_sim_file(
+            dm.simulation.simulationId, run_dir.join(res.resultsFile), res.resultsFile
```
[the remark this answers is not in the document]
mkeilman: [3096baa](https://github.com/radiasoft/sirepo/pull/5216/commits/3096baae8be5e9aa5d3b373d52c99f7fe1ad39fc)

sirepo/template/srw.py
mkeilman: [80c95c3](https://github.com/radiasoft/sirepo/pull/5216/commits/80c95c34ca203c39b621753679f9f84e0fb86eeb)
[85050ad](https://github.com/radiasoft/sirepo/pull/5216/commits/85050adcfd762b12f3dc087afbc7d7ce8e5476cd)
Running the bash script should not happen in mpi - rsopt takes care of it. If you try to run this parameters.py with mpi you get file descriptor errors

sirepo/template/srw.py
```diff
@@ -2608,8 +2653,10 @@ def _write(zip_file, path):
             ctx.readmeFileName,
             template_common.render_jinja(SIM_TYPE, ctx, ctx.readmeFileName),
         )
-        for f in ctx.libFiles:
-            z.write(f, f)
+        if data.report != _SIM_DATA.ML_REPORT:
```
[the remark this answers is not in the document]
mkeilman: [80c95c3](https://github.com/radiasoft/sirepo/pull/5216/commits/80c95c34ca203c39b621753679f9f84e0fb86eeb)
I was trying to reuse the zip stuff from exportRsOpt but that just complicated things

sirepo/package_data/static/js/srw.js
```diff
@@ -807,6 +807,80 @@ SIREPO.app.controller('BeamlineController', function (activeSection, appState, b
...
+                method: 'results_to_activait',
+                args: {
+                    file: self.resultsFile,
+                    id: appState.models.simulation.simulationId,
+                }
+            }
+        );
+    }
+
+    self.createActvaitSimulation = () => {
```
robnagler, line 835: typo

sirepo/template/srw.py
```diff
@@ -848,6 +860,17 @@ def stateful_compute_undulator_length(data):
     return compute_undulator_length(data.args["tabulated_undulator"])
 
 
+def stateless_compute_results_to_activait(data):
+    _SIM_DATA._sim_file_to_other_sim_lib(
```
robnagler, line 864: This can't be private, because it is called outside the module. It needs to be public.

sirepo/sim_data/__init__.py
```diff
@@ -622,6 +622,26 @@ def schema(cls):
     def sim_file_basenames(cls, data):
         return cls._sim_file_basenames(data)
 
+    # TODO(mvk):  use a server to provide access to sim db and library files
+    @classmethod
+    def _sim_file_to_other_sim_lib(
+        cls, sim_id, basename, other_sim_type, qcall=None, model_name=None, field=None
+    ):
+        cls._assert_server_side()
```
robnagler, line 630: I don't see how this would work. It's not server side.

sirepo/sim_data/__init__.py
```diff
@@ -622,6 +622,26 @@ def schema(cls):
     def sim_file_basenames(cls, data):
         return cls._sim_file_basenames(data)
 
+    # TODO(mvk):  use a server to provide access to sim db and library files
+    @classmethod
+    def _sim_file_to_other_sim_lib(
```
robnagler, line 627: Make this an srw specific method `srw_sim_file_to_other_sim_lib` and put in sim_data.srw

robnagler (changes requested):

sirepo/package_data/static/js/srw.js
mkeilman: [38d893b](https://github.com/radiasoft/sirepo/pull/5216/commits/38d893b4140332db04f8e193a82a0b0e8968b361)

sirepo/sim_data/__init__.py
mkeilman: It does work, though - `_cfg.lib_file_uri` is `None`. Does this imply something deeper is amiss?

sirepo/template/srw.py
mkeilman: [987d29e](https://github.com/radiasoft/sirepo/pull/5216/commits/987d29e2fd3306bb306756b8fe3228dad1047030)

sirepo/sim_data/__init__.py
mkeilman: [987d29e](https://github.com/radiasoft/sirepo/pull/5216/commits/987d29e2fd3306bb306756b8fe3228dad1047030)

I'm still not sure of this - I think I should use copy and not a GET request, but it's not clear to me...

mkeilman: [69fc7bc](https://github.com/radiasoft/sirepo/pull/5216/commits/69fc7bc86e07aabb4df966c0acf8a334a5668c8e)
This is better (is it?)

sirepo/sim_data/__init__.py
robnagler: Yes indeed https://github.com/radiasoft/sirepo/issues/5254

sirepo/sim_data/__init__.py
robnagler: This is template code (see also https://github.com/radiasoft/sirepo/issues/5254 because there is no qcall here), this won't work right. It would need be a server-side API. For now, let's do it the long way around with GET via sim_db_file. 

How much data is this file?

mkeilman: [3d46d8f](https://github.com/radiasoft/sirepo/pull/5216/commits/3d46d8f319ea38d4abbf613dee0b4bb0cef4b516)
To repeat my earlier comment, not sure of this, although in any case other sim_data functions can use it

The file size is highly variable but anywhere from a megabyte to gigabytes; really as much as the user wants to generate.

sirepo/package_data/static/js/srw.js
```diff
@@ -807,6 +807,82 @@ SIREPO.app.controller('BeamlineController', function (activeSection, appState, b
...
+    self.simScope = $scope;
+    self.resultsFile = null;
+    self.simComputeModel = 'machineLearningAnimation';
+    self.simState = persistentSimulation.initSimulationState(self);
+
+    function openActivait(data) {
+        requestSender.sendStatelessCompute(
+            appState,
+            d => {
+                $window.location.href = `activait#/data/${data.models.simulation.simulationId}`;
```
robnagler, line 823: This doesn't seem right at all. Why is there a hardwired URL? There are routines for formatting URLs and local redirects.

sirepo/package_data/static/js/srw.js
```diff
@@ -807,6 +807,82 @@ SIREPO.app.controller('BeamlineController', function (activeSection, appState, b
...
+                method: 'results_to_activait',
+                args: {
+                    file: self.resultsFile,
+                    id: appState.models.simulation.simulationId,
+                }
+            }
+        );
+    }
+
+    self.createActivaitSimulation = () => {
```
robnagler, line 835: I see the problem. createActivaitSimulation calls newSimulation. There's no need for openActivait to be called here at all. Rather, implement this in template.srw.new_simulation. See the code for api_newSimulation. This is documented in https://github.com/radiasoft/sirepo/wiki/Template-API. Might be worth a read.

This comes down to the rule: If you find yourself writing business logic in the GUI, something is wrong. The GUI is about presentation, and this is business logic about how simulations are created.

robnagler (changes requested):

sirepo/package_data/static/js/srw.js
mkeilman: That's where I started, but nothing gets you from one app to another app with a specific sim id - at best you can reach the sim list with `globalRedirect()`. I used the srw -> shadow conversion as a guide:
https://github.com/radiasoft/sirepo/blob/8fb85dc7befa7684ac4de4631b053c75d47a014a/sirepo/package_data/static/js/sirepo-components.js#L3036
which only provides a link of this same format for the user to click. I think an issue to cover this case in requestSender is warranted.

robnagler: If you see something that's "not right", then it's time to fix it, not copy it. Operations like this should always be server side, similar to the problem. For example #1972 is still a problem. The reason it isn't fixed is that it takes time. The GUI should not be responsible for operations like this. Please add an issue that the SRW-Shadow conversion code should be server-side, too.

> nothing gets you from one app to another app with a specific sim id

This is a bug, not a feature (or a fact). Our software is one ecosystem, there are no walls that prevent you from modifying the code to be better. If you have the time, and it seems to me you do, then please do it right. It's not complicated or complex.

mkeilman: I don't always have a good feel for this. If it's in the code, and not egregiously weird, then my assumption is that it is "right". It is, at least, "how we do it". I'll use this experience to expand what I regard as "weird".

At the time, I had no time, but at this time I have the time

robnagler: That's fair. A skill to improve. Here's a simple rule: If a server response to the GUI is to turn around and do something else immediately (not after a poll, like runStatus) on the server, that's weird. Why didn't the server just do both things? This has to do with data locality. Why would the data come all the way back just to be pushed to the same server again. 

Another rule: don't give the server something it already has. A good example of this is runSimulation. The data should already be saved on the server when runSimulation is called. Otherwise, there will be a collision with the data on disk vs what simulation is being run. Therefore, the only thing that should be sent is the identity of the data (including version/serial) to runSimulation.

mkeilman:
Latest merge incorporates changes for #5257, pending in #5270

sirepo/sim_data/__init__.py
mkeilman: Obviated

sirepo/sim_data/__init__.py
mkeilman: Obviated

sirepo/package_data/static/js/srw.js
mkeilman: [93f26d2](https://github.com/radiasoft/sirepo/pull/5216/commits/93f26d2abc840faddd2fc8d2684fad38c0908736)

sirepo/package_data/static/js/srw.js
mkeilman: [93f26d2](https://github.com/radiasoft/sirepo/pull/5216/commits/93f26d2abc840faddd2fc8d2684fad38c0908736)

mkeilman:
The latest substantial change is moving the sim creation details into activait's `new_simulation()` function. The file copy could no longer happen in sim_data so it's all in the template. Also using a symlink to the h5 file instead of a copy - we already copy it out of the run_dir so that's up to 3 large files. We'll worry about handling deletion of the source file later (unless #5226 comes first)

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2160,6 +2165,20 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, userAg
         $window.open(self.formatUrl(routeName, params), '_blank');
     };
 
+    self.openSimulation = (app, route, simId) => {
+        const m = {};
+        m[route] = routeMapLocal(route, `/${route}/:simulationId`);
```
robnagler, line 2170: Why does this link need to be hardwired and not gotten from the schema?

sirepo/template/srw.py
```diff
@@ -631,6 +636,13 @@ def post_execution_processing(
     success_exit=True, is_parallel=True, run_dir=None, **kwargs
 ):
     if success_exit:
+        # Should sim_type and sim_id be available in kwargs?
```
robnagler, line 639: I think these should be available somehow. The lack of an object is a problem. You shouldn't have to reverse engineer run_dir. That breaks the abstraction. You could fix job_cmd to pass msg.whatever-you-need as keyword args.

sirepo/template/srw.py
```diff
@@ -2182,12 +2219,24 @@ def _rotate_report(report, ar2d, x_range, y_range, info):
...
 
+def _export_rsopt_files():
+    files = PKDict()
+    for t in (
+        "py",
+        "sh",
+        "yml",
+    ):
+        files[f"{t}FileName"] = f"{_SIM_DATA.EXPORT_RSOPT}.{t}"
+    files["postProcFileName"] = f"{_SIM_DATA.EXPORT_RSOPT}_post.py"
```
robnagler, line 2230: use dot notation.

robnagler (changes requested):

sirepo/template/srw.py
mkeilman: [8ba900d](https://github.com/radiasoft/sirepo/pull/5216/commits/8ba900dc4ee272a5f02eb0b7a977a8421c31a9f4)

sirepo/package_data/static/js/sirepo.js
mkeilman: I considered that but opted for "expediency". Since this copies the changes in #5270, could we do it there instead, just to move this forward?

robnagler: Yes.

sirepo/template/srw.py
mkeilman: [631df7d](https://github.com/radiasoft/sirepo/pull/5216/commits/631df7def4a23af8aadaf4a6729506c730a39a66)
Something like that?

sirepo/template/srw.py
```diff
@@ -636,13 +636,9 @@ def post_execution_processing(
     success_exit=True, is_parallel=True, run_dir=None, **kwargs
 ):
     if success_exit:
-        # Should sim_type and sim_id be available in kwargs?
-        if _SIM_DATA.ML_REPORT in str(run_dir):
+        if _SIM_DATA.is_for_ml(kwargs.get("compute_model")):
```
robnagler, line 639: Here define `compute_model` as an argument, since it is a required argument. This code is not fail fast, which it should be. The reason for kwargs is to allow dispatched routines which do not use new parameters to continue to operate when new parameters are added to the interface. This future proofs the dispatchee's from changes to the dispatcher.

robnagler:
@mkeilman ping

sirepo/template/srw.py
mkeilman: I presume a named argument (`compute_model=None...assert(compute_model)`) or all the other templates have to change. Also, same treatment for sim_id?

robnagler: You don't need to add compute_model=None when all the other implementations of post_execution_processing have **kwargs. That's what kwargs is for. However, srw.post_execution_processing(success_exit, compute_model, ...., **kwargs) will fail if compute_model is not in kwargs at the dispatch point (in job_cmd). This way you can have a required parameter in some dispatchees that verifies that the dispatcher is supplying a specific result. You should not have:

def post_execution_processing(
    success_exit=True, is_parallel=True, run_dir=None, **kwargs
):

because these are *required* parameters and we want fast fail.

It's important to think through this: what does a dispatcher provide? What does a dispatched call expect? How does this interact with kwargs and named parameters?

mkeilman: I did not know you could pull named parameters out of kwargs to use as positionals - I thought they were mutually exclusive. You can't go the other way and _call_ a  function with a positional and kwargs that includes that same name. So this should be
```python
def post_execution_processing(
    compute_model, sim_id, success_exit, **kwargs,
):
```
If that's right, should I change all the other templates, or limit the scope?

robnagler: Yes, remove the defaults as they are incorrect, but you can leave the params as they are used.

robnagler: > I did not know you could pull named parameters out of kwargs to use as positionals - I thought they were mutually exclusive. You can't go the other way and call a function with a positional and kwargs that includes that same name. 

Yes the implementation is correct. I think the language needs to be clear.

The word "positional" is only relevant when talking about [arguments](https://docs.python.org/3/glossary.html#term-argument), that is, the values passed to a call of a function. There are no positional [parameters](https://docs.python.org/3/glossary.html#term-parameter). All parameters have a position in a definition and a name.

The parameter `**<anything>` (commonly called `kwargs`) collects any remaining unnamed *arguments*. And `*<anything>` (commonly `args`) collects all positional arguments that are mapped to named parameters.

mkeilman:
Will incorporate #5270 changes when ready

mkeilman:
ping - I think I have everything covered

sirepo/template/activait.py
```diff
@@ -175,6 +175,23 @@ def get_fft_report(run_dir, data):
     return w, plots, f"FFT", summaryData
 
 
+def new_simulation(data, new_sim_data, qcall, **kwargs):
+    if "sourceSimType" not in new_sim_data:
+        return
+    t_basename = f"{new_sim_data.sourceSimType}-{new_sim_data.sourceSimId}-{new_sim_data.sourceSimFile}"
```
robnagler, line 181: Remove the temporary. If you feel the need for a temporary, then use `f = data.models.dataFile`.

sirepo/template/activait.py
```diff
@@ -175,6 +175,23 @@ def get_fft_report(run_dir, data):
...
+    data.models.dataFile.dataOrigin = "file"
+    data.models.dataFile.file = t_basename
+    t = simulation_db.simulation_lib_dir(_SIM_DATA.sim_type(), qcall=qcall).join(
+        _SIM_DATA.lib_file_name_with_model_field("dataFile", "file", t_basename)
+    )
+    if t.exists():
+        return
+    s = simulation_db.simulation_dir(
```
robnagler, line 189: remove temporary

sirepo/template/activait.py
```diff
@@ -175,6 +175,23 @@ def get_fft_report(run_dir, data):
...
+        new_sim_data.sourceSimType, sid=new_sim_data.sourceSimId, qcall=qcall
+    ).join(new_sim_data.sourceSimFile)
+    t.mksymlinkto(s, absolute=False)
```
robnagler, line 192: You should note that this is hack. `#TODO(mkeilman) Only works on local installation, not sbatch/NERSC` or something like that.

mkeilman: Whoops, Paul was too fast.  I'll clean these up along the way.

PR #5218, opened 2022-11-30
Fix #5044 all getApplicationData remove & compute return files

### LINKED ISSUE #5044: 20221201 get_application_data translated to statefulCompute
get_application_data needs to happen in an agent. It should work because the entire user's directory is mapped in the agent so all is available. This will need to be moved to job_api so that it works without changing all the uses.

robnagler:
- getApplicationData was removed
- statefulCompute, statelessCompute, and analysisJob can return files
- srw samplePreview is statefulCompute

### DIFF
--- a/sirepo/http_reply.py
+++ b/sirepo/http_reply.py
@@ -87,16 +87,15 @@ def gen_file_as_attachment(qcall, content_or_path, filename=None, content_type=N
         filename = content_or_path.basename
     filename = re.sub(r"[^\w\.]+", "-", filename).strip("-")
     if re.search(r"^\.\w+$", filename):
-        # the safe filename has no basename, use "download" instead
+        # the safe filename has no basename, prefix with "download"
         filename = "download" + filename
-    if content_type is None:
-        content_type, _ = mimetypes.guess_type(filename)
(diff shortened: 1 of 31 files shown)


moellep:
In the madx app - if you select the "View MAD-X log" in the Simulation Status panel on the Visualization page, it causes a server error. If you select the Download "Raw Data File" from any Visualization report, you get a server error.

moellep:
Discard Changes to Example fails with a javascript error "requestSender.getApplicationData is not a function". It looks like there is a special hack in there related to SRW appDataService.appDataForReset()

moellep:
The appDataForReset could be removed - it is trying to clean up the user defined beam and undulator models when you revert the sim. That isn't important. The appDataService.appDataForReset() and calls to getApplicationData in resetSimulationModal should be removed.

robnagler:
madx download log test added and fixed code. remove appDataForReset

robnagler:
@moellep can't re-request a review from the github gui since first review is still pending. next time add a review instead of commenting in main thread. please review, thanks.

moellep:
I've retested this PR and everything works now. It can be merged.

PR #5231, opened 2022-12-05
fix radiasoft/raydata#39: History of 5 most recently executed runs in Queue Tab

rorour:


rorour:
@e-carlin Side note: I'm not loving the loading indicators. You were right about the constant appearing/disappearing being distracting. Having two scan tables on the same page really highlights this. I have some more ideas, I'm still partial to doing something with a timestamp similar to what I [showed here](https://github.com/radiasoft/sirepo/pull/4977#discussion_r1006266304). But maybe even subtler. And/or changing the text color to a lighter gray. Let me know your thoughts.

sirepo/package_data/static/html/raydata-analysis-queue.html
```diff
@@ -1,5 +1,6 @@
 <div class="container-fluid">
   <div class="col-md-12">
     <div data-basic-editor-panel="" data-view-name="analysisQueue"></div>
+    <div data-basic-editor-panel="" data-view-name="historyScans"></div>
```
e-carlin, line 4: Any reason why this is named historyScans? I think a name like recentlyExecutedScans would match up better with what the model represents and would be easier to track through the code (model name is similar to report title). Likewise it would look more similar to our other models representing similar data (executedScans and queuedScans)

sirepo/package_data/static/js/raydata.js
```diff
@@ -531,6 +534,9 @@ SIREPO.app.directive('scansTable', function() {
                 $scope.reverseSortScans = ! $scope.reverseSortScans;
             };
 
+            if ($scope.analysisStatus === 'history') {
```
e-carlin, line 537: I'd rather wait for a user to ask for this feature. We are only displaying 5 of the most recent scans so as long as they are in some sorted order (they are) the user should be able to look at all of them and grasp what is going on. If they want to change the sort order then they can click on one of the column headers and change the sort.

e-carlin (changes requested):
Couple nits. One bug, when I click on the scan in the "recently executed" report the modal pops up but the loading wheel just spins forever. If I click on the same scan in the "executed scans" report I get back the images.

sirepo/package_data/static/html/raydata-analysis-queue.html
rorour: 71d7b187ccf540a8b22a16f86cc3ece57952d92d

sirepo/package_data/static/js/raydata.js
rorour: 71d7b187ccf540a8b22a16f86cc3ece57952d92d

rorour:
@e-carlin Thanks for finding that bug with the analysis images. It was due to having 2 analysis modals on the same page with the same id. Fixed 9d2b8a84d5b6f28d1e6f31cd8f5c2e884785cf34

e-carlin (commented):
Code looks good. One bug, when I click to add a column on the recently executed table the "save changes" button appears in the analysis queue table and column is added to that table. If I refresh the page then it also appears in the recently executed table. It should appear in both when I added. Also, deleting a column deletes it from one table but not the other until the page is refreshed. Ideally these actions should update all tables.

e-carlin (approved):
Approved. The bug with columns will be addressed in https://github.com/radiasoft/raydata/issues/48

PR #5241, opened 2022-12-06
fix #5164: implement warppba in react

### LINKED ISSUE #5164: 20221214 WarpPBA GUI with React

garsuga:


react/src/data/wrapper.ts
```diff
@@ -52,12 +51,19 @@ export class FormStateWrapper extends AbstractModelsWrapper<FormModelState, Form
...
+            throw new Error("model could not be hooked because it was not found: " + modelName);
+        }
+        return m;
     }
 
     getFieldFromModel(fieldName: string, model: FormModelState): FormFieldState<unknown> {
-        return model[fieldName];
+        let fv = model[fieldName];
+        if(fv === undefined || fv === null) {
+            throw new Error(`field could not be found in model state: ${fieldName}, ${JSON.stringify(model)}`)
```
robnagler, line 64: We are tending towards errors where names are bound to values: `field=${fieldName} could not be found in model=${JSON.stringify(model)}`

react/src/data/wrapper.ts
```diff
@@ -94,21 +100,29 @@ export class ModelsWrapper extends AbstractModelsWrapper<ModelState, unknown> {
...
+            throw new Error("model could not be hooked because it was not found: " + modelName);
+        }
+        return m;
     }
 
     getFieldFromModel(fieldName: string, model: ModelState): unknown {
-        return model[fieldName];
+        let fv = model[fieldName];
+        if(fv === undefined || fv === null) {
+            throw new Error(`field could not be found in model state: ${fieldName}, ${JSON.stringify(model)}`)
```
robnagler, line 113: Is there a way to modularize this so that you don't have to write so much to describe the error?

react/src/layout/report.tsx
```diff
@@ -351,6 +360,11 @@ export class SimulationStartLayout extends Layout<SimulationStartConfig, {}> {
         let endSimulationButton = <Button variant="primary" onClick={endSimulation}>End Simulation</Button>;
```
robnagler, line 360: Is "primary" a concept we defined? If so, what are the choices? What happens if there's a typo?

The text `End Simulation` should come from somewhere. Sometimes it is `End Simulation` and other times it is something else, e.g. in Activait. Embedding strings throughout the app is going to be problematic, because they get repeated.

react/src/layout/report.tsx
```diff
@@ -351,6 +360,11 @@ export class SimulationStartLayout extends Layout<SimulationStartConfig, {}> {
...
         let startSimulationButton = <Button variant="primary" onClick={startSimulation}>Start Simulation</Button>
```
robnagler, line 361: Does it make sense to make a PrimaryButton that understands how to map startSimulation to `Start Simulation`?

react/src/layout/report.tsx
```diff
@@ -377,6 +391,7 @@ export class SimulationStartLayout extends Layout<SimulationStartConfig, {}> {
                             <Stack gap={2}>
                                 <span>{'Simulation Completed'}</span>
                                 <span>{`Elapsed time: ${elapsedTimeSeconds} ${'second' + (elapsedTimeSeconds !== 1 ? 's' : '')}`}</span>
```
robnagler, line 393: The good old "plural function". Again it seems like it would be better to modularize this. For something that runs for a few days, displaying n*86400 seconds is not very useful. This will come up so we'll want to be able to switch formats based on the data.

react/src/data/accessor.ts
```diff
@@ -12,6 +12,16 @@ export class ModelsAccessor<M, F> {
                 modelsWrapper.hookModel(modelName)
             ]
         }))
+
+        // expand wildcard
+        this.dependencies = this.dependencies.flatMap(d => {
+            if(d.fieldName === '*') {
```
robnagler, line 18: Magic syntax like this is hard to find. What about having a keyword, e.g. _ALL_. This will make explicitly coupling a bit easier and make the app less rigid.

robnagler (changes requested):

react/src/layout/report.tsx
e-carlin: In the current schema we have "strings" which get used to create these names. Ex for raydata https://github.com/radiasoft/sirepo/blob/master/sirepo/package_data/static/json/raydata-schema.json#L68

react/src/data/wrapper.ts
```diff
@@ -52,12 +51,19 @@ export class FormStateWrapper extends AbstractModelsWrapper<FormModelState, Form
...
+        let m = useSelector(this.formSelectors.selectFormState(modelName));
+        if(m === undefined || m === null) {
+            throw new Error("model could not be hooked because it was not found: " + modelName);
+        }
+        return m;
     }
 
     getFieldFromModel(fieldName: string, model: FormModelState): FormFieldState<unknown> {
-        return model[fieldName];
+        let fv = model[fieldName];
```
e-carlin, line 62: const and single letter

react/src/data/wrapper.ts
```diff
@@ -94,21 +100,29 @@ export class ModelsWrapper extends AbstractModelsWrapper<ModelState, unknown> {
...
+        let m = useSelector(this.modelSelectors.selectModel(modelName));
+        if(m === undefined || m === null) {
+            throw new Error("model could not be hooked because it was not found: " + modelName);
+        }
+        return m;
     }
 
     getFieldFromModel(fieldName: string, model: ModelState): unknown {
-        return model[fieldName];
+        let fv = model[fieldName];
```
e-carlin, line 111: const and single letter

react/src/layout/report.tsx
```diff
@@ -261,12 +261,21 @@ export function ReportAnimationController(props: { animationReader: AnimationRea
...
+    items: SchemaLayout[]
 }
 
 export class SimulationStartLayout extends Layout<SimulationStartConfig, {}> {
+    childLayouts: Layout[];
+    
+    constructor(config: SimulationStartConfig) {
+        super(config);
+
+        this.childLayouts = (config.items || []).map(LAYOUTS.getLayoutForSchema);
```
e-carlin, line 274: can SimulationStartConfig items be given a default value of [] so we don't have to have this `|| []` in react?

e-carlin (commented):
I was interested in looking at some of the react stuff so I gave it a review. I stopped commenting on const and single letter variable names but there are many places they should be addressed.

react/src/data/wrapper.ts
garsuga: I am not sure. Writing error logging in JS is sort of difficult and we should probably use a library. Other languages make getting the relative context for a line of code easier than what I can find in JS. Besides that, I do not know what would be better to describe this without just writing it out. If I have some error message like "key not found in object" it doesn't really help anyone besides actually failing fast because it doesn't describe anything.

react/src/layout/report.tsx
garsuga: Primary is a concept defined by Bootstrap. While I agree that we should not have too many strings floating around in our app there is no benefit in duplicating these definitions inside the schema. If we want to make this a parameter for the layout within the schema that is fine. The strings area of the schema should be used for things that potentially change but primary is just the style for the button. This is more suited as a global for the React app if anything. I believe primary is actually the default here though, so for now I would personally lean towards leaving it there. These are reusable components and their definitions will not be duplicated elsewhere so I think consistency is the most important thing here. All simulation start layouts will have the same kind of button and primary happens to be the name of it.

react/src/layout/report.tsx
garsuga: Perhaps, but the definition would be `(handler, text) => <Button variant="primary" onClick={handler}>{text}</Button>` and honestly the 'End Simulation' button should probably be `variant='danger'` since it ends something irreversibly. We can always add more abstraction but it just introduces more names and more digging in cases like these. They aren't conceptually the same here. The buttons are very different and just happen to have similar definitions here but I can see them diverging pretty quickly. Soon all that will be left of the abstraction is a function that looks like `let btn = (props, text) => <Button {...props}>{text}</Button>`. Having two almost repeated lines is not enough to justify making a function because it couples these two buttons that are not what they are for the same reasons (like how ending the simulation should probably be a red button with the danger variant).

react/src/layout/report.tsx
garsuga: Yes, I agree. This was a band-aid approach and I could really use any info you have on making things plural. English sucks so I wonder if if would be any better than `(n > 1) ? "seconds" : "second"`. Should this just be a function like `let plural = (n, plural, singular) => (n > 1) ? plural : singular`? Either way I need to make the time handling better here. Hours and minutes and the whole thing. Will update that on this PR.

garsuga: Is `(n > 1) => "seconds" : "second"` any worse than `plural(n, "seconds", "second")`? I feel like I am barely missing the path where we can abstract this. All I would be abstracting here is the condition of `n > 1`. It feels just as complex in either case. You still need to know the order of the params but you just replace `n > 1` with the name `plural` and add some import in the process.

react/src/data/accessor.ts
garsuga: So you want `model.ALL` vs `model.*`?

garsuga: My only thought after a second is that technically we could have a field name collide with our magic syntax here. Any field named `all` would fail and to prevent a very confusing error we would need a check for this in an unrelated part of the app.

react/src/data/accessor.ts
```diff
@@ -12,6 +12,16 @@ export class ModelsAccessor<M, F> {
...
+                let fieldNames = Object.keys(this.modelValues[d.modelName]);
```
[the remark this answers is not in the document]
garsuga: Yes, please point these out. As I re-wrote things I often ended up with leftover variables. Ill remove this.

react/src/layout/report.tsx
garsuga: It can but this would be just as long and would be placed on the line above. Types cannot have default values since they aren't classes and the objects following them aren't derived from the type, they just follow them. I could place `config.items = config.items || []` on line 273 but since this is a mutable type I would also need to do `config = {...config}` to create a copy first on line 272. Easier to work with the data we have instead of mutating it first thing.

react/src/data/wrapper.ts
robnagler: In the angular app we send the stack to the server along with the error. This would definitely be useful. Is there an error handler in React similar [to AngularJS's](https://github.com/radiasoft/sirepo/blob/355d5e66fb769dfe51a4f3406409ba163ef7893f/sirepo/package_data/static/js/sirepo.js#L2972)?

react/src/layout/report.tsx
robnagler: The signal to noise ratio of this line is high. There's a lot of syntax and the words button and simulation are repeated three times:

`let endSimulationButton = <Button variant="primary" onClick={endSimulation}>End Simulation</Button>;`

Our minds have a harder time with differences (what's the meaning here) when there is a lot of repetition.

The danger I see of `"primary"` is that it is interpreted at run-time(?), and if it is CSS, it would not fail fast if there was a typo. Is that true?

> All simulation start layouts will have the same kind of button and primary happens to be the name of it.

This suggests a default override. There are buttons with danger and primary, and the default is primary. The thing that defines layouts should default to primary (written once in a "parent" of some sort) and the override would happen when we needed it.

react/src/layout/report.tsx
robnagler: > Perhaps, but the definition would be (handler, text) => <Button variant="primary" onClick={handler}>{text}</Button>

I was thinking of something more like this:

```jsx
(handler) => <Button variant="primary" onClick={handler}>{labelFor(handler)}</Button>
```

This has many advantages including i18n and for us consistency of names.

> HTML and JSX are long because there is a lot of information. 

Actually, there is less information in a typical HTML and JSX doc vs say YAML. (A gzip on the file will prove this.) This is why YAML is easier to read (and thusly more popular) than XML for defining state.  

Going to the code, if there was a function called `button` which accepted handler, this code could go away:

```jsx
        let endSimulationButton = <Button variant="primary" onClick={endSimulation}>End Simulation</Button>;
        let startSimulationButton = <Button variant="primary" onClick={startSimulation}>Start Simulation</Button>
```

And, the use would look like:

```jsx
                        return (
                            <Stack gap={2}>
                                <span>{'Simulation Completed'}</span>
                                <span>{`Elapsed time: ${elapsedTimeSeconds} ${'second' + (elapsedTimeSeconds !== 1 ? 's' : '')}`}</span>
                                <div>{children}</div>
                                {button(startSimulation)}
                            </Stack>
                        )
```

The abstraction is clear and contains the minimum information necessary to describe what is going to happen.

react/src/layout/report.tsx
robnagler: I was thinking something [like this](https://github.com/plurals/pluralize/blob/master/pluralize.js). Not necessarily this library, but something that is an abstraction for taking the word and pluralizing. Again, there's an argument for modularizing the strings from tables with a bit of business logic.

react/src/data/accessor.ts
robnagler: My bad I edited my comment `_ALL_` or some variant that is searchable that would not collide with field names. This would be pretty easy, since we already restrict field name syntax. In the future, someone will want to search for the concept of `*`, which is very hard to find. Therefore, it's better to have a longer string that is globally unique and perhaps self documenting (even better `_ALL_FIELDS_`) for cases like this.

react/src/layout/report.tsx
robnagler: > Types cannot have default values since they aren't classes 

This is concerning. There's power in defaults including future proofing code. Is there no way around this?

e-carlin: >>Types cannot have default values since they aren't classes

>This is concerning. There's power in defaults including future proofing code. Is there no way around this?

Is there a reason to use a type here instead of a class?

react/src/layout/report.tsx
garsuga: > The danger I see of "primary" is that it is interpreted at run-time(?), and if it is CSS, it would not fail fast if there was a typo. Is that true?

Since it is typescript it would raise an error while compiling. The definition for `variant` probably looks like `let type ButtonVariant = "primary" | "secondary" | "danger" //... etc`. There are checks we can use to prevent typos for strings like this. Adding them to some constant like `BUTTON_PRIMARY_NAME = "primary"` would feel redundant and if your intended approach is something like an enum then the type definition above would be an example of how enums are created in typescript.

react/src/layout/report.tsx
garsuga: > Going to the code, if there was a function called button which accepted handler, this code could go away:

I agree that being able to use localization would be nice but that could be handled better than coupling it with the concept of a button and handler. The buttons are also not necessarily going to be updated together. My main point was that they are not two similar lines of code because they are conceptually the same. The end simulation button should probably have a different variant defined for it anyways. After you add this as an input to the button function it would just be the same as replacing `<Button>` with `button()` and add another function developers need to know and use. It doesn't save any complexity in the long run because those two lines might not be updated together.

react/src/layout/report.tsx
garsuga: > Is there a reason to use a type here instead of a class?

Yes, one of the big advantages of duck typing over classes is that we don't need to keep a dictionary of classes we should be using while reading in data. We could do it either way but I tend to avoid complicated approaches like this for things like defaults. Typescript also lets us mark which of the values could possibly be null with a ?.

react/src/data/wrapper.ts
garsuga: There doesn't seem to be an error handler within React for logging. There are ways we can handle errors during rendering but there isn't much I can find for logging using the React library. There are plenty of logging libraries though so we would just need to pick one.

react/src/data/wrapper.ts
garsuga: About every variable in the program could be a const. If that is what you and @robnagler want I can go around and change them all.

react/src/data/wrapper.ts
robnagler: Please create an issue for this. We need to get the errors server side. Otherwise, they are lost.

react/src/layout/report.tsx
robnagler: > My main point was that they are not two similar lines of code because they are conceptually the same. 

Conception is in the eye of the beholder. I see them as a the same. 

Buttons in our system (not React's) have variants, handlers, and labels as I see it. A label is a concept we own, which should be centralized so that when a button (or something else) needs a label, it gets it from somewhere which manages labels. We already modularize this to a great extent. This is a basic principle of UI architecture: separating strings from formatting. It's very easy to mix the two, but it is poor separation of responsibilities to mix the two. This is why we want to avoid any strings like this in the JSX. It should all be looked up in a string table.

Buttons should have names and things like variants should be looked up as well. It's easier to reason about variants in a single table (start is primary, end danger, cancel is danger, etc.) This allows clearer grouping.

Once a button has a name, it is easy to use it to index these various tables including handlers, possibly.

react/src/layout/report.tsx
e-carlin: > one of the big advantages of duck typing over classes is that we don't need to keep a dictionary of classes we should be using while reading in data

Can you elaborate more on that? Why/where would we need to keep a dictionary of classes? 

I don't feel like this is really duck typing either. If config could quack it would have items initialized.

> We could do it either way but I tend to avoid complicated approaches like this for things like defaults

My concern with this bit of code is that it is a pattern we have in our angularjs app that, while simple when looking at one case, becomes complex when we look at all cases. Every caller is going to have to do this check. By distributing that confusion around the code things get more complex. Obviously, in this case it only happens here. But, people copy these patterns and code tends to get used more and more as time goes on. If we solve the problem once for everything then all aspects of the code benefit.

PR #5251, opened 2022-12-08
Fix #5151 LDAP auth method for Sirepo

### LINKED ISSUE #5151: 20221202 LDAP: Create a new auth method (4 days)
Create a new sirepo auth method using ldap. This will involve adding a module and required methods to sirepo/auth. As well as the GUI features to support it.

Benbenbenin0:
Authenticates username/password, need to revisit approach to multi-factor.


e-carlin (changes requested):
Looking good. Some changes needed. We reviewed in person.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,44 @@
...
+    def api_authLdapAuthorized(self):
+        """Python-LDAP client authentication
+        Bind to server and authenticate
+        """
+        # Test User Credentials:
+        # username: Archimedes of Syracuse
+        # password: eureka
+
+        LDAP_SERVER = "ldap://10.10.10.10:389"
+        BASE_DN = "dc=example,dc=com"
```
robnagler, line 27: This must be configured with the default for dev being example and com. Did you look at the customer's example app? It may effect the object to search which is likely "user".

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,44 @@
...
+        OBJECT_TO_SEARCH = "objectClass=inetOrgPerson"
+        ATTRIBUTES_TO_SEARCH = ["uid"]
+
+        req = self.parse_post(type=False)
+
+        # initialize our connect to ldap-server
+        connect = ldap.initialize(LDAP_SERVER)
```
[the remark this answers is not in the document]
robnagler: Probably should be an object or cfg.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # bind to server and login
+        connect.simple_bind_s(
+            self._get_login(req.req_data.username), req.req_data.password
+        )
+
+        return PKDict()
+
+    def _get_login(self, username):
+        return "cn=" + username + ",ou=users,dc=example,dc=com"
```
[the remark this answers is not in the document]
robnagler: Syntax here should be guided by AD. It's likely it will be an email.

robnagler (changes requested):

sirepo/auth/ldap.py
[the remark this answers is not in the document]
robnagler: > Are there injection attacks that can happen with this?

No. This is an argument to something that validates it. However, we wouldn't want to accept a 1GB user name or something so we should protect the input as with all inputs. We expect it to be characterized a particular way (e.g. less than 256 chars).

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,44 @@
...
+AUTH_METHOD_VISIBLE = True
+
+
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapAuthorized(self):
+        """Python-LDAP client authentication
+        Bind to server and authenticate
+        """
+        # Test User Credentials:
```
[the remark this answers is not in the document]
Benbenbenin0: added test user vagrant|vagrant

sirepo/auth/ldap.py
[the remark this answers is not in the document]
Benbenbenin0: > Syntax here should be guided by AD. It's likely it will be an email.

@robnagler what do you mean by AD? The syntax now assumes the login dn has an email.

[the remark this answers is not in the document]
robnagler: AD = Active Directory

> The syntax now assumes the login dn has an email.

I don't see this code. In any event, LDAP requires [escaping special characters](https://www.ibm.com/docs/en/sva/9.0.2.1?topic=tasks-valid-characters-ldap-user-group-names).

robnagler (changes requested):

sirepo/auth/ldap.py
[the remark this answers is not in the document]
Benbenbenin0: Added front end warning when entering email/pass over 256 char, turned _get_login into _get_escaped and tested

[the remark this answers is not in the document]
robnagler: All web APIs should protect themselves. It's nice that the GUI let's the user know of the error, but the real problem is if someone hacks the API and submits invalid text.

sirepo/auth/ldap.py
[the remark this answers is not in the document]
Benbenbenin0: I understand putting the server address in the cfg so it can be changed, @robnagler are you saying to initialize the server in the cfg as well (c = ldap.initialize(LDAP_SERVER))?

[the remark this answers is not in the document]
robnagler: This is module level initialization for pkconfig. It would look something like:

```py
cfg = pykern.pkconfig.init(
    domain=pykern.pkconfig.RequiredUnlessDev("example.com", str, "LDAP domain"),
)
```

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,59 @@
...
+export PYKERN_PKDEBUG_WANT_PID_TIME=1
+
+# primary cause of failed wheel generation
+pip install --upgrade pip
+echo "upgraded pip"
+
+# install prerequisites
+sudo yum groupinstall "Development tools"
+sudo yum install openldap-devel python-devel
+echo "installed prerequisites"
```
robnagler, line 19: These types of echos are not useful. In general, things should be silent unless they fail.

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# secondary cause of failed wheel generation
+pip install --upgrade setuptools wheel
```
robnagler, line 22: Do not do this here. This is normal development, not specific to ldap.

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "installed wheel setup"
+
+# main installs
+pip install python-ldap
```
robnagler, line 26: Do not install this.

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,59 @@
...
+sudo yum -y install openldap-clients openldap-servers
+echo "installed ldap"
+
+sudo systemctl start slapd
```
robnagler, line 30: I'm not sure we want to enable slapd, start should be good enough. We can talk about making this script idempotent later.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
+"""LDAP login
+:copyright: Copyright (c) 2018-2022 RadiaSoft LLC.  All Rights Reserved.
```
robnagler, line 2: LDAP authentication support.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+
+        c = ldap.initialize(_cfg.ldap_server)
+
+        if(len(u) > 256 or len(p) > 256):
+            raise Exception("login user/pass greater than 256 chars")
+
+        c.simple_bind_s(self._get_escaped(u), p)
+        
+        raise sirepo.util.Redirect(sirepo.uri.local_route(req.req_data.simulation_type))
+        return PKDict()
```
robnagler, line 34: dead code.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+
+AUTH_METHOD_VISIBLE = True
+
+_cfg = None
+
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapAuthorized(self):
+
+        # TODO(BG): pass simulation type as method parameter to assign to 'type', currently passing simulation_type to req and leaving 'type' False
```
robnagler, line 21: Simulation_type is fine in parse_post. Just rename the post value to be simulationType, and it will be parsed automatically.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+        p = req.req_data.password
+
+        c = ldap.initialize(_cfg.ldap_server)
+
+        if(len(u) > 256 or len(p) > 256):
+            raise Exception("login user/pass greater than 256 chars")
+
+        c.simple_bind_s(self._get_escaped(u), p)
+        
+        raise sirepo.util.Redirect(sirepo.uri.local_route(req.req_data.simulation_type))
```
robnagler, line 33: This seems incorrect. It probably needs to be a call to auth.login as auth.email does.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+
+AUTH_METHOD = "ldap"
+
+AUTH_METHOD_VISIBLE = True
+
+_cfg = None
+
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapAuthorized(self):
```
robnagler, line 19: I don't think this is "authorized". This is not like oauth. Rather, it's more like authEmailLogin. This does the authorization, but it's all in one function, from the GUI perspective it is asking to login.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # TODO(BG): pass simulation type as method parameter to assign to 'type', currently passing simulation_type to req and leaving 'type' False
+        req = self.parse_post(type=False)
+        u = req.req_data.username
```
robnagler, line 23: temporaries are unnecessary.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+        p = req.req_data.password
+
```
robnagler, line 25: whitespace in general is unnecessary.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,56 @@
...
+        c = ldap.initialize(_cfg.ldap_server)
+
+        if(len(u) > 256 or len(p) > 256):
+            raise Exception("login user/pass greater than 256 chars")
```
robnagler, line 29: Never raise Exception. It's useless to the end user. Rather make this an AssertionError. It's probably a program error.

robnagler (changes requested):
An initial request. We can talk through some of these things.

sirepo/auth/ldap.py
Benbenbenin0: Changing to AssertionError.

I read that assertions stop executing the program when raised and exceptions do not. Is this correct and why exceptions are not usefull/fail fast?

sirepo/auth/ldap.py
Benbenbenin0: Changing to authEmailLogin.

Would it be useful to seperate the authorization into its own method?

sirepo/auth/ldap.py
Benbenbenin0: Changing to auth.login, currently always registers a new user.

Want to verify I understand, to implement this properly I need to first search auth_db for a user "model" (by LDAP email?) to pass to auth.login, logging in with that user or registering a new user if None?

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+from pykern.pkdebug import pkdlog, pkdp
+from pykern.pkcollections import PKDict
+import ldap
+import sirepo.quest
+
+AUTH_METHOD = "ldap"
+
+#: User can see it
+AUTH_METHOD_VISIBLE = True
```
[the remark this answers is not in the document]
robnagler: It is needed.

sirepo/auth/ldap.py
robnagler: AssertionError is a subclass of Exception. All exceptions will stop a program if not caught. This will be caught by uri_router and turned into a ServerError reply.

sirepo/auth/ldap.py
robnagler: It's not "authEmailLogin" It's authLdapLogin.

sirepo/auth/ldap.py
robnagler: This is a bit tricky. We do not necessarily want yet another table for users (auth_ldap_user_t), but I see no way around this. Perhaps it's ok, but there's code [that assumes METHOD_EMAIL](https://github.com/radiasoft/sirepo/blob/219a97bed9b610b27b12fc67a1058b3a0f3ef7e3/sirepo/auth_role_moderation.py#L82), which is (maybe) problematic. ldap could reuse email's model, because [there's other code](https://github.com/radiasoft/sirepo/blob/219a97bed9b610b27b12fc67a1058b3a0f3ef7e3/sirepo/pkcli/jupyterhublogin.py#L35) which relies on AuthEmailUser having emails. It's going to be tricky, but you could set ldap.UserModel = "AuthEmailUser".

I think you should authenticate with ldap to avoid the case where the user types the wrong email address. Once authenticated, you can then check to see if there's a user in auth_email_user_t. If there is, then you can use that uid to login (like api_authEmailLogin does).

Try this first.

etc/run-auth-ldap.sh
Benbenbenin0: Meaning install this in normal dev instead of in run-auth? The login API is written with python-ldap.

robnagler: python-ldap installs the libraries in /usr where the "system" (red hat) python is installed. The python we use is in `~/.pyenv` which doesn't look in /usr for python libraries. What you want to do is probably "pip install ldap"

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,59 @@
...
+export SIREPO_AUTH_METHODS='email:guest:ldap'
+export PYKERN_PKDEBUG_WANT_PID_TIME=1
+
+# primary cause of failed wheel generation
+pip install --upgrade pip
+echo "upgraded pip"
+
+# install prerequisites
+sudo yum groupinstall "Development tools"
+sudo yum install openldap-devel python-devel
```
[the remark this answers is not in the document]
e-carlin: I believe openldap-devel is needed. I get the following error when trying to install python-ldap without it

```
      gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DHAVE_SASL -DHAVE_TLS -DLDAPMODULE_VERSION=3.4.3 "-DLDAPMODULE_AUTHOR\
=python-ldap project" "-DLDAPMODULE_LICENSE=Python style" -IModules -I/home/vagrant/.pyenv/versions/py3/include -I/home/vagrant/.pyenv/versions/3.7.2/include/p\
ython3.7m -c Modules/LDAPObject.c -o build/temp.linux-x86_64-cpython-37/Modules/LDAPObject.o
      In file included from Modules/LDAPObject.c:3:
      Modules/common.h:15:10: fatal error: lber.h: No such file or directory
         15 | #include <lber.h>
            |          ^~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for python-ldap
Failed to build python-ldap
```

etc/run-auth-ldap.sh
e-carlin: I believe python-ldap is the correct package. 

`pip install ldap` won't work. See the docs https://pypi.org/project/ldap/ .

For me it seems `pip install python-ldap` installed in the right place
```
~$ find ~/.pyenv/ -name 'ldap'
~$ pip install python-ldap
<snip>
Successfully installed python-ldap-3.4.3
~$ find ~/.pyenv/ -name 'ldap'
/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/ldap
```

etc/run-auth-ldap.sh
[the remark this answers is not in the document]
robnagler: openldap-devel is fine. It's a bug in python-ldap that it doesn't require openldap-devel, but that's not our issue.

etc/run-auth-ldap.sh
robnagler: I'm sorry, I was confused. I thought "python-ldap" was a yum install. pip install is correct. ignore my comment.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapLogin(self):
+
+        # validates username/password, escapes special chars for ldap dn
+        def _validate_escape_credentials(req):
+            email = req.req_data.email.lower()
+            if len(email) > 256 or len(req.req_data.password) > 256:
```
robnagler, line 38: This is actually more complicated than 256 ([really 255](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc756101(v=ws.10)?redirectedfrom=MSDN#name-length-limitations-for-ldap-simple-bind-operations)). You will be adding base_dn. Moreover, the AD limits are a bit strange, and I think passwords [are limited to 127 characters](https://learn.microsoft.com/en-us/previous-versions/tn-archive/cc512606(v=technet.10)?redirectedfrom=MSDN) due to limitations in the Windows login dialogue. Technically, LDAP itself has no limits.

I would be more conservative here just because you don't want to deal with weirdnesses in the server, e.g. truncating passwords before hashing. Define a constant of MAX_ENTRY to be 127, and use that everywhere. 

You can check base_dn against this number as well and thrown an error in the config code

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+#: Well known alias for auth
+user_model = "AuthEmailUser"
+
+
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapLogin(self):
+
+        # validates username/password, escapes special chars for ldap dn
+        def _validate_escape_credentials(req):
```
robnagler, line 36: Functions are sorted alphabetically, put this after `_user`. The name is awkward. If you want something shorter, use `_validate_and_escape_creds`. The word `creds` in this context (nested function) is fine, because someone reading this will know what `creds` means here. For a public method, I would used credentials, but below I used `creds` for the object it returns.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+                    "invalid user and/or password",
+                    f"email={email} or password is none/zero length",
+                )
+            return (
+                "mail=" + re.sub(_ESCAPE_DN_MAIL, r"\\\1", email) + _cfg.base_dn,
+                email,
+                req.req_data.password,
+            )
+
+        def _authorize_ldap(dn, email, p):
```
robnagler, line 54: `_authorize_ldap` is redundant. Call this `_bind`, since it is an internal function that is describing what it is doing.

The parameter `p` is too short. Call it password. However, you will instead call a single argument here `creds` which is returned from `_validate_and_escape_creds`.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+            try:
+                c = ldap.initialize(_cfg.ldap_server)
```
robnagler, line 56: Remove temporary variable.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password greater than 256 chars",
+                )
+            if not email or not req.req_data.password:
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password is none/zero length",
+                )
+            return (
```
robnagler, line 48: Here's an example where a simple container would be more descriptive:

```py
return PKDict(
    dn=...
    email=email,
    password=password,
)
```

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+        def _user(email):
+            m = self.auth_db.model(user_model)
+            u = m.unchecked_search_by(unverified_email=email)
+            if not u:
+                u = m.new(unverified_email=email, user_name=email)
+                u.save()
+            return u
+
+        req = self.parse_post()
+        (dn, e, p) = _validate_escape_credentials(req)
```
robnagler, line 75: See above and pass single object to `_bind`

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,87 @@
...
+        _authorize_ldap(dn, e, p)
+        self.auth.login(
+            this_module, sim_type=req.type, model=_user(e), want_redirect=True
+        )
+
+
+def init_apis(*args, **kwargs):
+    global _cfg
+    _cfg = pkconfig.init(
+        ldap_server=("ldap://10.10.10.10:389", str, " ldap://ip:port"),
```
robnagler, line 85: `127.0.0.1` instead of 10.10.10.10, more portable. Use [RequiredUnlessDev](https://github.com/radiasoft/pykern/blob/34fdbfc58ed6aee5257a1129041bf9945bd363e5/pykern/pkconfig.py#L175) and better names:

```py
_cfg = pkconfig.init(
    server=pkconfig.RequiredUnlessDev("ldap://127.0.0.1:389", str, " ldap://ip:port"),
    dn_suffix=pkconfig.RequiredUnlessDev(",ou=users,dc=example,dc=com", str, "ou and dc values of dn"),
)
```

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3013,6 +3013,74 @@ SIREPO.app.directive('emailLoginConfirm', function(requestSender, $route) {
...
+                  </div>
+                  <div class="sr-input-warning" data-ng-show="showWarning">{{ warningText }}</div>
+                  <p class="help-block">By signing up for Sirepo you agree to Sirepo\'s <a href="en/privacy.html">privacy policy</a> and <a href="en/terms.html">terms and conditions</a>, and to receive informational and marketing communications from RadiaSoft. You may unsubscribe at any time.</p>
+                </div>
+              </div>
+            </form>
+        `,
+        controller: function($scope) {
+            function handleResponse(data) {
+                if (data.state == 'ok') {
```
robnagler, line 3053: I think data.state should always be ok here. If it isn't, it should throw an error.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3013,6 +3013,74 @@ SIREPO.app.directive('emailLoginConfirm', function(requestSender, $route) {
...
+                    $scope.showWarning = true;
+                    $scope.warningText = data.formError;
+                    $scope.$broadcast('sr-clearDisableAfterClick');
+                }
+            }
+
+            $scope.login = function() {
+                if(!$scope.email || !$scope.password) {
```
robnagler, line 3061: This code is repeated and it's complicated. Make it function

robnagler (changes requested):
Much better.

sirepo/package_data/static/js/sirepo-components.js
Benbenbenin0: Changing to display 'contact support@sirepo.com' error seen in [emailLogin](https://github.com/radiasoft/sirepo/blob/b93ebec93ed5f83cce9f20980ff7be02c8e78556/sirepo/package_data/static/js/sirepo-components.js#L2963)

Benbenbenin0:
@robnagler I don't think most environments will have a LDAP server setup, so writing a test for binding doesn't seem useful unless there's a central LDAP server to test with. Tests for searching/saving users to auth_db doesn't seem helpful either.

Calling the API to try every exception case (besides binding with invalid credentials) and checking escaping for expected output seem useful, what are your thoughts?

etc/ldap_ldifs/admin_access.ldif
```diff
@@ -0,0 +1,4 @@
+dn: olcDatabase={1}monitor,cn=config
+changetype: modify
+replace: olcAccess
+olcAccess: {0}to * by dn.base="gidNumber=0+uidNumber=0,cn=peercred,cn=external, cn=auth" read by dn.base="cn=admin,dc=example,dc=com" read by * none
```
robnagler, line 4: missing newline (and many other files)

etc/run-auth-ldap.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+export SIREPO_AUTH_METHODS='guest:ldap'
+
```
robnagler, line 3: space isn't really necessary here. only two lines.

etc/temp-ldap-setup.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
robnagler, line 1: Why is this file called temp-ldap-setup.sh? Why not setup-ldap.sh?

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+
+AUTH_METHOD = "ldap"
+
+AUTH_METHOD_VISIBLE = True
+
+_cfg = None
+
+_ESCAPE_DN_MAIL = re.compile(r"([,\\#<>;\"=+])")
+
+_MAX_ENTRY = 127
```
robnagler, line 24: Document this a bit.

robnagler (changes requested):
Great! minor changes

etc/temp-ldap-setup.sh
Benbenbenin0: Didn't expect to merge, named temp because would've been deleted

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password greater than 127 chars",
+                )
+            if len(_cfg.dn_suffix) > _MAX_ENTRY:
+                raise sirepo.util.UserAlert(
+                    "invalid LDAP dn",
+                    f"dn={_cfg.base_dn} greater than 127 chars",
+                )
+            if not email or not password:
```
robnagler, line 68: This test should be first.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+
+        def _validate_and_escape_creds(req):
+            email = req.req_data.email.lower()
+            password = req.req_data.password
+            if len(email) > _MAX_ENTRY or len(password) > _MAX_ENTRY:
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password greater than 127 chars",
+                )
+            if len(_cfg.dn_suffix) > _MAX_ENTRY:
```
robnagler, line 63: This test is an invariant. It's part of cfg. Instead of "str" for pkconfig.init, create a function that tests the value. For examples search around for "def _cfg_" which are examples of pkconfig validators.

From a security perspective, do not return configuration errors to the user. Security permeates code. It's things like this that slip by reviewers (as it did the first time with me). You never want to return a configuration knowledge (invalid dn is knowledge) to an end user unless it is completely safe to do so.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+                    f"email={email} or password is none/zero length",
+                )
+            return PKDict(
+                dn=("mail=" + re.sub(_ESCAPE_DN_MAIL, r"\\\1", email) + _cfg.dn_suffix),
+                email=email,
+                password=password,
+            )
+
+        req = self.parse_post()
+        creds = _validate_and_escape_creds(req)
```
robnagler, line 80: Here I would recommend `c` instead of creds. It's a localized temporary variable, and the `c` will stand out as such, i.e. we know we don't use it except in the next few lines.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+        def _user(email):
+            m = self.auth_db.model(user_model)
+            u = m.unchecked_search_by(unverified_email=email)
+            if not u:
+                u = m.new(unverified_email=email, user_name=email)
+                u.save()
+            return u
+
+        def _validate_and_escape_creds(req):
+            email = req.req_data.email.lower()
```
robnagler, line 56: We like temporaries to be single letters. There are exceptions, e.g. `req`. One option is to create a temporary variable `res = PKDict(email=...)`. Then refer to `res.email`. Alternatively, name the values `e` and `p`, again so they stand out as temporaries.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+                )
+            if len(_cfg.dn_suffix) > _MAX_ENTRY:
+                raise sirepo.util.UserAlert(
+                    "invalid LDAP dn",
+                    f"dn={_cfg.base_dn} greater than 127 chars",
+                )
+            if not email or not password:
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password is none/zero length",
```
robnagler, line 71: This should not be an f-string. The logging arguments to UserAlert (and most other sirepo.util exceptions) are passed verbatim to pkdlog so use a format string `"email={} or password is falsey"`

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,96 @@
...
+                u.save()
+            return u
+
+        def _validate_and_escape_creds(req):
+            email = req.req_data.email.lower()
+            password = req.req_data.password
+            if len(email) > _MAX_ENTRY or len(password) > _MAX_ENTRY:
+                raise sirepo.util.UserAlert(
+                    "invalid user and/or password",
+                    f"email={email} or password greater than 127 chars",
```
robnagler, line 61: See below: use a format string .

On second thought, it might be more logical to have code that looks like this:

```py
_validate_entry(res, "email")
_validate_entry(res, "password")
```

This would simplify and clarify the log messages. 

```py
def _validate_entry(creds, field):
    if not creds[field]:
        e = "falsey"
    elif len(creds[field]) > _MAX_ENTRY
       e = "too long"
    else:
        return
    raise sirepo.util.UserAlert("invalid user and/or password", "{} field={}; email={}", e, field, creds.email)
```

robnagler (changes requested):
I missed some things last time.

sirepo/auth/ldap.py
Benbenbenin0: Unsure if done correctly, replaced 'str' in 'dn_suffix' with function that raises AssertionError similar to pkconfig.raise_error().

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,105 @@
...
+    global _cfg
+
+    def _valid_suffix(value):
+        if not value:
+            e = "falsey"
+        elif len(value) > _MAX_ENTRY:
+            e = "over max chars"
+        else:
+            return value
+        raise AssertionError(e + " value={}".format(value))
```
robnagler, line 96: You can use an f-string here. This type of error message is more controlled. It's not user input but server input so cfg messages are "safe" when using f-string.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,105 @@
...
+        _bind(res)
+        self.auth.login(
+            this_module, sim_type=req.type, model=_user(res.email), want_redirect=True
+        )
+
+
+def init_apis(*args, **kwargs):
+    global _cfg
+
+    def _valid_suffix(value):
```
robnagler, line 89: Name this _cfg_dn_suffix and move outside. pkconfig holds on to this declaration so we don't want it at the module level (not nested) so that the closure (stack) is not held on to.

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,105 @@
...
+        if not value:
```
robnagler, line 90: pkconfig converts zero length strings to None, and None is not passed unless [`@parse_none`](https://github.com/radiasoft/pykern/blob/34fdbfc58ed6aee5257a1129041bf9945bd363e5/pykern/pkconfig.py#L314) is decorating the cfg parser. This allows you to simplify the code greatly.

robnagler (changes requested):
Some lesson about how pkconfig works here.

sirepo/auth/ldap.py
Benbenbenin0: I see

sirepo/auth/ldap.py
Benbenbenin0: > pkconfig converts zero length strings to None...

Can you elaborate? When I initialize dn_suffix with an empty string as below, _cfg.dn_suffix successfully returns an empty string.

`dn_suffix=pkconfig.RequiredUnlessDev(
            "", _cfg_dn_suffix, "ou and dc values of dn"
        )`

robnagler:
> @robnagler I don't think most environments will have a LDAP server setup, so writing a test for binding doesn't seem useful unless there's a central LDAP server to test with. Tests for searching/saving users to auth_db doesn't seem helpful either.
> 
> Calling the API to try every exception case (besides binding with invalid credentials) and checking escaping for expected output seem useful, what are your thoughts?

I think we can mock the ldap part, and then we get an end to end test without ldap. I can show you how to do this.

sirepo/auth/ldap.py
robnagler: Leave the pkconfig.RequiredUnlessDev as it was. This is what I meant:

```py
 def _cfg_dn_suffix(value):
        if len(value) > _MAX_ENTRY:
            raise AssertionError(f"value={value} is too long (>{_MAX_ENTRY})")
        return value
```

Benbenbenin0:
@robnagler Can't re-request review, made changes:

- sirepo.util.Response() no longer exists after master merge, replaced with UserAlert() in ldap.py/_bind().
- Added validation and incorrect credentials test, wondering if appropriate to use pkunit.pkok() for both

sirepo/auth/ldap.py
```diff
@@ -40,9 +40,12 @@ def _bind(creds):
                 m = "Unable to contact LDAP server"
                 if isinstance(e, ldap.INVALID_CREDENTIALS):
                     m = "Incorrect email or password"
-                else:
-                    pkdlog("email={} exception={}", creds.email, e)
-                raise sirepo.util.Response(self.reply_ok(PKDict(formError=m)))
+                raise sirepo.util.UserAlert(
```
robnagler, line 43: Use `raise sirepo.util.SReplyExc(self.reply_ok(PKDict(formError=m)))`

SReplyExc replaced Response, because Response and Reply had very different meanings.

sirepo/auth/ldap.py
```diff
@@ -60,7 +63,7 @@ def _validate_entry(creds, field):
             else:
                 return
             raise sirepo.util.UserAlert(
-                "invalid user and/or password",
+                "Invalid user and/or password",
```
robnagler, line 66: I think this should be an SReply as well. We want to use form_error vs UserAlert to show the user errors.

I would argue that it would be better style to do this:

```py
r = _validate_entry(res, "email") and _validate_entry(res, "password") and _bind(res)
if r:
     return r
```

And convert the raises into returns. There's no need to raise an exception in this case, and it's complicated. The api naturally returns an SReply, and you should implement it that way.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
+                simulationType="myapp",
+                email="badEmail",
+                password="badPassword",
+            ),
+        )
+        pkunit.pkok("error" in r.content_as_str(), "did not return error state")
+
 
 def initialize(*args, **kwargs):
     """Simulate ldap.initialize"""
```
robnagler, line 86: I think we should use the word "Mock" since the word simulation is over used.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
     from pykern.pkcollections import PKDict
 
-    return PKDict(simple_bind_s=lambda x, y: None)
+    # Simulate ldap.simple_bind_s
```
robnagler, line 89: Again Mock.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
 """
 import pytest
 
+# module must have ldap.INVALID_CREDENTIALS
+INVALID_CREDENTIALS = Exception
```
robnagler, line 10: You need a unique test here or the use of INVALID_CREDENTIALS will match any exception, changing the semantics of the code. Change the comment to use Mock and move this down to where initialize is so all the mock stuff is together.

```py
class INVALID_CREDENTIALS():
    pass
```

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
+    sys.modules["ldap"] = pkinspect.this_module()
+
+    with srunit.quest_start(cfg={"SIREPO_AUTH_METHODS": "ldap"}) as qcall:
+        from pykern import pkunit
+
+        r = qcall.call_api(
+            "authLdapLogin",
+            data=PKDict(
+                simulationType="myapp",
+                email="badEmail",
```
robnagler, line 78: I would test only one of these. Also the word "bad" is a value judgment. Rather, say "not-an-email" or something like that. For the password, you can say `any password`. The validation will not check the password.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
+
 
 def test_happy_path():
-    from sirepo import srunit
     from pykern.pkcollections import PKDict
     from pykern import pkinspect
+    from sirepo import srunit
     import sys
 
     sys.modules["ldap"] = pkinspect.this_module()
```
robnagler, line 19: This code is copied three times. It's very specific and hard to get right as well as distracting from the core part of the test. Refactor.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
+    sys.modules["ldap"] = pkinspect.this_module()
+
+    with srunit.quest_start(cfg={"SIREPO_AUTH_METHODS": "ldap"}) as qcall:
+        from pykern import pkunit
+
+        r = qcall.call_api(
+            "authLdapLogin",
+            data=PKDict(
+                simulationType="myapp",
+                email="",
```
robnagler, line 56: Here I would test invalid password and have a valid email so that it goes all the way though to ldap.bind.

tests/auth/ldap_test.py
```diff
@@ -6,33 +6,92 @@
...
+                password="",
+            ),
+        )
+        pkunit.pkok("error" in r.content_as_str(), "did not return error state")
```
robnagler, line 60: You should look for form_error explicitly. There are many ways to get an error, and you are looking for a specific one.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3096,12 +3096,7 @@ SIREPO.app.directive('ldapLogin', function (requestSender, errorService) {
         `,
         controller: function ($scope) {
             function handleResponse(data) {
-                if (data.state == 'ok') {
-                    showWarning(data.formError);
-                }
-                else {
-                    showWarning('Server reported an error, please contact support@sirepo.com.');
-                }
+                showWarning(data.error);
```
robnagler, line 3099: Let's continue to use formError. data.error is too coarse and conflicts with other uses in sirepo.

robnagler (changes requested):
Testing is tricky. Try to

sirepo/auth/ldap.py
Benbenbenin0: Noting this style, makes much more sense for api

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,102 @@
...
+class API(sirepo.quest.API):
+    @sirepo.quest.Spec("require_cookie_sentinel")
+    def api_authLdapLogin(self):
+        def _bind(creds):
+            try:
+                ldap.initialize(_cfg.server).simple_bind_s(creds.dn, creds.password)
+            except Exception as e:
+                m = "Unable to contact LDAP server"
+                if isinstance(e, ldap.INVALID_CREDENTIALS):
+                    m = "Incorrect email or password"
```
robnagler, line 42: In reading this more carefully, let's keep all the errors the same so make it a constant `_INVALID_CREDNTIALS = "Invalid user and/or password"` that's sahred between `_validate` and `_bind` so here:

```py
m = _INVALID_CREDENTIALS
```

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,102 @@
...
+                pkdlog("{} email={}", e, creds.email)
+                return sirepo.util.SReplyExc(self.reply_ok(PKDict(form_error=m)))
```
robnagler, line 44: ```return m``` (see below)

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,102 @@
...
+
+        def _validate_entry(creds, field):
+            if not creds[field]:
+                e = "falsey"
+            elif len(creds[field]) > _MAX_ENTRY:
+                e = "over max chars"
+            else:
+                return
+            pkdlog("{} field={}; email={}", e, field, creds.email)
+            return sirepo.util.SReplyExc(
```
robnagler, line 62: `return _INVALID_CREDENTIALS` see below

sirepo/auth/ldap.py
```diff
@@ -0,0 +1,102 @@
...
+            + re.sub(_ESCAPE_DN_MAIL, r"\\\1", req.req_data.email)
+            + _cfg.dn_suffix,
+        )
+        r = (
+            _validate_entry(res, "email")
+            or _validate_entry(res, "password")
+            or _bind(res)
+        )
+        if r:
+            raise r
```
robnagler, line 80: The previous review said to drop the raise for return. What was I meant was `return` from the whole api method. An api returns a reply (or a dict, but we want "ok" in this case) so code would just be:

`return self.reply_ok(PKDict(form_error=r))`

tests/auth/ldap_test.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+import pytest
+from pykern.pkdebug import pkdlog, pkdp
+
+
+def test_happy_path():
+    from pykern.pkcollections import PKDict
+    from pykern import pkunit
+
+    res = PKDict(
```
robnagler, line 15: `res` stands for result in our style. Some people use `rv` (return value) for the same thing. `res` here is therefore incorrect. Moreover, you don't need the temporary as it is only used once. 

Once you refactor that for all these methods, you will notice that _call_login could accept email, password and construct the PKDict.

tests/auth/ldap_test.py
```diff
@@ -0,0 +1,88 @@
...
+def test_cred_validation():
+    from pykern.pkcollections import PKDict
+    from pykern import pkunit
+
+    res = PKDict(
+        email="any-user",
+        password="",
+    )
+    r = _call_login(res)[0]
+    pkunit.pkok("form_error" in r.content_as_str(), "did not return form_error")
```
robnagler, line 34: This is pointing out an interesting problem for our tests when calling apis directly. What's missing is comparing json to a struct. In this case, we want something like this:

```py
pkunit.pkre("invalid user and", r.data.form_error)
```

You don't have this, but how would you get it?

BTW, you generally do not want a message, because this obscures the values. The nice thing pkunit is that it outputs the line and file where the error occurs so you can just jump to that part of the test to see what is wrong. In general, pkok should be avoided for pkre, because the latter shows what the re and the value in the error message, which is useful for debugging the test. The actual error location is output so strings like "did not return form_error" aren't all that useful.

tests/auth/ldap_test.py
```diff
@@ -0,0 +1,88 @@
...
+def test_incorrect_creds():
+    from pykern.pkcollections import PKDict
+    from pykern import pkunit
+
+    res = PKDict(
+        email="not-a-user",
+        password="any-password",
+    )
+    r = _call_login(res)[0]
+    pkunit.pkok("form_error" in r.content_as_str(), "did not return form_error")
```
robnagler, line 46: Same as above. And, you don't need the temporary.

tests/auth/ldap_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+        r = qcall.call_api(
+            "authLdapLogin",
+            data=PKDict(
+                simulationType="myapp",
+                email=res.email,
+                password=res.password,
+            ),
+        )
+        m = qcall.auth_db.model("AuthEmailUser")
```
robnagler, line 66: You don't need the temporaries.

robnagler (changes requested):
More elaboration that I didn't catch the first time, plus some notes on testing and refactoring.

tests/auth/ldap_test.py
Benbenbenin0: > What's missing is comparing json to a struct. In this case, we want something like this: `pkunit.pkre("invalid user and", r.data.form_error)`

What's an example of a JSON and struct we would want to compare for testing?

I'm aware of the specific struct type in C, what is a struct in this context? Do we refer to PKDict as a struct, and therefore the _SReply we get from calling api's as well?

My basic understanding is we use JSON for application/code schema and storing/sharing simulation data (I think). In this case the expected is just a string "invalid user...", what JSON would need to be referenced in authentication testing (or other testing)?

May not be good questions, feels like I'm missing context so trying to orient my understanding of what's being discussed.

robnagler: There's actually no good word for the result of a JSON deserialization. JSON is a string. We could just call it an "object", which it is, but that's not too descriptive. Calling it a PKDict is also incorrect as it could be a list or a string, etc. That's why I used "struct". Sorry for the confusion.

robnagler (approved):

Issue #5255, opened 2022-12-09
20221209 react missing report panel behaviors


garsuga:
from @moellep 
- Clear reports when "Start SImulation" is selected.
- When you run a new Start Simulation and it finishes, it isn't showing the new plots, but still displays the old results. In JSPEC you can see this by changing the "Total Simulation Time" and "Total Number of Steps" from 180 to 300 or 50 and run again - the plots should have a longer x axis.
- Change the run status pollInterval from 500ms to 2000ms. Otherwise the client can overwhelm the server with plot requests if they take awhile to compute. This won't be a problem when we have websockets in the future.
- When you run Start Simulation and run-status is checking for results, it should plot anything available while the status is "running". It shouldn't wait until "completed" to show the plots.

robnagler:
> Clear reports when "Start SImulation" is selected.

Please consider changing this to clear plots on save changes to data. This happens for "automatic" plots. Once you change the data, the graphs become invalid.

> Change the run status pollInterval from 500ms to 2000ms. 

This number comes in the reply from the server.

garsuga:
> This number comes in the reply from the server.

I will work on a commit to make use of that

> Please consider changing this to clear plots on save changes to data. This happens for "automatic" plots. Once you change the data, the graphs become invalid.

This is already the behavior of automatic reports. For manual reports they are all controlled centrally but automatic reports make their own subscriptions to data changes. The central control for manual reports does not look for changes in the data right now but it could in the future.

robnagler:
Please create an issue about clearing data for "manual reports" when data for it changes.

garsuga:
Just to be clear, should be clear the panels or should be add some UI element to the panels to signal that they are outdated like how github puts "Outdated" by commented code that has been changed in PRs?

robnagler:
That is an interesting idea. Definitely worth trying. The server knows when reports are outdated. This should be consistent for automatic and manual reports.

What would happen on a page refresh? Would you get all graphs even when they are outdated?

garsuga:
I will open a new issue for this since this one should be closed.

PR #5256, opened 2022-12-09
fix #5234, #5255: better report state handling

### LINKED ISSUE #5234: 20221215 react update plots as simulation is running
- clear plots when "start new simulation" is selected
- when frames are available from runStatus, update the plots rather than waiting for the full simulation to finish

garsuga:
also fixes:
- reports will 'carry on' polling and showing status after refresh
- heatplots do their min/max math correctly now
- reports will seek latest frame on refresh
- theres a WIP type for the status returned by requests like run-status
- additional states can be handled by the start simulation layout like missing and error

### DIFF
--- a/react/src/data/report.ts
+++ b/react/src/data/report.ts
@@ -87,7 +87,6 @@ export class ReportEventManager {
             models,
             simulationId,
             report,
-            pollInterval: 500,
             forceRun: true,
             callback: (simulationData) => {
                 this.handleSimulationData(report, simulationData);
--- a/react/src/data/report.ts
+++ b/react/src/data/report.ts
@@ -1,54 +1,77 @@
-import { getSimulationFrame, pollRunReport, runStatus } from "../utility/compute";
+import { getSimulationFrame, pollRunReport, ResponseHasState, getRunStatusOnce } from "../utility/compute";
 import { v4 as uuidv4 } from 'uuid';
 import React from "react";
 import { ModelStates } from "../store/models";
+import { mapProperties } from "../utility/object";
 
 export const CReportEventManager = React.createContext<ReportEventManager>(undefined);
 
-export type ReportEventSubscriber = (simulationData: any) => void;
+export type ReportEventSubscriber = {
+    /**
+     * Called once for every simulationData received (including during completion)
+     */
+    onReportData?: (simulationData: ResponseHasState) => void;
+    /**
+     * Called once when reports are starting to be generated
+     */
+    onStart?: () => void;
+    /**
+     * Called once when reports are finished being generated
+     */
+    onComplete?: () => void;
+}
+
+type RunStatusParams = {
+    appName: string,
+    models: ModelStates,
+    simulationId: string,
+    report: string
+}
 
 export class ReportEventManager {
-    reportEventListeners: {[reportName: string]: ReportEventSubscriber[]} = {}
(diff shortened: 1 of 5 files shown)


react/src/data/report.ts
```diff
@@ -1,49 +1,96 @@
...
+        report: string
     }) => {
-        runStatus({
+        pollRunReport({
             appName,
             models,
             simulationId,
             report,
-            forceRun: false,
+            pollInterval: 500,
```
robnagler, line 90: This comes from the server.

react/src/data/report.ts
```diff
@@ -1,49 +1,96 @@
...
+     * Called once when reports are starting to be generated
+     */
+    onStart?: () => void;
+    /**
+     * Called once when reports are finished being generated
+     */
+    onComplete?: () => void;
+}
+
+type RunStatusParams = {
```
robnagler, line 24: These params are returned from the server in nextRequest from runSimulation and runStatus. They should be treated opaquely by the client. nextRequestSeconds is also returned, and that should be used for the poll on the next polling. Different simulations return different amounts.

Please look at the [angular code](https://github.com/radiasoft/sirepo/blob/8fb85dc7befa7684ac4de4631b053c75d47a014a/sirepo/package_data/static/js/sirepo.js#L2558) to see how to handle requests from the server.

robnagler (dismissed):

react/src/data/report.ts
garsuga: fixed in latest

react/src/data/report.ts
garsuga: In order to "pick up where we left off" on reports during a page refresh these are sometimes manually provided. The client should also probably know about these so we know which requests can share parameters (like when running a report also begins polling).

robnagler: This isn't how the current app works. It simply calls runSimulation with all the data and the server decides if the simulation needs to run. This same logic could be used.

moellep:
Works much better - here are a few things I noticed with JSPEC:

- When the manual report is "running" but is at 0% complete, show the progress bar at 100% and animated. Some codes take many minutes to set up before data is returned and looking at a empty progress bar may look like it isn't doing anything.
- In the JSPEC "DC Cooling Example", only the IBS & Cooling Rates panel is showing partial updates. The RMS Ion Beam Evolution report should also be showing partial results. Set the Total Number of Steps and Total SImulation Time to 500 to run the report across a longer time period to test.
- In the JSPEC "Booster Ring" example, the IBS & Cooling Rates is showing the next/prev animation buttons, but that report does not have frames.

garsuga:
@moellep I set the default to 100% when there has not been a successful status and tried out a minimum percentage at 5% for the bar to aid with looking empty. Would you rather have it set so 0 shows as 100% or have a minimum percent?

react/src/data/report.ts
garsuga: Would this be a case where forceRun=false?

moellep:
Yes, 5% minimum would be fine, and also non 0% would also be OK.

react/src/data/report.ts
robnagler: Correct. The forceRun=True is for manual starts.

moellep: It looks like the nextRequestSeconds has been addressed, so I'll resolve this item.

PR #5268, opened 2022-12-13
fix #5267 react improvements

### LINKED ISSUE #5267: react app improvements
- show HTML title as "appName - Sirepo"
- format elapsed time as (days) HH:MM:SS
- show/hide form fields
- larger boolean checkbox element
- add ColumnBreakLayout
- show progress bar at 100% when 0% is completed

moellep:
- show HTML title as "appName - Sirepo"
- format elapsed time as (days) HH:MM:SS
- show/hide form fields
- larger boolean checkbox element
- add ColumnBreakLayout
- show progress bar at 100% when 0% is completed

### DIFF
--- a/react/src/component/app.js
+++ b/react/src/component/app.js
@@ -1,4 +1,4 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect } from "react";
 import { configureStore } from "@reduxjs/toolkit"
 import { modelsSlice } from "../store/models";
 import { formStatesSlice } from "../store/formState";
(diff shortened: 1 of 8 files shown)


garsuga:
Looks good. I've noticed the elapsed time does not update at the correct speed on the UI. This is because the Stopwatch class is not capable of causing a refresh for the component so it only updates the time during each response received from the server. We should add a manual refresh.

Something like:
```
let [s. updateS] = useState({});

useEffect(() => {
    if(lastSimState && lastSimState.state === 'running') setTimeout(() => updateS({}), 500) 
})
```

PR #5270, opened 2022-12-14
Fix #5257 - added openSimulation to requestSender

### LINKED ISSUE #5257: Add function to redirect to specific sim id of another app
`requestSender.globalRedirect()` strips everything after the hash so `<app>#/<section>/...` becomes `<app>`. Add a new function or modify this one so we can go directly to a simulation.

mkeilman:
Also modifies simList to use it


sirepo/package_data/static/js/sirepo.js
```diff
@@ -2160,6 +2165,20 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, userAg
         $window.open(self.formatUrl(routeName, params), '_blank');
     };
 
+    self.openSimulation = (app, route, simId) => {
+        const m = {};
+        m[route] = routeMapLocal(route, `/${route}/:simulationId`);
```
mkeilman, line 2170: Look up in schema instead

mkeilman: Update: `/simulation-schema` works with the ajax call but not with sendRequest. In the latter case the flask request contains no form data. Other routes like `simulation-list` appear to work. Not sure where or why the data is getting lost...

robnagler: Need more info here. Is the ajax a normal form post and sendRequest is json post?

robnagler:
@mkeilman ping

sirepo/package_data/static/js/sirepo.js
mkeilman: I see the problem: simulation-schema and the corresponding API are designed to use form data through ajax. I think a new route and api are the way to go then.

mkeilman: Latest commit adds simulationRoute which gets a specific local route for an app. I don't think we need the whole schema.

Not sure about the proper Spec or the naming; 'route' is used a lot. Maybe 'appSection' is better.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2161,6 +2166,31 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, userAg
         $window.open(self.formatUrl(routeName, params), '_blank');
     };
 
+    self.openSimulation = (app, route, simId) => {
```
robnagler, line 2169: All this information is in the GUI in the route map. I don't think this should (or does) change from app to app.  You definitely don't need a new API.

Why do you need the schema at this point?

robnagler (changes requested):

sirepo/package_data/static/js/sirepo.js
mkeilman: We want the "data" route, which does not exist in srw's localMap. I was previously assuming the form of the route (`/${route}/:simulationId`). Maybe that's actually a safe assumption since addRoute references it:
https://github.com/radiasoft/sirepo/blob/fe4a0b7512432acd62db91d6ee02deec8ff4b67d/sirepo/package_data/static/js/sirepo.js#L204-L207
but using the schema seems to be "the right way". One way or another the route has to get added to localMap.

robnagler: Yes, I see the problem now. I think what we want is a generic local route for this so we avoid stuff like this:

$scope.route || 'visualization',

We could have a route alias that is know globally. However, that's even more complicated.

Perhaps this does need to be a new entrypoint, but with all the logic server side, generating the local route and redirecting at the end, e.g.

window.open(self.formatUrl("simulationRedirect", {code: "srw", simId: "xyz"}, '_blank');

Thoughts?

mkeilman: Sounds good; I'll flesh that out

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2161,6 +2161,17 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, userAg
...
+        $window.open(
+            self.formatUrl('simulationRedirect', {
+                simulation_type: app,
+                route: route,
```
robnagler, line 2168: I think this should be local_route/localRoute (everywhere), since that's what it is.

robnagler (changes requested):
Much better, thanks. Minor naming issue.

sirepo/package_data/static/js/sirepo.js
mkeilman: > I think this should be local_route/localRoute (everywhere), since that's what it is.

Done almost everywhere;  I excluded e.g. arguments to `reply_redirect_for_local_route` since they have "local _route" in the function name

robnagler (changes requested):
Much simpler. Thanks for working through it.

robnagler (approved):

PR #5284, opened 2022-12-20
Fix #5171 radia default to designer

### LINKED ISSUE #5171: Default to designer tab
...except for imported dump files or some current examples, which have no designer tab

mkeilman:
This uses the existing appModes in the schema along with a field on the simulation model to default to a section on a per-sim basis. Also sets the designer page as the default for radia, except for imported models, which still go to the viz page.


moellep:
In this branch, if I open the Wiggler or the Dipole example sims, it causes a javascript error:
```
angular.js:15567 TypeError: Cannot read properties of undefined (reading 'forEach')
    at addShapesForObject (radia.js?20230102.214402:699:22)
```

moellep:
This is now closer, but if you visit the Dipole example and select "discard changes to examples", you get the "forEach" javascript error. If you return to the simulations page, and revisit the sim, it works correclty.

mkeilman:
The latest change works but the applicationMode sticks around in the url after discarding changes to examples. Looks a little weird...

PR #5294, opened 2022-12-27
Fix #4882 Remove werkzeug exceptions

### LINKED ISSUE #4882: 20221021 Normalize exception handling
- remove werkzeug exceptions, handle directly as 404, etc.
- UserDirNotFound is now an exception (was a call)
- Create specific 404 exception that's not noisy
- https://github.com/radiasoft/sirepo/issues/2553
- https://github.com/radiasoft/sirepo/issues/3592
- https://github.com/radiasoft/sirepo/issues/4616
- https://github.com/radiasoft/sirepo/issues/2477
- https://github.com/radiasoft/sirepo/issues/2085

robnagler:
- All exceptions raised are subclass of sirepo.util.Reply 
- Removed all references to werkzeug except in sirepo.pkcli.service.flask. 
- Rewrote util.secure_filename with a restricted implementation
- Replaced util.safe_path with util.validate_path with a much restricted interface to match our use
- Added run-auth-bluesky.sh and run-auth-basic.sh

### DIFF
--- a/sirepo/util.py
+++ b/sirepo/util.py
@@ -85,6 +85,12 @@ class Reply(Exception):
         return self.__repr__()
 
 
+class BadRequest(Reply):
+    """Raised for bad request"""
+
+    pass
+
+
 class OKReply(Reply):
     """When a Reply exception is a successful response"""
 
--- a/sirepo/util.py
+++ b/sirepo/util.py
@@ -15,6 +15,7 @@ import hashlib
 import importlib
 import inspect
 import numconv
+import posixpath
 import pykern.pkinspect
 import pykern.pkio
 import pykern.pkjson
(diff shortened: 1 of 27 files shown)


sirepo/util.py
```diff
@@ -325,22 +351,6 @@ def json_dump(obj, path=None, pretty=False, **kwargs):
     return res
 
 
-def raise_bad_request(*args, **kwargs):
-    _raise("BadRequest", *args, **kwargs)
-
```
moellep, line 330: There is a call to raise_bad_request in http_request.py which needs to be replaced

moellep:
Looks good.

sirepo/util.py
robnagler: The code was not executed because content_type_eq was blowing up so the result was a server error. There's a test and this should be fixed.

PR #5295, opened 2022-12-27
fmt for black 22.12

rorour:
depends on https://github.com/radiasoft/pykern/pull/332

### DIFF
--- a/sirepo/auth/__init__.py
+++ b/sirepo/auth/__init__.py
@@ -490,7 +490,6 @@ class _Auth(sirepo.quest.Attr):
             pass
         elif s == _STATE_LOGGED_IN:
             if m in _cfg.methods:
-
                 f = getattr(_METHOD_MODULES[m], "validate_login", None)
                 if f:
                     pkdc("validate_login method={}", m)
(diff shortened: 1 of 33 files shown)


robnagler:
I thought the difference was between 22.12 and 23.x?

rorour:
What version of black have you been using? Before any of these changes I was running 21.6. I assume we would all have to be on the same or very close. I think we have just been lucky to not run into issues with black versions so far.

When working on the ci changes for pykern I discovered a bug that was patched in 22.12, and this pr contains the changes to match fmt with 22.12.

The reason for the fixed version required in pykern would be future proofing. Not sure if that's the right term someone with 23.x vs 22.12 would have different outputs for fmt which would cause issues passing ci.

robnagler:
This is a real problem, because we can't have battling formatter versions. Our code base will be a mess. This is not something I would have anticipated when "buying" black. I thought they would have figured all this out already, and be done with the formatting changes. The [definition of "black" changes from year to year](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy). I wonder if we wait until the new year that we'll have a 23.x version that will be stable all year.

We should require a fixed version for this reason. Thank you @rorour for pointing that out. I don't know if it is future proofing as much as chaos proofing. 

In any event, we should fix the version with `black~=22.0` so we get bug fixes within the year. We'll have to figure out how to deal with broken PRs on Sirepo, because the version change has to coincide with the sirepo-ci image change.

robnagler (approved):

PR #5223, opened 2022-12-01
Fix #4989: all tests passing on fedora36

e-carlin:
Synergia will be removed (https://github.com/radiasoft/sirepo/issues/5222). So, skip the synergia tests for now.

As part of this fix #4993 (json decoding) and fix #5005 (changes to openpmd).

PR #5318, opened 2023-01-02
Fix #2447: CancelledError inherits from BaseException on > py 3.8

### LINKED ISSUE #2447: asyncio.CancelledError inherits from BaseException in > py 3.8.x
This doesn't need to happen soon, but it will break things, and I wanted to put a placeholder. It may be that we do this when we upgrade to fedora/3x

e-carlin:
Change the code to:
- Catch CancelledError explicitly when needed in cases where we were relying on it to be a subclass of Exception previously.
- Remove any code that does nothing but raise on CancelledError. This pattern was used to not smother CancelledError in cases where we had an `except Exception`. Those will no longer smother the error.

In addition, in sirepo.job_supervisro.run re-raise the CancelledError and remove check for return of False from _send_op. This has no actual behavior change. We don't await _run so it doesn't matter if the CancelledError makes it out. But, it is more "correct" to not smother CancelledError and have it propagate all the way out so a cancel happens fully.

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -12,10 +12,12 @@ import importlib
 import pykern.pkio
 import re
 import sirepo.auth
+import sirepo.const
 import sirepo.events
 import sirepo.sim_db_file
 import sirepo.simulation_db
 import sirepo.tornado
+import sirepo.util
 import tornado.gen
 import tornado.ioloop
 import tornado.locks
@@ -273,7 +275,7 @@ class DriverBase(PKDict):
                 # POSIT: Canceled errors aren't smothered by any of the below calls
                 await self.kill()
                 await self._do_agent_start(op)
-            except Exception as e:
+            except (Exception, sirepo.const.ASYNC_CANCELED_ERROR) as e:
                 pkdlog("{} error={} stack={}", self, e, pkdexc())
                 self.free_resources(internal_error="failure starting agent")
                 raise
--- a/sirepo/job_driver/docker.py
+++ b/sirepo/job_driver/docker.py
@@ -124,10 +124,6 @@ class DockerDriver(job_driver.DriverBase):
             await self._cmd(
                 ("stop", "--time={}".format(job_driver.KILL_TIMEOUT_SECS), self._cname),
             )
-        except sirepo.util.ASYNC_CANCELED_ERROR:
-            # CanceledErrors need to make it back out to be handled
-            # by callers (ex job_supervisor.api_runSimulation)
-            raise
         except Exception as e:
             if not c and "No such container" in str(e):
                 # Make kill response idempotent
--- a/sirepo/job_supervisor.py
+++ b/sirepo/job_supervisor.py
@@ -7,7 +7,6 @@
 from __future__ import absolute_import, division, print_function
 from pykern import pkcollections
 from pykern import pkconfig
-from pykern import pkinspect
 from pykern import pkio
 from pykern.pkcollections import PKDict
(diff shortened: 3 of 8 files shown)


e-carlin:
This depends on https://github.com/radiasoft/sirepo/pull/5223 which is waiting on the fedora 36 migration. So, tests won't currently pass.

sirepo/job_driver/__init__.py
```diff
@@ -273,7 +274,7 @@ async def _agent_start(self, op):
                 # POSIT: Canceled errors aren't smothered by any of the below calls
                 await self.kill()
                 await self._do_agent_start(op)
-            except Exception as e:
+            except (Exception, *sirepo.util.ASYNC_CANCELED_ERROR) as e:
```
robnagler, line 277: I read about `except*`. This nested/multi-exception thing seems very complicated. Do we need it here? It seems like these are two unrelated exceptions.

sirepo/job_driver/docker.py
```diff
@@ -124,10 +124,6 @@ async def kill(self):
             await self._cmd(
                 ("stop", "--time={}".format(job_driver.KILL_TIMEOUT_SECS), self._cname),
             )
-        except sirepo.util.ASYNC_CANCELED_ERROR:
```
robnagler, line 127: This is an implicit coupling on the meaning of ASYNC_CANCELED_ERROR. concurrent.futures.CancelledError is a subclass of Exception so this changes the semantics. Is concurrent.futures.CancelledError not called in 3.10? If it is, then we need to handle that case specially to re-raise like it was. 

I think ASYNC_CANCELED_ERROR belongs in const, but it could be that it should be removed and carefully coded in all cases.

sirepo/tornado.py
```diff
@@ -51,6 +51,7 @@ async def get(self):
             if x:
                 try:
                     r = x.result()
+                # TODO(e-carlin): I think this shouldn't catch CancelledError too. In py 3.8 it won't which I think is the right behavior
```
robnagler, line 54: I think it needs to catch and reraise.

sirepo/job_supervisor.py
```diff
@@ -912,8 +912,7 @@ async def _send_op(op, compute_job_serial, prev_db):
 
         op.task = asyncio.current_task()
         op.pkdel("run_callback")
-        if not await _send_op(op, compute_job_serial, prev_db):
```
robnagler, line 915: I think we need to catch the top level await on a task. It's not clear to me that this will work as written.

robnagler (changes requested):
Perhaps we should go over this together. I need to understand why cancelled can't be caught. We may have to solve this properly, that is, not having cancel. This might be possible.

sirepo/job_driver/__init__.py
e-carlin: 2f7887c01

sirepo/job_driver/docker.py
e-carlin: 2f7887c01

sirepo/tornado.py
e-carlin: 2f7887c01

sirepo/job_supervisor.py
e-carlin: 2f7887c01

e-carlin:
I squashed and updated the commit message to provide more context for what we talked about [0b60bd3](https://github.com/radiasoft/sirepo/pull/5318/commits/0b60bd3062f1447e0668daf1f87524ae5334fdf9).

robnagler (approved):
Thanks for walking me through the code.

PR #5322, opened 2023-01-03
fix #5321 minor react improvements

### LINKED ISSUE #5321: minor react improvements
- format numeric input fields in scientific notation if < 0.001 or > 10000
- numeric input fields align right
- add singleFrameAnimation config option to manual plots so animation control can be hidden
- improve animation buttons UI

moellep:
- format numeric input fields in scientific notation if < 0.001 or > 10000
- numeric input fields align right
- add singleFrameAnimation config option to manual plots so animation control can be hidden
- improve animation buttons UI

### DIFF
--- a/react/src/component/app.scss
+++ b/react/src/component/app.scss
@@ -43,3 +43,7 @@ a {
 .app-body {
     margin-top: 25px;
 }
+
+.btn-light {
+  --bs-btn-border-color: #ccc;
+}
--- a/react/src/component/login/login.scss
+++ b/react/src/component/login/login.scss
@@ -2,3 +2,9 @@
 .sr-navbar-auth {
     padding: 10px;
 }
+
+.sr-login-avatar {
+  border-radius: 3px;
+  width: 20px;
+  height: 20px;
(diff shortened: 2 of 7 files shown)


garsuga:
>add singleFrameAnimation config option to manual plots so animation control can be hidden

Before this PR we were doing this by checking frameCount and hiding the controls if there was only a single frame. If this was not enough then I think we should fix the frameCount value on the server instead of making this change permanent. It seems to duplicate information available in the frameCount value.

garsuga (approved):

moellep:
The singleFrameAnimation is for plots which update as the simulation is running, but show only the most recent results. The "RMS Ion Beam Evolution" and "IBS & Cooling Rates" plots in jspec are an example. They have no real frame, but a change in frame value indicates that new data is available to plot. We could add new reply info to remove the frameCount for these reports and replace it with a last-modified time.

garsuga:
That seems good. Maybe we can find a better name for the value but I don't know what that name would be.

PR #5325, opened 2023-01-03
Fix #5206: min max scale ranges

### LINKED ISSUE #5206: 20221230 min max scaler should accept range input
When the Min-Max Scaler is selected, additional fields for min and max values should be shown as input values, defaulting to 0.0 to 1.0.

gurhar1133:
* Fit transform and scale x and y
* Feature ranges available for MinMaxScaler
* Error handling for when y scaler outputs out of range


moellep:
When I'm testing with the image set below, I tried setting both scalers to min/max 0/1 (which is similar to hard-coded settings on master), but it fails with an error:
```Model training failed due to: Incompatible shapes: [64,4] vs. [128,2]```

Here is the sim on alpha: https://alpha.sirepo.com/activait#/data/hHEBkwV3

You could export it and bring it into your dev environment for testing.

**Comment by moellep** [comment]
It should produce output similar to this:
tes5000

**Comment by gurhar1133** [comment]
@moellep I think it is fixed

File: sirepo/package_data/template/activait/loadImages.py.jinja [hunk 1]
```diff
@@ -56,35 +33,60 @@ class SirepoHDF5Sequence(keras.utils.Sequence):
...
+            return
+        if scaler == MinMaxScaler:
+            return scaler(feature_range=({{ feature_min }}, {{ feature_max }}))
+        return scaler()
+
+    def _fit(self):
+        if not self.inScaler and not self.outScaler:
+            return
+        with h5py.File(self.filename, "r", libver="latest", swmr=True) as f:
+            self.original_shape_x = f[self.x.key].shape
```
moellep, line 55: We don't need the extra libver and swmr settings for both the h5py.File() calls - the file isn't being written while the readers are running.

moellep:
This works well. I tested with several different scalers and they worked as expected.

PR #5343, opened 2023-01-09
5334 sim status

moellep:
This is the step toward a new plotting result format to be used by the react apps. The server provides which plots are available, and whether they have frames or a file modified time. Eventually, all the values will be within the "result" container.

This changes removes the client config frameCountFieldName and singleFrameAnimation and removes the need for the client to track which reports apply to the selected settings. The changes don't interfere with the existing angular client.

### DIFF
--- a/sirepo/template/jspec.py
+++ b/sirepo/template/jspec.py
@@ -90,138 +90,25 @@ def analysis_job_compute_particle_ranges(data, run_dir, **kwargs):
     )
 
 
-def _v2_add_report_mtime(result, model_name, path):
-    if path.exists():
-        result.append(
-            PKDict(
-                modelName=model_name,
-                lastUpdateTime=int(float(os.path.getmtime(str(path)))),
-            )
-        )
-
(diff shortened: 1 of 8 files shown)


sirepo/template/jspec.py
```diff
@@ -583,3 +529,77 @@ def _set_mass_and_charge(ion_beam):
         ion_beam.mass, ion_beam.charge_number = _SPECIES_MASS_AND_CHARGE[
             ion_beam.particle
         ]
+
+
+def _v2_add_report_mtime(result, model_name, path):
+    if path.exists():
+        result.append(
```
robnagler, line 536: Let's remove result. It's not the right concept here. This should just be reports.

robnagler (changes requested):

robnagler (approved):

robnagler:
I updated the branch. I did not auto merge because wasn't sure if other code needed on react side.

PR #5346, opened 2023-01-10
Fix #5205: invert scaling for plot

### LINKED ISSUE #5205: 20221230 show prediction vs ground truth and unscaled ranges in fit reports
The fit plots should label which axis is the prediction and which is the ground truth. Also, the ranges on both axis need to be reverted from the scaled values to the actual values.

gurhar1133:
Ground truth and prediction are labeled on the fit plot which now also includes unscaled values.

Also, for cases where regression is being used to classify (discrete outputs and prediction size matching range of outputs) np.argmax is used to pick highest likelihood class


sirepo/package_data/template/activait/train.py.jinja
```diff
@@ -35,7 +36,36 @@ model.fit(
...
+def _prediction_shape(prediction, testy):
+    return prediction.shape[-1] == _domain_len(testy)
+
+def _classification_case(prediction, testy):
+    return {{ discreteOutputs }} and _prediction_shape(prediction, testy)
+
+def _predictions_final(p, testy):
+    if _classification_case(p, testy):
+        p = np.argmax(p, axis=1)
+    return p
```
mkeilman, line 54: I think it's best to avoid changing the input parameter, so:
```python
def _predictions_final(p, testy):
    return np.argmax(p, axis=1) if _classification_case(p, testy) else p
```

mkeilman (commented):
Minor code comment.

sirepo/package_data/template/activait/train.py.jinja
gurhar1133: Good call

moellep:
Works well with both control setting sims and image classification.

Issue #5280, opened 2022-12-16
Shapes sometimes do not get updated after points get updated


mkeilman:
The calculation of points happens in a stateless compute, which takes a moment to complete. Meanwhile the shapes can update with no point information and thus not show up in the workplace. Need to add a callback to update the shapes again. Probably a good time to centralize shape management, which is scattered around various controllers and services.

PR #5347, opened 2023-01-10
Fix #5329 radia extruded center

### LINKED ISSUE #5329: Extruded objects do not change position
Changing the values of Center for an extruded object does not change the object's position.

(see difference in y center in mouseover of the object to actual position)
image

mkeilman:
Main issue here was to redraw the shapes once the new center values are saved. Some duplicated code here but it's fine in the short term (but see https://github.com/radiasoft/sirepo/issues/5280#issue-1500983276).


mkeilman:
This is more than I like to merge myself but it's fairly isolated and the upcoming alpha should include it

PR #5352, opened 2023-01-10
Fix #4844: handle bad auth request

### LINKED ISSUE #4844: 20221018 auth_role_moderation must be robust against bad data
```txt
05:00:07 sirepo/uri_router.py:110:call_api api={'base_uri': 'save-moderation-reason', 'cls': <class 'sirepo.auth_role_moderation.API'>, 'decl_ur
i': '/save-moderation-reason', 'func': <function Spec.__call__.<locals>._wrapper at 0x7f023d00e158>, 'func_name': 'api_saveModerationReason', 'name': 'saveModerationReason', 'params': []} exception='PKDict' object has no attribute 'reason' stack=Traceback (most recent call last):
05:00:07 sirepo/uri_router.py", line 102, in call_api
```

If no reason is given, then a uniqueness constraint is raised, which doesn't make sense:

```txt
05:00:13 sirepo/auth_role_moderation.py:140:api_saveModerationReason Error=(sqlite3.IntegrityError) UNIQUE constraint failed: user_role_invite_t
.uid, user_role_invite_t.role
05:00:13 [SQL: INSERT INTO user_role_invite_t (uid, role, status, token, moderator_uid) VALUES (?, ?, ?, ?, ?)]
05:00:13 [parameters: ('<u1>, 'sim_type_jupyterhublogin', <ModerationStatus.PENDING: 'pending'>, '<token>', None)]
```

First step is to create a test for this.

gurhar1133:
UI:
* allow empty reason
* disable button after submit

Server:
* user alert if no reason


robnagler:
@rorour you worked on this code so would be great if you could review.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3636,20 +3636,24 @@ SIREPO.app.directive('moderationRequest', function(appState, errorService, panel
               <label for="requestAccessExplanation">Please describe your reason for requesting access:</label>
               <textarea data-ng-show="!submitted" data-ng-model="data.reason" id="requestAccessExplanation" class="form-control" rows="4" cols="50" required></textarea>
             </div>
-            <button data-ng-show="!submitted" type="submit" class="btn btn-primary" data-ng-click="submitRequest()">Submit</button>
+            <button data-ng-disabled="disableSubmit" data-ng-show="!submitted" type="submit" class="btn btn-primary" data-ng-click="submitRequest()">Submit</button>
```
e-carlin, line 3639: Can you use disableAfterClick? https://github.com/radiasoft/sirepo/blob/bf88fb821bb4190771590fc2b6b12fc44070bbc2/sirepo/package_data/static/js/sirepo-components.js#L519

rorour (approved):
Looks good to me!

sirepo/auth_role_moderation.py
```diff
@@ -150,6 +150,8 @@ def _send_request_email(info):
                 )
 
         l = self.absolute_uri(self.uri_for_api("admModerateRedirect"))
+        if not req.req_data.get("reason"):
```
robnagler, line 153: This is more robust `len(get("reason", "").strip()) == 0`. When validating user input, make sure it's not a blank.

sirepo/package_data/static/js/sirepo-components.js
robnagler: What we want is to enable the button only when the user enters something, similar to the way "Save changes" works.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3636,20 +3636,24 @@ SIREPO.app.directive('moderationRequest', function(appState, errorService, panel
...
             $scope.submitted = false;
+            $scope.disableSubmit = false;
             $scope.submitRequest = function () {
                 const handleResponse = (data) => {
                     if (data.state === 'error') {
                         errorService.alertText(data.error);
                     }
                     $scope.submitted = true;
                 };
+                if (!$scope.data.reason) {
```
robnagler, line 3654: This needs to be even more robust than the server-side, which can allow certain junk. We may want the user to enter more than "x", for example.

robnagler (changes requested):

sirepo/package_data/static/js/sirepo-components.js
gurhar1133: What type of validation would you like for the reason? A min length? Something else?

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3659,6 +3663,23 @@ SIREPO.app.directive('moderationRequest', function(appState, errorService, panel
                     }
                 );
             };
+
+            function reasonOk(reason) {
+                if (reason && reason.length > 4) {
```
robnagler, line 3668: `return reason && reason.length > 4` except that you need to strip reason, because it might contain blanks on either end.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3659,6 +3663,23 @@ SIREPO.app.directive('moderationRequest', function(appState, errorService, panel
...
+                    return true;
+                }
+                return false;
+            }
+
+            function validateReason() {
+                if (reasonOk($scope.data.reason) && !$scope.submitted) {
```
robnagler, line 3675: `$scope.disableSubmit = !reasonOk($scope.data.reason) || $scope.submitted`

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3636,20 +3636,24 @@ SIREPO.app.directive('moderationRequest', function(appState, errorService, panel
...
         controller: function(requestSender, $scope) {
             $scope.data = {};
             $scope.submitted = false;
+            $scope.disableSubmit = true;
             $scope.submitRequest = function () {
                 const handleResponse = (data) => {
                     if (data.state === 'error') {
                         errorService.alertText(data.error);
                     }
                     $scope.submitted = true;
```
robnagler, line 3652: Seems like this should be set on the send side, not the response. As soon as you validate the reason, set submitted to true. submitted is from the button perspective not from actually sending to the server. We want to avoid duplicate sends.

robnagler (changes requested):

sirepo/package_data/static/js/sirepo-components.js
gurhar1133: Good point

sirepo/package_data/static/js/sirepo-components.js
gurhar1133: Gotcha

robnagler (approved):

PR #5360, opened 2023-01-11
fix #5359 react improvements

### LINKED ISSUE #5359: minor react improvements
- align form labels with input text
- remove internal Containers
- use bootstrap classes for form layout
- add shown config to FieldGrid
- improve field grid layout

moellep:
- align form labels with input text
- remove internal Containers
- use bootstrap classes for form layout
- add shown config to FieldGrid
- improve field grid layout


react/src/layout/form.tsx
```diff
@@ -79,17 +79,20 @@ export class FieldGridLayout extends Layout<FieldGridConfig, {}> {
         let schema = useContext(CSchema);
         let store = useStore();
 
+        if (! useShown(this.config.shown, true, formState, ValueSelectors.Fields)) {
```
garsuga, line 82: I am concerned with using hooks inside conditionals. Especially as we have new people working on React. This works because there are no other elements of the conditional but things like short-circuiting will cause an error. I think for hooks specifically it might be better to have them on their own lines

garsuga (dismissed):

react/src/layout/form.tsx
moellep: Yes, that makes sense - I've changed this with b239f85

PR #5368, opened 2023-01-12
fix #5355: use schema routes to format urls

### LINKED ISSUE #5355: 20230118 react use schema local routes and global routes for url formatting

garsuga:
@robnagler I am adding you to see if this approach for routing works better for you. The format we used in the schema happens to be the same as the one used by ReactRouter so we can input our routes directly into that library. I think it is a step in the right direction

@moellep can you help me look for regressions? I noticed the report viewing on the JSPEC file dropdown is (no longer)? working but it might be better solved in another issue.


garsuga:
I added the react prefix removal here. It was a very small change and required this PR to work anyways.

moellep:
I think the jspec file dropdown is an independent bug - I have some changes related to that in an omega branch which I'll create a pull request for soon.

The only problem I could find with the routing is the "Python Source" link in the cog wheel is no longer working.

garsuga:
Yes good catch, I need to fix that. That one uses one of the urls that has two optional parameters I believe. Maybe it would be more efficient to leave it as a regression until I can fix those in my next PR.

moellep:
Yes, that's fine to leave the python source broken for now. I'll approve this PR

PR #5401, opened 2023-01-18
fix #5400 show alert on javascript module load error

### LINKED ISSUE #5400: show error message if javascript library fails to load
The app fails silently and later crashes if a javascript library fails to load. Instead, it should show a user alert asking for the user to reload the page.

moellep:


moellep:
I'm closing this PR without merging. The root cause of the error is on the server, and pushing the problem to the client doesn't help. We will monitor the server logs for this failure instead.

PR #5405, opened 2023-01-18
Fix #5397 save spa session data in memory by user (not session_id)

### LINKED ISSUE #5397: 20230118 don't write spa_session
Drop the feature for now.

#5390 #5393

robnagler:


moellep:
When I log the contents of _DB after each request on dev, I'm seeing the size grow. Logging in I see 2 entries. Logging out and then back in as a different user, I see 8 entries. Logging out again and back in I see 14. Is that normal behavior?

moellep:
Also, reloading the app, or leaving the app and coming back via the app list creates 2 new _DB entries. Maybe that is expected behavior because it is stored in the header and not the cookie.

robnagler:
Yes, I was testing, too, and it is. Angular reloads on login so that clears the SPA. Each SPA (tab) has to have a different id for this to work in the way it was intended. There needs to be active garbage collection. With websockets that will be easy, if a socket drops, the spa_session will be removed.

For the purposes of beginSession though, we can manage this by uid. I will change the PR to work that way for the _DB case.

robnagler:
@moellep please review (you didn't actually review the first time so I can't rerequest). I'm going to merge and push to alpha.

PR #5416, opened 2023-01-19
radiasoft/raydata#65: authenticate requests with scan monitor

e-carlin:


sirepo/template/raydata.py
```diff
@@ -38,10 +38,16 @@ def stateless_compute_scan_fields(data):
...
 def _request_scan_monitor(data):
+    c = sirepo.feature_config.for_sim_type(SIM_TYPE)
     try:
         r = requests.post(
-            sirepo.feature_config.for_sim_type(SIM_TYPE).scan_monitor_url,
+            c.scan_monitor_url,
             json=data,
+            headers=PKDict(
+                {
+                    sirepo.util.AUTH_HEADER: f"{sirepo.util.AUTH_HEADER_SCHEME_BEARER} {c.scan_monitor_api_secret}"
```
robnagler, line 48: Why is it in sirepo.util?

robnagler (changes requested):

rorour (approved):
Looks good to me

sirepo/template/raydata.py
e-carlin: It is shared with sim_db_file so it needs to be somewhere both can access. Is there somewhere else you think would be better?

robnagler: sirepo.const might be better since it is a constant, not a utility. Also, you might say HTTP_AUTH_SCHEME_BEARER. It's not technically a "header", rather a value. A bit arbitrary, and not important.

e-carlin: Ok, ended up making a function for it in sirepo.util and moving the const's to sirepo.const 2f0e408c9

e-carlin:
@robnagler ping

sirepo/const.py
```diff
@@ -4,8 +4,23 @@
...
+AUTH_HEADER = "Authorization"
+
+#: http auth header scheme bearer
+AUTH_HEADER_SCHEME_BEARER = "Bearer"
+
+#: POSIT: Matches anything generated by `sirepo.job.unique_key`
+_UNIQUE_KEY_CHARS_RE = r"\w+"
+
+AUTH_HEADER_RE = re.compile(
+    AUTH_HEADER_SCHEME_BEARER + r"\s(" + _UNIQUE_KEY_CHARS_RE + ")",
```
robnagler, line 20: `\s+`. make f-string

sirepo/template/raydata.py
robnagler: The "POSIT" bothers me. These are all coupled so they should be in one place. I didn't realize the generation was going on. Seems like unique_key and the rest of it should be in util. It's a bit arbitrary, but the context matters.

robnagler (changes requested):
see comment

sirepo/const.py
e-carlin: 268b2c331

e-carlin:
> The "POSIT" bothers me. These are all coupled so they should be in one place. I didn't realize the generation was going on. Seems like unique_key and the rest of it should be in util. It's a bit arbitrary, but the context matters.

I think you were looking at old code. Take a look at the updated and let me know if the problem still applies.

sirepo/const.py
```diff
@@ -4,8 +4,23 @@
...
+import re
 from pykern.pkcollections import PKDict
 
+#: Http auth header name
+AUTH_HEADER = "Authorization"
+
+#: http auth header scheme bearer
+AUTH_HEADER_SCHEME_BEARER = "Bearer"
+
+#: POSIT: Matches anything generated by `sirepo.job.unique_key`
```
robnagler, line 16: This posit is the one that bothers me

e-carlin: Thanks 2197e6080

e-carlin:
@rorour @robnagler I revived this now that the scan_monitor lives in sirepo. Please review again.

sirepo/raydata/scan_monitor.py
```diff
@@ -11,9 +11,9 @@
 import databroker
 import databroker.queries
 import io
-import os
 import requests
 import sirepo.feature_config
+import sirepo.feature_config
```
rorour, line 16: duplicate import

sirepo/util.py
```diff
@@ -30,11 +30,23 @@
 AUTH_HEADER = "Authorization"
 
 #: http auth header scheme bearer
-AUTH_HEADER_SCHEME_BEARER = "Bearer"
+_AUTH_HEADER_SCHEME_BEARER = "Bearer"
```
robnagler, line 33: I feel like this should be somewhere else. It's very specific to http so maybe http_util.py.

rorour (changes requested):
looks good to me aside from the duplicate import

sirepo/util.py
```diff
@@ -447,6 +463,10 @@ def to_comma_delimited_string(arr):
     return ",".join([str(x) for x in arr])
 
 
+def unique_key():
```
robnagler, line 466: This deserves a comment since it's been more widely published. Clients can assume it is a "reasonably sized" web- and file-system safe token that is randomly generated from enough bits that is globally unique. It is our version of UUID4 without the structure. We might make some guarantee, e.g. less than 128 characters, so it can be stored in a database with some type of structure.

robnagler (changes requested):

sirepo/util.py
e-carlin: Is it worth creating http_util.py just for this? It is only used in this file. Maybe move it and the uses to http_util.py?

sirepo/util.py
e-carlin: 97e3e0c25

sirepo/raydata/scan_monitor.py
e-carlin: 97e3e0c25

rorour (approved):

robnagler:
@e-carlin is this relevant?

e-carlin:
Yes, but the raydata code has changed quite a bit since I last reviewed. I'll update and report back.

e-carlin:
https://github.com/radiasoft/sirepo/issues/7142

PR #5422, opened 2023-01-20
Fix #5271 no rotation on symm shape

### LINKED ISSUE #5271: no object rotation applied to symmetry object
When applying a rotation to a racetrack coil and then applying a symmetry transform to the group that the coil is a part of the symmetry visualization in the Design tab does not show the coil rotated properly.
image
Original, rotated coil on the right and improperly rotated reflection coil on the left.
It looks like this is probably just a visualization bug as 3D Visualization shows the expected result.
image

mkeilman:
This is a hefty update that reworks shape drawing. The salient addition is the concept of "object views" that keep track of shapes of objects as viewed along each axis. They also pull the shape transform handling out of RadiaSourceController.

Regarding this issue specifically, the svg "transform"  attribute is not sophisticated enough for our purposes. We use affine transforms of the 3d points here. That said, some notes:

- "Linked" shapes (like those in symmetries) no longer update live as the user drags the parent shape (this will be re-added)
- It turns out that "correctly" ordering the points on an arbitrary polygon is np-complete. Thus for now rotations of shapes around axes other than 'z' are not reproduced when viewed along other axes - the views are just projections of the "footprint"


moellep:
When I created a simple geometry with transforms from scratch, I'm seeing the error "'error
': ''PKDict' object has no attribute 'model''" in the server logs when it tries to render the geometry. Attached is my sim zip.
[testx.zip](https://github.com/radiasoft/sirepo/files/10492236/testx.zip)

mkeilman:
> When I created a simple geometry with transforms from scratch, I'm seeing the error "'error ': ''PKDict' object has no attribute 'model''" in the server logs when it tries to render the geometry. Attached is my sim zip. [testx.zip](https://github.com/radiasoft/sirepo/files/10492236/testx.zip)

Appears to be limited to clones. Fix incoming

mkeilman:
ping - I'd like the next alpha to include this

PR #5423, opened 2023-01-21
fix radiasoft/raydata#42: download PDF of analysis

rorour:
depends on https://github.com/radiasoft/raydata/pull/68


sirepo/feature_config.py
```diff
@@ -154,6 +154,11 @@ def b(msg, dev=False):
             "codes that contain proprietary information and authorization to use is granted through oauth",
         ),
         raydata=dict(
+            pdf_temp_dir=(
+                "/home/vagrant/src/radiasoft/raydata/run/",
```
rorour, line 158: How do I do this correctly? I couldn't get it to work without hard coding the whole path

robnagler: For development, you should be relative to srdb.root, which should be setup like it is done in pkcli.service. The default value could be RequiredUnlessDev("run") or something like that. For production, the value would absolutely configured. The configuration parser for pdf_temp_dir would look to see if the directory is relative, then it would be joined with srdbroot. If absolute, it would be used as is. In both cases, the directory must exist(?).

rorour: ad8603b44b747bc7417cb275935fae525a2aa5e7

moellep:
@rorour could you check on this branch - it says it needs to resolve conflicts within the raydata files.

rorour:
https://github.com/radiasoft/sirepo/pull/5488 is higher priority and will also cause conflicts. I will resolve conflicts here when that is merged

e-carlin:

rorour:
@e-carlin awaiting review from you I thought

e-carlin:
Ok, sorry for the delay. Can you fix the conflicts and then I'll review?

rorour:
@e-carlin ready for review

sirepo/package_data/static/js/raydata.js
```diff
@@ -377,11 +380,14 @@ SIREPO.app.directive('scansTable', function() {
             $scope.images = null;
             $scope.noScansReturned = false;
             $scope.orderByColumn = 'start';
+            $scope.pdfSelectAllScans = false;
+            $scope.pdfSelectedScans = [];
```
e-carlin, line 384: I think making this a dict instead of an array will make the operations on it easier.

sirepo/package_data/static/js/raydata.js
```diff
@@ -305,10 +305,12 @@ SIREPO.app.directive('scansTable', function() {
         template: `
             <div data-show-loading-and-error="" data-model-key="scans">
               <div>
-                <button class="btn btn-info btn-xs" data-ng-click="addColumn()" style="float: right;"><span class="glyphicon glyphicon-plus"></span></button>
+                <button class="btn btn-info btn-xs" data-ng-click="addColumn()" style="float: right; margin:5px;"><span class="glyphicon glyphicon-plus"></span></button>
+                <button class="btn btn-info btn-xs"  data-ng-show="showPdfColumn" data-ng-click="downloadSelectedAnalyses()" style="float: right; margin:5px;">Download Selected Analysis PDFs</button>
```
e-carlin, line 309: I think this should only become visible when there are pdfs selected. The UI is a little confusing now where you can select all pdfs and hit this button but nothing happens.

Screen Shot 2023-03-14 at 4 41 04 PM

sirepo/pkcli/job_cmd.py
```diff
@@ -105,6 +105,14 @@ def _op(expect_file):
         ).does_api_reply_with_file(msg.api, msg.data.method)
         if x:
             with simulation_db.tmp_dir(chdir=True) as d:
+                if sirepo.feature_config.for_sim_type(template.SIM_TYPE).pdf_tmp_dir:
```
e-carlin, line 108: Since this is generic let's make the name for the dir something more generic. Maybe file_reply_tmp_dir instead of pdf_tmp_dir.

In addition, I don't think this needs to be nested under simulation_db.tmp_dir. In this case we are using our own dir so we don't need the simulation_db.tmp_dir

sirepo/template/raydata.py
```diff
@@ -4,9 +4,13 @@
 :copyright: Copyright (c) 2021 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+from pykern import pkcompat
```
e-carlin, line 7: not used

sirepo/package_data/static/js/raydata.js
```diff
@@ -377,11 +380,14 @@ SIREPO.app.directive('scansTable', function() {
...
             $scope.noScansReturned = false;
             $scope.orderByColumn = 'start';
+            $scope.pdfSelectAllScans = false;
+            $scope.pdfSelectedScans = [];
             $scope.reverseSortScans = false;
             $scope.runLogModalId = 'sr-view-log-iframe-' + $scope.analysisStatus;
             $scope.runLogScanId = null;
             $scope.scans = [];
             $scope.selectedScan = null;
+            $scope.showPdfColumn = $scope.analysisStatus === 'executed';
```
e-carlin, line 390: This addition made me look at this entire directive and wonder if there is a way to split it up. The controller is 267 lines long which is big enough to be fairly unwieldy. We don't have to do it now but as you are working in the code it would be good to keep in the back of your mind how to split things up.

One idea I have is to split some of the business logic and presentation logic up. For example, we could have a wrapping directive that knows how to get the scans (sendScanRequest). Then we could pass the scans into a directive that knows how to present them (show the table, sort the columns, etc). 

I don't have a good idea how to separate out things like the pdf. Only some tables want it so it would be nice to separate out the logic/presentation for it into something that knows just how to deal with pdfs. Then it could be injected/attached to the scans table in that instance. Angular doesn't provide much in the way for doing things like this...

sirepo/package_data/static/js/raydata.js
```diff
@@ -488,6 +499,26 @@ SIREPO.app.directive('scansTable', function() {
                 $scope.saveColumnChanges();
             };
 
+            $scope.downloadSelectedAnalyses = () => {
+                if (! $scope.pdfSelectedScans.length) {
```
e-carlin, line 503: This check can be removed once we only show the button to download once scans are selected.

sirepo/package_data/static/js/raydata.js
```diff
@@ -547,6 +586,27 @@ SIREPO.app.directive('scansTable', function() {
                 $scope.reverseSortScans = ! $scope.reverseSortScans;
             };
 
+            $scope.togglePdfSelectAll = () => {
+                $scope.pdfSelectedScans = [];
+                if (! $scope.pdfSelectAllScans) {
```
e-carlin, line 591: This logic took me a second to understand (and I'm still not sure I have it right). But I think that line 598 should be moved to the top and then the `!` can be removed here. I think that will make it clearer that in the case where pdfSelectAllScans is true we want to add all scans to pdfSelectedScans.

sirepo/template/raydata.py
```diff
@@ -4,9 +4,13 @@
...
+from pykern import pkio
 from pykern import pkjson
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp, pkdlog, pkdformat, pkdexc
+from sirepo.template import template_common
+import base64
```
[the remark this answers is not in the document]
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

sirepo/package_data/static/js/raydata.js
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

sirepo/pkcli/job_cmd.py
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

sirepo/template/raydata.py
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

sirepo/package_data/static/js/raydata.js
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

sirepo/package_data/static/js/raydata.js
rorour: 34d5c6d1b0aa5be7fb841e47a6fbf0fc34ecd9e9

e-carlin (approved):
Looks good!

PR #5427, opened 2023-01-25
Fix #5181: version check on startup

### LINKED ISSUE #5181: 20230131 Add version check on startup
When running standalone in Docker (sirepo.run) or other unmanaged configs, the server should check its version against the public server's version. The GUI should notify the (usually guest) user that there's a newer version available, if there is.

gurhar1133:
URL for api request needs to be unhardcoded. But now version check will alert the user if version not up to date.


robnagler:
This is no longer a priority.

Issue #5326, opened 2023-01-04
20230228 tornado sending data without blocking


robnagler:
Serving data:

https://bhch.github.io/posts/2017/12/serving-large-files-with-tornado-safely-without-blocking/

And using sendfile:

https://github.com/giampaolo/pyftpdlib/blob/6165a40fd4f7c7cd2d549dbb816b3208091ffdff/pyftpdlib/handlers.py#L658

Issue #5199, opened 2022-11-27
fork all APIs to make them async


robnagler:
Tornado requires all APIs to be async. We will need to fork copies of all APIs until Tornado is in production. We can name them something like async_job_api, async_auth_api, etc.

PR #5436, opened 2023-01-27
Fix #4388 run tornado or flask based server

### LINKED ISSUE #4388: Replace Flask with Tornado
Do not change supervisor structure. Just replace Flask in server.py.

robnagler:
I think this is safe to merge, and completes the spirit of the task of replacing flask with tornado so I bound it to #4388:

* Run: `SIREPO_PKCLI_SERVICE_TORNADO=1 sirepo service http`
* Test: SIREPO_PKCLI_SERVICE_TORNADO=1 bash test.sh
* This does not fix runSimulation (#5199) or blocking I/O (#5326)


sirepo/reply.py
```diff
@@ -367,13 +377,105 @@ def render_static_jinja(self, base, ext, j2_ctx, cache_ok=False):
...
+        return r
+
+    def _disposition(self, disposition):
+        self.header_set(
+            _DISPOSITION, f'{disposition}; filename="{self.__attrs.download_name}"'
         )
         return self
 
+    def _download_name(self, filename):
+        f = re.sub(r"[^\w\.]+", "-", filename).strip("-")
```
e-carlin, line 473: A name for the rexeg would be helpful to understand why it is being performed.

sirepo/request.py
```diff
@@ -18,38 +19,116 @@
...
+        r = internal_req.request
+        sreq = _SRequest(
+            http_authorization=_parse_authorization(r.headers.get("Authorization")),
+            http_headers=r.headers,
+            http_method=r.method,
+            http_request_uri=r.path,
+            http_server_uri=f"{r.protocol}://{r.host}/",
+            internal_req=internal_req,
+            remote_addr=r.remote_ip,
+            _body_as_bytes=lambda: internal_req.request.body,
```
e-carlin, line 62: Why are these public and the other fields aren't? Why does something like http_authorization not need an accessor method but  body_as_bytes does?

e-carlin (approved):
LGTM. A couple minor comments / questions.

sirepo/request.py
robnagler: > Why are these public and the other fields aren't? 

_body_as_bytes, etc. are only referenced inside request.py. The others are accessed outside request.py.

> Why does something like http_authorization not need an accessor method but body_as_bytes does?

body_as_bytes could be an accessor. A refactoring that could be done, but I'll leave as now. Perhaps documents better.

I like that _parse_authorization is always called to validate the header on every request at request time (before other code gets to execute). flask.request.authorization validates at request startup, too. 

I was striking a balance between too much wrapping so I could get things done. In general internal_req should be used sparingly and that's why it is called awkwardly internal_req.

robnagler: I simplied body_as_bytes d57937d0

sirepo/reply.py
robnagler: d57937d0

sirepo/reply.py
```diff
@@ -367,13 +377,105 @@ def render_static_jinja(self, base, ext, j2_ctx, cache_ok=False):
...
+                    if c[k]:
+                        r.append("Secure")
+                else:
+                    raise AssertionError(f"unhandled cookie attr={k}")
+            resp.set_header("Set-Cookie", "; ".join(r))
+
+        def _file(resp):
+            a = self.__attrs
+            c = a.content
+            resp.write(c.handle.read())
```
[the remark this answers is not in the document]
robnagler: Good call. Added it to destroy d57937d0

sirepo/auth/__init__.py
```diff
@@ -80,22 +80,22 @@
...
 
 
-def init_quest(qcall, top_level_call_api=False):
+def init_quest(qcall, internal_req=None):
     """Under development
 
     Only use as `init_quest(qcall)`.
 
-    top_level_call_api only used by uri_router.
+    internal_req used by uri_router.
```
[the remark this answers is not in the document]
robnagler: d57937d0

sirepo/request.py
e-carlin: > _body_as_bytes, etc. are only referenced inside request.py. The others are accessed outside request.py.

I believe `http_headers` is only accessed inside request.py. The others are accessed outside but through a method that does nothing else than call `self._<name>()`. So, I'm still not sure what the difference is between the two? Why not have them all be public or private?

> In general internal_req should be used sparingly

It would be helpful to document this somewhere in the code.

e-carlin: Sorry I left my review before seeing your update. But I think my points still stand.

robnagler: http_headers should be made private. `http_headers_uget` is not the same `body_as_bytes`. In this case, there'd be lambda's in three places that would do the gets. `_http_headers` implements `get`, but the object is different in the three cases.

PR #5445, opened 2023-02-01
fix #4959: create layouts for arranging child layouts

### LINKED ISSUE #4959: react layout handling
we need a way to handle the layout of panels and abnormal layouts like large 3d displays on the page in react. this needs to be able to handle hiding/showing of panels and other elements

garsuga:
Adds a few useful layouts like vertical and horizontal stacks and a waterfall layout to arrange the panels.

Also formats all of the react schema, they were getting very messy with all the edits.


garsuga:
CI tests got rate limited, haha

robnagler:
> CI tests got rate limited, haha

Need to fix https://github.com/radiasoft/download/issues/406 because GitHub rate limits on IPs. 

Workaround: rerun the job manually. I did this for you.

moellep:
I'm testing this PR now and it looks really good. I think the breakpoints are too small, and I think it should be sm:1, md:2, xl:3 which matches the existing app. That way you don't get into bad layouts like this:
x
or this:
y

moellep:
There are a few other small issues, like the "Ion Particle Animation" panel appearing twice in JSPEC, but I think this can be changed in the config. 

I'll make a few other layout changes to JSPEC now and check-in to this branch this morning.

PR #5450, opened 2023-02-03
Fix #5449 auth_db.rename_table works

### LINKED ISSUE #5449: expiration column db upgrade fails
Running the _20220609_add_expiration_column_to_user_role_t fails when starting the dev http server:
```TypeError: execute() got an unexpected keyword argument 't'```

To reproduce remove the entry from the db_upgrade_t table, and rename the existing user_role_t.expiration to another name (our version of sqllite does not support "drop column").

### PR DESCRIPTION
**Pull request by robnagler**
- db_upgrade test
- auth_db.drop_table
- _20230203_drop_spa_session replaces _20221120_rename_session_to_spa_session

### CODE REVIEW

**Comment by moellep** [comment]
Also the line:
```python
def _execute_sql(self, text, **kwargs):
        return self.execute(sqlalchemy.text(text + ";"), **kwargs)
```
should be changed to remove the kwargs. That line is wrong, even if we did want kwargs to be passed, because it should be passed as kwargs (a dict), not **kwargs.

I did some testing with sqlalchemy and found that the bind params seem to only work with parts of the sql statement. 
This works: ```execute('SELECT * FROM user_role_t WHERE uid = :uid', {uid = '00000000'})```
but this fails: ```execute('ALTER TABLE :t ADD foo DATETIME', {t='user_role_t'})```
with the error: 
```
(sqlite3.OperationalError) near "?": syntax error
[SQL: ALTER TABLE ? ADD foo3 DATETIME]
```

PR #5452, opened 2023-02-03
Fix #5451 replies need manual memory management

### LINKED ISSUE #5451: tornado home page files fail with no handle

robnagler:
Flask accepts the file handle on replies and doesn't need management. Tornado does not, and we process. However we need to free the handle after it is read. With a nested call_api, the handle is carried across qcalls and therefore needs to be manually freed. Required modifying all call_api calls. 

Also fixed exception handling during the tornado write calls.


sirepo/template/srw.py
```diff
@@ -609,10 +614,14 @@ def import_file(req, tmp_dir, qcall, **kwargs):
             except Exception:
                 pass
         raise
+    finally:
+        if r:
+            r.destroy()
+            r = None
```
rorour, line 620: r isn't used again. Why is `r = None` necessary?
Even in cases where r is set to another response (L574 & 596) why is `r = None` necessary?

rorour:
Do I need to use a flag for tornado to run with these changes?

robnagler:
> Do I need to use a flag for tornado to run with these changes?

Yes:

SIREPO_PKCLI_SERVICE_TORNADO=1 sirepo service http

sirepo/template/srw.py
rorour: @robnagler

rorour:
Runs fine. I was waiting on a response to the comment above

sirepo/template/srw.py
robnagler: I was just repeating a garbage collection pattern that was used above. It's important that r not be used again. If new code is added some day, I wanted `r` to be invalid at that point.

rorour: What would you think about having `destroy()` return `None` and doing `r = r.destroy()` ? Something about needing 2 actions for proper garbage collection in every use of r seems terribly inefficient to me. But I'm guessing you might think my suggestion is unintuitive and that it's better to explicitly show what's happening to r in every case

robnagler: Yes, I think in this case, it's clearer to have two lines, even if it is extra code. This particular function is quite odd should be done differently. All file imports should run in the agent. When it was written, we didn't have stateless compute. What it's doing is creating a simulation so it can call run simulation which does the actual import.

PR #5460, opened 2023-02-07
Fix #5380: SDDS Plotting Refactor

### LINKED ISSUE #5380: 20230215 Common SDDS plotting prep
Many apps (elegant, opal, jspec, omega) need to be able plot data from a sdds file, either as a line plot or a heatmap.

The code could be consolidated into a utility class. Something like:

```python
class SDDSUtil():
    def __init__(filename):
        ...
    def heatmap(plot_attrs):
        ...
    def lineplot(plot_attrs):
        ...
```
where plot_attrs is a PKDict of values, x, x_label, title, ...

For example, in template.jspec, much of the sim_frame_particleAnimation() method could be replaced with this:
```python
    ...
    return SDDSUtil(filename).heatmap(PKDict(
        x=xfield,
        x_label=_field_label(xfield),
        y=yfield,
        y_label=_field_label(yfield),
        title="Ions at time {:.2f} [s]".format(time),
    )
```

Consolidating the code should make it easier to fix bugs in one place. For example, only template.opal sim_frame_plot2Animation() correctly trims points to match the values from the first "x" read (the file could be growing as read, so subsequent "y" reads could have more values).

x_label could be omitted and the label could be constructed from the label and units in the sdds file. Also, units could be automatically added from the sdds file if present. 

The SDDSUtil() class should use the sdds_util and template_common libraries, but eventually it should replace them.

rorour:


sirepo/template/sdds_util.py
```diff
@@ -41,6 +42,90 @@
...
+            col = extract_sdds_column(self.filename, col_name, 0)
+            if col.err:
+                return col.err
+            plot = PKDict(
+                points=col["values"],
+                label=plot_attrs.frame_args[f],
+                col_name=col_name,
+            )
+            if f == "x":
+                plot.x = True
```
rorour, line 99: I don't like this, but I was trying to find a way to handle the different case for formatting x in _sdds_report

moellep: This is good progress. Here are a few items which should clean it up a bit:

The API to line plot is called like this:
```python
    return sdds_util.SDDSUtil(str(frame_args.run_dir.join(_OPAL_SDDS_FILE))).lineplot(
        PKDict(
            x=("x",),
            y=("y1", "y2", "y3"),
            ...
        )
    )
```
That may confuse people into thinking there could multiple "x" values allowed which isn't the case. Instead, the API should probably be changed to something like this, which is very explicit that there is one x values allowed and multiple y values.
```python
    return sdds_util.SDDSUtil(str(frame_args.run_dir.join(_OPAL_SDDS_FILE))).lineplot(
        PKDict(
            x_field="x",
            y_fields=("y1", "y2", "y3"),
            ...
        )
    )
```

I've done some testing with jspec and the bug related to sorting vtrans seems to be fixed in the physics code, and I can't find a case where the V_trans values are out of order. So we can remove ```def _resort_vtrans(...)``` from template.jspec and remove the ```v_trans_hook``` parameter from SDDSUtil.

The code 
```python
    if f == "x":
        plot.x = True
```
can be removed from SDDSUtil if the line is changed in template.jspec from:
```python
    def _format_plot(plot, sdds_units):
        if "x" in plot and plot.x:
            ....
```
to
```python
    def _format_plot(plot, sdds_units):
        if x_field == plot.col_name:
            ....
```

also, in SDDSUtil, the line:
```python
            if f == "x":
                x = plot
```
could be changed to this, avoiding the static literal:
```python
            # after plot_attrs.x is renamed to plot.attrs.x_field
            if f == plot_attrs.x_field:
                x = plot
```

moellep (changes requested):

moellep:
Here's another thought - I think we should remove the "frame_args" from the SDDSUtil interface and replace it with "model" which can be constructed from the frame_args. That way the SDDSUtil could be used for plots which don't come through the sim_frame methods. (Automatic plots don't have a frame args, just a model - manual plots have a frame_args which have values which may be different from the underlying model, which is why the code merges it together).

So we could have a new template_common helper function:
```python
def model_from_frame_args(frame_args):
    res = frame_args.sim_in.models[frame_args.frameReport]
    res.update(frame_args)
    return res
```

And the lineplot interface would look like this:
```python
    return sdds_util.SDDSUtil(filename).lineplot(
        PKDict(
            x_field="x",
            y_fields=("y1", "y2", "y3"),
            format_col_name=_map_field_name,
            model=template_common.model_from_frame_args(frame_args),
            format_plot=_format_plot,
        )
    )
```

Also, I'd make "format_col_name" optional in SDDSUtil, because in many cases the field value will match up with the sdds column name.

sirepo/template/sdds_util.py
rorour: bc2d5a6844bd5910a78960ca4e509e4f80fa3deb

moellep:
This looks great - I've added a unit test which can be expanded as we find failures.

Try refactoring template.elegant _extract_report_data() and see if the new module can be applied. One thing which will need to be added is a page_index attribute, because many elegant plots have paged datasets (see the elegant sim "bunchComp - fourDipoleCSR" plots for an example).

Another thing to consider would be to simplify the interface by using default values for x_field and y_fields. Defaulting to "x" and ("y1", "y2", "y3") in SDDSUtil would remove those args from jspec and opal. Basically, you want to make the API calls as sparse as possible if there is duplication across apps. Add some documentation to SDDSUtil.lineplot which documents the available arguments and default values.

moellep (approved):

PR #5469, opened 2023-02-09
fix #4945 react colorbar

### LINKED ISSUE #4945: 20230205 react heatplot colorbar
Add on to #4927
Add a colorbar to the side of the heatmap plot, except for the "high contrast" color map, in which case it should be hidden.

moellep:
- added AxisRight support to axis.ts
- compute zRange and dataId for HeatPlotConfig
- cache heatplot canvas
- added colorbar component and use w/heatplot
- added sirepo color maps
- improve axis number format
- reduced axis margins
- fix graph2d clipping bug using wrong clipPath
- pass report model down to the plots


moellep:
@garsuga please review, especially the parts in layout/report.tsx where the report model is combined into currentFrame.data.

garsuga:
I refactored a couple things that were causing me issues. It helps to migrate the remaining js files to typescript. Could you check and make sure everything still works? Were there any other places you wanted me to look at besides the model stuff?

garsuga:
Also with the HeatPlotImage component, we should definitely look at a different approach. With the way it is written it will always we a little slow. I think we should try to separate the "moving" from where we raster the image using something like portals.

garsuga (approved):

PR #5487, opened 2023-02-15
fix #5471 new API for bluesky sims

### LINKED ISSUE #5471: 20230215 run bluesky simulation without agent
Bluesky would be much faster avoiding the agent and return the result data file directly from the call.

moellep:


robnagler:
This will be solved with #6784.

PR #5488, opened 2023-02-16
fix radiasoft/raydata#75: run analysis on any scan

rorour:
depends on https://github.com/radiasoft/raydata/pull/76


sirepo/package_data/static/js/raydata.js
```diff
@@ -14,6 +14,9 @@ SIREPO.app.config(() => {
         <div data-ng-switch-when="RecentlyExecutedScansTable" class="col-sm-12">
           <div data-scans-table="" data-model-name="modelName" data-analysis-status="recentlyExecuted"></div>
         </div>
+        <div data-ng-switch-when="RunAnalysisTable" class="col-sm-12">
+          <div data-scans-table="" data-model-name="modelName" data-analysis-status="all"></div>
```
robnagler, line 18: This type of code is problematic, because the value `all` is not unique. Maybe `allStatuses`

sirepo/package_data/static/js/raydata.js
```diff
@@ -502,6 +513,22 @@ SIREPO.app.directive('scansTable', function() {
...
+            $scope.runAnalysis = (scan, event) => {
+                event.stopPropagation();
+                requestSender.sendStatelessCompute(
+                    appState,
+                    (json) => {},
+                    {
+                        method: 'run_analysis',
+                        args: {
+                            catalogName: appState.models.catalog.catalogName,
+                            uid: scan.uid,
```
robnagler, line 525: `uid` is another confusing value. It has nothing to do with our normal `uid` values. Too much to change, but reading the code is confusing for a newbie.

robnagler (changes requested):

sirepo/package_data/static/js/raydata.js
rorour: 9919d9721389a3c534885a7bb02158f5c7e72625

e-carlin:

rorour:
Awaiting review

robnagler:
I didn't have time. lgtm, but I don't have context so @e-carlin you should review.

e-carlin (changes requested):
The code looks good. I noticed a few usability problems while clicking around. Lmk if you can't reproduce any of them and we can troubleshoot together:

The executed tab seems to be broken. I entered the times, clicked save changes, and got back this error.
```
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/tornado/web.py", line 1713, in _execute
    result = await result
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 218, in post
    self.write(await self._incoming(PKDict(pkjson.load_any(self.request.body))))
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 215, in _incoming
    return getattr(self, "_" + req.pkdel("method"))(req)
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 287, in _get_scans
    if m.start() >= a.searchStartTime and m.stop() <= a.searchStopTime:
TypeError: '<=' not supported between instances of 'tuple' and 'int'
```

On the Run Analysis tab the status is a little “i”. When I hover over it I don’t get a tooltip telling me what that icon means. If you hover over the status on the other tabs you’ll get a tooltip with what the status is. I think this should be added to the run analysis tab.

On the Run Analysis tab you shouldn’t be able to click on a row for a scan that has never been run and get the output.


I can queue the scan to be run for analysis multiple times. I think we should probably only allow a scan to be queued if it is in one of the stopped states. Otherwise things like the status won’t be right (a scan can be running and queued).

There is some sort of asynchrony bug going on. I queued the same scan multiple times (~10), refreshed the “Run Analysis” tab, then clicked on “View Log” for the scan I ran analysis for. It took ~15 seconds for the log to come back which I think is too long. Also, if I changed the start/stop time on the Run Analysis tab it took ~10s for the “Save Changes” button to appear. In addition after it came back I saw this error
```
HTTPServerRequest(protocol='http', host='127.0.0.1:9001', method='POST', uri='/scan-monitor', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/tornado/web.py", line 1713, in _execute
    result = await result
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 218, in post
    self.write(await self._incoming(PKDict(pkjson.load_any(self.request.body))))
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 215, in _incoming
    return getattr(self, "_" + req.pkdel("method"))(req)
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 239, in _analysis_run_log
    p = _Analysis.analysis_output_dir(req.uid).join("run.log")
  File "/home/vagrant/src/radiasoft/raydata/raydata/pkcli/scan_monitor.py", line 109, in analysis_output_dir
    return cfg.db_dir.join(uid)
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/py/_path/local.py", line 336, in join
    strargs = [fspath(arg) for arg in args]
  File "/home/vagrant/.pyenv/versions/3.7.2/envs/py3/lib/python3.7/site-packages/py/_path/local.py", line 336, in <listcomp>
    strargs = [fspath(arg) for arg in args]
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I think the “Run Analysis” button should be grayed out after click until the status changes to queued. Then the button should either continue to be disabled until the scan finishes or it should be hidden altogether.

robnagler (approved):
I will leave it to you or @e-carlin to merge.

rorour:
> On the Run Analysis tab the status is a little “i”. When I hover over it I don’t get a tooltip telling me what that icon means. If you hover over the status on the other tabs you’ll get a tooltip with what the status is. I think this should be added to the run analysis tab.

Scans that haven't been added to the _Analysis database don't currently have a status. What would you like to call the status for these scans?

e-carlin:
Hmm maybe just "Not Analyzed"? If you have something better go with that.

rorour:
Something about having this status strikes me as odd. (Not just the name Not Analyzed but the status in general.) The _AnalysisStatus enum was created to describe the states of scans in the database, and now it's also describing scans that are not in the database. "Not analyzed" might be the most helpful description to read on the tooltip but it's not really clear as a member of _AnalysisStatus because scans that are pending, error etc could also be described as not analyzed. We need to describe that the scan has not been interacted with for analysis in any way, and something like "None" or "NA" for not applicable is the only thing that makes sense for me as an "analysis status" for these scans. I'm going with "None" right now, but if you have any thoughts on this let me know.

e-carlin:
[Naming, one of the hardest things in programming...](https://martinfowler.com/bliki/TwoHardThings.html). Thanks for giving it some thought. None sounds good!

rorour:
@e-carlin Addressed your comments in https://github.com/radiasoft/sirepo/pull/5488/commits/dbe98ffef9c57a5ac305af256ad28cdbbd5a5210 and https://github.com/radiasoft/raydata/pull/76/commits/822119265ce37f37995ab2892a93a66d3ed63271. Except for the asynchrony problem- is this not the same issue [identified here](https://github.com/radiasoft/sirepo/pull/4977#issuecomment-1292732746)? If you would like me to fix it on this PR I think it would be helpful to meet to discuss it more.
I also wasn't able to reproduce the TypeError.

sirepo/package_data/static/js/raydata.js
```diff
@@ -370,10 +381,13 @@ SIREPO.app.directive('scansTable', function() {
         `,
         controller: function(appState, errorService, panelState, raydataService, requestSender, $scope, $interval, $timeout) {
             $scope.analysisModalId = 'sr-analysis-output-' + $scope.analysisStatus;
+            // POSIT: Matches _AnalysisStatus.NONE in scan monitor
+            $scope.analysisStatusNone = "none";
```
e-carlin, line 385: I don't believe this needs to be on scope. Also, let's move it to a const (UPPER_CAMEL_CASE) inside of raydataService. 

The same can be done for defaultColumns and the ["pending", "running"] on line 507.

sirepo/package_data/static/js/raydata.js
```diff
@@ -488,6 +502,17 @@ SIREPO.app.directive('scansTable', function() {
                 $scope.saveColumnChanges();
             };
 
+            $scope.disableRunAnalysis = (scan) => {
```
e-carlin, line 505: I think this can be reduced to just `return ["pending", "running"].includes(scan.status)`.

Then in runAnalysis instead of adding the scan to disabledRunButtonScans we can set the scan.status to pending. 

That way we eliminate the logic for keeping track of disabledRunButtonScans.

sirepo/package_data/static/js/raydata.js
```diff
@@ -522,7 +564,7 @@ SIREPO.app.directive('scansTable', function() {
 
             $scope.setSelectedScan = (scan) => {
                 $scope.selectedScan = scan;
-                if ($scope.selectedScan !== null) {
+                if ($scope.selectedScan !== null && $scope.selectedScan.status !== $scope.analysisStatusNone) {
```
e-carlin, line 567: I think this should include pending scans too.

sirepo/package_data/static/js/raydata.js
rorour: This doesn't work - the front end receives another response that resets the scan status before the server has changed it to pending, which re-enables the button for a moment. I think the front end needs to keep track of this.

sirepo/package_data/static/js/raydata.js
rorour: 18d9639913cbfbaea48046856468f9dcbb7714aa

sirepo/package_data/static/js/raydata.js
rorour: 18d9639913cbfbaea48046856468f9dcbb7714aa

rorour:
regarding the type error scan: I still haven't been able to recreate this, but based on the stack trace it seems like the request to analysis_run_log is not sending a valid uid in your case. I'm not sure why that is, but maybe the check I added in viewLogIframeWrapper will fix it?

sirepo/package_data/static/js/raydata.js
e-carlin: Ah right. Sorry about that.

One concern with this code is if a scan doesn't enter an `ANALYSIS_STATUS_NON_STOPPED` state in time for the GUI to pick up on it it will never be removed from disabledRunButtonScans. For example, this could happen if it goes straight to error w/o a call to run_status in between to pick up on the temporary running/pending state.

Here is a strategy I think might work: On sending to run_analysis add the scan to a dict of `disabledRunButtonScans` (maybe rename to pendingRunAnalysis or something like that). On reply from the server (error or success) remove the scan from the dict, set its status to pending. On the next scans request the status will be updated with the actual status of the scan (what the scan monitor says). In $scope.disableRunAnalysis() check pendingRunAnalysis. If it is in pendingRunAnalysis then return true. Else return raydataService.ANALYSIS_STATUS_NON_STOPPED.includes(scan.status).


Also, a dict should be used instead of an array to keep track of the disable scans. The access pattern is to always look for a certain scan id. So, we can have that be the key and the value doesn't matter (ex true for everything). That way we do constant time lookups/deletions instead of linear time (includes, splice).

e-carlin:
> regarding the type error: I still haven't been able to recreate this, but based on the stack trace it seems like the request to analysis_run_log is not sending a valid uid in your case. I'm not sure why that is, but maybe the check I added in viewLogIframeWrapper will fix it?

I can't reproduce it now either. Even if I remove the check I can't reproduce... Let's get rid of the check for now and we can just keep an eye out for the error. My concern with the check is that it masks some problem. I don't think there should ever be a case where we open the modal and don't have a scanId.

e-carlin (changes requested):
Looks good. Two comments (disabledRunButtonScans and the wrapper check).

sirepo/package_data/static/js/raydata.js
rorour: 260c46d618791faf0cbf7814cf1b64c872b72c51

e-carlin (approved):
Looks good!

Issue #5483, opened 2023-02-14
20230216 Alpha slider error


mkeilman:
Javascript error when opening visualization tab:
```javascript
TypeError: Cannot read properties of null (reading 'readout')
    at updateReadout (sirepo-components.js?20230131.232553:3818:83)
    at update (sirepo-components.js?20230131.232553:3806:17)
    at sirepo-components.js?20230131.232553:3859:17
    at m.$broadcast (angular.js:19720:15)
    at init (radia.js?20230131.232553:2715:24)
    at radia.js?20230131.232553:2948:17
    at m.$emit (angular.js:19648:15)
    at $scope.init (sirepo-plotting-vtk.js?20230131.232553:2955:24)
    at controller (sirepo-plotting-vtk.js?20230131.232553:3016:20)
    at Object.invoke (angular.js:5141:16)
```
The functionality is apparently not affected, so we probably calling the update before the UI is ready

PR #5489, opened 2023-02-16
Fix #5310 radia align shapes in groups

### LINKED ISSUE #5310: Allow alignment of grouped shapes
Add buttons in the group editor to align the centers or edges of shapes in the group, in each dimension.

mkeilman:
Notes:
- Adds centering and alignment buttons to groups in the object list that appear on hover
- Objects in the group align with the first
- Alignment is done relative to the current coordinate plane (xy, yz, zx)
- Objects cannot belong to multiple groups so the copy button for groups has been removed
- Adds colors to the object list to make it easier to associate with the shapes


moellep:
I'm not sure if this is an error related to this branch, or a separate issue:

When I select "new simulation", and drag a cuboid and an l-bend onto the grid, and then visit the visualization page, it shows an error notice "Cannot read properties of null (reading 'readout')". Sometimes it works, OK, or sometimes it shows that notice if I'm on the Visualization tab and refresh the browser page.

moellep:
Also, maybe a separate issue:

If I create a group of 2 items and then apply a rotation transform to the group, the group outline rotates, but the subitems do not. When viewed on the Visualization page, the subitems do appear correctly rotated.

mkeilman:
> When I select "new simulation", and drag a cuboid and an l-bend onto the grid, and then visit the visualization page, it shows an error notice "Cannot read properties of null (reading 'readout')". Sometimes it works, OK, or sometimes it shows that notice if I'm on the Visualization tab and refresh the browser page.

See #5483

mkeilman:
> If I create a group of 2 items and then apply a rotation transform to the group, the group outline rotates, but the subitems do not. When viewed on the Visualization page, the subitems do appear correctly rotated.

Yes, separate issue

PR #5493, opened 2023-02-20
Fix #5491 nersc sbatch script

### LINKED ISSUE #5491: Check for NERSC account fails incorrectly
In `job_agent._sbatch_script()` this call is made:
```python
o = subprocess.check_output(["hpssquota"], text=True)
assert re.search(
    r"{}\s+\d+\.".format(p), o
), f"sbatchProject={p} is invalid. hpssquota={o}"
```
The `hpssquota` command on NERSC yields e.g.:
```bash
+-----------------------------------------+------------+-------------+----------------+------------+-------------+----------------+
|                             File system | Space used | Space quota | Space used (%) | Inode used | Inode quota | Inode used (%) |
+-----------------------------------------+------------+-------------+----------------+------------+-------------+----------------+
|                                    home |  822.75MiB |    40.00GiB |           2.0% |     30.84K |       1.00M |           3.1% |
|                                cscratch |    3.68MiB |    20.00TiB |           0.0% |     124.00 |      10.00M |           0.0% |
| mkeilman usage on HPSS charged to m4091 |      0.00B |     7.20TiB |           0.0% |          - |           - |              - |
+-----------------------------------------+------------+-------------+----------------+------------+-------------+----------------+
```
which does contain the valid account m4091 but is not found by the regex. The fancy table is new; `re.search(r"{}\s+\|?\s+\d+\.".format(p), o)` catching the pipe (optionally) works, or running `hpssquota --space` to output the old format and keep the regex the same. There are also options to output JSON (`hpssquota -J`) and other formats if that is more robust

mkeilman:
First whack at this. Notes:
- New nersc package as suggested
- `hpssquota` is called with the project/account as an argument, and we look for an error string in the output
- `hpssquota` itself calls another subprocess that exits normally even on errors so we can't catch System.Exit or look at the return code (except inside the stdout string - which maybe we should do?)
- Avoiding json since all we care about is whether the account exists at all
- Added nersc_test


sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    assert not re.search(_ACCOUNT_NOT_FOUND, res.stdout), invalid_project_msg(project)
+    return sbatch_account(project)
+
+
+def sbatch_account(project):
+    return f"#SBATCH --account={project}"
+
+
+def invalid_project_msg(project):
```
robnagler, line 39: doesn't need to be exported and would be just fine inline.

sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
...
+        _test_res(project)
+        if pkconfig.channel_in_internal_test()
+        else subprocess.run(
+            ("hpssquota", project),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
+    )
+    assert not re.search(_ACCOUNT_NOT_FOUND, res.stdout), invalid_project_msg(project)
```
robnagler, line 31: We've tended towards:

```py
if re.search(...):
    raise Assertion()
```

Python will strip out `assert` statements when run with certain flags, and this is one we don't want to strip.

However, you probalby want a UserAlert here, because the project is in error, you want to let the user know with a clearly worded error message.

sirepo/pkcli/job_agent.py
```diff
@@ -751,16 +751,7 @@ async def _prepare_simulation(self):
...
-            p = self.msg.sbatchProject
-            if not p:
-                return ""
-            o = subprocess.check_output(["hpssquota"], text=True)
-            assert re.search(r"^[-\w]+$", p), f"invalid NERSC project={p}"
-            assert re.search(
-                r"{}\s+\d+\.".format(p), o
-            ), f"sbatchProject={p} is invalid. hpssquota={o}"
-            return f"#SBATCH --account={p}"
+        import sirepo.nersc
```
robnagler, line 754: from sirepo import nersc when in a function. However, it doesn't need to be in a function, since this is a simple module.

sirepo/pkcli/job_agent.py
```diff
@@ -775,7 +766,7 @@ def _processor():
 #SBATCH --constraint={_processor()}
 #SBATCH --qos={self.msg.sbatchQueue}
 #SBATCH --tasks-per-node={self.msg.tasksPerNode}
-{_assert_project()}"""
+{sirepo.nersc.assert_project(self.msg.sbatchProject)}"""
```
robnagler, line 769: assert_project doesn't tell the whole story. Now that it is a public function it should say what it does: sbatch_project_option()

sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
...
+
+VALID_TEST_ACCOUNT = "VALID_TEST_ACCOUNT"
+
+
+def assert_project(project):
+    if not project:
+        return ""
+    res = (
+        _test_res(project)
+        if pkconfig.channel_in_internal_test()
```
robnagler, line 23: Unit tests can "reach inside" modules. Define a function `_hpssquota(project)` that returns the returns the subprocess result. Remove the `if`. In the test, import the module and use monkeypatch like this:

```py
from sirepo import nersc
monkeypatch(nersc, "_hpssquota", lambda p: return mock_output[p])
```

`mock_output` will be a map of the different project values you want to test and the output from hpssquota modified to test two cases.

sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""Run Python processes in background
+
+:copyright: Copyright (c) 2016 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+import re
+import subprocess
+from pykern import pkconfig
```
robnagler, line 9: No config so delete.

sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""Run Python processes in background
```
robnagler, line 2: Copy and paste

sirepo/nersc.py
```diff
@@ -0,0 +1,44 @@
...
+
+:copyright: Copyright (c) 2016 RadiaSoft LLC.  All Rights Reserved.
```
robnagler, line 4: correct date

tests/nersc_test.py
```diff
@@ -0,0 +1,21 @@
...
+def test_nersc_project(fc):
+    from sirepo import nersc
+    from pykern.pkunit import pkeq, pkok, pkre, pkfail, pkexcept
+
+    _NO_SUCH_PROJECT = "NOT_" + nersc.VALID_TEST_ACCOUNT
+
+    with pkexcept(nersc.invalid_project_msg(_NO_SUCH_PROJECT)):
+        nersc.assert_project(_NO_SUCH_PROJECT)
+
+    pkeq(
```
robnagler, line 18: Tests should dernormalize this state. You don't need to match exactly. You want to know that the logic path is followed so you pkre and partially match what you expect.

robnagler (approved):

robnagler:
@mkeilman @moellep I did not mean to approve and I don't see how to revert.

sirepo/pkcli/job_agent.py
mkeilman: [6fb1a0e](https://github.com/radiasoft/sirepo/pull/5493/commits/6fb1a0edbd1d45106f8b767799ec29c8fef45527)

sirepo/pkcli/job_agent.py
mkeilman: [ee288f3](https://github.com/radiasoft/sirepo/pull/5493/commits/ee288f301b4e638df09ad32e740bbb3566224860)

sirepo/nersc.py
mkeilman: [edebf8a](https://github.com/radiasoft/sirepo/pull/5493/commits/edebf8ac2b64f6de408bcb80d56d5324c8b57181)

sirepo/nersc.py
mkeilman: [edebf8a](https://github.com/radiasoft/sirepo/pull/5493/commits/edebf8ac2b64f6de408bcb80d56d5324c8b57181)

sirepo/nersc.py
mkeilman: [720fe4a](https://github.com/radiasoft/sirepo/pull/5493/commits/720fe4ae533accbd555782bae31eda9a95fbe056)

sirepo/nersc.py
mkeilman: [03195d3](https://github.com/radiasoft/sirepo/pull/5493/commits/03195d350b9a432a7a4a3e9444798628b1ac172f)

tests/nersc_test.py
mkeilman: [f5c7773](https://github.com/radiasoft/sirepo/pull/5493/commits/f5c7773f6e8928ded548737d1aacb5fb99dd8471)

sirepo/nersc.py
mkeilman: [f5c7773](https://github.com/radiasoft/sirepo/pull/5493/commits/f5c7773f6e8928ded548737d1aacb5fb99dd8471)

mkeilman:
Changes in place. I used some other examples in sirepo as a model for the monkeypatching - lmk if that's the right approach

sirepo/nersc.py
```diff
@@ -0,0 +1,29 @@
...
+
+ACCOUNT_NOT_FOUND = "no such fileset"
+
+
+def sbatch_project_option(project):
+    if not project:
+        return ""
+    res = _hpssquota(project)
+    if re.search(ACCOUNT_NOT_FOUND, res.stdout):
+        raise sirepo.util.UserAlert(f"Account {project} not found on NERSC")
```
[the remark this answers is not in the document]
mkeilman: I don't know if there is a difference - the comment in the script is `#SBATCH --account=...` but the UI calls it "project". [daac54b](https://github.com/radiasoft/sirepo/pull/5493/commits/daac54b634107fa505c45103273be7930990e55a) makes it uniform (except for the comment)

mkeilman:
A couple of quick notes: first, as per @robnagler I will change this to be "pessimistic" and look for a good result and not any specific error text. Also, it turns out that you can call hpssquota with any valid project code, even one you are not a part of, and it will return successfully (though the sbatch script will fail). If you leave it out you get only those projects assigned to you, so that is what I will do, plus I'll move to json as discussed earlier

moellep (commented):
in _hpssquota(), I don't see the project arg getting passed along to hpssquota - so I don't see how the project would show up in the "fs" field of the results.

moellep:
When I call "hpssquote -N -J" with no project, it gives me one entry, which has "fs": "moellep usage on HPSS charged to m4091", but I don't see how that gets parsed.

mkeilman:
> When I call "hpssquote -N -J" with no project, it gives me one entry, which has "fs": "moellep usage on HPSS charged to m4091", but I don't see how that gets parsed.

That's an error - in the previous commit I was passing the project, in which case there is an entry with "fs": <project> - but we can't do only that because hpssquota will happily show info for projects you don't belong to. So I will have to add a check for the project. I'll also update the testing monkeypatch to more closely hew to the actual output.

moellep:
@robnagler could you take another look at this PR. I don't know if some users have multiple project codes on NERSC or how this would handle it.

Issue #5474, opened 2023-02-13
20230227 don't argmax output from image to image training


moellep:
For image-to-image training, keep the final output linear, and show ground-truth/predicted images in grayscale. For example a 3 color mask would use white, gray and black as the values, and the predicted value would be gray values somewhere on that scale.

PR #5494, opened 2023-02-20
Fix #5315 #5474: image to image files

### LINKED ISSUE #5315: ML image to mask with unet
Allow image-to-image h5 input files. A good example would be the oxford pets database:

https://keras.io/examples/vision/oxford_pets_image_segmentation/

gurhar1133:
Activait works with image to image example. Also no more classifying each pixel for image to image case.


moellep (changes requested):
Looks good - a few comments:
- could we put the Segments panel inline with the Fit panels? The Segments could be shown to the left of the "Fit 1" on the pets example
- The csv 2019 World Happiness regression now shows the javascript error after the training completes.
```
TypeError: Cannot read properties of undefined (reading '0')
    at self.imageToImage (activait.js?20230221.205821:416:26)
```

gurhar1133:
@moellep changes have been made. LMK if anything else

moellep (approved):

PR #5499, opened 2023-02-21
Fix #5409: remove watchpoints

### LINKED ISSUE #5409: srw remove watchpoints from generated optical names list
Watchpoints in the generated optical element list have no operation ("pass") and should not be included in the "names" array or have a case in the set_optics() method.

For example, the Sample watchpoint in the CHX example should not be in the generated code in these places:

```python
def set_optics(v, names=None, want_final_propagation=True):
    # ...
   if not names:
        names = ['S0', 'S0_HDM', 'HDM', 'HDM_S1', 'S1', 'S1_S2', 'S2', 'S2_CRL1', 'CRL1', 'CRL2', 'CRL2_KLA', 'KLA', 'KL', 'KL_S3', 'S3', 'S3_Sample', 'Sample']

        # ...
        elif el_name == 'Sample':
            # Sample: watch 48.7m
            pass
    # ...

def main():
    # ...
    names = ['S0','S0_HDM','HDM','HDM_S1','S1','S1_S2','S2','S2_CRL1','CRL1','CRL2','CRL2_KLA','KLA','KL','KL_S3','S3','S3_Sample','Sample']
    # ...
```

gurhar1133:
Removed Sample and Watchpoint generation in names list and the associated `pass` blocks


moellep (changes requested):
You don't want to use item.name as the test because that could be anything the user types. Instead look at item.type == 'watch'

sirepo/package_data/template/srw/beamline_optics.py.jinja
```diff
@@ -38,9 +38,13 @@ v.op_{{ item.name }}_{{ nameMap.get(name, name) }}
 {% if loop.index == 1 %}
         if el_name == '{{ item.name }}':
 {% else %}
+        {% if item.type not in ('watch', 'sample') %}
         elif el_name == '{{ item.name }}':
```
moellep, line 42: Only for item.type == 'watch'. The "sample" type is a different beamline element which we do want to include.

moellep: I know this is confusing because sometimes watchpoints have the name "Sample" which is the position of the sample in the actual beamline.

moellep: Also try the Example/Sample from Image SRW example to make sure it works normally.

moellep: It looks like the generated code for the beamlineAnimation model depends on watches being present in names:
```python
        for n in beamline_info.names:
            names.append(n)
            if n in beamline_info.watches:
                #...
```
So there will need to be some care in removing watches from the generated output, but preserving it in the internal data structures used to generated the code.

gurhar1133: > Also try the Example/Sample from Image SRW example to make sure it works normally.

made the changes. Re the Sample from Image example, i get the error message in run.log:
```

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 628739 RUNNING AT v.radia.run
=   EXIT CODE: 9
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Killed (signal 9)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```
But I get that error on master too

gurhar1133:
@moellep made changes. LMK if any more are needed 👍

moellep:
Looks good!

gurhar1133:
@moellep unblock merge?

moellep (approved):

PR #5505, opened 2023-02-23
fix #5504 multipanel layout

### LINKED ISSUE #5504: 20230222 multi report panel
Allow on panel to show multiple reports, with an optional animation controller which controls all sub reports.

moellep:
separate out some report components:
- useAnimationReader hook for responding to report data and frame navigation
- AnimationControlButtons
add a new multipanel layout which supports multiple reports tied to one set of animation controls


react/src/layout/multipanel.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    getFormDependencies = () => {
+        return [...(this.items)].map(childLayout => childLayout.getFormDependencies()).flat();
+    }
+
+    component = (props: LayoutProps<{}>) => {
+        let { reportName, reportGroupName, frameIdFields } = this.config;
+        let modelsWrapper = useContext(CModelsWrapper);
+        let title = useInterpolatedString(modelsWrapper, this.config.title, ValueSelectors.Models);
+        //TODO(pjm): without this call, heatplots don't respond to window resize
```
garsuga, line 37: This is fine, I think we should think about hooks as having side-effects. This is very declarative and works.

react/src/layout/multipanel.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Col sm={12} className="px-3 py-2">
+                <Panel title={title} panelBodyShown={true}>
+                    <Row>
+                        <ReportAnimationController
+                            animationReader={animationReader}
+                            showAnimationController={showAnimationController}
+                            currentFrameIndex={undefined}
+                        >
+                            {
+                                (currentFrame: any) => {
```
garsuga, line 62: Put 'unknown' for the type if unknown

react/src/layout/report.tsx
```diff
@@ -105,6 +103,77 @@ export class AutoRunReportLayout extends Layout<AutoRunReportConfig, {}> {
     }
 }
 
+export function useAnimationReader(reportName, reportGroupName, frameIdFields) {
```
garsuga, line 106: Define types in function header

garsuga (commented):
Looks good, just some type stuff. Try to define types whenever possible in functions. Especially return types, things often just "work" with implied return types and explicit types can be really useful for catching errors. All functions at the top level in a file should be strongly typed since they can be shared/exported. This also helps us keep track of when we bring in too many types from other areas of the program that should maybe have ownership of the code we are writing instead. A great example of this is /components/reusable/form.tsx which is no longer reusable since it directly deals with schema concepts.

moellep:
Now uses types in header and arguments, replacing "any" and "unknown" 3b1c3e0

Issue #1151, opened 2018-04-10
h5py causes initial server request to hang


moellep:
When running on freshly installed server software, if a new user loads an app, the simulation list response never makes it back to the client. Reloading the browser works OK. 
The cause is somewhere in the h5py init code. Importing h5py from server.py works around the problem for now. We use h5py 2.7.1 - it looks like this problem may not happen with h5py 2.8.0

Closes #1168.

moellep:
Latest version of h5py doesn't fix it if HDF5_MPI=ON is enabled.

robnagler:
Maybe there's a way to turn off MPI dynamically. Would be good to submit an issue to h5py.

mrakitin:
Closes #1168.

robnagler:
Resolved by server.py import but also by avoiding template imports by server.

PR #5509, opened 2023-02-24
fix #5381: HDF5 Plotting Refactor

### LINKED ISSUE #5381: 20230301 Common H5 plotting prep
This is similar to #5380 but deals with hdf5 files. Apps which plot from hdf5 files include (opal, warppba, warpvnd, silas, omega). After #5380 is completed, a new class HDF5Util could be created which supports line and heat plots. This could possibly be expanded to include other common hdf5 file manipulation.

rorour:


sirepo/template/hdf5_util.py
```diff
@@ -0,0 +1,71 @@
...
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdp
+from sirepo.template import template_common
+import numpy as np
+
+
+class HDF5Util:
+    # TODO(rorour): add docs for heatmap & lineplot
+    def __init__(self):
+        pass
```
moellep, line 16: Here it should take the hdf5 file in the constructor, and then when plotting, the file should be opened with h5py and the file handled passed to the callbacks if needed.

sirepo/template/hdf5_util.py
```diff
@@ -0,0 +1,71 @@
...
+                continue
+            plots[f] = PKDict(
+                label=plot_attrs.model[f],
+                dim=f,
+                points=[],
+                name=name_and_field[0],
+                index=plot_attrs.index(name_and_field),
+            )
+        plot_attrs.format_plots(plots)
+        x = plots.get(x_field)
```
moellep, line 60: here format_plots() could be passed the open file handle as an argument, so the caller doesn't need to open the hdf5 file beforehand.

moellep (changes requested):

sirepo/template/hdf5_util.py
rorour: I'm not totally understanding your vision here- let me know if this still needs to change

sirepo/template/hdf5_util.py
rorour: d27e19ab27a41125f8de5d4f719ec9b485904168

sirepo/template/hdf5_util.py
```diff
@@ -0,0 +1,71 @@
...
+        x_field = "x"
+        y_field = "y"
+        plots = PKDict()
+        for f in (x_field, y_field):
+            n = plot_attrs.model[f]
+            plots[f] = PKDict(
+                name=n,
+                points=np.array(
+                    plot_attrs.file["/Step#{}/{}".format(plot_attrs.frame_index, n)]
+                ),
```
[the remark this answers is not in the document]
rorour: d27e19ab27a41125f8de5d4f719ec9b485904168

rorour:
Do you know why this is failing ci? I saw this line in the log:
`import h5py found in: sirepo/template/hdf5_util.py`
would that be causing a failure?

For warppba extract_particle_report, I refactored to use template_common, but I wasn't sure about converting to use HDF5Util. What do you think is the best way to approach that?

moellep:
The "import h5py found" can be fixed by adding hdf5_util to the exception list in test.sh:
https://github.com/radiasoft/sirepo/blob/master/test.sh#L27

That check was added due to problems loading apps if h5py wasn't imported in server.py
original bug: https://github.com/radiasoft/sirepo/issues/1151

In this case, we aren't adding apps, but moving the "import h5py" into the utility class, so it shouldn't be a problem.

rorour:
@moellep Thanks, tests are passing now

moellep (approved):

PR #5525, opened 2023-03-03
Fix #5408: multiElectronAnimation export improve

### LINKED ISSUE #5408: srw multi-electron exported code improvements
Rather than set the "wm" argument within the argument definition list, it should be set explicitly within the main() function, similar to the way the single-electron code is generated. This would allow the same script to be easily used for single-electron cases by modifying the main() method.

```python
    # ...
    #Multi-Electron (partially-coherent) Wavefront Propagation
    ['wm', '', '', 'calculate multi-electron (/ partially coherent) wavefront propagation', 'store_true'],
    # ...

def main():
    v = srwl_bl.srwl_uti_parse_options(srwl_bl.srwl_uti_ext_options(varParam), use_sys_argv=True)
    names = ['S0','S0_HDM','HDM','HDM_S1','S1','S1_S2','S2','S2_CRL1','CRL1','CRL2','CRL2_KLA','KLA','KL','KL_S3','S3','S3_Sample','Sample']
    op = set_optics(v, names, True)
    v.wm = True
```

gurhar1133:
When export from cogwheel: `v.wm = False`. When exported from the Partially Coherent tab: `v.wm = True`

### DIFF
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -1991,9 +1991,12 @@ def _generate_srw_main(data, plot_reports, beamline_info):
             ]
         )
     else:
-        if report == "multiElectronAnimation":
-            # TODO (gurhar1133): source type and run_all need to be in condition??
-            content.append("v.wm = True")
+        if report == 'multiElectronAnimation':
+            # TODO (gurhar1133): need to work out these conditions
+            if run_all:
+                content.append('v.wm = True')
+            else:
+                content.append('v.wm = False')
         if (run_all and source_type != "g") or report == "intensityReport":
             content.append("v.ss = True")
             if plot_reports:
--- a/sirepo/template/srw.py
+++ b/sirepo/template/srw.py
@@ -1992,11 +1992,10 @@ def _generate_srw_main(data, plot_reports, beamline_info):
         )
     else:
         if report == 'multiElectronAnimation':
-            # TODO (gurhar1133): need to work out these conditions
-            if run_all:
-                content.append('v.wm = True')
-            else:
-                content.append('v.wm = False')
+            if not run_all:
+                content.append("v.wm = True")
+        else:
+            content.append("v.wm = False")
         if (run_all and source_type != "g") or report == "intensityReport":
             content.append("v.ss = True")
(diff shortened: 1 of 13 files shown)


sirepo/template/srw.py
```diff
@@ -1991,6 +1991,11 @@ def _generate_srw_main(data, plot_reports, beamline_info):
             ]
         )
     else:
+        if report == "multiElectronAnimation":
```
robnagler, line 1994: It seems odd to not set v.wm in all cases. What is it in the third case?

robnagler (changes requested):

sirepo/template/srw.py
gurhar1133: @moellep if `report == "multiElectronAnimation"` and  `run_all == True`  do we want to set `v.wm == False`?

moellep: When you export the python source from the cogwheel, all the options should be explicit in main() - all sequential reports should be True, and ```v.wm = False```. I think ```run_all == True``` in that case. Each plot can export its own python source, and in that case, the only setting which should be on, is the specific plot - so ```v.wm = False``` should not be present in that case.

Try it out on each plot in the UI to verify.

moellep (changes requested):
When you select "Export Python Code" from the Partially Coherent Intensity plot on the Partially Coherent Results tab, ```v.wm = True``` is getting generated which is good, but the "wm" parameter in the parameter list should have the value "" rather than "1":
```
    #Multi-Electron (partially-coherent) Wavefront Propagation
    ['wm', '', '1', 'calculate multi-electron (/ partially coherent) wavefront propagation', 'store_true'],
```

moellep:
On the Partially Coherent Results tab, in the Partially Coherent Intensity editor, enable the "Calculate Degree of Coherence in Addition to Intensity" field. Then run the simulation. Three plots will display - they should all have the same python export, ex. ```v.wm = True``` should be set on for all the Export Python Code for those plots. Those two extra plots are coherenceXAnimation and coherenceYAnimation

moellep (approved):

PR #5526, opened 2023-03-03
Fix #5508: remove min on diffractionOrder

### LINKED ISSUE #5508: Grating diffraction order needs to allow negative integers
> In "Grating -> Orientation" panel, the "Diffraction Order" input box doesn't seem to allow for entering negative order numbers (-1, -2, etc.), however the use of gratings in this orientation is quite common. I don't know at this point, maybe some tests and fixes will be required in this relation in SRW; but Sirepo should allow entering negative orders.
> One could add validation that user's input in this dialog is limited to integer numbers only, but this is less important.

gurhar1133:
Negative diffraction order allowed now

### DIFF
--- a/tests/template/srw_shadow_data/to_shadow_grating.out/shadow.json
+++ b/tests/template/srw_shadow_data/to_shadow_grating.out/shadow.json
@@ -83,7 +83,7 @@
             "offx": 0,
             "offy": 0,
             "offz": 0,
-            "order": -1.0,
+            "order": 1,
             "param": 0.0,
             "phot_cent": 1000.0,
             "position": 55,
--- a/tests/template/srw_shadow_data/to_shadow_grating.out/shadow.json
+++ b/tests/template/srw_shadow_data/to_shadow_grating.out/shadow.json
@@ -83,7 +83,7 @@
             "offx": 0,
             "offy": 0,
             "offz": 0,
-            "order": 1,
+            "order": -1.0,
             "param": 0.0,
             "phot_cent": 1000.0,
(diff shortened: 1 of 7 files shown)


moellep (changes requested):
Also, change the field type from Float to Integer.

moellep (approved):

tests/template/srw_shadow_data/to_shadow_grating.out/shadow.json
```diff
@@ -83,7 +83,7 @@
             "offx": 0,
             "offy": 0,
             "offz": 0,
-            "order": -1.0,
+            "order": 1,
             "param": 0.0,
```
moellep, line 87: This test case is a little odd that it changed sign from -1.0 to 1. Could you look at the conversion in srw_shadow.py and make sure it is doing the right thing?

gurhar1133: In `srw_shadow.py`, if the type for the field is Float in the schema then the value will be multiplied by `scale` which is -1 for diffractionOrder:

```python
if schema[from_f][1] == "Float":
        v = float(v) * scale
```

But the type was changed to Integer. Do we still want to scale by -1 now that it is integer?

moellep: Yes, I think that should be changed to check if it is Float or Integer, apply the scale. Shadow needs the diffraction order to be negative to agree with SRW.

PR #5530, opened 2023-03-04
Fix #5309 radia improved shapes

### LINKED ISSUE #5309: 20230228 Shapes should better reflect the actual object shape
For example, racetracks and cylinders currently present their rectangular footprint. If possible include bevels and fillets. STL shapes are probably too complex for this.

mkeilman:
Notes:
- Appropriate face-on shapes for cylinders and racetracks (no "hole" in the racetrack though)
- Along axes other than the extrusion axis, this uses d3's `hull` to calculate a polygon out of the projected points
- Does not include shape modifications (bevels, fillets)


sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -244,7 +246,7 @@ class PlotPolygon extends AbstractPlotShape2D {
     getSizeCoords() {
         return this.getCoords(this.size);
     }
-
+    
     setPoints(arr) {
```
moellep, line 250: @mkeilman can you check if your editor supports "remove whitespaces in blank lines" feature? Maybe something along these lines:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/205803079-Configure-editor-to-remove-whitespaces-in-blank-lines

mkeilman: Yes - I've turned that on. Also turned on "ensure every saved file ends with a line break" while I was at it

PR #5548, opened 2023-03-08
Fix #5420: compute pgm exception

### LINKED ISSUE #5420: compute PGM value should not log exception
The srwlib SRWLOptMirTor may throw a TypeError if the find_orient() fails for the values.  template.srw should look at the exception in _compute_PGM_value() and if it is a TypeError for NoneType, it shouldn't be logged.

gurhar1133:
Now wont log errors associated with bad inputs into grating


mkeilman (commented):
Minor commment

mkeilman (approved):
looks fine to me

PR #5554, opened 2023-03-08
Fix #5491 nersc account error

### LINKED ISSUE #5491: Check for NERSC account fails incorrectly

mkeilman:
Trying this again. We need to parse the "stdout" attribute of the subprocess, not the subprocess itself. I've tested this on NERSC itself so it should behave

### DIFF
--- a/sirepo/nersc.py
+++ b/sirepo/nersc.py
@@ -30,4 +30,4 @@ def _hpssquota(project=None):
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         text=True,
-    )
+    ).stdout
--- a/sirepo/nersc.py
+++ b/sirepo/nersc.py
@@ -23,7 +23,7 @@ def sbatch_project_option(project):
 
 
 # Note: project is only used for unit tests
-def _hpssquota(project=None):
+def _hpssquota():
     # -N excludes home and scratch file systems; -J outputs json
     return subprocess.run(
         ("hpssquota", "-N", "-J"),
--- a/sirepo/nersc.py
+++ b/sirepo/nersc.py
@@ -23,7 +23,7 @@ def sbatch_project_option(project):
 
 
 # Note: project is only used for unit tests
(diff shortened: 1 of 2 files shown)


sirepo/nersc.py
```diff
@@ -30,4 +30,4 @@ def _hpssquota(project=None):
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         text=True,
-    )
+    ).stdout
```
e-carlin, line 33: Can the project arg be removed? It doesn't look like it is used.

e-carlin (approved):

PR #5559, opened 2023-03-09
fix #3933 reraise if module isn't same name

### LINKED ISSUE #3933: 20221230 Refine ModuleNotFound in import_submodule

robnagler:
- removed some single letter variables that weren't helpful

### DIFF
--- a/sirepo/util.py
+++ b/sirepo/util.py
@@ -294,7 +294,7 @@ def import_submodule(submodule, type_or_data):
     from sirepo import feature_config
     from sirepo import template
 
-    t = template.assert_sim_type(
+    sim_type = template.assert_sim_type(
         type_or_data.simulationType
         if isinstance(
             type_or_data,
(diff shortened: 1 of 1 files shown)


sirepo/util.py
```diff
@@ -294,25 +294,30 @@ def import_submodule(submodule, type_or_data):
     from sirepo import feature_config
     from sirepo import template
 
-    t = template.assert_sim_type(
+    sim_type = template.assert_sim_type(
```
rorour, line 297: Why did you change the var name?

rorour (approved):
Looks good to me

sirepo/util.py
robnagler: Internal discussion.

PR #5574, opened 2023-03-15
Fix #5117: image preview improvement

### LINKED ISSUE #5117: Improve prev/next for image samples
The prev/next image buttons for activait could be improved and share with the column pager already available in that app:

The next/prev currently looks like this:
Untitled 2

It would look better if it followed the layout of the column pager and provided counts, totals and enabled/disabled the next/prev buttons depending on the state:

gurhar1133:
Page counter added to image preview panel.


mkeilman (commented):
Code looks fine - I think "page i of n" would look better if centered, but I don't feel strongly about it

gurhar1133:
@moellep any thoughts?

PR #5583, opened 2023-03-16
Fix #5570 Add feature_config trust_sh_env for non-RadiaSoft shell environments

### LINKED ISSUE #5570: Decouple bashrc and pyenv from subprocess execution
pkcli.service assumes pyenv is the way python executes, but this should be configurable.

job_agent assumes bashrc can be executed, but this can be problematic in local mode. job environment should be configurable to not execute bashrc and instead just pass incoming process environment. 

Make this a single configuration variable in feature_config

robnagler:
- trust_sh_env is false by default
- In non-RadiaSoft environments, will not source bashrc/profile and assume Python is correct (no pyenv)

### DIFF
--- a/sirepo/job.py
+++ b/sirepo/job.py
@@ -1,15 +1,15 @@
 # -*- coding: utf-8 -*-
 """Common functionality that is shared between the server, supervisor, and driver.
 
-:copyright: Copyright (c) 2019 RadiaSoft LLC.  All Rights Reserved.
+:copyright: Copyright (c) 2019-2023 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
-from __future__ import absolute_import, division, print_function
 from pykern import pkconfig
 from pykern.pkcollections import PKDict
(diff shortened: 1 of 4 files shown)


sirepo/job.py
```diff
@@ -159,11 +159,13 @@ def agent_cmd_stdin_env(cmd, env, uid, cwd=".", source_bashrc=""):
         cwd (str): directory for the agent to run in (will be created if it doesn't exist)
 
     Returns:
-        tuple: new cmd (tuple), stdin (file), env (PKDict)
+        tuple: new cmd (tuple), stdin (file), env (PKDict or None)
     """
     import os
     import tempfile
 
+    if sirepo.feature_config.cfg().trust_sh_env:
```
e-carlin, line 167: Why also have this check if it is an arg? Can't the caller protect?

e-carlin (approved):

sirepo/job.py
robnagler: The caller doesn't need to understand the abstraction. A lot is going on in side. It's not an arg.

e-carlin (approved):

PR #5613, opened 2023-03-28
Fix #5612: Remove RequiredUnlessDev for raydata.file_reply_tmp_dir

### LINKED ISSUE #5612: raydata.file_reply_tmp_dir
https://github.com/radiasoft/sirepo/issues/5590#issuecomment-1487471883

RequiredUnlessDev isn't the best choice for this configuration parameter. In cases where raydata isn't a configured sim_type it will still be a required bit of configuration. I think the logic in _tmp_dir (it's type) is enough.

e-carlin:
On systems w/o raydata as a configured sim_type this parameter shouldn't be required.

### DIFF
--- a/sirepo/feature_config.py
+++ b/sirepo/feature_config.py
@@ -174,7 +174,7 @@ def _init():
             "codes that contain proprietary information and authorization to use is granted through oauth",
         ),
         raydata=dict(
-            file_reply_tmp_dir=pkconfig.RequiredUnlessDev(
+            file_reply_tmp_dir=(
                 "raydata_file_reply_tmp_dir",
                 _tmp_dir,
                 "directory to share analysis pdfs between scan monitor and supervisor",


sirepo/feature_config.py
```diff
@@ -174,7 +174,7 @@ def b(msg, dev=False):
             "codes that contain proprietary information and authorization to use is granted through oauth",
         ),
         raydata=dict(
-            file_reply_tmp_dir=pkconfig.RequiredUnlessDev(
+            file_reply_tmp_dir=(
```
robnagler, line 177: I do not understand why this is here.

e-carlin: Where would be a better place? It is used abstractly inside of job_cmd so adding the configuration there doesn't make sense. Should we separate configuration for sim_types out of feature_config and into a module for each sim_type?

PR #5631, opened 2023-03-31
fix #5432: move raydata code into sirepo

### LINKED ISSUE #5432: 20230204 move raydata code into Sirepo
The only proprietary part of raydata is the notebooks. Initially we wanted to distribute them with the Docker container but this is going to change. We want them mounted so they can be changed by the physicists without having to cut a new release. So, all of the raydata code can be moved into Sirepo which will make development and deployment simpler.

rorour:


sirepo/pkcli/raydata.py
```diff
@@ -4,9 +4,147 @@
...
+import sirepo.feature_config
+import sirepo.srtime
+import sqlalchemy
+import sqlalchemy.ext.declarative
+import sqlalchemy.orm
+import time
+import tornado.websocket
+import uuid
+import zipfile
+
```
rorour, line 34: I removed an import from replay that seemed unnecessary  https://github.com/radiasoft/raydata/blob/c7b02306b9111ac2584ebb256a3a3615361494d5/raydata/pkcli/replay.py#L18

etc/run-raydata.sh
e-carlin, line 1: I know I said to create this but after seeing it let's remove it and use the script in raydata/etc. Better to reference a public repo from a private one than vice versa

sirepo/pkcli/raydata.py
```diff
@@ -18,5 +156,455 @@ def run_background(cfg_dir):
...
+            data=PKDict(
+                scans=sorted(scans, key=lambda e: e.start),
+                cols=[k for k in scans[0].keys() if k not in _NON_DISPLAY_SCAN_FIELDS]
+                if scans
+                else [],
+            )
+        )
+
+    sirepo.srtime.init_module()
+    pkio.mkdir_parent(sirepo.feature_config.cfg().raydata.db_dir)
```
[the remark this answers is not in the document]
e-carlin: That loop was a "fail fast" check to make sure that we had a notebook for each catalog we are monitoring. If you dig down into the call in pykern you'll see we raise not found if the file doesn't exist.

I think the check is worthwhile. But, it could be put in a clearly named function to make it easier for readers to understand what its use is.

sirepo/feature_config.py
```diff
@@ -174,6 +174,22 @@ def b(msg, dev=False):
             "codes that contain proprietary information and authorization to use is granted through oauth",
         ),
         raydata=dict(
+            catalog_names=(frozenset(), set, "list of catalog names to monitor"),
+            concurrent_analyses=(
```
e-carlin, line 178: I think these all belong in pkcli.raydata (or sirepo/raydata see below) since that is the only place they are used.  Raydata is unlike other codes so configuring it differently I think is ok.

scan_monitor_url can stay here though since that is used in sirepo.template.raydata which works like other codes.

sirepo/pkcli/raydata.py
e-carlin, line 1: I think this file is too big. It makes it hard to understand what's going on and move around the code. I think there are two options:
1. Create a sirepo/pkcli/raydata/ folder and then move the individual files into there. This has the downside that you have to call them like `sirepo raydata.foo bar`. The `.` in the name is weird.
2. create a sirepo/raydata directory. Move the logic into modules in it. Then make sirepo/pkcli/raydata.py just a small file that calls functions in modules in sirepo/raydata/. 

I like option 2 the best. The code in the raydata repo was already starting to get unwieldy and could've used breaking into more modules. So, I think it would be good to make that change as we are making this one. What do you think?

sirepo/feature_config.py
```diff
@@ -174,6 +174,22 @@ def b(msg, dev=False):
...
+                2,
+                int,
+                "max number of analyses that can run concurrently",
+            ),
+            db_dir=(
```
e-carlin, line 183: I don't think we should hard code these paths. Use sirepo/srdb.py to get the run dir and then put them under that. They could be set pkconfig.RequiredUnlessDev which will help me remember I need to configure them in higher environments. You can also make sure they are relative paths in dev and absolute paths everywhere else.

e-carlin (changes requested):

e-carlin:
> While working on this, I ran into an error that [you mentioned previously](https://github.com/radiasoft/sirepo/pull/5488#pullrequestreview-1328979439). I created an issue for it since it seems unrelated to this change. [radiasoft/raydata#86](https://github.com/radiasoft/raydata/issues/86)

Thanks for catching and making an issue!

sirepo/pkcli/raydata.py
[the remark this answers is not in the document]
rorour: e799cdc8671eaf3a2fa69713cd0682b70a4ea7bb

etc/run-raydata.sh
rorour: e799cdc8671eaf3a2fa69713cd0682b70a4ea7bb

sirepo/feature_config.py
rorour: e799cdc8671eaf3a2fa69713cd0682b70a4ea7bb

sirepo/pkcli/raydata.py
rorour: e799cdc8671eaf3a2fa69713cd0682b70a4ea7bb

sirepo/feature_config.py
rorour: I'm not entirely sure I understood all this, let me know if something needs to change. e799cdc8671eaf3a2fa69713cd0682b70a4ea7bb

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+            2,
+            int,
+            "max number of analyses that can run concurrently",
+        ),
+        db_dir=pkconfig.RequiredUnlessDev(
+            os.path.join(sirepo.srdb.root(), "raydata"),
+            pkio.py_path,
+            "root directory for db",
+        ),
+        notebook_dir=pkconfig.RequiredUnlessDev(
```
e-carlin, line 509: This can be just `notebook_dir=pkconfig.Required(pkio.py_path, "directory for analysis notebooks")`. We can expect all dev users to run using the script and in higher environments it will be configured.

e-carlin (changes requested):
Let's go over this together. Some of the bits for configuration will be easier to describe in realtime.

sirepo/raydata/scan_monitor.py
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+            os.path.join(sirepo.srdb.root(), "raydata/package_data"),
+            pkio.py_path,
+            "root directory for proprietary files",
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+            int,
+            "max number of analyses that can run concurrently",
+        ),
+        db_dir=pkconfig.RequiredUnlessDev(
+            os.path.join(sirepo.srdb.root(), "raydata"),
+            pkio.py_path,
+            "root directory for db",
+        ),
+        notebook_dir=pkconfig.RequiredUnlessDev(
+            os.path.join(sirepo.srdb.root(), "raydata/package_data"),
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+    global cfg
+    cfg = pkconfig.init(
+        catalog_names=(frozenset(), set, "list of catalog names to monitor"),
+        concurrent_analyses=(
+            2,
+            int,
+            "max number of analyses that can run concurrently",
+        ),
+        db_dir=pkconfig.RequiredUnlessDev(
+            os.path.join(sirepo.srdb.root(), "raydata"),
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+        return PKDict(
+            data=PKDict(
+                scans=sorted(scans, key=lambda e: e.start),
+                cols=[k for k in scans[0].keys() if k not in _NON_DISPLAY_SCAN_FIELDS]
+                if scans
+                else [],
+            )
+        )
+
+    global cfg
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/raydata/scan_monitor.py
```diff
@@ -0,0 +1,522 @@
...
+
+#: path scan_monitor registers to receive api requests
+_URI = "/scan-monitor"
+
+cfg = None
+
+engine = None
+
+
+def scan_monitor():
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/pkcli/raydata.py
```diff
@@ -18,5 +28,9 @@ def run_background(cfg_dir):
     _run()
 
 
+def scan_monitor():
+    sirepo.raydata.scan_monitor.scan_monitor()
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/pkcli/raydata.py
```diff
@@ -5,8 +5,18 @@
...
+import sirepo.raydata.replay
+import sirepo.raydata.scan_monitor
+
+
+def create_scans(num_scans, catalog_name, delay=True):
+    sirepo.raydata.create_scans.create_scans(num_scans, catalog_name, delay)
+
+
+def replay(source_catalog, destination_catalog, num_scans):
+    sirepo.raydata.replay.replay(source_catalog, destination_catalog, num_scans)
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

sirepo/pkcli/raydata.py
```diff
@@ -5,8 +5,18 @@
...
 from pykern.pkcollections import PKDict
-from pykern.pkdebug import pkdp, pkdlog
 from sirepo.template import template_common
+import sirepo.raydata.create_scans
+import sirepo.raydata.replay
+import sirepo.raydata.scan_monitor
+
+
+def create_scans(num_scans, catalog_name, delay=True):
+    sirepo.raydata.create_scans.create_scans(num_scans, catalog_name, delay)
```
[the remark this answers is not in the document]
rorour: b2193636ae2f3070921e6c0eabdce8ec9b2fdc8b

e-carlin (approved):
LGTM

PR #5647, opened 2023-04-06
fix radiasoft/raydata#82: disable running automatically

rorour:


e-carlin (approved):

sirepo/raydata/scan_monitor.py
```diff
@@ -507,6 +509,11 @@ def start():
     def _init():
         global _cfg
         _cfg = pkconfig.init(
+            automatic_analysis=(
+                False,
```
e-carlin, line 513: Default to true. Ideally we are using this feature all of the time.

e-carlin (changes requested):

PR #5649, opened 2023-04-06
Fix #4564 vtk to v26.9.14

### LINKED ISSUE #4564: Update and test latest vtk.js (25.0.0)
According to the release notes there are breaking changes...

mkeilman:
Notes:
- Issue was for originally v25.0.0 but got put off
- Picker handling had to change a bit
- Orienting the view with XYZ buttons may cause a javascript warning (not an error) "Resetting view-up since view plane normal is parallel". (thrown by `resetCamera()`) There is no change of functionality as far as I can tell.
- Tested everywhere we have 3d


moellep:
Something seems to have changed with font anti-aliasing. In SRW on the 3D Beamline tab, the text looks pixelated:
x
on alpha, it looks correct:
y

mkeilman:
> Something seems to have changed with font anti-aliasing. In SRW on the 3D Beamline tab, the text looks pixelated: x on alpha, it looks correct: y

It looks fine to me, except on Safari where I see just black rectangles. I'll see if there is a fix...

mkeilman:
The Safari problem is "fixed" by turning off WebGL2.0 (at least 14.0) - it's an experimental feature so imho it's OK to ignore for this PR.  I still see no aliasing here but it is os- and browser-dependent. If it looks bad enough for others we can put this pr on hold again, but we are a couple of major versions behind at this point so I'd prefer to move forward if possible

mkeilman:
>  I still see no aliasing here but it is os- and browser-dependent

The reason I saw no aliasing appears to be related to the screen itself. When I use a large monitor the aliasing ie evident, once I zoom out a sufficient amount. On my laptop this does not happen - the text is smooth at any zoom level.

moellep:
I can reproduce the behavior on the old vtk by turning off "interpolate" on the vtkTexture. In srw.js, this line:
```javascript
            vtk.Rendering.Core.vtkTexture.newInstance({
                interpolate: true,
                inputData: [labelImage(itemText(item), labelSize)],
            }));
```

Somehow that flag isn't working with the current vtk.

PR #5651, opened 2023-04-07
Fix #4803 adapt to NERSC perlmutter 

### LINKED ISSUE #4803: 20230228 switch nersc to perlmutter
Prerequisite https://github.com/radiasoft/containers/issues/106

### LINKED ISSUE #5639: tasksPerNode needs to be dynamic, coming from the server
The default tasks per node is 128 on cpu-only nodes on perlmutter:

>            `"tasksPerNode": ["Processes per node", "Integer", 32],`

We can probably solve this with configuration. It shouldn't be hardwired in the schema.

See also #4803

### LINKED ISSUE #5640: hundli_console.py needs to use mpi.restrict_op_to_first_rank
Currently has:

`if os.environ.get("OMPI_COMM_WORLD_RANK", os.environ.get("PMI_RANK", "0")) != "0"`

This doesn't work with perlmutter which uses `PALS_RANK`. 

See #4803

robnagler:
- Hardwired move to perlmutter. Cori will be deprecated shortly. Likely needed for Fedora 36.
- Fix #5640 use restrict_op_to_first_rank for main
- Fix #5639 tasksPerNode is 128 on perlmutter


mkeilman (approved):
All looks fine to me

PR #5672, opened 2023-04-12
Fix #5670: abcd propagator parameters

### LINKED ISSUE #5670: add support for ABCD matrix propgation
When the "ABCD LCT" option is selected for crystal propagation, the A B C D values should be shown and editable. These should be added as crystal fields in the schema. 

See this for an example sim which uses the ABCD propagator.

https://github.com/radiasoft/rslaser_scratch/blob/main/notebooks/wavefront/nonlinear_kick04_interpolate.ipynb

The datafiles required for that notebook are create here https://github.com/radiasoft/rslaser_scratch/blob/main/notebooks/wavefront/nonlinear_kick04_interpolate.ipynb

gurhar1133:
A, B, C, D added to the crystal propagation for abcd_lct propagator. I also turned on the l_scale parameter for abcd_lct propagation because if it is too high you will run out of memory. 0.001 is a good value for l_scale to not run out of memory


gurhar1133:
Change default l_scale to 0.001

PR #5678, opened 2023-04-13
Fix #5669: silas drag and drop bug

### LINKED ISSUE #5669: drag and drop into wrapped beamline doesn't work
In silas, when you have a long beamline (25+ elements), which wraps across rows, you can't move or drop an item into the first row.

To recreate the problem, create a new sim, and add many elements until it has wrapped a few of them, and then experiment with adding a new item to the first row.

Also, when trying to drop into the first row, the UI become very jittery as it tries to resize the beamline.

x

gurhar1133:
Think this fixes the problem on silas


moellep:
I found a case where the drop-zone would flicker if the browser width was right at the overflow size, so I made the drop zone always fixed size for silas.

PR #5683, opened 2023-04-14
fix raydata issue 86

moellep:
- trigger modal with angularjs toggle so scanId has been digested
- release jquery modal trigger when the scope is destroyed


moellep:
There were two ways the analysis_run_log compute could be called incorrectly.
- when the jquery call was used to show the modal, angularjs may have not completed the digest, so the scan id didn't make it to the subcomponent, even with the $timeout(). Instead a boolean toggle is set on the scope, and the modal is shown when that propagates down into the log component
- calling document.on() but never releasing it could cause multiple modal windows to hang around even after their UI was removed. That caused multiple calls to analysis_run_log when you would switch between tabs and open a new log. Moving the handler to $element.on() and calling $element.off() on $destroy fixed tha.

rorour (approved):
This is great, thanks for fixing this!

PR #5692, opened 2023-04-17
fix #5690 added ml export maxOutpuDimension field

### LINKED ISSUE #5690: set max scale for export to activait
In SRW, there should be an option to set the max scale for the exported ML data, otherwise the images are too large to train against.

moellep:
- Added exportRsOpt.maxOutputDimension to constrain ML training data


mkeilman:
Just noticed you requested a review on this; fwiw looks fine!

PR #5695, opened 2023-04-18
Fix #4926 added message-id on sending

### LINKED ISSUE #4926: add message-id to outgoing smtp
I thought the smtp library or postfix would add the message-id, but it doesn't. The message id should be formatted like `<datetime.random@sending-host>`.

robnagler:


robnagler:
@garsuga ping

garsuga (approved):

PR #5727, opened 2023-04-22
Fix #5704 cloudmc source from file

### LINKED ISSUE #5704: 20230501 External Source files
Allow a user to upload existing (pre-generated) external source files, per https://docs.openmc.org/en/stable/usersguide/settings.html?highlight=photon%20transport#file-based-sources

An intermediate solution is to remove the limit on the number of tabular energy bins a user can define. By my count the current limit is 19.

mkeilman:
This adds the option to load a source from an openmc file. I tested with the following:
```python
p = openmc.SourceParticle()
openmc.write_source_file([p], "openmc_src.h5")
```
then importing that file. A simulation ran without error but I'm not sure what to expect from the results. A "real" file would make a better test.


moellep:
Ok - I've found a way to test this feature and it worked correctly. When you run a simulation with a run-mode of eigenvalue, it will include a source_bank entry in each statepoint file. You can then use one of the statepoint output files as a source input to a simulation. This worked ok for me ZPRE example.

PR #5748, opened 2023-04-25
Fix #5705 cloudmc eigenvalue reporting

### LINKED ISSUE #5705: 20230501 Eigenvalue reporting
When the eigenvalue run mode is selected, we should display the eigenvalue information from the log file in a new window.

The requested information should populate a new eigenvalue window with the text at the bottom of the log file, a la:

> ============================> RESULTS <============================
>
> k-effective (Collision) = 0.92282 +/- 0.00172
> k-effective (Track-length) = 0.92209 +/- 0.00142
> k-effective (Absorption) = 0.92309 +/- 0.00183
> Combined k-effective = 0.92230 +/- 0.00114
> Leakage Fraction = 0.60136 +/- 0.00077

mkeilman:
Notes:
- Adds a table of the k value, average, and standard deviation as read from the log file
- The values are current, not a history


moellep:
I think we want a history of the eigenvalues - something like the last 5 or 10 values shown in the table. Also the values in the RESULTS section of the log should be shown. The RESULTS section will be different for eigenvalue or fixed source simulations.

mkeilman:
@moellep ready for another look. Latest changes add the eigenvalue history with configurable length and the results for both run modes.

PR #5783, opened 2023-05-01
Fix #5725: on different branch

### LINKED ISSUE #5725: Create an "app" renaming utility
The new accel app should be renamed to LLRF (sim_type llrf).

gurhar1133:
Basic implementation for a sirepo app renamer


gurhar1133:
@moellep lmk what you think

PR #5787, opened 2023-05-02
Fix #5769: reshape import

### LINKED ISSUE #5769: Import a neural network model using Reshape Layer fails
If you try to import a neural net model which includes a Reshape Layer, it will fail silently in the UI. It should show an "import failed" message in that case.

The error on the server is:
```
template/activait.py", line 1528, in _set_fields_by_layer_type PKDict(
KeyError: 'Reshape'
```

gurhar1133:
Now can import Reshape layers and will get UI error if user tries to import unsupported error

### DIFF
--- a/sirepo/package_data/static/js/activait.js
+++ b/sirepo/package_data/static/js/activait.js
@@ -2132,6 +2132,11 @@ SIREPO.viewLogic('mlModelView', function(appState, panelState, requestSender, $s
                     args: {
                         file: appState.models.mlModel.modelFile
                     }
+                },
+                {
+                    onError: data => {
+                        throw new Error(data.error);
+                    }
                 }
             );
         }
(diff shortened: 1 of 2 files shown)


mkeilman:
Is there an example of such a model that we can import to test?

moellep:
@mkeilman here is a sim which has a neural network which uses reshape. If you "Download unweighted model" and import into a different sim using the Model Type "From Model Archive File", it should be able to handle it with this PR.
https://alpha.sirepo.com/activait#/regression/hkaRfhOF

mkeilman:
> @mkeilman here is a sim which has a neural network which uses reshape. If you "Download unweighted model" and import into a different sim using the Model Type "From Model Archive File", it should be able to handle it with this PR. https://alpha.sirepo.com/activait#/regression/hkaRfhOF

Thanks - that works

Issue #5718, opened 2023-04-19
ajax calls need to be more robust


robnagler:
[Fetching the schema](https://github.com/radiasoft/sirepo/blob/630d5ee3939cd21c2c31fbf430e9a60331faa4b5/sirepo/package_data/static/js/sirepo.js#L122) can fail, and the err gets lost:

```txt
schema load failed:  SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
    at JSON.parse (<anonymous>)
    at n.parseJSON (jquery-2.2.4.min.js?20230418.194640:4:6340)
    at Ab (jquery-2.2.4.min.js?20230418.194640:4:8286)
    at z (jquery-2.2.4.min.js?20230418.194640:4:11743)
    at XMLHttpRequest.<anonymous> (jquery-2.2.4.min.js?20230418.194640:4:15619)
sirepo.js?20230418.194640:141 Uncaught TypeError: err.match is not a function
    at Object.error (sirepo.js?20230418.194640:141:25)
    at i (jquery-2.2.4.min.js?20230418.194640:2:27151)
    at Object.fireWith [as rejectWith] (jquery-2.2.4.min.js?20230418.194640:2:27914)
    at z (jquery-2.2.4.min.js?20230418.194640:4:12084)
    at XMLHttpRequest.<anonymous> (jquery-2.2.4.min.js?20230418.194640:4:15619)
```

`err` must be a different type than string. The error that the data are not json, but html.  You can have custom converters, which might be helpful. In any event, the textStatus should be checked for different errors. I don't think checking for forbidden works correctly either.

PR #5791, opened 2023-05-03
fixed a minor bug in ux on forbidden requests to simulation schema

hyperrealist:
Forwarding to /forbidden from loading screen would fail due to a bug in the regex matching of the Error object. I stumbled upon this while trying to host sirepo on a remote node.

Issue: match() is performed on a non-string Error object
Fix: Perform regex matching on a string representation of the Error object

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -138,7 +138,7 @@ angular.element(document).ready(function() {
         error: function(xhr, status, err) {
             if (! SIREPO.APP_SCHEMA) {
                 srlog("schema load failed: ", err);
-                if (err.match(/forbidden/i)) {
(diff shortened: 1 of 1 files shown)


e-carlin:
Thanks for opening this PR @hyperrealist! 

Any chance you can provide steps to reproduce? In my testing I always got back a string (ex "FORBIDDEN")

robnagler:
@e-carlin this happened to me. This fixes #5718.

e-carlin:
Ok. The ajax docs are pretty confusing. The function signature says string but then the text says object or string. In any event toString exists on both string and object so it should work either way https://api.jquery.com/jquery.ajax/

robnagler:
Thanks @ hyperrealist!

hyperrealist:
@e-carlin sorry about the late reply, but great to see it merged!

PR #5823, opened 2023-05-10
Fix #5716 rename hellweg

### LINKED ISSUE #5716: Rename rshellweg to hellweg
The hellweg app should have the url "/hellweg" and be referred to as Hellweg, not RSHellweg.

Benbenbenin0:
Mostly name/field value changes.

Tested functionality by running all panels in 'Space Charge' and 'Emittance Conservation' examples.


sirepo/pkcli/hellweg.py
```diff
@@ -12,7 +12,7 @@
 from sirepo.template import template_common
 import copy
 import py.path
-import sirepo.template.rshellweg as template
+import sirepo.template.hellweg as template
```
robnagler, line 15: This is not correct style. This should import sirepo.template, and when it needs hellweg, it should:

```py
t = sirepo.template.import_module("hellweg")
```

sirepo/pkcli/hellweg.py
```diff
@@ -12,7 +12,7 @@
 from sirepo.template import template_common
```
robnagler, line 12: Two lines up (line 10), change to:

```py
import rshellweg.solver
```

It's only used once so it doesn't need to be unqualified.

Also remove the `from __future__`

robnagler (changes requested):
Minor issues. Also ./tests/animation_test.py needs to be updated.

robnagler (approved):

PR #5828, opened 2023-05-11
script to find cut off runs

rorour:
takes an argument for MAX_SECS and optionally TIME_WINDOW (defaults to 5).

### DIFF
--- a/calculate_run_time.py
+++ b/calculate_run_time.py
@@ -1,0 +1,16 @@
+import csv
+import glob
+import json
+import sys
+
+MAX_SECS = int(sys.argv[1])
+TIME_WINDOW = int(sys.argv[2]) if len(sys.argv) > 2 else 5
+
+with open('out.csv', 'w') as f:
+    w = csv.writer(f)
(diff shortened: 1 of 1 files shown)


robnagler:
This was interesting. At first I thought, "please use pykern utilities". This turned out to not be so useful, because I was running this on CentOS 7 with /usr/bin/python which is version 2.7.5 so your code worked without changes.

Here's the script I finally ended up with:

```py
import csv
import datetime
import glob
import json
import sys

MAX_SECS = int(sys.argv[1])
TIME_WINDOW = int(sys.argv[2]) if len(sys.argv) > 2 else 5

with open('../out.csv', 'w') as f:
    w = csv.writer(f)
    w.writerow(["file", "type", "start"])
    for n in sorted(glob.glob('*.json')):
        j = json.load(open(n))
        for x in j['history']:
            if "status" not in x:
                continue
            d = int( x['lastUpdateTime'])-int(x['computeJobStart'])
            if x['status'] == 'canceled' and d + TIME_WINDOW >= MAX_SECS and d <= MAX_SECS:
                w.writerow([n, j['simulationType'], datetime.date.fromtimestamp(int(x['computeJobStart'])).strftime("%Y-%m-%d")])
```

PR #5838, opened 2023-05-15
fix raydata#89: csx workflow

rorour:
depends on https://github.com/radiasoft/raydata/pull/90

### DIFF
--- a/sirepo/raydata/scan_monitor.py
+++ b/sirepo/raydata/scan_monitor.py
@@ -215,7 +215,7 @@ class _Req(_JsonPostRequestHandler):
             )
 
         def _filename_and_json(path):
-            return PKDict(filename=str(path.basename), json=pkio.read_text(path))
+            return PKDict(filename=path.basename, json=pkjson.load_any(path))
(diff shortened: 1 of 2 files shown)


sirepo/raydata/scan_monitor.py
```diff
@@ -214,11 +214,20 @@ def _filename_and_image(path):
                 ),
             )
 
-        def _paths(uid):
+        def _filename_and_json(path):
+            return PKDict(filename=str(path.basename), json=pkio.read_text(path))
```
e-carlin, line 218: I think path.basename is already a string

Instead of read_text if you call pkjson.load_any then you can remove the JSON.parse in the GUI

e-carlin (approved):
Looks good! Just a couple nits. You can merge once fixed.

sirepo/raydata/scan_monitor.py
rorour: f9683c34b2bc41b6f042087a75e103166fba7053

PR #5840, opened 2023-05-15
Fix #5810: shifter self test

### LINKED ISSUE #5810: Shifter MPI self-test function
Should run SRW in a simple way in MPI (a few particles) and validate the output. If it fails, produce diagnostics. If it succeeds, prints SUCCESS or some sentinel.

gurhar1133:


sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+_SEQUENTIAL_TEST_BASH_FILE = "sequential_test.sh"
+_SEQUENTIAL_TEST_JSON = "nersc_sequential.json"
+_RUN_DIR = "sirepo_run_dir"
+_SEQUENTIAL_RESULT_FILE = "nersc_test_result.json"
+
+
+def sequential():
+    s = pkio.py_path(_RUN_DIR)
```
robnagler, line 22: Add a docstring explaining what the command does, it's environment and output.

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,51 @@
...
+from pykern import pkio
+from pykern import pkjson
+from pykern import pksubprocess
+import sirepo.sim_data
+
+
+_SEQUENTIAL_TEST_BASH_FILE = "sequential_test.sh"
+_SEQUENTIAL_TEST_JSON = "nersc_sequential.json"
+_RUN_DIR = "sirepo_run_dir"
+_SEQUENTIAL_RESULT_FILE = "nersc_test_result.json"
```
robnagler, line 18: Keep the name consisent (indeed share it in a constant). The output might not be json so call it `nersc_sequential.log`.

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+"""Nerc test suite
+
+:copyright: Copyright (c) 2023 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkcollections import PKDict
+from pykern.pkdebug import pkdc, pkdexc, pkdp, pkdlog
+from pykern import pkio
```
robnagler, line 9: Do fully qualified imports (pykern.pkio, etc.). We are inconsistent, but the default should be qualified.

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,51 @@
...
+    b = f"""
+#!/bin/bash
+set -e
+cd 'sirepo_run_dir'
+export SIREPO_MPI_CORES='1'
+export PYKERN_PKCONFIG_CHANNEL='dev'
+export PYKERN_PKDEBUG_WANT_PID_TIME='1'
+export PYTHONUNBUFFERED='1'
+export SIREPO_AUTH_LOGGED_IN_USER='someuser'
+export SIREPO_JOB_MAX_MESSAGE_BYTES='200000000'
```
robnagler, line 35: This can be dropped.

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,51 @@
...
+export SIREPO_SIMULATION_DB_LOGGED_IN_USER='someuser'
+export SIREPO_SRDB_ROOT=$PWD
```
robnagler, line 37: Try dropping this one

robnagler (changes requested):

sirepo/package_data/template/srw/nersc_sequential.json
```diff
@@ -0,0 +1,887 @@
+{
```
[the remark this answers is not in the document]
gurhar1133: best way to resolve path to `package_data/nersc_test/nersc_sequential.json`?
Cant use
```python
sirepo.sim_data.resource_path(_SEQUENTIAL_TEST_JSON)
```
or 
```python
sirepo.sim_data.get_class("nersc_test").resource_path(_SEQUENTIAL_TEST_JSON)
```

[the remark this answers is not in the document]
robnagler: sirepo.resource_path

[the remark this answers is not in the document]
gurhar1133: wait, what do you mean? There doesnt seem to be a sirepo.resource_path()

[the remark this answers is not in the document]
robnagler: look in sirepo/resource.py

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,62 @@
...
+def sequential():
+    """Sequential testing of job_cmd for nersc. Generates bash script
+    that invokes job_cmd on a srw in.json with one MPI core.
+    """
+    try:
+        s = pykern.pkio.py_path(_RUN_DIR)
+        pykern.pkio.unchecked_remove(s)
+        s.ensure(dir=True)
+        o = s.join(_SEQUENTIAL_RESULT_FILE)
+        shutil.copyfile(
```
robnagler, line 36: The copy should not be necessary. Rather, use pkjinja.render_resource. There's a subtlety here, which creates a problem that we need for another issue. We should use sirepo.resource, but it doesn't have a render_file. It should so create an issue for that. sirepo.resource manages package_path, which we need to conform to.

sirepo/package_data/nersc_test/sequential_test.sh.jinja
```diff
@@ -0,0 +1,13 @@
...
+cd sirepo_run_dir
+export SIREPO_MPI_CORES=1
+export PYKERN_PKCONFIG_CHANNEL=dev
+export PYKERN_PKDEBUG_WANT_PID_TIME=1
+export PYTHONUNBUFFERED=1
+export SIREPO_AUTH_LOGGED_IN_USER={{user}}
+export SIREPO_SIMULATION_DB_LOGGED_IN_USER={{user}}
+export PYTHONPATH=
+export PYTHONSTARTUP=
+perl -p -e "s<{{sirepo_run_dir}}><$PWD>" '{{json_in_path}}' > 'jobCmdIn.json'
```
robnagler, line 12: This is not correct, because there is no `{{sirepo_run_dir}}` in the json file. Indeed, you should simply generate jobCmdIn.json (or just call it sequential_test.json) in nersc_test.py.

robnagler (changes requested):

sirepo/pkcli/nersc_test.py
```diff
@@ -0,0 +1,62 @@
...
+        pykern.pksubprocess.check_call_with_signals(
+            ["bash", s.join(_SEQUENTIAL_TEST_BASH_FILE)],
+            output=str(o),
+        )
+        r = pykern.pkjson.load_any(o)
+        if r.state != "completed":
+            raise RuntimeError(f"incomplete result {r.state}")
+        return "nersc_test.sequential PASS"
+    except Exception as e:
+        return e
```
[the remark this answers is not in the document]
gurhar1133: Declare `s` before the try? return it in a PKDict with the exception or something?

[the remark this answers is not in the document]
robnagler: The "return" here is to an end-user, probably at NERSC. Therefore, it should be text, not a dict. We should describe the error so we can help debug. We will not have access to the environment to reproduce the error. You might provide more information that is obvious, e.g. what unix user id the process is running as.

tests/pkcli/nersc_test_test.py
```diff
@@ -0,0 +1,36 @@
...
+]
+
+
+def test_sequential():
+    w = pykern.pkunit.work_dir()
+    with pykern.pkio.save_chdir(w):
+        pykern.pksubprocess.check_call_with_signals(
+            ["sirepo", "nersc_test", "sequential"]
+        )
+        o = list(sorted([w.bestrelpath(f) for f in pykern.pkio.walk_tree(w)]))
```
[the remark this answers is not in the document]
gurhar1133: For a deviance test we could add kwargs or optional params to `sequential`. Then pass something that triggers an error?

gurhar1133:
@robnagler made changes

robnagler (commented):
I decided to code my comments in f5f03e5f3

robnagler:
I think it would now good to test a failure case. This is tricky to do, but I think it might be possible if we supply an invalid in.json to the test. sequential could take an argument with the name of the in.json to use instead of JOB_CMD_FILE. It would be a small if case which is probably worth it.

tests/pkcli/nersc_test_test.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+"""Test pkcli.nersc_test
+
+:copyright: Copyright (c) 2023 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+import pytest
+import pykern.pkio
```
robnagler, line 8: We avoid imports at the global scope, because the import may fail, which is better caught in a test.

pytest does support classes for testing. It's better if the two cases are separate entry points that pytest can see, e.g. test_conformance and test_deviance. The words "success" and "failure" are confusing in the context of a test of a test so stick to stricter terms.

The class I think here is overkill but not unreasonably so. I would recommend simplifying and sharing in a different way. For example, exec_cmd could hold the value CmD_BASIC. "result.log" could be part of "check_result" (not here, but could be). Then you could simplify the two cases.

robnagler (changes requested):

robnagler (commented):
I made comments in cb935871604c49011711dd4f53210d858722955b. Try to understand all the changes, and set up a meeting so we can discuss your observations.

gurhar1133:
@robnagler anything else needed for this one?

robnagler:
Update branch and merge. We can then look at implementing a parallel test with shifter in sbatch, which is trickier.

gurhar1133:
> Update branch and merge. We can then look at implementing a parallel test with shifter in sbatch, which is trickier.

unblock the merge?

robnagler (approved):

PR #5848, opened 2023-05-17
Fix #5596 Add file_lock concurrency protection

### LINKED ISSUE #5596: Use sirepo.file_lock and limit unnecessary locking
Look at shared memory locking, fnctl locking, etc.

Locking should be simplified:

- [x] #5740
- [x] #5741
- [x] #5742
- [x] #5743
- [x] #5744
- [x] #5745
- [x] #5746
- [x] #5747
- [x] #5749
- [x] #5750 
- [x] #5751 
- [x] #5758

### LINKED ISSUE #5740: simulation_db._serial_new should take old serial
The current `_serial_new` code is problematic for Tornado. A global lock (#5596) won't fix the issue. It's highly unlikely that we'll have collisions on the same sdata file, which is all that matters. The simulation id is already unique so really it's just best to make sure the new serial is bigger than value in the file, which can be read just for this purpose.

### LINKED ISSUE #5742: Restrict critical sections
Global lock sections should be avoided.

### LINKED ISSUE #5743: simulation_db.migrate_guest_to_persistent_user lock guest_uid dir
The guest uid directory must exist so lock it instead of locking globally.

### LINKED ISSUE #5744: simulation_db.simulation_dir lock user_dir

### LINKED ISSUE #5745: simulation_db validate_serial should be in save_simulation_json
The simplest approach would be to validate all input so try that first. May be that needs to be a parameter on save_simulation_json. 

Prereq #5741

### LINKED ISSUE #5746: auth.unchecked_get_user does not need a lock
It's a single query

### LINKED ISSUE #5747: auth.complete_registration should pass display_name to user_registration
It doesn't need to lock.

### LINKED ISSUE #5749: auth_db does not need locking
After re-reading the [SQLite Transaction Control](https://sqlite.org/lang_transaction.html) and [File Locking and Concurrency](https://sqlite.org/lockingv3.html), I'm quite sure we don't need locking for sqlite operations. One of the big fixes to our code was to not use a global connection. Connections are bound to qcalls, and they are definitely single threaded. Nested call_apis commit before calling to ensure a single thread of control (asyncio task).

### LINKED ISSUE #5750: spa_session only needs thread locking in flask

### LINKED ISSUE #5751: factor out is_mock from auth_login
Complicates testing because login needs to be async with file locking (#5596)

robnagler:
This should not be merged until 6/12/23 to avoid issues with USPAS. @moellep and @e-carlin tests pass, and I did some click around testing, but both of you testing would be useful, especially with on and off `SIREPO_PKCLI_SERVICE_TORNADO`. I'm mostly testing with tornado on.

- Fix #5740 ensure serial greater than incoming and on disk
- Fix #5742 restricted critical sections and removed most
- Fix #5743 migrate_guest_to_persistent_user locks new and guest uid
- Fix #5744 simulation_dir uses lock_user
- Fix #5745 validate version in save_simulation_json
- Fix #5746 removed lock on auth.unchecked_get_user
- Fix #5747 complete_registration passed display_name
- Fix #5749 removed auth_db locking, verified all critical sections
- Fix #5750 thread locking only in flask
- Fix #5751 add auth.create_and_login_user for testing


e-carlin (approved):
Some comments on style and questions but everything looks good. I ran it in threaded uwsgi mode and didn't see any errors.

sirepo/file_lock.py
```diff
@@ -0,0 +1,67 @@
...
+                try:
+                    os.close(f)
+                except Exception:
+                    pass
+            time.sleep(_LOOP_SLEEP)
+        raise RuntimeError(f"fail to flock path={self._path} timeout={_cfg.timeout}")
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._lock:
+            os.unlink(self._path)
```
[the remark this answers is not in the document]
robnagler: Tradition. It maps to the many C examples for flock.

sirepo/simulation_db.py
```diff
@@ -344,32 +359,29 @@ def lib_dir_from_sim_dir(sim_dir):
...
-            dir_path = os.path.dirname(path)
-            new_dir_path = dir_path.replace(guest_uid, to_uid)
-            pkdlog("{} -> {}", dir_path, new_dir_path)
-            pkio.mkdir_parent(new_dir_path)
-            os.rename(dir_path, new_dir_path)
-
-
-def open_json_file(sim_type, path=None, sid=None, fixup=True, save=False, qcall=None):
+    with user_lock(uid=guest_uid, qcall=qcall) as g:
+        with user_lock(uid=uid, qcall=qcall):
```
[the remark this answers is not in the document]
robnagler: 7aa555d21

sirepo/file_lock.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, path):
+        p = pykern.pkio.py_path(path)
+        if p.check(dir=True):
+            p = p.join("lock")
+        else:
+            p += ".lock"
+        self._path = str(p)
+
+    def __enter__(self):
+        for i in range(_LOOP_COUNT):
```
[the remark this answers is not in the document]
robnagler: 7aa555d21

sirepo/file_lock.py
```diff
@@ -0,0 +1,67 @@
...
+                f = os.open(self._path, os.O_RDWR | os.O_CREAT | os.O_TRUNC)
+                fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                # Verify open file and path on disk are same file
+                # https://stackoverflow.com/a/18745264
+                if os.stat(f).st_ino == os.stat(self._path).st_ino:
+                    self._lock = f
+                    return None
+            except (IOError, OSError, FileNotFoundError):
+                pass
+            if f:
```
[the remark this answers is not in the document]
robnagler: 57de145c

sirepo/file_lock.py
```diff
@@ -0,0 +1,67 @@
...
+        self._path = str(p)
+
+    def __enter__(self):
+        for i in range(_LOOP_COUNT):
+            try:
+                f = os.open(self._path, os.O_RDWR | os.O_CREAT | os.O_TRUNC)
+                fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                # Verify open file and path on disk are same file
+                # https://stackoverflow.com/a/18745264
+                if os.stat(f).st_ino == os.stat(self._path).st_ino:
```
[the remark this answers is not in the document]
robnagler: 57de145c

sirepo/auth_db/user.py
```diff
@@ -34,40 +34,37 @@ def add_roles(self, roles, expiration=None):
...
 
     def add_role_or_update_expiration(self, role, expiration):
-        with sirepo.util.THREAD_LOCK:
-            if not self.has_role(role):
-                self.add_roles(roles=[role], expiration=expiration)
-                return
-            r = self.search_by(uid=self.logged_in_user(), role=role)
-            r.expiration = expiration
-            r.save()
+        if not self.has_role(role):
```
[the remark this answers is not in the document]
robnagler: auth_db executes an explicit [sqlalchemy.orm.Session.begin](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.begin):

> Begin a transaction, or nested transaction, on this [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session), if one is not already begun.

[7.0 Transaction Control At The SQL Level](https://www.sqlite.org/lockingv3.html) describes how SQLite handles this:

> The SQL command "BEGIN TRANSACTION" (the TRANSACTION keyword is optional) is used to take SQLite out of autocommit mode. Note that the BEGIN command does not acquire any locks on the database. After a BEGIN command, a SHARED lock will be acquired when the first SELECT statement is executed. A RESERVED lock will be acquired when the first INSERT, UPDATE, or DELETE statement is executed. No EXCLUSIVE lock is acquired until either the memory cache fills up and must be spilled to disk or until the transaction commits. In this way, the system delays blocking read access to the file until the last possible moment.

This explains the behavior that you were seeing a few days ago. The commit is what will cause the integrity error, and it doesn't happen except at the last possible moment. In normal databases, it would detect the integrity error in a transaction manager, but SQLite doesn't have any shared state except the file itself.

If two applications write simultaneously, one will get a constraint violation, possibly later than you would expect, which is why we should switch to a real database.

sirepo/server.py
```diff
@@ -13,6 +13,7 @@
 import re
 import sirepo.const
 import sirepo.feature_config
+import sirepo.file_lock
```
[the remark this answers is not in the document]
robnagler: 95b5a6c9e

robnagler:
@e-carlin note [the has_role change](https://github.com/radiasoft/sirepo/pull/5848/files#diff-913b3c8d56b5839a703e3c0b41a7fee071aa9fc1da2d7861cdb56a8b75341f04R39), because it was a merge conflict with code you wrote.

sirepo/auth_db/user.py
```diff
@@ -33,38 +33,40 @@ def all_roles(self):
...
-        with sirepo.util.THREAD_LOCK:
-            u = self.logged_in_user()
-            for r in roles:
-                if self.has_role(r):
-                    continue
-                self.new(uid=u, role=r, expiration=expiration).save()
-            sim_data.audit_proprietary_lib_files(qcall=self.auth_db.qcall)
+        u = self.logged_in_user()
+        for r in roles:
+            try:
```
e-carlin, line 38: The try/except can be removed. The error won't be raised here. All we need is the if check.

e-carlin (commented):
One comment on the merge conflict. Other than that looks good. Approving but the comment needs to be fixed.

Thanks for adding functions/comments to file_lock.py. I found the code easier to understand.

sirepo/auth_db/user.py
robnagler: There's a race condition so you need both. I thought [my comment explained](https://github.com/radiasoft/sirepo/pull/5848/files/9702d537fd4d5b1a5a0d429268513bb0fe66ba67) the reason for the `if`, which is superfluous with a "real" db.

e-carlin: typo: through -> throw

I believe it doesn't throw errors are the point of the `save` either. `save` just adds to the session (still not sure what that exactly means but it doesn't seem to do any checks). It is the `commit` that raises the error. So the try/except here won't be used. I created the following script to highlight this. But, the script doesn't have any race condition so I may be missing that key part of your comment.

```python
from pykern.pkdebug import pkdc, pkdlog, pkdp
import sqlalchemy
import sqlalchemy.ext.declarative
import sqlalchemy.orm

_engine = None


@sqlalchemy.ext.declarative.as_declarative()
class UserDbBase:
    def new(self, **fields):
        return self.auth_db.model(self.__class__.__name__, **fields)

    def save(self):
        self.auth_db.session().add(self)


class _AuthDb(PKDict):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self._orm_session = None

    def commit(self):
        self._commit_or_rollback(commit=True)

    def create_or_upgrade(self):
        self.metadata().create_all(bind=_engine)

    def metadata(self):
        return UserDbBase.metadata

    def model(self, class_name_, **fields):
        x = UserRegistration(**fields)
        x.auth_db = self
        return x

    def new(self, **fields):
        return self.auth_db.model(self.__class__.__name__, **fields)

    def session(self):
        if self._orm_session is None:
            # New in sqlalchemy 2.0 autobegin, which we should set to False
            self._orm_session = sqlalchemy.orm.Session(bind=_engine)
            self._orm_session.begin()
        return self._orm_session

    def _commit_or_rollback(self, commit):
        if self._orm_session is None:
            return
        s = self._orm_session
        self._orm_session = None
        if commit:
            s.commit()
        else:
            s.rollback()
        s.close()


class UserRegistration(UserDbBase):
    __tablename__ = "user_registration_t"
    uid = sqlalchemy.Column(sqlalchemy.String(8), primary_key=True)

    def add_records(self):
        try:
            for _ in range(3):
                self.new(uid="12345678").save()
        except sqlalchemy.exc.IntegrityError:
            raise AssertionError(
                "This should never be raised. IntegrityError happens in commit."
            )


def main():
    global _engine
    _engine = sqlalchemy.create_engine(
        f"sqlite:///test.sqlite",
        # We ensure single threaded access by not sharing connections.
        # TODO(robnagler) set pool_class=NullPool with sqlalchemy 2.0
        connect_args={
            "check_same_thread": False,
            "timeout": 2,
        },
        # echo=True,
        # echo_pool=True,
    )
    a = _AuthDb()
    a.create_or_upgrade()
    a.session()
    a.model("UserRegistration").add_records()
    pkdp("integrity error is not raised")
    a.commit()
    raise AssertionError("we will never get here")


main()
```

robnagler: > I believe it doesn't throw errors are the point of the save either. save just adds to the session (still not sure what that exactly means but it doesn't seem to do any checks). It is the commit that raises the error. So the try/except here won't be used.

Security cannot be baked in after the fact, neither can robustness. Correct coding of sqlalchemy requires that we catch the integrity contraint at the time of the insert. With a "real" database, e.g. Postgres, DML hangs when there is a conflict so that you can write code like the try/except, and it'll do the right thing. SQLite "cheats", because there is no fine grained locking [like with Postgres](https://www.postgresql.org/docs/current/transaction-iso.html). The `if` is a hack that is not incorrect (rather, it's goofy), because there is no LBFYL with a db with a transaction manager. The db delays the DML, because there's an active transaction (**potential** conflict). Once the commit happens in the other client (other process doing a simultaneous insert, for example), the DML of the current process continues with an integrity constraint exception which bubbles up to sqlalchemy so it is true EAFP. This type of behavior is crucial for scale of ACID (as opposed to eventually consistent, which Google uses for scale).

Perhaps a whiteboard would help.

e-carlin: I appreciate your detailed thoughts.

> Security cannot be baked in after the fact, neither can robustness. Correct coding of sqlalchemy requires that we catch the integrity contraint at the time of the insert

This smells like security theater. I was curious so I tested it and it turns out it is (with the large caveat that I could easily have a bug in my code). Using the above script but switching to postgres (`postgresql+psycopg2://postgres:abc123@127.0.0.1:5432/rs_test`) it raises at the `commit` not at the `save` as does sqlite. Furthermore, it raises `psycopg2.errors.UniqueViolation` not `IntegrityError`.

```
sirepo (5596-file-lock)$ python sql.py
/home/vagrant/src/radiasoft/sirepo/sql.py:10: MovedIn20Warning: Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environ\
ment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
  @sqlalchemy.ext.declarative.as_declarative()
sql.py:87:main integrity error is not raised
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1885, in _execute_context
    self.dialect.do_executemany(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 982, in do_executemany
    context._psycopg2_fetched_rows = xtras.execute_values(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/psycopg2/extras.py", line 1299, in execute_values
    cur.execute(b''.join(parts))
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "user_registration_t_pkey"
DETAIL:  Key (uid)=(12345678) already exists.


The above exception was the direct cause of the following exception:

  File "/home/vagrant/src/radiasoft/sirepo/sql.py", line 92, in <module>
    main()
  File "/home/vagrant/src/radiasoft/sirepo/sql.py", line 88, in main
    a.commit()
  File "/home/vagrant/src/radiasoft/sirepo/sql.py", line 25, in commit
    self._commit_or_rollback(commit=True)
  File "/home/vagrant/src/radiasoft/sirepo/sql.py", line 54, in _commit_or_rollback
    s.commit()
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 1454, in commit
    self._transaction.commit(_to_root=self.future)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 832, in commit
    self._prepare_impl()
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 811, in _prepare_impl
    self.session.flush()
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 3449, in flush
    self._flush(objects)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 3589, in _flush
    transaction.rollback(_capture_exception=True)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/util/langhelpers.py", line 70, in __exit__
    compat.raise_(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/util/compat.py", line 211, in raise_
    raise exception
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 3549, in _flush
    flush_context.execute()
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/unitofwork.py", line 456, in execute
    rec.execute(self)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/unitofwork.py", line 630, in execute
    util.preloaded.orm_persistence.save_obj(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/persistence.py", line 245, in save_obj
    _emit_insert_statements(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/orm/persistence.py", line 1097, in _emit_insert_statements
    c = connection._execute_20(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1710, in _execute_20
    return meth(self, args_10style, kwargs_10style, execution_options)
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/sql/elements.py", line 334, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1577, in _execute_clauseelement
    ret = self._execute_context(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1948, in _execute_context
    self._handle_dbapi_exception(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 2129, in _handle_dbapi_exception
    util.raise_(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/util/compat.py", line 211, in raise_
    raise exception
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1885, in _execute_context
    self.dialect.do_executemany(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 982, in do_executemany
    context._psycopg2_fetched_rows = xtras.execute_values(
  File "/home/vagrant/.pyenv/versions/py3/lib/python3.9/site-packages/psycopg2/extras.py", line 1299, in execute_values
    cur.execute(b''.join(parts))
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "user_registration_t_pkey"
DETAIL:  Key (uid)=(12345678) already exists.

[SQL: INSERT INTO user_registration_t (uid) VALUES (%(uid)s)]
[parameters: ({'uid': '12345678'}, {'uid': '12345678'}, {'uid': '12345678'})]
(Background on this error at: https://sqlalche.me/e/14/gkpj)
```

e-carlin: I hacked together some threaded code and it also looks like it is raised in the commit

```python
    def add_record(self):
        try:
            self.new(uid="00005678").save()
        except sqlalchemy.exc.IntegrityError:
            raise AssertionError(
                "This should never be raised. IntegrityError happens in commit."
            )


def main():
    global _engine
    _engine = sqlalchemy.create_engine(
        f"postgresql+psycopg2://postgres:abc123@127.0.0.1:5432/rs_test",
        # We ensure single threaded access by not sharing connections.
        # TODO(robnagler) set pool_class=NullPool with sqlalchemy 2.0
        # echo=True,
        # echo_pool=True,
    )

    def a():
        pkdp("begin")
        a = _AuthDb()
        a.create_or_upgrade()
        a.session()
        a.model("UserRegistration").add_record()
        a_added_record_event.set()
        b_commited_event.wait()
        pkdp("integrity error is not raised")
        try:
            a.commit()
        except:
            pkdp("error in commit stack={}", pkdexc())
        pkdp("done")

    def b():
        pkdp("begin")
        a_added_record_event.wait()
        a = _AuthDb()
        a.create_or_upgrade()
        a.session()
        a.model("UserRegistration").add_records()
        pkdp("integrity error is not raised")
        try:
            a.commit()
        except:
            pkdp("error in commit stack={}", pkdexc())
        pkdp("done")
        b_commited_event.set()

    a_added_record_event = threading.Event()
    b_commited_event = threading.Event()
    a = threading.Thread(target=a)
    b = threading.Thread(target=b)
    a.start()
    b.start()
    a.join()
    b.join()
```

e-carlin: I retract my previous statement about the error type. The stack trace confused me but the error is IntegrityError if you catch it in the commit.

robnagler: This is my misunderstanding of how sqlalchemy works. It seems to think it can substitute for the dbms, which it shouldn't do. sqlalchemy has a concept of "flushing", which is equivalent to Unix sync for file systems.

Try your example with this one additional line:

```py
            for _ in range(3):
                self.new(uid="12345678", name=f"name{_}").save()
            self.auth_db.session().query(self.__class__).all()
```

The IntegrityError will happen here when the query is executed (`all`), because [autoflush](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.no_autoflush) will send the commands to the dbms.

To fix sqlalchemy, we simply have to do:

````py
def save(self):
        self.auth_db.session().add(self)
        self.auth_db.session().flush()
```

The `flush()` will cause sqlalchemy to send the DML to the db on `save`, always -- as it should imiho. We may want it on all DML calls, e.g. delete and delete_all. I don't understand why delete_all_for_column_by_values calls execute directly (which probably flushes implicitly), and maybe now I do: someone probably needed the DML to flush and execute probably flushes.

e-carlin: That makes sense. I think adding flush is the right strategy. I tested and it produced exceptions/deadlock in the ways I would expect.

If we add `flush` to the `save` can we remove the `if not self.has_role(r):` check?

I tried doing that and ran into the below error:
```
sirepo/auth_db/user.py:42:add_roles role exists
sirepo/quest.py:41:start c=True
sirepo/quest.py:106:destroy destroy failed attr={'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': Fal\
se}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': \
{<SNIP>}, 'auth_db': {<SNIP>}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth\
': {<SNIP>}, 'auth_db': {<SNIP>}}}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, 'au\
th': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, \
'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'i\
n_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNI\
P>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<\
SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket': \
{<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {'in_pkcli': False}, 'auth': {'_logged_in_method': None, '_logged_in_user': None, 'qcall': {'_bucket\
': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}, 'auth_db': {'_orm_session': None, 'qcall': {'_bucket': {<SNIP>}, 'auth': {<SNIP>}, 'auth_db': {<SNIP>}}}}}}}}}}} stack=Traceback (most recent call last):
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/quest.py", line 104, in destroy
    v.destroy(commit=commit)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth_db/__init__.py", line 182, in destroy
    self._commit_or_rollback(commit=commit)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth_db/__init__.py", line 228, in _commit_or_rollback
    s.commit()
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 1454, in commit
    self._transaction.commit(_to_root=self.future)
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 830, in commit
    self._assert_active(prepared_ok=True)
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/sqlalchemy/orm/session.py", line 604, in _assert_active
    raise sa_exc.PendingRollbackError(
sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (sqlite3.IntegrityError) UNIQUE constraint failed: user_role_t.uid, user_role_t.role
[SQL: INSERT INTO user_role_t (uid, role, expiration) VALUES (?, ?, ?)]
[parameters: ('MpoAhWin', 'sim_type_jupyterhublogin', None)]
(Background on this error at: https://sqlalche.me/e/14/gkpj) (Background on this error at: https://sqlalche.me/e/14/7s2a)

Exception was printed at:

sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (sqlite3.IntegrityError) UNIQUE constraint failed: user_role_t.uid, user_role_t.role
[SQL: INSERT INTO user_role_t (uid, role, expiration) VALUES (?, ?, ?)]
[parameters: ('MpoAhWin', 'sim_type_jupyterhublogin', None)]
(Background on this error at: https://sqlalche.me/e/14/gkpj) (Background on this error at: https://sqlalche.me/e/14/7s2a)
  File "/home/vagrant/.pyenv/versions/py3/bin/sirepo", line 33, in <module>
    sys.exit(load_entry_point('sirepo', 'console_scripts', 'sirepo')())
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/sirepo_console.py", line 18, in main
    return pkcli.main("sirepo")
  File "/home/vagrant/src/radiasoft/pykern/pykern/pkcli/__init__.py", line 158, in main
    res = argh.dispatch(parser, argv=argv)
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/argh/dispatching.py", line 179, in dispatch
    for line in lines:
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/argh/dispatching.py", line 290, in _execute_command
    for line in result:
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/argh/dispatching.py", line 273, in _call
    result = function(*positional, **keywords)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/jupyterhublogin.py", line 66, in create_user
    return PKDict(email=email, jupyterhub_user_name=n)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/contextlib.py", line 126, in __exit__
    next(self.gen)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/quest.py", line 42, in start
    qcall.destroy(commit=c)
{'email': 'x@x.x', 'jupyterhub_user_name': 'x'}
```

It looks like we have to do some cleanup of the session after we do the flush because a failing flush does a rollback. In the case of jupyterhublogin create-user (add_roles) we are just ignoring the IntegrityError and trying to commit at the end. Without fixing the session the commit at the end of the quest fails (sqlalchemy.exc.PendingRollbackError). Possibly further interactions with the session/db before the commit would fail too. I didn't test. 

I didn't test this on postgres because I ran into many unrelated errors getting the db setup for a test. I can resolve those and test on a postgres if you think it would be of value? But, my intuition is saying that regardless of the db if we add `flush` we need to start handling the `PendingRollbackError`. 


FWIW [this is the code](https://github.com/radiasoft/sirepo/files/11752574/test.patch) I was using to test. I called `bash etc/run-jupyterhub.sh` to produce the error.

robnagler: https://github.com/radiasoft/sirepo/issues/5977

e-carlin (approved):
lgtm. Thanks for the disccusion on sqlalchemy

PR #5876, opened 2023-05-23
Fix #5862: check model state

### LINKED ISSUE #5862: framePeriod needs to check model state
I saw an error "Cannot read properties of undefined (reading 'framesPerSecond')" when playing through a hellweg animation, and selecting the "Simulations" link. The code in FrameCache.getFrame() needs to check when the response is handled - is the simulation still loaded? Is it the same simulation that made the request? The response can come back after the user has clicked away.

gurhar1133:
Sim id consistency and loaded is checked now


sirepo/package_data/static/js/sirepo.js
```diff
@@ -1320,6 +1321,12 @@ SIREPO.app.factory('frameCache', function(appState, panelState, requestSender, $
                     '<frame_id>': self.frameId(modelName, index),
                 },
                 function(data) {
+                    const c = appState.models.simulation.simulationId;
+                    if (!appState.isLoaded()) {
+                        if (! c || c != i) {
```
mkeilman, line 1326: Use double equal `c !== i`

mkeilman (commented):
Minor comments

PR #5891, opened 2023-05-24
fix radiasoft/raydata#62: remove unused options (Simulation Documentation URL, Export as Zip, Python Source) in raydata settings submenu

rorour:


e-carlin (changes requested):
Looks good.

One fix: A user can still click "export as zip" in the simulations list menu. This should be disabled as well
Screen Shot 2023-05-30 at 11 03 04 AM

rorour:
@e-carlin ping

e-carlin (approved):
Looks good! Sorry for the delay reviewing.

PR #5906, opened 2023-05-29
Fix #5905: db_upgrade to deploy flash update

### LINKED ISSUE #5905: deploy flash changes
Need to audit_proprietary_lib_files with `force=True` so users get the latest flash.tar.gz (https://github.com/radiasoft/download/pull/479)

e-carlin:
Update users flash.tar.gz because it contains an update https://github.com/radiasoft/download/pull/479

### DIFF
--- a/sirepo/db_upgrade.py
+++ b/sirepo/db_upgrade.py
@@ -168,10 +168,6 @@ def _20210211_upgrade_runner_to_job_db(qcall):
             pkdlog(s, **k)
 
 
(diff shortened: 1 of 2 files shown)


robnagler (approved):

robnagler:
> db_upgrade_work/db/auth.txt

I wonder if this is useful. If you fix this, you should also pull the old upgrade since it doesn't make sense to execute three times.

e-carlin:
I think it's worthwhile. Better to be safe than sorry with a db upgrade.

PR #5923, opened 2023-06-01
Fix #5879 radia cylindrical divisions

### LINKED ISSUE #5879: Add cylindrical divisions
We currently allow the user to subdivide objects in a rectangular grid. Add an explicit interface to Radia's cylindrical divisions, which we implicitly call when constructing cylinders and fillets.

mkeilman:
This expanded a bit to clean some things up. Notes:
- Users can specify planar or cylindrical segmentation
- Radia performs cylindrical segmentation in such a way that the "number of segments in the radial direction" is not meaningful. That field is hidden in favor of the radial segment width
- The cylinder origin can be set to track the object's center
- The cleanup includes a change in the jinja from a big if-then to using dispatch in `radia_util` and replacing single with double quotes in the generated python


mkeilman:
@moellep should be ready now

PR #5956, opened 2023-06-08
Fix #5931: error if no epics server on start monitor

### LINKED ISSUE #5931: Show error message if initial epics connection fails
In the accel app, if you start the monitor but there is no epics process running, it waits a bit and then stops, but doesn't show an error message. It should show a failure-to-connect message, with the host and port shown.

gurhar1133:
Error goes to sim status if no epics server on start monitor


mkeilman (approved):
Approved!

PR #5958, opened 2023-06-08
Fix #5949: thermal lensing optional

### LINKED ISSUE #5949: Allow turning thermal lensing off
And option "Include Thermal Lensing" should be added to the Propagation tab. If not selected, "Transverse Variation of Thermal Lensing" should be hidden.

gurhar1133:
made thermal lensing optional


moellep:
We need more information for this issue - just hiding another checkbox based on a new checkbox value seems odd.

gurhar1133:
> We need more information for this issue - just hiding another checkbox based on a new checkbox value seems odd.

do we know what we want to do here?

moellep:
We should open a new PR for this when we have more details on what gets enabled/disabled when thermal lensing is optional. I'm closing the PR for now.

PR #5967, opened 2023-06-14
fix #5414 added precision and removed quotes from elegant csv output

### LINKED ISSUE #5414: 202303 elegant csv file output
This was reported by a user.
The CSV files produced for Sirepo output have significant rounding which can be very problematic. Particularly for `t` in distribution files which require high precision to be meaningful in many cases.

While checking this issue I noticed that even all-numerical CSV files have their values individually enclosed in double quotes, which makes parsing the file much more difficult. It would be ideal to write the CSV files without these double quotes.

moellep:


sirepo/template/elegant.py
```diff
@@ -742,7 +742,14 @@ def _sdds(filename):
                 f"invalid suffix={options.suffix} for download path={path}"
             )
         out = elegant_common.subprocess_output(
-            ["sddsprintout", "-columns", "-spreadsheet=csv", str(path)],
+            [
+                "sddsprintout",
+                "-noTitle",
+                "-columns",
+                "-spreadsheet=delimiter=\\,",
+                "-formatDefaults=float=%1.9e,double=%1.16e,long=%1ld,short=%1hd",
```
robnagler, line 750: I think `float=%1.8e` is better. There are at most 9 significant digits in single precision. I didn't check the short and long. 

I'm a bit concerned with the delimiter. Is there no possibility for quotes or commas in titles?

robnagler (changes requested):

sirepo/template/elegant.py
moellep: You can't put a quote or comma in an element name, and I think that is all that would show up in the sdds output columns. I tested on several of the common elegant files with OSX Numbers and google docs. The default for spreadsheet=csv is to quote every value, which causes problems for users.

robnagler: I tested it and inspected the data. Looked fine to me.

robnagler (approved):
btw what is the 5414-csv-fix branch?

moellep:
5414-csv-fix might have been Ben's branch when he was looking at this issue 2 months ago

PR #5971, opened 2023-06-15
Fix #5943: error message non premium

### LINKED ISSUE #5943: silas didn't display non-premium user time limit
When running a long silas simulation on alpha with a non-premium user, the simulation stopped, but no error message was displayed. This bug may also affect SRW. It should show a message about "upgrade to premium".

gurhar1133:
Added timeout error message for silas/srw


sirepo/package_data/static/js/sirepo-beamline.js
```diff
@@ -843,6 +843,9 @@ SIREPO.app.directive('beamlineAnimation', function(appState, frameCache, persist
         scope: {},
         template: `
           <div class="col-sm-3">
+            <div data-canceled-due-to-timeout-alert="simState"></div>
+            <div data-simulation-stopped-status="simState" data-ng-if="simState.stateAsText() != 'Missing' "></div>
```
e-carlin, line 847: I think we should add "isStateMissing" and check the string in there. I'm also not convinced we need it. If the state is missing then getCanceledAfterSecs should be falsey which would mean the showAlert check in simulationStoppedStatus is false. I haven't tested this code so I could be missing something.

In addition do we need to add a check for isStopped?

sirepo/package_data/static/js/sirepo-beamline.js
```diff
@@ -843,6 +843,9 @@ SIREPO.app.directive('beamlineAnimation', function(appState, frameCache, persist
...
+            <div class="col-sm-12" data-ng-show="simState.hasTimeData()" data-simulation-status-timer="simState"></div>
```
e-carlin, line 848: simulationStatusTimer already checks hasTimeData

e-carlin (changes requested):

sirepo/package_data/static/js/sirepo-beamline.js
gurhar1133: I think that you were right about `simState.stateAsText() != 'Missing'` being redundant. Do you want me to add a isStateMissing to persistentSimulation?

Also, not sure what you mean by adding a check for isStopped? Do you mean adding another state checking function to persistentSimulation?

e-carlin: No need to add isStateMissing since we don't need it.

I was curious if we need to check for isStopped like [we do in elegant](https://github.com/radiasoft/sirepo/blob/1a3676c3aeb588c97f6129520eb7a70d90128f44/sirepo/package_data/static/html/elegant-visualization.html#L25)

e-carlin: Sorry I'm not being clear - I'm wondering if we need to protect the call to `simulation-stopped-status` like we do in other places. Ex:
- guarded with frameCount > 0 https://github.com/radiasoft/sirepo/blob/be9e4bcae98faeb284428421b450eb42450ed10a/sirepo/package_data/static/js/sirepo-components.js#L4389
- guarded with isStopped https://github.com/radiasoft/sirepo/blob/be9e4bcae98faeb284428421b450eb42450ed10a/sirepo/package_data/static/js/srw.js#L2542
- guarded with isStopped https://github.com/radiasoft/sirepo/blob/be9e4bcae98faeb284428421b450eb42450ed10a/sirepo/package_data/static/html/elegant-visualization.html#L23

e-carlin:
In error it shows the error twice which I think looks a bit odd

Screen Shot 2023-06-23 at 2 28 32 PM

gurhar1133:
> In error it shows the error twice which I think looks a bit odd
> 
> Screen Shot 2023-06-23 at 2 28 32 PM

Odd, I'll fix

gurhar1133:
@e-carlin  i think the data-simulation-stopped-status made this
```html
<div data-ng-show="simState.isStateError()" class="col-sm-9" style="margin-top: 1ex">
        Error: {{ simState.getError() }}
</div>
```
redundant

e-carlin (approved):

PR #5973, opened 2023-06-15
fix #5734 fixes for email login on react

### LINKED ISSUE #5734: react email login does not work
Clicking login link from email does not complete email login. Prompts user to log in again

moellep:
- added back removed code from sirepo.uri for reactRoutes overrides
- don't allow login without valid email
- don't allow submitting an empty displayName


sirepo/uri.py
```diff
@@ -58,14 +59,22 @@ def local_route(sim_type, route_name=None, params=None, query=None):
     s = simulation_db.get_schema(sim_type)
     if not route_name:
         route_name = default_local_route_name(s)
-    parts = s.localRoutes[route_name].route.split("/:")
+    n = "localRoutes"
```
robnagler, line 62: It seems localRoutes and reactRoutes are the same except for the fact that `source` is the only one different. I don't see a reason why source couldn't be the same for React. The SPA will have to redirect old URIs anyway so it could do this manually by changing the anchor form to the non-anchor form of the uri. Perhaps `/source/:sid` doesn't work right in React?

I think we only want to check `is_react_sim_type` and ignore whether the route is in reactRoutes. If it is not in reactRoutes, it still needs to be a localRoute. If React doesn't implement that route, it should either error or do something intelligent in the SPA. `uri` should only need to know that the route needs an anchor or not.

The code would look cleaner with a function like this in any event:

```py
    def _uri(route_table):
        parts = s[route_table][route_name].route.split("/:")
        u = parts.pop(0)
	for p in parts:
            if p.endswith("?"):
		p = p[:-1]
		if not params or p not in params:
                    continue
            u += "/" + _to_uri(params[p])
	return u
```

`route_table` might go away if my conjecture above is true.

react/src/component/login/email.tsx
```diff
@@ -22,20 +22,21 @@ export const LoginEmailConfirm = (props) => {
...
-        .then(() => navigate(`/react/${appName}`))
+        .then(() => navigate(`/${appName}`))
         )
     }
+
     return (
         <Container>
             {
-                needCompleteRegistration ? (
+                needCompleteRegistration === '1' ? (
```
robnagler, line 34: This is going to get us in trouble. Is there no utility module for evaluating booleans from the schema?

react/src/component/login/email.tsx
```diff
@@ -22,20 +22,21 @@ export const LoginEmailConfirm = (props) => {
...
                     <LoginExtraInfoForm onComplete={(data) => completeLogin(data)}/>
                 ) : (
                     <>
-                        <p>Click the button complete the login process.</p>
-                        <Button onClick={() => completeLogin()}>Continue</Button>
+                        <p>Please click the button below complete the login process.</p>
```
robnagler, line 38: I have asked numerous times for us to use string tables. This embedded stuff is going to get harder and harder to maintain. The style of "Please click" vs "Click" and "Continue" vs "Confirm" should be auditable, easily, in a string file.

react/src/component/login/email.tsx
```diff
@@ -70,7 +71,7 @@ export const LoginWithEmail = (props) => {
                         <Form.Control placeholder="email@example.com" value={email} onChange={(e) => updateEmail(e.target.value)}/>
                     </Col>
                     <Col sm="2">
-                        <Button variant="primary" onClick={() => doLogin(email)}>Continue</Button>
+                        <Button variant="primary" onClick={() => doLogin(email)} disabled={! (email && email.match(/^.+@.+\..+$/))}>Continue</Button>
```
robnagler, line 74: I would like to see us avoid code that looks like this. It's a rabbit hole of copied conditionals and broken regexes. I would think this logic could be conditional in Javascript, not JSX. There should be a function that returns whether an email address is valid and another function or value that says whether this button is disabled when an email is invalid. Indeed, any form OK button should disabled unless the fields are disabled. And here, the value is "Continue". It should be a standard and in a string table (with abstractions for sharing).

react/src/component/login/login.tsx
```diff
@@ -236,11 +236,13 @@ export const LoginExtraInfoForm = (props: { onComplete: ({displayName}) => void
                 <Form.Label column sm="3">
                     Your full name
                 </Form.Label>
-                <Col sm="9">
+                <Col sm="7">
```
robnagler, line 239: The "7" vs "9" should be encapsulated. This is going to be another rabbit hole of layout management. The math should be done at some higher level that counts columns and Form fields.

react/src/component/login/login.tsx
```diff
@@ -236,11 +236,13 @@ export const LoginExtraInfoForm = (props: { onComplete: ({displayName}) => void
...
                     <Form.Control value={displayName} onChange={(e) => updateDisplayName(e.target.value)}/>
                 </Col>
+                <Col sm="2">
+                    <Button variant="primary" onClick={() => onComplete({ displayName })} disabled={! displayName}>Continue</Button>
```
robnagler, line 243: displayName is valid should be a function. And OK button should only be clickable when all form fields are valid.

robnagler (dismissed):
We need to build the infrastructure around forms so that we can layout automatically based on whether form fields are visible or not, add centralized "OK button" management, and ensure forms are valid before they are submittable. I've written specific comments here. I really would like us to have serious discussion about this. There are tradeoffs, but I see a lot of duplicated concepts, and this is only one form. We also don't have centralized error handling: form fields can be in error and they should have an error shown in/around the form field so the user can associate the error with a specific field.

moellep:
I agree with all the comments. I think all forms should have a schema model underneath which handles the types, validation and labels.

robnagler:
We can merge this, but we need a plan for getting there. I do think the sirepo.uri change needs to be modified before the merge. It shouldn't fall through when there is no reactRoute.

sirepo/uri.py
moellep: I was able to remove reactRoutes from schema-common.json and reverted the change in sirepo.uri. It still needs to use the # urls for angular and normal urls for react.

sirepo/package_data/static/react-json/common-schema.json
```diff
@@ -404,8 +403,7 @@
         "completeRegistration": "/complete-registration",
         "login": "/login",
         "logout": "/logout",
-        "loginConfirm": "/login-confirm/:method/:token/:needCompleteRegistration",
-        "loginWithEmailConfirm": "/login-confirm/email/:token/:needCompleteRegistration",
+        "loginWithEmailConfirm": "/login-with-email-confirm/:token/:needCompleteRegistration",
         "source": "/:simulationId",
```
robnagler, line 407: I think this is going to be broken. This needs to be `/source/:simulationId`.

sirepo/uri.py
robnagler: Great, thanks.

PR #5986, opened 2023-06-19
Fix #5981: energy spectrum from percentage

### LINKED ISSUE #5981: Hellweg Simulation Settings: Energy spectrum
In Hellweg's Visualization-->Simulation Settings panel, line 4 of Results is 'Energy Spectrum (FWHM)'. The reported value, which is, e.g., "0.25 % ", is confusing:  0.25% of what?  My guess is, of the average energy.  It would be good to verify this and modify the output format accordingly.

gurhar1133:
Converts Energy Spectrum FWHM from percentage of mean to actual Energy Spectrum FWHM value

### DIFF
--- a/sirepo/package_data/static/js/hellweg.js
+++ b/sirepo/package_data/static/js/hellweg.js
@@ -381,7 +381,7 @@ SIREPO.app.directive('appHeader', function(appState) {
     };
 });
 
-SIREPO.app.directive('summaryTable', function() {
+SIREPO.app.directive('summaryTable', function(appState) {
     return {
         restrict: 'A',
         scope: {
@@ -404,10 +404,17 @@ SIREPO.app.directive('summaryTable', function() {
             function parseSummaryRows(summaryText) {
                 var text = summaryText.replace(/^(\n|.)*RESULTS\n+==+/, '').replace(/==+/, '');
                 var label = null;
+                var meanEnergy;
                 $scope.summaryRows = [];
                 text.split(/\n+/).forEach(function(line) {
                     line.split(/\s*=\s*/).forEach(function(v) {
                         if (label) {
+                            if (label == 'Average Energy') {
+                                meanEnergy = Number(v.replace(/MeV/, ''));
+                            }
+                            if (label == 'Energy Spectrum (FWHM)') {
+                                v = energySpectrumFromPercentage(v, meanEnergy);
+                            }
                             $scope.summaryRows.push([label, v]);
                             label = null;
                         }
@@ -419,6 +426,15 @@ SIREPO.app.directive('summaryTable', function() {
                         }
                     });
                 });
+            }
+
+            function energySpectrumFromPercentage(value, meanEnergy) {
+                return String(
+                    appState.formatFloat(
+                        ((Number(value.replace(/ %/, '')) * meanEnergy) / 100) * 1000,
+                        6,
+                    )
+                ) + "KeV";
             }
 
             function updateSummaryInfo(e, summaryText) {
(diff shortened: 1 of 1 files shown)


sirepo/package_data/static/js/hellweg.js
```diff
@@ -404,10 +404,17 @@ SIREPO.app.directive('summaryTable', function() {
             function parseSummaryRows(summaryText) {
                 var text = summaryText.replace(/^(\n|.)*RESULTS\n+==+/, '').replace(/==+/, '');
                 var label = null;
+                var meanEnergy;
```
mkeilman, line 407: Use `let meanEnergy`. In general, when we need to fiddle with older code we should update it with our current standards (within reason). So I would also use `const text...` and `let label...`

sirepo/package_data/static/js/hellweg.js
```diff
@@ -404,10 +404,17 @@ SIREPO.app.directive('summaryTable', function() {
...
                 $scope.summaryRows = [];
                 text.split(/\n+/).forEach(function(line) {
                     line.split(/\s*=\s*/).forEach(function(v) {
```
mkeilman, line 410: In the spirit of the previous comment, `.forEach(line => {...})` etc.

sirepo/package_data/static/js/hellweg.js
```diff
@@ -404,10 +404,17 @@ SIREPO.app.directive('summaryTable', function() {
...
                         if (label) {
+                            if (label == 'Average Energy') {
```
mkeilman, line 412: Use triple equals `label === 'Average Energy'`

sirepo/package_data/static/js/hellweg.js
```diff
@@ -404,10 +404,17 @@ SIREPO.app.directive('summaryTable', function() {
...
+                                meanEnergy = Number(v.replace(/MeV/, ''));
+                            }
+                            if (label == 'Energy Spectrum (FWHM)') {
+                                v = energySpectrumFromPercentage(v, meanEnergy);
+                            }
```
mkeilman, line 417: Is 'Average Energy' guaranteed to come before 'Energy Spectrum'? If not `meanEnergy` may not be defined.

sirepo/package_data/static/js/hellweg.js
```diff
@@ -421,6 +428,15 @@ SIREPO.app.directive('summaryTable', function() {
                 });
             }
 
+            function energySpectrumFromPercentage(value, meanEnergy) {
+                return String(
+                    appState.formatFloat(
+                        ((Number(value.replace(/ %/, '')) * meanEnergy) / 100) * 1000,
```
mkeilman, line 434: I think this can be more permissive in the regex: `value.replace(/\s%/, '')` in case the amount of whitespace varies

sirepo/package_data/static/js/hellweg.js
```diff
@@ -421,6 +428,15 @@ SIREPO.app.directive('summaryTable', function() {
...
+                        6,
+                    )
+                ) + "KeV";
```
mkeilman, line 437: Single quotes

mkeilman (commented):
A few style comments

sirepo/package_data/static/js/hellweg.js
gurhar1133: Good point. I'll write it to get meanEnergy and set Energy Spectrum after that first loop

mkeilman (commented):
Tiny nitpick

sirepo/package_data/static/js/hellweg.js
```diff
@@ -419,6 +419,32 @@ SIREPO.app.directive('summaryTable', function() {
                         }
                     });
                 });
+                const meanEnergy = getMeanEnergy($scope.summaryRows);
```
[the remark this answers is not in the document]
gurhar1133: Good point

mkeilman (approved):
lgtm

moellep:
I think the correct fix for this would be in the hellweg source code, rather than working-around the confusing terms in the javascript.

PR #6015, opened 2023-06-23
Fix #5948 n0 n2 from fenics

### LINKED ISSUE #5948: new crystal tab Propagation Calculation
Add a new crystal tab "Propagation Calculation". The tab should show a field for "Calculation Type" (Calculated by Fenics, Analytical). Below the field should be a "calculate" button which will run the computation on the server (it may take awhile), and as a result the n0 and n2 fields should be shown read-only below.

gurhar1133:


gurhar1133:
this will mean a slight update in rslaser to [test_sirepo_compatability](https://github.com/radiasoft/rslaser/blob/main/tests/pulse_test.py) will be needed

PR #6027, opened 2023-06-27
Fix #6017 cloudmc 3d toolbar

### LINKED ISSUE #6017: improve cloudmc 3D toolbar
I think there could be several changes to improve the toolbar in the cloudmc app:
x

I think the first two buttons (arrow and pointer) could be removed, and change the text to "Click and drag to rotate. Double-click to reset camera. Control-click an object to select". When it is in the "arrow" mode, it seems to support all three, so just use that mode by default.

Replace the -90, -45 and 45, 90 buttons with two rotate buttons (forward and reverse), and rotate 45 degrees when clicked. Possibly change the "Reset" button to a home icon.

Move the perspective button inline with the other toolbar buttons.

Move the "Toggle Marker" checkbox to the panel's editor dialog.

When clicking on a dimension button (x, y, z), the buttons shift slightly due to the difference in the text size. Set a min-width on the button to prevent this.

mkeilman:
This addresses the comments in the issue.  I opted for a "circled dot" icon rather than "home" for the reset button - I think it conveys the function a little better, but I'm flexible.


moellep:
This looks great.  
x

I think the rotate is backwards - the "spin-left" icon should be on the left and rotate -45 degrees, and the "spin-right" should be on the right and rotate 45 degrees.

Rotate -45 degrees should rotate 45 degrees counterclockwise, and Rotate 45 degrees should be clockwise.
x2

mkeilman:
> I think the rotate is backwards - the "spin-left" icon should be on the left and rotate -45 degrees, and the "spin-right" should be on the right and rotate 45 degrees.
> 
> Rotate -45 degrees should rotate 45 degrees counterclockwise, and Rotate 45 degrees should be clockwise.

In physics, positive rotations are typically right-handed: "thumbs up" with the right hand, thumb pointing along the axis, and the fingers curl in the direction of positive rotation. Thus counter-clockwise is positive. Engineers may feel differently. How about we leave it and get feedback from the users?

robnagler:
I think we should have it be the "natural" direction so we can switch on users every few years. ;-)

Electrical engineers think about the [right hand rule](https://en.wikipedia.org/wiki/Right-hand_rule), which oddly, I remembered. 

Software engineers probably think what looks right to them.

What does a typical CAD program do?

moellep:
ParaView uses these icons to orient, and the rotation is clockwise for +90, counterclockwise for -90. They do show rotate-right as the first icon.

x

PR #6035, opened 2023-06-28
Fix #6032 - try small instead of subscript

### LINKED ISSUE #6032: Increase readability of cylindrical segmentation fields
The "az" (azimuthal) and "ax" (axial) subscripts are a bit hard to read due to their font and small size

mkeilman:
This seems better to me. We can also use sans-serif for the subscript if the italicized font is still a problem


moellep:
Yes, that looks more readable.

PR #6048, opened 2023-06-30
fix #5829 implement parallel geometry extraction

### LINKED ISSUE #5829: parallelize volume extraction to vtk
In cloudmc, the volume extraction can take a long time creating the individual volume vtk files. This could be done in parallel.

For example, the ZPRE example currently takes about 5 minutes to package up 16 volumes.

moellep:
- avoid file writes and reads to get points and triangles, use pymoab to get info directly
- assemble vti files via multiprocessing
- remove pkcli.cloudmc.run_background


sirepo/template/cloudmc.py
```diff
@@ -198,6 +198,20 @@ def write_parameters(data, run_dir, is_parallel):
         run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
         _generate_parameters_file(data, run_dir=run_dir),
     )
+    if is_parallel:
+        return template_common.get_exec_parameters_cmd()
+
+
+def _dagmc_animation_python(filename):
+    return f"""
+from sirepo.pkcli.cloudmc import extract_dagmc
```
robnagler, line 207: I don't see the need for the unqualified import.

sirepo/template/cloudmc.py
```diff
@@ -198,6 +198,20 @@ def write_parameters(data, run_dir, is_parallel):
...
+import sirepo.simulation_db
+
+sirepo.simulation_db.write_json(
+    '{_VOLUME_INFO_FILE}',
```
robnagler, line 211: These can be double quotes inside a triple quote.

sirepo/pkcli/cloudmc.py
```diff
@@ -7,41 +7,28 @@
 from pykern import pkio
 from pykern.pkcollections import PKDict
 from pykern.pkdebug import pkdp, pkdlog
+from pymoab.rng import Range
```
robnagler, line 10: I do not see the value of an unqualified import here or the next line. I had to search around for this unnecessarily, thinking Range might be a local class.

sirepo/pkcli/cloudmc.py
```diff
@@ -105,155 +97,99 @@ def _array_from_list(arr, arr_type):
...
+    mb.load_file(dagmc_filename)
+    _write_vti(vol_id, _get_points_and_polys(mb, volumes))
+
+
+def _get_tag_value(mb, tag, handle):
+    return str(mb.tag_get_data(tag, handle).flat[0])
+
+
+def _get_points_and_polys(mb, volumes):
+    verticies = Range()
```
robnagler, line 111: A class would be good here, too.

robnagler (changes requested):
I have only one "important" change: encapsulate multiprocessing.Pool call in sirepo.mpi. In general, I'm concerned about the coding style of lots of disconnected functions in one file. Better to form classes that groups them, since almost all are single use. This helps the reader understand what's going on, I think.

sirepo/pkcli/cloudmc.py
```diff
@@ -74,29 +61,34 @@
...
+            mb,
+            mb.tag_get_handle(pymoab.types.GLOBAL_ID_TAG_NAME),
+            volumes[0],
+        )
+        res[name] = PKDict(
+            name=name,
+            volId=vol_id,
+        )
+        work_items.append((dagmc_filename, vol_id, volumes))
+    work_items.sort(key=lambda v: -len(v[2]))
```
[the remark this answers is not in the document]
moellep: The chunk size should have been left off - the default is size 1, I think. I moved the mutiprocessing logic into sirepo.mpi, and it could be replaced with threading if that makes more sense.

moellep:
@robnagler please take another look at this PR - I've moved the multiprocessing code to sirepo.mpi, and created 2 classes for grouping and extraction.

robnagler:
Thanks for the changes. I didn't see a test for this so I just committed some changes rather than writing comments (which are in the commit msg).

sirepo/mpi.py
```diff
@@ -69,6 +73,28 @@ def restrict_op_to_first_rank(op):
     return res
 
 
+def restrict_ops_to_first_node(ops):
```
e-carlin, line 76: I think a different name would be helpful here. Specifically, "first_node". I don't know what that means. First in terms of what? Then the docstring says "current node". Are those the same thing? How does "first node", "current node", and "first rank" all relate to each other?

moellep: I think we can take this discussion to a new issue.

PR #6051, opened 2023-06-30
Fix #6046: "completed reports" instead of "animation"

### LINKED ISSUE #6046: Improve text for beamline simulation progress
The beamline simulation progress shows "X animation frames". When it should show "X completed reports", or something similar. "Animation frames" isn't the right word to use in this case.

x

gurhar1133:


sirepo/package_data/static/js/silas.js
```diff
@@ -106,6 +106,10 @@ SIREPO.app.controller('SourceController', function (appState, frameCache, persis
              ? 'parameter' : '3d';
     };
 
+    self.simCompletionState = (completionState) => {
+        return completionState.replace("animation frames", "completed reports");
```
e-carlin, line 110: It's been a while since I've used this code but I think you want to override the "completionState" field in the silas schema. Here is what you will be overriding from schema-common.json https://github.com/radiasoft/sirepo/blob/d15784a34527b04ae868836d42d8fee45ddfbfda/sirepo/package_data/static/json/schema-common.json#L497

e-carlin (changes requested):

sirepo/package_data/static/js/silas.js
e-carlin: Here is an example from activait https://github.com/radiasoft/sirepo/blob/d15784a34527b04ae868836d42d8fee45ddfbfda/sirepo/package_data/static/json/activait-schema.json#L461

gurhar1133: nice, thanks. Ill change it

Issue #6059, opened 2023-07-03
raydata: start and stop query easy selection


e-carlin:
Give users the option to set the start and stop search fields to some pre-defined times. We can start with last hour and last day

rorour:
Do you have any thoughts on what the UI will look like? I could see this as a few buttons ("Last Hour, Last Day") underneath the start/stop fields. Or possibly a drop down menu with those options.

e-carlin:
I don't have ideas. But I think the few buttons below the start stop fields sounds fine. Coming up with layouts such as this is not my strong suit. Paul may have ideas.

PR #6082, opened 2023-07-06
fix #6058: raydata start and stop search times pre-populated

### LINKED ISSUE #6058: raydata: start and stop fields pre-populated
Users suggested when they first enter the app that the start and stop fields are pre-populated. Currently they are empty. Maybe set them to search in the past hour?

rorour:


sirepo/package_data/static/js/raydata.js
```diff
@@ -624,6 +630,20 @@ SIREPO.app.directive('scansTable', function() {
...
             };
 
+            $scope.setDefaultStartStopTime = () => {
+                const m = appState.models[$scope.modelName];
+                if (!m.searchStartTime && !m.searchStopTime) {
+                    $scope.setSearchTimeLastHour();
+                };
+            };
+
+            $scope.setSearchTimeLastHour = () => {
```
rorour, line 640: I separated this out because I anticipate using it for https://github.com/radiasoft/sirepo/issues/6059. It would probably be a good idea to get this merged quickly so I can start that issue with this change

e-carlin (changes requested):
When I open the app the "save changes / cancel " buttons are now present. Even if I haven't modified the time. I don't think they should appear.

Screen Shot 2023-07-06 at 5 19 53 PM

In addition, if I set one of the times, go back to the simulation list, and then back to the simulation only one of the times (the time I set in the first step) is set.

Screen Shot 2023-07-06 at 5 22 28 PM

rorour:
@e-carlin ready for re-review

e-carlin (approved):
LGTM

PR #6088, opened 2023-07-07
Fix #5911: sirepo.resource.render_resource()

### LINKED ISSUE #5911: sirepo.resource needs a render_file
sirepo.resource doesn't have a render_file but it should

gurhar1133:
also updated sirepo.pkcli.nersc_test to use this


tests/resource_test.py
```diff
@@ -17,3 +18,22 @@ def test_static_files():
...
+
+    d = pkunit.data_dir()
+    pkunit.file_eq(
+        d.join("expect.py"),
+        actual_path=resource.render_resource(
+            "actual.py",
+            d,
+            pkunit.work_dir(),
+            PKDict(
+                x="x",
```
robnagler, line 35: Use something that's not the same thing as the key, e.g. x=1. Also, there really only needs to be one value to make sure the parameter gets passed. The fact that it has two values complicates the test without adding value.

tests/resource_test.py
```diff
@@ -17,3 +18,22 @@ def test_static_files():
...
     )
+
+
+def test_render_resource():
+    from sirepo import resource
+    from pykern import pkunit
+
+    d = pkunit.data_dir()
+    pkunit.file_eq(
+        d.join("expect.py"),
```
robnagler, line 29: I think the fact that these are Python files confuses things. Having them be text files is better, because there's no implied semantics for someone reading this in the "tests as documentation" mode.

tests/resource_test.py
```diff
@@ -17,3 +18,22 @@ def test_static_files():
...
+        actual_path=resource.render_resource(
+            "actual.py",
+            d,
```
robnagler, line 32: This parameter is actually a bug. It shouldn't work this way. Resources should always come from the package_data directory. In this case, you'll need to construct something so that it is found with package_path.

To make this simpler, you could pick an arbitrary sirepo resource like test_static_files does. One way to do this is to create something like `sirepo/package_data/resource_test_data` that is small and then you don't have to setup package_path (which is unnecessarily complicated). This costs a little bit in complexity of package_data having test data, but it is a more realistic test. The file in resource_test_data could even be a README.txt.jinja that includes some documentation (and a single jinja expansion) that tells you why this particular file is there.

You don't necessarily need to have a `tests/resource_data` directory for this test. Just test some text that you know should be there along with the jinja value expansion, e.g. `pkre("var=abc")`. The file_eq tends to be a bit heavy for these types of problems, because there's too much copying and not enough context within the python part of the test.

Try this suggestion without thinking too hard about it. If it doesn't work, we can talk about it.

robnagler (changes requested):

sirepo/resource.py
```diff
@@ -44,6 +45,27 @@ def glob_paths(*paths):
     )
 
 
+def render_resource(filename, resource_dir, run_dir, jinja_params):
```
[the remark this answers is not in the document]
gurhar1133: The name `render_resource` is redundant?

gurhar1133:
@robnagler I think I addressed most of what you requested to be changed. If more, maybe this would be a good one for a software team huddle?

robnagler:
I'd like you to re-read the comments and try to come up with more changes.

gurhar1133:
ping @robnagler I think this conforms better to the resource.py module and is tailored to the service not the use.

sirepo/resource.py
```diff
@@ -44,6 +45,27 @@ def glob_paths(*paths):
     )
 
 
+def render(*paths, target_dir=None, j2_ctx=None):
```
robnagler, line 48: `render_jinja`

sirepo/resource.py
```diff
@@ -44,6 +45,27 @@ def glob_paths(*paths):
...
+    """Render a jinja resource into target_dir
```
robnagler, line 49: Render a resource template file with Jinja into target_dir.

sirepo/resource.py
```diff
@@ -44,6 +45,27 @@ def glob_paths(*paths):
...
+
+    Args:
+        paths (str): Path components of resource file without pykern.pkjinja.RESOURCE_SUFFIX
+        target_dir (py.path): target directory for rendered file
+        j2_ctx (PKDict): parameters to jinja file
+
+    Returns:
+        py.path: path to rendered file
```
robnagler, line 57: output path which is ``target_dir.join(paths[-1])``.

tests/resource_test.py
```diff
@@ -5,6 +5,7 @@
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
 import pytest
+from pykern.pkcollections import PKDict
```
robnagler, line 8: Move this import inside test_render_resource. We avoid all imports in py.test to ensure we can control module initialization order.

sirepo/resource.py
```diff
@@ -44,6 +45,27 @@ def glob_paths(*paths):
...
+    """
+    f = paths[-1]
+    t = target_dir.join(f)
```
robnagler, line 60: Change `t`  to `res`, since we generally use that as a convention for the return value.

robnagler (changes requested):
Looks good. Minor comments on comments.

robnagler (approved):

PR #6096, opened 2023-07-11
fix 6095 increase server_test timeout

moellep:


tests/server_test.py
```diff
@@ -36,7 +36,8 @@ def test_elegant_data_file(fc):
         run = fc.sr_post("runStatus", run.nextRequest)
         time.sleep(1)
     else:
-        pkunit.pkfail("runStatus: failed to complete: {}", run)
+        if run.state != "completed":
```
robnagler, line 39: Increase the timeout, don't do this. It doesn't make sense.

PR #6105, opened 2023-07-12
fix #6059: raydata start and stop query easy selection

### LINKED ISSUE #6059: raydata: start and stop query easy selection

rorour:


e-carlin (approved):
Perfect

PR #6108, opened 2023-07-13
Fix #6011: thermal transport equations and other thermal transport tab improvements

### LINKED ISSUE #6011: Thermal Transport Display clarifications
- [x] Include the repetition rate and energy of the pump laser, as well as whether the crystal is pumped from left or right or both
- [x] Parameters below the equation need descriptions, update notations to:
    - T_0 notation is fine
    - δT notation is fine
    - ξ should be “w_p”
    - Add z_0 (half-length of crystal [cm])
    - Add alpha (inverse absorption length [1/cm])
- [x] Equation display clean up (current equation isn’t correct):
When selecting a different pump pulse profile (gaussian, higher order gaussian, top hat), the display should update what equation is shown
    - "gaussian": T = T_0 + dT * exp(-2*(x^2+y^2)/(wp^2)) * exp(-alpha * (z - z0))
    - "higher order gaussian" : T = T_0 + dT * exp(-((x^2+y^2)/(2*wp^2))^P) * exp(-alpha * (z - z0))
    - "top hat": If x^2+y^2 <= w_p, then T = T_0 + dT * alpha * exp(-alpha * (z - z0))/(1 - exp(2 * alpha * z0)); If x^2+y^2 > w_p, then T = T_0

gurhar1133:


moellep:
I made a few changes:
- moved the constant thermal equations out of the model def and into the "strings" section of the schema
- hide values which aren't used in the selected equation

PR #6120, opened 2023-07-18
Fix #6023 srw watchpoint panels

### LINKED ISSUE #6023: Allow access to frame panels before data is available
In SRW, the beamline watchpoint panels become available when they have data available. This means you can't change the "max plot dimensions" of the data until it is first shown at full resolution, which can take a long time. We should consider a way to make the panels visible, but without a graph while the simulation is running.

mkeilman:
Notes:
- Now building the array of reports as they are provided by the template, not necessarily when the data is available
- We set a flag, falsey by default, to wait for the data via SIREPO.nonDataFileFrame

For srw, once the initial intensity report is done, the watchpoints appear with empty panels. We might be able to add a loading message in there.

silas does not provide watchpoint reports until the data is ready, so the panels still do not show up. If we want the same empty but visible reports for silas we'll have to tweak that.


moellep:
This looks good. I think we should add a "Waiting for simulation results" or similar message to the panels which are shown, but have no data. It is OK that silas does not have the feature because it doesn't deal with rescaling or other slower operations to get the plot data.

Also, the blank panels should be shown only when the simulation is running. If you cancel a simulation midway, it should hide the panels which have no data. Otherwise it could be confusing if you view a beamline tab for a canceled simulation.

mkeilman:
@moellep see how it looks now

moellep:
Looks good - for me the "Waiting for Data" is white text on a white background. Removing the "color: white" from the ".sr-panel-waiting .sr-panel-wait" entry shows it like this, but I'm not sure if that was what you wanted. I use the "NSLS-II CSX-1 beamline" for testing because it has many watchpoints.

x

mkeilman:
> Looks good - for me the "Waiting for Data" is white text on a white background. Removing the "color: white" from the ".sr-panel-waiting .sr-panel-wait" entry shows it like this, but I'm not sure if that was what you wanted. I use the "NSLS-II CSX-1 beamline" for testing because it has many watchpoints.

I could have sworn it was white on grey like "requesting data" - I'll get that in place

PR #6128, opened 2023-07-18
Fix #4826 check_origin: default to Tornado behavior

### LINKED ISSUE #4826: Remove job_supervisor._AgentMsg.check_origin
[check_origin](https://github.com/radiasoft/sirepo/blob/b2acdebf16b727913374c4711da91dae3ff41682/sirepo/pkcli/job_supervisor.py#L84) always returns true. It seems like this is not necessary, since the Origin header is not necessarily sent? Has this been tested?

robnagler:


robnagler:
@gurhar1133 ping

gurhar1133 (approved):

PR #6130, opened 2023-07-19
fix #6073: purge_non_premium needs to ignore UserDirNotFound

### LINKED ISSUE #6073: purge_non_premium needs to ignore UserDirNotFound
It should remove the job runs for users that don't exist.

rorour:


sirepo/job_supervisor.py
```diff
@@ -176,7 +176,7 @@ def init_module(**imports):
         purge_non_premium_task_secs=(
             None,
             pkconfig.parse_seconds,
-            "when to clean up simulation runs of non-premium users (%H:%M:%S)",
+            "time interval to clean up simulation runs of non-premium users (%H:%M:%S)",
```
rorour, line 179: The names of these two variables (`purge_non_premium_after_secs` & `purge_non_premium_task_secs`) are vague. I think this would be an improvement:
`purge_non_premium_after_secs` -> `purge_non_premium_task_lifetime`
`purge_non_premium_task_secs` -> `purge_non_premium_task_interval`

robnagler: https://github.com/radiasoft/sirepo/issues/6133

robnagler (approved):

PR #6138, opened 2023-07-20
Fix #6113 exports to links

### LINKED ISSUE #6113: Export dump, pythonSource, etc. should be links
It opens a new window, but I think it can be just like download data file and be an data-ng-href

mkeilman:
Notes:
- This replaces `data-ng-click` with `data-ng-href` wherever an export operation exists (that I could find)
- Accordingly, `exportArchive()` is now `exportArchiveUrl()` (etc.)
- Everything gets shunted to `urlForExport()` in panelState


mkeilman:
One comment: exportJupyterNotebook was only used by webcon which is now defunct. Should we remove it here?

moellep:
I think we can have a separate issue to remove unused methods like exportJupyterNotebook and exportJupyterNotebook2, and others.

PR #6145, opened 2023-07-20
fix #6066: raydata better status icons

### LINKED ISSUE #6066: raydata: better status icons
The status icons are hard to differentiate. Most importantly, between the "info" icon and the "error" icon. We should also consider adding colors to the icons (ex red for error).

Which icons and colors we use to denote status should be generic concepts in Sirepo. Work with @mkeilman to come up with status/colors and add them to the design language.

rorour:


e-carlin (approved):
Looks great.

PR #6149, opened 2023-07-20
fix #6133: rename job_supervisor.cfg.purge_non_premium_*

### LINKED ISSUE #6133: Rename job_supervisor.cfg.purge_non_premium_*
[@rorour writes](https://github.com/radiasoft/sirepo/pull/6130/files#r1267417513):
> The names of these two variables (`purge_non_premium_after_secs` & `purge_non_premium_task_secs`) are vague. I think this would be an improvement:
> `purge_non_premium_after_secs` -> `purge_non_premium_task_lifetime`
> `purge_non_premium_task_secs` -> `purge_non_premium_task_interval`

I agree the names are confusing. How about: 
- purge_non_premium_after_secs -> purge_non_premium_run_dir_lifetime
- purge_non_premium_task_secs -> purge_non_premium_check_interval

I really don't like purge_none_premium, because it dominates. `purge_check_interval` is sufficient. How about run_dir_lifetime and leave the purge_non_premium implicit?

rorour:


sirepo/job_supervisor.py
```diff
@@ -168,16 +168,16 @@ def init_module(**imports):
                 "maximum run-time for sequential job",
             ),
         ),
-        purge_non_premium_after_secs=(
-            0,
-            pkconfig.parse_seconds,
-            "expiration period for purging non-premium users simulations",
-        ),
-        purge_non_premium_task_secs=(
+        purge_check_interval=(
```
robnagler, line 171: Please add [ReplacedBy](https://github.com/radiasoft/sirepo/blob/dae139c45ebe9d820c5d8e2e2558d7e0da731353/sirepo/mpi.py#L149) for the old name.

sirepo/job_supervisor.py
```diff
@@ -168,16 +168,16 @@ def init_module(**imports):
...
             None,
             pkconfig.parse_seconds,
             "time interval to clean up simulation runs of non-premium users (%H:%M:%S)",
```
robnagler, line 174: `(%H:%M:%S)` is implicit from `pkconfig.parse_seconds`. What we need to do is have `pkconfig` output a help file from the descriptions, names, defaults, and types (parse_seconds).

I would add that `0` means no checks are performed.

sirepo/job_supervisor.py
```diff
@@ -168,16 +168,16 @@ def init_module(**imports):
...
         ),
+        run_dir_lifetime=(
+            0,
+            pkconfig.parse_seconds,
+            "expiration period for purging non-premium users simulations",
```
robnagler, line 179: "simulations" => "simulation run output". We overuse "simulation".

sirepo/job_supervisor.py
```diff
@@ -168,16 +168,16 @@ def init_module(**imports):
...
-            "expiration period for purging non-premium users simulations",
-        ),
-        purge_non_premium_task_secs=(
+        purge_check_interval=(
             None,
             pkconfig.parse_seconds,
             "time interval to clean up simulation runs of non-premium users (%H:%M:%S)",
         ),
+        run_dir_lifetime=(
+            0,
```
robnagler, line 177: I think 0 is a poor choice, because this expires immediately. This could yield unexpected results. I think the minimum should be twice the simulation run period. We don't need to assert this, but maybe put it at 1d here.

robnagler (changes requested):

sirepo/job_supervisor.py
rorour: 796209d1cbcd5b1c7da7d74982d9aeeb0a88e993

sirepo/job_supervisor.py
rorour: 796209d1cbcd5b1c7da7d74982d9aeeb0a88e993

sirepo/job_supervisor.py
rorour: 796209d1cbcd5b1c7da7d74982d9aeeb0a88e993

e-carlin (approved):

PR #6154, opened 2023-07-21
fix #6153 add genesis time dependence simulations

### LINKED ISSUE #6153: time-depenence genesis simulations
Add the time-dependence fields and the ITDP flag to enable that mode.

moellep:


sirepo/template/genesis.py
```diff
@@ -232,6 +244,16 @@ def _get_col(col_key):
...
 
 
+def validate_file(file_type, path):
+    if file_type == "io-partfile":
+        if _is_text_file(path):
+            return "The PARTFILE should be a binary file. Use the DISTFILE to import a text file."
+    elif file_type == "io-distfile":
+        if not _is_text_file(path):
+            return "The DISTFILE should be a text file with columns: X PX Y PY T P."
+    return None
```
robnagler, line 254: Do we assume that if it doesn't match a known type, it's correct?

sirepo/template/genesis.py
```diff
@@ -327,6 +358,15 @@ def _get_frame_counts(run_dir):
     return res
 
 
+def _is_text_file(path):
+    try:
+        pkio.read_text(path)
```
robnagler, line 363: If the files are large, this is expensive (reads the whole file). If the file is not large, probably fine.

we may want to consider binaryornot, which is based on Perl's algorithm.

sirepo/template/genesis.py
```diff
@@ -327,6 +358,15 @@ def _get_frame_counts(run_dir):
...
+        return True
+    except UnicodeDecodeError:
+        pass
```
robnagler, line 366: `return False` here so less code.

sirepo/template/genesis.py
```diff
@@ -367,14 +407,13 @@ def _parse_namelist(data, text):
...
-    # TODO(pjm): remove this if scanning or time dependence is implemented in the UI
+    # TODO(pjm): remove this if scanning is implemented in the UI
     dm.scan.iscan = "0"
-    dm.timeDependence.itdp = "0"
     return data
 
 
 def _z_title_at_frame(frame_args, nth):
-    _, s = _get_lattice_and_slice_data(frame_args.run_dir)
+    _, s = _get_lattice_and_slice_data(frame_args.run_dir, 0)
```
robnagler, line 416: I tend to prefer: 

```py
s = _get_lattice_and_slice_data(frame_args.run_dir, 0)[1]
```

robnagler (changes requested):
Looks fine to me, just some style comments.

sirepo/template/genesis.py
moellep: That's correct - I added input file checking for 2 common cases, which caused confusion in the omega webinar prep.

sirepo/template/genesis.py
moellep: Yes, I looked at binaryornot but we do not currently have it as part of the base install, and genesis input files are not overly large.

PR #6160, opened 2023-07-25
Fix #6021: Return content too large

### LINKED ISSUE #6021: get_data_file() should return a file-to-large error rather than not-found
The simulation below creates very large output files which exceed the sirepo limit. When you try the "Raw Data File" option on the 1st watchpoint report, it results in a 404 "Not Found", but instead should respond with a 413 "Content Too Large" or similar error.

https://www.sirepo.com/srw#/beamline/D093hGau

e-carlin:
When requesting a data file the op (OP_IO) may be cancelled because it exceeded max_run_secs. The most likely reason is because the file requested is too large. When this happens we now provide a better message to the user than the previous 404.


Benbenbenin0 (changes requested):
Still redirecting to '404 Not Found' for me, didn't raise `sirepo.util.DownloadDataFileTimeout` because `r.state = completed`.

I think the bug might have to do with the href created for the download being wrong, I'll investigate but let's discuss Monday.

sirepo/job_api.py
```diff
@@ -100,6 +100,12 @@ async def api_downloadDataFile(
                     req_data=req.req_data,
                     suffix=s,
                 )
+                if r.state == "canceled":
+                    # POSIT: Users can't cancel donwloadDataFile. So canceled means there was a
+                    # timeout (max_run_secs exceeded).
+                    raise sirepo.util.DownloadDataFileTimeout(
+                        f"sim_type={req.type} sid={req.id} report={req.req_data.report}"
```
robnagler, line 107: Do we want to provide this information to the user or is this logging? If it is logging, don't use an fstring.

sirepo/job_api.py
```diff
@@ -100,6 +100,12 @@ async def api_downloadDataFile(
                     req_data=req.req_data,
                     suffix=s,
                 )
+                if r.state == "canceled":
```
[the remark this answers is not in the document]
e-carlin: > Should we add a "cancel reason"

I think that might be useful. But I want to sort out https://github.com/radiasoft/sirepo/issues/6182 before coming up with an approach. 

> Also, _Supervisor.op_run_timeout seems to not be necessary.

I believe it is. _receive_api_beginSession (on _Supervisor) creates a sequential op. All sequential ops have the ability to time out and _Supervisor.op_run_timeout will be called in this case.

sirepo/job_api.py
e-carlin: 5083fc773

robnagler:
We had a discussion about this, and @e-carlin will be making changes.

e-carlin:
I made the changes. Please review.

sirepo/job_api.py
```diff
@@ -100,7 +108,14 @@ async def api_downloadDataFile(
...
                     suffix=s,
                 )
-                assert not r.state == "error", f"error state in request=={r}"
+                if r.state == sirepo.job.CANCELED:
+                    # POSIT: Users can't cancel donwloadDataFile. So canceled means there was a
+                    # timeout (max_run_secs exceeded).
+                    _raise_content_too_large(req)
+                if r.state == sirepo.job.ERROR:
+                    if r.get("errorCode") == sirepo.job.ERROR_CODE_RESPONSE_TOO_LARGE:
+                        _raise_content_too_large(req)
```
robnagler, line 117: I think `raise _content_too_large(req)` is probably better, because the stack trace will be more useful. It was a mistake to have the `raise_` functions in sirepo.util.

sirepo/job_api.py
```diff
@@ -100,7 +108,14 @@ async def api_downloadDataFile(
                     req_data=req.req_data,
                     suffix=s,
                 )
-                assert not r.state == "error", f"error state in request=={r}"
+                if r.state == sirepo.job.CANCELED:
+                    # POSIT: Users can't cancel donwloadDataFile. So canceled means there was a
+                    # timeout (max_run_secs exceeded).
+                    _raise_content_too_large(req)
+                if r.state == sirepo.job.ERROR:
+                    if r.get("errorCode") == sirepo.job.ERROR_CODE_RESPONSE_TOO_LARGE:
```
robnagler, line 116: This is fine, but it is also problematic, because every place error is checked, we have to check errorCode. Ideally, we would centralize this type of error handling. For now, it's "fine".

sirepo/pkcli/job_cmd.py
```diff
@@ -161,6 +166,9 @@ def _do_analysis_job(msg, template):
 
 
 def _do_download_data_file(msg, template):
+    def _error_file_too_large():
```
robnagler, line 169: Consistent name: `_response_too_large`

sirepo/pkcli/job_cmd.py
```diff
@@ -176,15 +184,16 @@ def _do_download_data_file(msg, template):
         if u is None:
             u = r.filename.basename
         c = r.get("content")
+        if _is_over_response_size_limit(
+            len(c) if c else os.path.getsize(str(r.filename))
```
robnagler, line 188: `len(c) if c else r.filename.size()`

sirepo/pkcli/job_cmd.py
```diff
@@ -299,6 +308,12 @@ def _do_stateless_compute(msg, template):
     return _dispatch_compute(msg, template)
 
 
+def _error_if_response_too_large(payload):
+    if _is_over_response_size_limit(len(payload)):
+        return PKDict(state=job.COMPLETED, error="Response is too large to send")
```
robnagler, line 313: It's unclear that "completed" is useful here. Rather, return "error" and "errorCode" exactly as in the download (put) case. They could share code. I think the _dispatch_compute case might want to be able to get a file back just like download_data_file, but that's too much change right now.

robnagler (changes requested):

sirepo/pkcli/job_cmd.py
e-carlin: 11aa3733a

sirepo/pkcli/job_cmd.py
e-carlin: 11aa3733a

e-carlin:
@robnagler ping

sirepo/pkcli/job_cmd.py
```diff
@@ -91,8 +91,7 @@ def _op(expect_file):
                 state=job.ERROR,
                 error=f"method={template.SIM_TYPE}.{msg.jobCmd}_{msg.data.method} unexpected return=JobCmdFile uri={r.uri}",
             )
-        e = _validate_msg(r.content)
-        if e:
+        if e := _error_if_response_too_large(r.content):
```
robnagler, line 94: Something new to me.

robnagler (approved):
I'll let you merge. Not sure what else is required.

PR #6168, opened 2023-07-26
fix #6144 and fix #6165 add telescope and splitter optical elements

### LINKED ISSUE #6144: Telescope_lct features
- [x] Support the features in the notebook (lens.Telescope_lct for example). Create an sirepo example for this.
    https://github.com/radiasoft/rslaser-examples/blob/main/laser_propagation/amplifier_100TW_Bella.ipynb
- [x] Same for beamsplitter

### LINKED ISSUE #6165: silas mirror icon and descriptive text
The issue is that we are implementing two mirrors at an angle to each other, which flips the laser fields
But the image on the GUI shows a single mirror, which at best represents a direct reflection of the fields without flipping them.

Show 2 mirrors in the icon, something like this:

mirror

A description should be added to the mirror editor with wording like "The pulse is reflected off of two mirrors, resulting in a flip of the electric fields. Since 1D amplifier geometry is assumed, the mirrors' small, off-axis angles are ignored."

moellep:


gurhar1133 (approved):
Looks good

PR #6177, opened 2023-07-29
fix #6142 implement new thermal transport simulation

### LINKED ISSUE #6142: The Thermal Transport tab needs to generate a sim which uses the ThermoOptic class
- [ ] The Thermal Transport tab needs to generate a sim which uses the ThermoOptic class in rslaser, rather than a stand-alone fenics sim. See the rslaser_thermooptic.ipynb notebook in rslaser_scratch. There are many different ways to invoke, depending on the pump rep rate, the pump profile, and the method (analytic or fenics). Either a steady state calc, or an "over time" calculation should be used depending on the report rate <> 100.

moellep:


gurhar1133 (approved):
Fixed small bug where if radial and longitudinal had slightly diff dimension then numpy would not be able to load into one ndarray. So used h5 instead.

PR #6183, opened 2023-07-31
Fix #6180: rep rate error

### LINKED ISSUE #6180: Pump rep rate error message
- [ ] Parse and display thermal simulation error message, especially the low rep-rate error
- [ ] also add colormap option to temperature profile heatmap report

gurhar1133:
The previous regex in `_parse_silas_log()` was not picking up on errors (since expected litteraly "*")


mkeilman:
This looks fine. One note: since we know ahead of time if the rep rate and profile are incompatible, it would be nice to invalidate the form so the user can't save the model in such a state. Maybe not worth the effort at this stage, however...

mkeilman (approved):
I had one comment but it's not in the way of approval

gurhar1133:
> This looks fine. One note: since we know ahead of time if the rep rate and profile are incompatible, it would be nice to invalidate the form so the user can't save the model in such a state. Maybe not worth the effort at this stage, however...

I'll ask the physicists at the meeting today

gurhar1133:
@mkeilman  you were right about just restricting the input domain. I will make that change and re request review

gurhar1133:
@mkeilman after more discussion of this with the physicists, later changes to SILAS will address input restriction. So I am merging this.

PR #6186, opened 2023-08-01
Fix #6022 srw scale wavefront during sim

### LINKED ISSUE #6022: Avoid timeout on SRW frame request for large files
The simulation below creates large intermediary wavefront files during the simulation process.

When the sim_frame() request is made by the client, the wavefront is read into memory on the server, and then converted into an appropriate SRW output file based on the characteristic requested and written to disk. Then that datafile is read and scaled using scipy.ndimage and returned to the client. The request takes too long and can cause a time-out on the server, or potentially on the client.

Instead, and extra step should be take to make a copy of the wavefront during the simulation process, and reduce the scale at that point. Then the sim_frame() request will be working with a much reduced set of data, and should complete quickly.

https://www.sirepo.com/srw#/beamline/D093hGau

mkeilman:
Putting this in for review though I'm not quite sure about it. It does load the resized plots quicker. Further notes:
- Because all the logic for resizing the data comes from the intensity files, this copies that functionality into the jinja template
- For the same reason the wavefront files behave as they did previously, not as per the posited pseudo-code
- The copying is done with `inspect` - copied instead of replaced because other 3d plots use the functions
- The copied functions refer to pkdc and numpy, so the imports are duplicated in a few places
- There were race conditions between writing and reading the intensity; the reads are now under the same retry block as the other reports

One other item: I had trouble with the example provided in the issue. The 1st watchpoint plot renders, some time after which I get a possible out of memory error.  I can't test in production because the plot fails, so I don't know if it's due to an error I've introduced. I'm not sure how to find that out.


mkeilman:
To draft while I fix tests

sirepo/package_data/template/srw/parameters.py.jinja
```diff
@@ -437,6 +437,71 @@ def epilogue():
     pass
     {% endif %}
 
+{% if isBeamlineAnimation %}
```
robnagler, line 440: This is too much code here.

sirepo/template/srw.py
```diff
@@ -2398,6 +2426,14 @@ def _set_parameters(v, data, plot_reports, run_dir, qcall=None):
     v.setupMagneticMeasurementFiles = (
         plot_reports or is_for_rsopt
     ) and _SIM_DATA.srw_uses_tabulated_zipfile(data)
+    v.isBeamlineAnimation = "beamlineAnimation" in report
+    v.maxMsgBytes = _MAX_MESSAGE_BYTES
+    v.canvasMaxSize = _CANVAS_MAX_SIZE
+    v.jsonExpandFactor = _JSON_MESSAGE_EXPANSION
+    v.extendPlot = inspect.getsource(_extend_plot)
```
robnagler, line 2433: This is not a good precedent.

robnagler (changes requested):
I think we have to be clear about parameters.py for download and parameters.py for execution. The former should be sparse, and for SRW, not require anything from sirepo or pykern repos. The latter is the case here, and parameters.py doesn't need all this code injected. Rather, put this in sirepo.template.srw.run_epilogue_epilogue. It's important to protect against parallel execution, which run_epilogue does with restrict_op_to_first_rank.

Write unit tests for the scaling code. If you need a lot of data, then should consider setting up a separate repo with said data so we don't bloat radiasoft/sirepo.

If this is not clear, please setup a meeting.

mkeilman:
> The latter is the case here, and parameters.py doesn't need all this code injected. Rather, put this in sirepo.template.srw.run_epilogue_epilogue.

Thank you! I did not know about that.

mkeilman:
> > The latter is the case here, and parameters.py doesn't need all this code injected. Rather, put this in sirepo.template.srw.run_epilogue_epilogue.
> 
> Thank you! I did not know about that.

I'm seeing that this won't work well here as-is. We don't show the plots or even count them until they are resized, so the entire simulation will have to complete before the user sees anything. We need to resize in the watchpoint loop in parameters.py. Would it be better to add a call to a templated function (requiring an extra import of sirepo.template) or throw some kwargs into `run_epilogue` and invoke that? I think the latter.

robnagler:
> We need to resize in the watchpoint loop in parameters.py

Makes sense. Just add a template function that gets called and protects itself. If it needs to be, call it from run_epilogue as well. However, it doesn't quite fit to run_epilogue with a parameter (this would be an interface change, since run_epilogue is dispatched to all codes). The function needs to isolate to the first rank like run_epilogue.

I think the synchronization in restrict_op_to_first_rank is probably ok, since that's how SRW works. All this stuff has to be thought through we each code, unfortunately.

robnagler:
pkio.atomic_write if this is on the side of writing within sirepo (not the export py).

mkeilman:
@robnagler this is not done, but is `template.run_on_server` the right idea? I presume we want something general, and `run_epilogue` could then invoke it.

mkeilman:
I'm not sure how to write a test for this. Is there a way to set up a non-example sim for testing? I don't see value in creating an example just for this case.

sirepo/package_data/template/srw/parameters.py.jinja
```diff
@@ -437,6 +437,15 @@ def epilogue():
...
+def process_watch(**kwargs):
+    {% if in_server %}
+    import sirepo.template
+    sirepo.template.run_on_server("srw", "process_watch", **kwargs)
```
robnagler, line 444: This is too much code. We don't need to dispatch. I don't think there needs to be a separate function. See below.

sirepo/template/srw.py
```diff
@@ -2398,6 +2448,7 @@ def _set_parameters(v, data, plot_reports, run_dir, qcall=None):
     v.setupMagneticMeasurementFiles = (
         plot_reports or is_for_rsopt
     ) and _SIM_DATA.srw_uses_tabulated_zipfile(data)
+    v.isBeamlineAnimation = "beamlineAnimation" in report
```
robnagler, line 2451: No need for this.

sirepo/template/srw.py
```diff
@@ -2528,6 +2579,8 @@ def _template_text(text):
 
 
 def _update_report_range(report, ar2d, x_range, y_range):
+    import numpy as np
```
robnagler, line 2582: I don't think this is necessary as np is imported above.

robnagler (changes requested):
> I'm not sure how to write a test for this. 

The functions can be independently testable, which would be the best approach. You can call private (underscored) functions from module tests so there's no need to make functions public.

sirepo/template/srw.py
mkeilman: [5e2a07d](https://github.com/radiasoft/sirepo/pull/6186/commits/5e2a07dea579641f9b481f897f742ccf2a53bd0f)

sirepo/template/srw.py
mkeilman: [25d3a7f](https://github.com/radiasoft/sirepo/pull/6186/commits/25d3a7f080bc2c87175e2548e98ee2930cc39eaf)

sirepo/template/srw.py
```diff
@@ -1989,6 +2025,9 @@ def _generate_srw_main(data, plot_reports, beamline_info):
                 content.append("op = set_optics(v, names, {})".format(is_last_watch))
                 if not is_last_watch:
                     content.append("srwl_bl.SRWLBeamline(_name=v.name).calc_all(v, op)")
+                    content.append(f"process_watch(wid={prev_watch})")
```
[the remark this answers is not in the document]
mkeilman: [2709c4b](https://github.com/radiasoft/sirepo/pull/6186/commits/2709c4b629a1f360ae3379ab3406e14329a3f003)

tests/template/srw_resize_3d_test.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+"""PyTest for :mod:`sirepo.template.srw`
+
+:copyright: Copyright (c) 2023 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+from pykern.pkcollections import PKDict
```
robnagler, line 7: Remove these imports and add an import pytest

sirepo/template/srw.py
```diff
@@ -145,6 +145,8 @@
 
 _JSON_MESSAGE_EXPANSION = 20
 
+_MAX_MESSAGE_BYTES = sirepo.job.cfg().max_message_bytes
```
robnagler, line 148: Do not cache this.

tests/template/srw_resize_3d_test.py
```diff
@@ -0,0 +1,36 @@
...
+        SIREPO_JOB_MAX_MESSAGE_BYTES=str(_N_BINS * _N_BINS),
+    )
+
+
+def test_srw_resize_3d(fc):
+    from sirepo.template import srw
+    import numpy
+
+    a, xr, yr = srw._reshape_3d(
+        numpy.random.rand(srw._MAX_MESSAGE_BYTES),
```
robnagler, line 27: Use job.cfg directly.

tests/template/srw_resize_3d_test.py
```diff
@@ -0,0 +1,36 @@
...
+        [0, 0, 0, 0.0, 1.0, _N_BINS, 0.0, 1.0, _N_BINS],
+        PKDict(),
+    )
+    pkok(
```
robnagler, line 31: For tests, it's better to keep things separate so that you can know which parameter failed.

```py
pkok(xr[2] < _N_BINS, "xr[2]={}", xr[2])
pkok(yr[2] < _N_BINS, "yr[2]={}", yr[2])
```

In general, giving the information this way is more useful, because it is shorter. You have to look at the test to see what the error is, but the text is easy to find.

tests/template/srw_resize_3d_test.py
```diff
@@ -0,0 +1,36 @@
...
+    os.environ.update(
+        SIREPO_JOB_MAX_MESSAGE_BYTES=str(_N_BINS * _N_BINS),
+    )
+
+
+def test_srw_resize_3d(fc):
+    from sirepo.template import srw
+    import numpy
+
+    a, xr, yr = srw._reshape_3d(
```
robnagler, line 26: I agree that this is the reasonable test.

tests/template/srw_resize_3d_test.py
```diff
@@ -0,0 +1,36 @@
...
+def setup_module(module):
+    import os
+
+    # Set max to something reasonable for testing
+    os.environ.update(
+        SIREPO_JOB_MAX_MESSAGE_BYTES=str(_N_BINS * _N_BINS),
+    )
+
+
+def test_srw_resize_3d(fc):
```
robnagler, line 22: You don't need `fc`. Then you can use     

```py
pkconfig.reset_state_for_testing(
        dict(
            SIREPO_JOB_MAX_MESSAGE_BYTES=str(_N_BINS * _N_BINS), 
    )
)
```

This keeps all the data local to the function.

robnagler (changes requested):
Looks good. Thanks for adding the test. A few minor comments.

tests/template/srw_resize_3d_test.py
mkeilman: [05493a0](https://github.com/radiasoft/sirepo/pull/6186/commits/05493a02bfb2ad3e47d3ac48ad9082ad242f5665)

tests/template/srw_resize_3d_test.py
mkeilman: [d695f0d](https://github.com/radiasoft/sirepo/pull/6186/commits/d695f0d9392f805ae540aa759eba134644d93717)

tests/template/srw_resize_3d_test.py
mkeilman: [851a07b](https://github.com/radiasoft/sirepo/pull/6186/commits/851a07bb4fdf6b2f875836aa9f0f36c01b0bd1cb)

sirepo/template/srw.py
mkeilman: [851a07b](https://github.com/radiasoft/sirepo/pull/6186/commits/851a07bb4fdf6b2f875836aa9f0f36c01b0bd1cb)

tests/template/srw_resize_3d_test.py
mkeilman: I see `import pytest` all over but it is rarely used. Also I still have to do the original imports in the test function. What does pytest do?

robnagler: The import pytest brings in pytest (the testing framework) as the first thing in the program. It may be superstition at this point, but it used to do goofy things with "fixtures" when you didn't import pytest. Perhaps we can dispense with it. Let's try for this test.

moellep:
This saves the single-electron intensity at each stage, but really we need the full wavefront so any characteristic can be extracted (single-electron intensity, multi-electron intensity, single-electron phase, single-electron real part, single-electron imaginary). Right now, changing the plot "Characteristic to be Extracted" doesn't change the plot from the intensity values.

x

So it would need to scale the raw wavefront, not the extracted values.

mkeilman:
draft while fixing

mkeilman:
Latest changes:
- Changing only the mesh of the wavefront did not work (as suspected). A dive into srw netted `ResizeElecFieldMesh()` which seems to be the right tool. As the comments to that function explain, it modifies the input wavefront in place
- Resizing the wavefront ignores various plot parameters such as `intensityPlotsWidth` so that the wavefront has enough information to reproduce everything. This implies that only the size changes around the max message length affect the size of the wavefront file

moellep:
This is looking really good. I think we can be a bit more aggressive on the scaling - I'll look at this and update this PR.

moellep:
@mkeilman I've made a few changes to the max dimension calculations with 1c2841f 
Give it a test with your development environment and then it is ready to go.

mkeilman:
> @mkeilman I've made a few changes to the max dimension calculations with [1c2841f](https://github.com/radiasoft/sirepo/commit/1c2841fe24fc6d83ab16a74a15a893f7d3b6b05f) Give it a test with your development environment and then it is ready to go.

"sim with large watchpoint files" is giving me "response too large to send" for the 2nd plot with max plot dimension set to original size

moellep:
Yes, you have to set the SIREPO_JOB_MAX_MESSAGE_BYTES=2g (which matches production), and also make sure the fix for #6312 is in your source tree.

mkeilman:
> Yes, you have to set the SIREPO_JOB_MAX_MESSAGE_BYTES=2g (which matches production), and also make sure the fix for #6312 is in your source tree.

got it, looks good

PR #6201, opened 2023-08-03
Fix #6200: Don't verify request's TLS certs in dev

### LINKED ISSUE #6200: don't verify calls to request in dev
We use self-signed certs so the verification fails

e-carlin:
We use self-signed certs so they fail verification

### DIFF
--- a/sirepo/crystal.py
+++ b/sirepo/crystal.py
@@ -160,7 +160,12 @@ def _get_server_data(energy, material, h, k, l):
         "df1df2": -1,
         "modeout": 1,
     }
-    r = requests.get(X0H_SERVER, params=payload, timeout=5)
+    r = requests.get(
+        X0H_SERVER,
+        params=payload,
+        timeout=5,
+        verify=job.cfg().verify_tls,
+    )
     content = r.text
     content = content.split("\n")
     return content
--- a/sirepo/crystal.py
+++ b/sirepo/crystal.py
@@ -160,12 +160,7 @@ def _get_server_data(energy, material, h, k, l):
         "df1df2": -1,
         "modeout": 1,
     }
-    r = requests.get(
-        X0H_SERVER,
-        params=payload,
-        timeout=5,
-        verify=job.cfg().verify_tls,
-    )
+    r = requests.get(X0H_SERVER, params=payload, timeout=5)
     content = r.text
     content = content.split("\n")
     return content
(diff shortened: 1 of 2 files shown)


e-carlin:
The jupyter change is necessary. I'm trying to update jupyter (and all related deps). The best way I've found to do this is to run an instance of jupyter using rsconf (v3 and v4) and to update the dev image and test. I'm running into TLS certificate issues that will be fixed by https://github.com/radiasoft/sirepo/pull/6201/files#diff-da04b971b9169f79fc9b01927fb279648bdd2d09be697647b87d96f225639024R75 (we use self-signed certs in rsconf that fail the check). So, then I thought why not just fix this across the board so I never run into this issue again. Really I'd like to wrap requests so I can put in a stronger guarantee (in the abstraction) that I don't run into this. But, that seemed like overkill at the moment. So, then I thought "all requests code should always be written to handle this verfication problem". So, my intention with this PR was to do that (similar to how we always put a security comment when we set `shell=True` in subprocess in case someone copies the code). So, the crystal change isn't strictly necessary but seemed like a goo "hygiene" fix.

robnagler:
This is not a good approach for 3rd party connections. The default should be deny, even in dev. Imagine if someone used this to play with GitHub's API, and there was a MITM at some coffee shop. The developer's creds would be captured.

verify_tls should be the default unless we know a good reason that it shouldn't be. We will have bugs in development that will be confusing with the benefit of baking in proper security by default.

Many security defects are created with this type of change.

e-carlin:
583888127 makes sense for 3rd party.

What do you think of leaving it for jupyter since that is a system we control? Or should I fix rsconf to not use TLS in dev? Or something else?

robnagler:
> rsconf to not use TLS in dev

This is a much better solution, because it is in the place in the code that has more exposure in security audits. Indeed, it need not be a default, but in the rsconf component yaml for dev it could have a `verify_tls: False` with an assertion inside rsconf that would make sure this was only the dev channel. This way you wouldn't have confusion with, say, [sirepo's radia-run](https://github.com/radiasoft/sirepo/blob/a66d8b1337ec8b2d2fd4d07a6da8e9ce7ff6fc5e/container-conf/radia-run.sh#L11), which doesn't use rsconf but does rely on dev mode. For normal development, we don't use TLS so we only need verify_tls false for something like Docker in dev. 

An even better solution would be to create a certificate authority, add it to /etc/pkcli (and other places, maybe), and then sign all dev certs with that CA. This is much more work, though.

e-carlin:
Sounds good.

PR #6207, opened 2023-08-08
fix #6204 many silas enhancements and bug fixes

### LINKED ISSUE #6204: silas enhancements
Several silas changes are required:
 - [x] simulation error should show in the Simulation Status panel, not as a User Alert
 - [x] Choose Crystal widget layout is incorrect on the amplifier tab
 - [x] javascript rep-rate error is being triggered when the crystal popup editor is closed
 - [x] rep-rate of 1.0 should be allowed
 - [x] Sim method is already part of the crystal parameters and should be moved into the crystal panel from the simulation settings panel
 - [x] Sim method needs to always disable if repo rate is <= 1.0
 - [x] Sim method should be "Analytical" or "Simulated", remove Fenics text
 - [x] When crystal is initialized in the generated code with the call to calc_n0n2(), the ABCD values should also be updated 
 - [x] Switch axes for the Radial/Axial Temp Profile plot so Longitudinal is the X axis
 - [x] Show error message if simulation fails on Laser Pulse tab (ex. invalid pulse width)

moellep:
- add enhancements in #6204
- added error messages to animation simulations
- hide crystal.calc_type if pump_rep_rate <= 1, always use fenics for that case
- removed thermalCrystal.crystalSimType, already present on crystal.calc_type
- autocompute ABCD matrix values and pulse.l_scale during simulations


gurhar1133 (approved):
lgtm

Issue #6132, opened 2023-07-19
Energy filter breaks plotting


sjcrs:
The EnergyFilter option actually multiplies the number of tally arrays in the statefile, based on the number of energy bins.  Currently, if a user employs the energy filter in addition to a mesh filter for a tally, it cannot be plotted within the visualization tab.  

- [ ] If the EnergyFilter is employed, there should be a drop box in the pencil menu to select which energy bin is shown.  This could be tricky, because some examples use ~500 bins (see https://nbviewer.org/github/openmc-dev/openmc-notebooks/blob/main/flux-spectrum.ipynb).  I'm open to suggestions for alternatives to a drop box.  Perhaps a double-sided sliding window that can sum adjacent energy bins? 

It would also be a useful feature if a user could select a bin and produce a 1d spectrum. Please see the notebook at `EFilterExample.ipynb` for an example. (Moved to #6527)

moellep:
We discussed in the meeting and decided to use a range selector for the energies - selecting the upper and lower boundary. Then show a summed mesh in 2d and 3d.

PR #6225, opened 2023-08-14
Fix #6081 cloudmc 2d tally controls

### LINKED ISSUE #6081: 2d tally plot controls improvements
Some rough edges that need smoothing:

- [x] Make sure the slider controlling the display plane is updated properly. Currently the range and value do not get reset when the axis of the plane changes, for example
- [x] If the slider has a single value in its range, disable it
- [x] Show the units (cm) with the slider value and the min/max fields

mkeilman:
Notes:
- The clipping range boxes are now two-handled sliders courtesy of jQueryUI
- The plane position slider also uses jQueryUI for consistency
- The controls update properly when the axis changes

The layout and labels could use some tweaking. It will also be worthwhile to put the sliders into a custom field, but that can wait to be part of #6132.

Of further note: the jQueryUI javascript file contains *only* the slider and its dependencies (see the [download builder](https://jqueryui.com/download/))


mkeilman:
Another note: because of the way the aspect ratio works, changing the clip ranges can make the plot change size or distort. We should make a new issue to more intelligently decide the plot size and aspect ratio (probably fix height or width depending on the viewable area).

moellep:
I really like the Plane position widget, but I don't like the Clip options because of the distortion you mentioned and that using the mouse zoom is much more natural. We can have the Clip if you really prefer, but it should be hidden in a sub-option. When the Plane position is changed, it should preserve the current zoom settings, so you can view a section across frames.

The heatmap should show the difference between a missing tally value, and a 0.0 tally value - but this could be opened as a new issue.

moellep:
The heatmap should use the same axes arrangement as the defaults on the 3D plot:
y: x by z (current is z by x)

mkeilman:
> I really like the Plane position widget, but I don't like the Clip options because of the distortion you mentioned and that using the mouse zoom is much more natural. We can have the Clip if you really prefer, but it should be hidden in a sub-option. When the Plane position is changed, it should preserve the current zoom settings, so you can view a section across frames.

Clip was meant to complement zoom for cases where the area of interest is in a small area,  but I'm not rabid for it. I'll remove it.

mkeilman:
To draft while I fix some things

mkeilman:
> The heatmap should show the difference between a missing tally value, and a 0.0 tally value - but this could be opened as a new issue.

Seems pretty involved, as the heatmap has no concept of "don't plot anything here". For that reason ad because it would affect all apps, I agree it should be its own issue.

mkeilman:
Most items addressed; one extra thing I added is to rotate the default 3d view (y axis pointing in, x axis increasing leftt too right) so it matches the 2d view.

I made new issues to omit missing tally values and to preserve zoom when changing slices.

PR #6228, opened 2023-08-15
fix #6171 unlimited omega sub-simulations

### LINKED ISSUE #6171: omega should support more than 4 chained simulations
The omega app should be updated to support any number of chained simulations.

moellep:


sirepo/template/omega.py
```diff
@@ -283,11 +253,49 @@ def _generate_parameters_file(data):
...
+        return PKDict(
+            simCount=sim_count,
+            modelName=model_name,
+            reportCount=report_count,
+            modelKey=f"{model_name}{sim_count}-{report_count}",
+        )
+
+    res = []
+    idx = 0
+    while True:
```
gurhar1133, line 267: maybe something like this
```python
sim_dir = run_dir.join(f"run{idx + 1}")
while sim_dir.exists() and _has_file():
    ...
    idx += 1
    sim_dir = run_dir.join(f"run{idx + 1}")
```
instead of the `break`

gurhar1133 (approved):
Looks good

sirepo/template/omega.py
moellep: Yes, that's a good idea. The repetition of f"run{idx + 1}" around the module isn't good either, so I'll replace that with a method as well.

PR #6232, opened 2023-08-16
fix #6230 added work-around for Safari browser zoom bug

### LINKED ISSUE #6230: plot zoom does not work in Safari
The mouse zoom feature for plots does not appear to work anymore in the Safari browser.

moellep:


mkeilman (approved):
lgtm

PR #6245, opened 2023-08-24
Fix #6239: genesis omega field dist

### LINKED ISSUE #6239: show genesis field distribution with other omega plots
When genesis is selected as one of the omega workflow sims, the particle, field and parameter plots should all be shown. Currently the field distribution is missing.

gurhar1133:
Field distribution plot added to genesis visualization in omega

### DIFF
--- a/sirepo/template/omega.py
+++ b/sirepo/template/omega.py
@@ -338,6 +338,7 @@ def _plot_beam(sim_type, frame_args):
 def _plot_field_dist(sim_type, frame_args):
     import sirepo.template.genesis
 
+    frame_args.frameIndex = 1
     return sirepo.template.genesis.sim_frame_fieldDistributionAnimation(frame_args)
 
 
--- a/sirepo/template/omega.py
+++ b/sirepo/template/omega.py
@@ -274,12 +274,6 @@ def _output_info(run_dir):
     idx = 0
     sim_dir = _sim_dir(run_dir, idx + 1)
     while sim_dir.exists() and _has_file(sim_dir):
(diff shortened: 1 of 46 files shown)


sirepo/template/omega.py
```diff
@@ -285,12 +287,26 @@ def _report_info(sim_count, model_name, report_count):
...
+                    _report_info(idx + 1, "simFieldDistributionAnimation", 1),
+                ]
+            )
         idx += 1
         sim_dir = _sim_dir(run_dir, idx + 1)
 
     return res
 
 
+def _is_genesis(run_dir, index):
```
mkeilman, line 302: Might move inside `_output_info`

sirepo/template/omega.py
```diff
@@ -285,12 +287,26 @@ def _report_info(sim_count, model_name, report_count):
...
+    t, i = _sim_info(
```
mkeilman, line 303: If you don't need `i` you can replace it with an underscore as a hint to the reader that it isn't used.

mkeilman (commented):
Minor comments

sirepo/template/omega.py
gurhar1133: Oh oops. I thought it was the radiasoft convention that we didn't do this.

mkeilman: I guess I don't know about the convention. I do see it in a number of places...

gurhar1133: I agree with you that `_` should be the convention for an unused unpacked tuple variable

PR #6275, opened 2023-09-05
Fix #6272 radia download dmp

### LINKED ISSUE #6272: radia.create_archive should be download data file?
@mkeilman there is special code for radia to export the dump file as an archive. This seems like this should be download data file or something similar. Please explain the usage.

mkeilman:
This removes the custom create_archive code, which was ultimately unnecessary and which only radia used. Now using downloadFile.


mkeilman:
to draft - downloadFile does not work because the dump file is not a lib file

robnagler:
downloadDataFile is for this purpose. It's a different API.

mkeilman:
> downloadDataFile is for this purpose. It's a different API.

I'm putting that in now, but if I may, I get those two confused a lot. Nothing about the names distinguishes them. "Data files" **are** files. Conversely **all** files can be thought of as "data files". I'd prefer "downloadLibFile" or something...

robnagler:
> I'm putting that in now, but if I may, I get those two confused a lot. Nothing about the names distinguishes them. "Data files" are files. Conversely all files can be thought of as "data files". I'd prefer "downloadLibFile" or something...

Yes, better names would be great. Feel free to put up a proposal for names. The word "data" should be stricken from our vocab. :(

mkeilman:
Latest change uses downloadDataFile. Also we've been attaching the dmp file "geometry.dat" as <simulation name>.dat - downloadDataFile does not account for a "save as", so there is a file copy in `get_data_file()`. The template should define the filename that way to begin with, but it touches a lot of places due to `get_g_id()`. Outside this scope I believe.

moellep:
I changed the download slightly so the file copy isn't needed - you can set the uri to override downloaded filename.

mkeilman:
> I changed the download slightly so the file copy isn't needed - you can set the uri to override downloaded filename.

Thanks, didn't know about that

PR #6286, opened 2023-09-07
fix #6285 raydata added search text, paging and UI improvements

### LINKED ISSUE #6285: raydata scan paging, search, and UI improvements

moellep:
List of changes in 0fa526f

Sorting/paging is only implemented on the main analysis scans page, so the other pages will need additional work.

This PR only works with a databroker with a mongodb back-end. Searching isn't supported on the "inplace" unpacked data. Instead use the "mongo_normalized" option.

New bugs may have been introduced, please test your normal scenarios.


e-carlin (approved):
Looks great.

sirepo/package_data/static/js/raydata.js
```diff
@@ -367,105 +375,71 @@ SIREPO.app.directive('scansTable', function() {
...
+                <button class="btn btn-info btn-xs raydata-button pull-right" data-ng-show="showPdfButton()" data-ng-click="downloadSelectedAnalyses()">Download Selected Analysis PDFs</button>
                 <table class="table table-striped table-hover">
                   <thead>
                     <tr>
-                      <th style="width: 20px; height: 20px;" data-ng-show="showPdfColumn"><input type="checkbox" data-ng-checked="pdfSelectAllScans" data-ng-click="togglePdfSelectAll()"/> PDF</th>
-                      <th data-ng-if="analysisStatus === 'allStatuses'" style="width: 50px; height: 40px;"></th>
-                      <th style="width: 50px; height: 40px;"></th>
-                      <th data-ng-repeat="column in columnHeaders track by $index" data-ng-mouseover="hoverChange($index, true)" data-ng-mouseleave="hoverChange($index, false)" data-ng-click="sortCol(column)" style="width: 100px; height: 40px;">
+                      <th style="width: 20px; height: 40px; white-space: nowrap" data-ng-show="showPdfColumn"><input type="checkbox" data-ng-checked="pdfSelectAllScans" data-ng-click="togglePdfSelectAll()"/> <span style="vertical-align: top">PDF</span></th>
```
rorour, line 390: Are the notebooks still producing PDFs or can we get rid of this column?

rorour (approved):
Looks great!

sirepo/package_data/static/js/raydata.js
e-carlin: That's a good question. I think they will continue to be produced. They will just be written out to a configured location with the rest of the analysis and we will have to retrieve them from there.

moellep: Good question - the old chx notebook we use does use the scan.pdf column, but it looks like csx does not. @e-carlin do you know if the current notebooks use this feature?

e-carlin: From my understanding the latest notebooks both write out pdf's. So scan.pdf should be True for them.

PR #6295, opened 2023-09-11
Fix #6041 cloudmc 2d geometry

### LINKED ISSUE #6041: include geometry in 2D plots
This package looks very promising.
https://github.com/fusion-energy/dagmc_geometry_slice_plotter

Also, look at the OpenMC plot explorer for examples:
https://github.com/openmc-dev/plotter

mkeilman:
Notes:
- This saves the *decimated* meshes at the point of volume extraction using `dagmc_geometry_slice_plotter` library. One can also slice the entire dagmc file, but this allows control of which volumes we display. 
- We generate all the possible volume outlines for each dimension after the openmcAnimation completes
- On the client side, this adds an overlay to the heatmap to display arbitrary svg paths

And for further consideration:
- We might avoid saving graveyard meshes
- We could also change the decimation factor of the saved meshes to reduce the number of points
- It might make more sense to calculate the outlines when the tally mesh is saved


moellep:
The 2D display looks really nice. I'm comparing with openmc plotter, and found some differences in our heatmap values (which is unrelated to this, I'll open a different issue). One thing I did notice is if I create a oversimplified 3x3 mesh, the sirepo overlay isn't changing between settings. The overlay did change correctly on larger tallies.

Here's an example:
https://www.sirepo.com/cloudmc#/geometry/dOJOh0Kc

sirepo (with SIREPO.PLOTTING_HEATPLOT_FULL_PIXEL = true) to match openmc-plotters alignment, which may not be correct.
x

openmc-plotter
x2

moellep:
The problem with the 2D geometry comes up if you change the size of the mesh and run a new simulation - it is still using the previously cached slices. So if you simulate on a 3x3 mesh, then change it 6x6 and run again, it doesn't recompute the new 2D geometry slices.

mkeilman:
We also need to invalidate the simulation once the settings change, or the data will not match the mesh. Should that be a new issue?

mkeilman:
> We also need to invalidate the simulation once the settings change, or the data will not match the mesh. Should that be a new issue?

Answering my own question: a true reset seems to require a bit of work (#6297) and is beyond the scope here.

moellep:
I updated the code so the reports always work off the state at the time the simulation was run.

PR #6300, opened 2023-09-12
Fix #6268 sirepo.reply defers marshalling until send

### LINKED ISSUE #6157: Remove unused app and server methods
We have some cruft APIs which are not being used, for example exportJupyterNotebook and exportJupyterNotebook2. These could be removed.

### LINKED ISSUE #6268: msgpack not json in sirepo.reply
It may be the conversion to JSON is happening too early.

### LINKED ISSUE #6274: api_downloadFile returns json on non-existent file
`{"state":"error","error":"Simulation library file \"sample.tif\" does not exist"}`

The source of the problem is lib_file_abspath raises a UserAlert. However, it doesn't know the context. It should probably throw a NotFound, but that will break the application for other GET operations.

I think the solution the problem is to render it as a server error if it is a GET, and include the error message in the rendered HTML.

### LINKED ISSUE #6288: api_serverStatus calls time.sleep, needs async sleep

### LINKED ISSUE #6291: admJobs & ownJobs pass req_data and sim_data
`Sep 08 20:00:57 982823     0 ../../sirepo/job_api.py:296:_request_api {'api': 'api_admJobs', 'req_data': {'simulationType': 'srw'}, 'serverSecret': <REDACTED>, 'sim_data': <class 'sirepo.sim_data.srw.SimData'>, \
'type': 'srw'}`

### LINKED ISSUE #6293: api_stateless/fulCompute need proper jids

robnagler:
- restructured sirepo.reply to avoid early json serialization
- restructured srunit to handle replies more consistently between websocket and http
- Fix #6291 restructured job_api to simplify parameters
- Fix #6293 clean up stateless/fulCompute to pass correct parameters
- Fix #6288 api_serverStatus calls asyncio.sleep
- Fix #6157 removed exportJupyterNotebook, webcon, and cruft in create_archive
- Fix #6274 avoid returning json for GET on errors, include detail in HTML
- HTML is always properly rendered (formally didn't include doctype)


sirepo/job_supervisor.py
```diff
@@ -574,7 +573,8 @@ def set_situation(self, op, situation, exception=None):
     @classmethod
     def __db_file(cls, computeJid):
         return _DB_DIR.join(
-            computeJid + sirepo.const.JSON_SUFFIX,
+            sirepo.simulation_db.assert_sim_db_file_path(computeJid)
```
e-carlin, line 576: I'm confused by this. I don't believe __db_file (db/supervisor-job) has anything to do with sim_db_file (db/user/<uid>/<sim_type>/<sid>) so I'm not sure why that method would be called here. Maybe it needs to be renamed to capture multiple uses?

sirepo/job_supervisor.py
```diff
@@ -588,7 +588,7 @@ def __db_init_new(cls, data, prev_db=None):
             queueState="queued",
             canceledAfterSecs=None,
             computeJid=data.computeJid,
-            computeJobHash=data.computeJobHash,
+            computeJobHash=data.get("computeJobHash"),
```
e-carlin, line 591: This is better than unused but it seems to me like an object here where if this field is accessed we blow up would be better. I worry about us comparing None's somewhere else and code working in an incorrect way.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4616,7 +4607,7 @@ SIREPO.app.service('plotToPNG', function() {
 
 });
 
-SIREPO.app.service('fileUpload', function(msgRouter) {
+SIREPO.app.service('fileUpload', function(msgRouter, requestSender) {
```
e-carlin, line 4610: requestSender looks to be unused.

e-carlin (changes requested):
A few things I ran into while testing:

- [x] sbatch login doesn't work
```
Sep 14 14:47:31 3987232     0 sirepo/uri_router.py:282:parse_msg start ws_id=1 req_seq=22 uri=/sbatch-login
Sep 14 14:47:31 3987232     0 sirepo/uri_router.py:443:_call_api api=sbatchLogin exception=invalid simulation type=None stack=Traceback (most recent call last):
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/uri_router.py", line 434, in _call_api
    await getattr(qcall, qcall.uri_route.func_name)(**kwargs)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/job_api.py", line 210, in api_sbatchLogin
    r = self._request_content(
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/job_api.py", line 382, in _request_content
    d = self.parse_post(
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/quest.py", line 151, in parse_post
    return http_request.parse_post(self, PKDict(kwargs))
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/http_request.py", line 100, in parse_post
    res[n] = f(v)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/http_request.py", line 71, in _type
    qcall.auth.check_sim_type_role(v)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/auth/__init__.py", line 159, in check_sim_type_role
    t = sirepo.template.assert_sim_type(sim_type)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/__init__.py", line 12, in assert_sim_type
    return sirepo.util.assert_sim_type(*args, **kwargs)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/util.py", line 237, in assert_sim_type
    assert is_sim_type(sim_type), f"invalid simulation type={sim_type}"
AssertionError: invalid simulation type=None
```

- [ ] The mail message for moderation redirect is missing a slash between the port and path `http://127.0.0.1:8080adm-moderate-redirect`

- [ ] The forbidden page is formatted incorrectly
Screen Shot 2023-09-14 at 8 51 10 AM

- [x] The mail message for moderation clarification is formatted oddly. There is an `=` in "information". No clue why. It could be something with emacs but not sure.
Screen Shot 2023-09-14 at 8 54 10 AM

robnagler:
sbatch login not working is fixed by 879ac6ab3

robnagler:
> The mail message for moderation clarification is formatted oddly.

This is formatted quoted printable, which looks odd when literal text.

sirepo/job_supervisor.py
robnagler: This is poor naming of assert_sim_db_file_path, but it is the function I want to use. I don't want the computeJid to be something that's unsafe.

I'm going to ignore this for now. sim_db_file is a complicated name.

sirepo/job_supervisor.py
robnagler: I can't fix this. computeJobHash=None is correct. There is no way to compute the job has, because there is not specification for the "job" since it is a method that does arbitrary things. The code in job_supervisor always compares with equals, and only for simulations, e.g. in downloadDataFile.

We are reusing "job" here for statelessCompute, and this is the complication. There might need to be a subclass SimulationJob, but that would be even more complicated.

sirepo/package_data/static/js/sirepo-components.js
robnagler: 2692403

robnagler:
@e-carlin

e-carlin:
I'll review this afternoon

sirepo/package_data/static/html/http-not-found.html
```diff
@@ -2,7 +2,7 @@
   <div class="row">
     <div class="col-sm-6 col-sm-offset-2">
       <h1>Not Found</h1>
-      The requested URL was not found on this server.
+      <p>The requested URL was not found on this server.</p>
```
e-carlin, line 5: is not found? too-large.html was changed for was to is.

sirepo/package_data/static/json/schema-common.json
```diff
@@ -131,15 +127,13 @@
         "deleteSimulation": "/delete-simulation",
         "downloadDataFile": "/download-data-file/<simulation_type>/<simulation_id>/<model>/<frame>/?<suffix>",
         "downloadFile": "/download-file/<simulation_type>/<simulation_id>/<filename>",
-        "errorLogging": "/error-logging",
```
e-carlin, line 134: This is a bit odd. I get that we never look up this value so we don't need it here. But, now we have routes that aren't listed in the schema but that exist which seems odd. 

While looking at this code I came across this comment which I don't think makes any sense https://github.com/radiasoft/sirepo/blob/423a03e8c432a9a9a2cca8adc589955e9927d740/sirepo/package_data/static/js/sirepo.js#L3310

e-carlin (approved):
2 very minor style nits. Everything else looked and worked well.

sirepo/package_data/static/html/http-not-found.html
robnagler: This is for 404.

sirepo/package_data/static/json/schema-common.json
robnagler: e4183ec27

PR #6311, opened 2023-09-15
Fix #4806 radia magnet viewer window

### LINKED ISSUE #4806: magnet view collapse and fullscreen don't work
In radia, if you select the collapse "^" icon on the Magnet Viewer panel, it doesn't minimize the panel. Also the fullscreen panel button doesn't work on that panel. Maybe those items shouldn't be shown on that panel.

mkeilman:
Notes:
- Show/hide was not being handled due to panelState missing from $scope
- Added code to redraw axes properly after show/hide
- Removed fullscreen - arranging the controls along with the magnet viewer is complicated enough to  have its own issue (if necessary)


moellep:
I found the 3d panel doesn't display correctly if you minimize the 3d panel, resize the browser window, and then show the panel. This is a general 3d bug, because it also happens in the cloudmc app.

x

PR #6330, opened 2023-09-19
fix #6323 fix slice position and ui image to display alignment

### LINKED ISSUE #6323: align openmc plots on-the-pixel
The openmc mesh is all within the mesh dimensions, rather than having the boundaries on the half-pixel. In the UI, this value needs to be set:
SIREPO.PLOTTING_HEATPLOT_FULL_PIXEL = true;

and on the 2d geometry slices, the values will need to be adjusted centerward, and should match the openmc-plotter.

It looks like the 3d plots are already plotting the correct way.

Corrected 2D:
Tally-Results-t1-heating-mean-4

3D:
Tally-Results-t1-heating-mean-5

moellep:
2D slices now match openmc-plotter

### DIFF
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -4,6 +4,7 @@ var srlog = SIREPO.srlog;
 var srdbg = SIREPO.srdbg;
 
 SIREPO.app.config(() => {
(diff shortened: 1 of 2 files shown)


mkeilman (approved):
All good

PR #6332, opened 2023-09-19
fix #6198: update srwpy imports

### LINKED ISSUE #6198: 20230922 Update srwpy imports
After https://github.com/radiasoft/download/pull/512, imports in sirepo will need to be updated to get rid of warning message

### LINKED ISSUE #6217: 20230922 update srw generated to to use srwpy
The generated python should use the new srwpy library.
For example, 
```python
import srwpy
# ...
srwpy.srwl_bl.SRWLBeamline()
```
instead of
```python
import srwl_bl
# ...
srwl_bl.SRWLBeamline()
```

rorour:
Fixes #6217 
Fixes #6198


robnagler (approved):
lgtm. The real test will be on alpha.

PR #6337, opened 2023-09-21
Fix #6282 final particle and field plots

### LINKED ISSUE #6282: genesis final particle and field output reports
Genesis creates a .dpa and a .dfl file for the final particle/field distributions if the IDUMP field is set to 1. These reports should always be created by sirepo, and included on the visualization page, shown  before the particle and field distribution time slice plots. The data format for these output files are slightly different than the .par and .fld files.

gurhar1133:


moellep:
@gurhar1133 I made a few more changes with ccb2729. Please take a test and if it looks good we can merge this.

PR #6340, opened 2023-09-22
fix #6290 use analysisJob for polling and pkio module for file compare

### LINKED ISSUE #6290: Change statefulCompute/statelessCompute that access runDir
Code [like this](https://github.com/radiasoft/sirepo/blob/aae7bf6ff265e3c7123411429612b38734023f42/sirepo/template/epicsllrf.py#L64) should not be written:

```py
def stateless_compute_read_epics_values(data, **kwargs):
    _PREV_EPICS_FILE = "prev-status.json"
    # TODO(pjm): hacked in animation directory
    run_dir = pkio.py_path(
        re.sub(r"/unused$", "/animation", str(simulation_db.simulation_run_dir(data)))
    )
    p = run_dir.join(_PREV_EPICS_FILE)
```

This is an analysisJob, not a statelessCompute.

BTW, globals (`_PREV_EPICS_FILE`) should be defined globally (with `global`), but in this case, it can be local to make the code clearer:

```py
    run_dir.join("prev-status.json")
```

Also, 

```py
    shutil.copyfile(str(e), str(p))
```

This is not necessary. The reason we chose [py.path](https://py.readthedocs.io/en/latest/path.html) is that it solves this problem for us:

```py
    e.copy(p)
```

And this should be encapsulated in pkio:

```py
        and e.exists()
        and p.exists()
        and filecmp.cmp(str(e), str(p), False)
```

See https://github.com/radiasoft/sirepo/issues/6289 for why.

moellep:
depends on radiasoft/pykern/pull/398

### DIFF
--- a/sirepo/package_data/static/js/epicsllrf.js
+++ b/sirepo/package_data/static/js/epicsllrf.js
@@ -161,7 +161,7 @@ SIREPO.app.controller('epicsllrfController', function (appState, epicsllrfServic
                     return;
                 }
                 inRequest = true;
-                requestSender.sendStatelessCompute(
+                requestSender.sendAnalysisJob(
                     appState,
                     function (data) {
                         inRequest = false;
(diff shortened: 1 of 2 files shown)


robnagler (approved):
I changed the copy to pass `stat=True` to get the effect you were trying to achieve.

PR #6346, opened 2023-09-25
Fix #6334 collapsed vtk reports

### LINKED ISSUE #6334: Collapsed vtk reports not drawn if expanded after browser resize
See also https://github.com/radiasoft/sirepo/pull/6311#issuecomment-1727886747.  From which note
```
[.WebGL-0x10400d08100] GL_INVALID_VALUE: Texture dimensions must all be greater than zero
...
[.WebGL-0x10400d08100] GL_INVALID_FRAMEBUFFER_OPERATION: Draw framebuffer is incomplete
```

mkeilman:
Notes:
- Various attempts to use offscreen rendering (useOffscreen in vtk.js, OffscreenCanvas in javascript) were not very successful
- Instead this uses `data-ng-if="panelState.doRender()"` *in addition to* the existing `data-ng-hide=panelState.Hidden()`
- A panel will render if a) the panel is not hidden and b) it has not yet been rendered in this session. That lets us avoid reloading the entire panel when unhiding
- We also have to disable listening to the vtk render's resize when hidden
- Includes special handing for silas and srw, but not warpvnd or zgoubi since they are basically defunct. It's easy enough to add if necessary

### DIFF
--- a/sirepo/package_data/static/html/genesis-visualization.html
+++ b/sirepo/package_data/static/html/genesis-visualization.html
@@ -8,14 +8,21 @@
         <div data-sim-status-panel="visualization.simState"></div>
       </div>
     </div>
-    <div class="clearfix"></div>
     <div class="col-md-6 col-xl-4" data-ng-if="visualization.simState.hasFrames()">
       <div data-report-panel="parameter" data-model-name="parameterAnimation"></div>
     </div>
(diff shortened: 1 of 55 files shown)


moellep:
#6359 is a different PR for this bug fix. Please take a look - it has the advantage that no new APIs were added, and apps can mark that they have 3d content on a view in the schema, so no app javascript is required.

PR #6347, opened 2023-09-25
fix #6273 fix autoSave error cases

### LINKED ISSUE #6273: autoSave callback confusion
[autoSave](https://github.com/radiasoft/sirepo/blob/8fcc5e530387c9b31e4d0f6206ff3816dc339425/sirepo/package_data/static/js/sirepo.js#L398) has some code that doesn't quite make sense:

```js
                    successCallback: function (resp) {
                        if (resp.error && resp.error == 'invalidSerial') {
                            srlog("...snip...")
                            refreshSimulationData(resp.simulationData);
                            errorService.alertText("...snip...")
                        }
                        else {
                            lastAutoSaveData = self.clone(resp);
                            ['simulationSerial', 'name', 'lastModified'].forEach(f => {
                                savedModelValues.simulation[f] =
                                    self.models.simulation[f] = lastAutoSaveData.models.simulation[f];
                            });
                        }
                        if ($.isFunction(callback)) {
                            callback(resp);
                        }
                    },
                    errorCallback: function (resp) {
                        if ($.isFunction(errorCallback)) {
                            errorCallback(resp);
                        }
                    },
```

If there is an error and it is not invalidSerial, the data are saved even though the response was "error". I suspect that the only case error would be returned is invalidSerial in practice. However, I want to move invalidSerial to an srException, since it's an oddball case for util.Error.

The other thing that happens in an error case is that callback is called, not errorCallback. I'm not sure that makes sense either.

@moellep what do you think about calling errorCallback in an error case, always, and not saving the models.

moellep:
- remove uncalled errorCallback from autoSave()
 - fixed bug if resp.error is present in autoSave response, but error is not invalidSerial
 - update fileManager sim from server response when renaming
 - if an invalidSerial occurs on the server, return the on-disk simulation data to be refreshed on the client

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -395,7 +395,7 @@ SIREPO.app.factory('appState', function(errorService, fileManager, requestQueue,
         return false;
     };
 
-    self.autoSave = function(callback, errorCallback) {
+    self.autoSave = function(callback) {
         if (! self.isLoaded() ||
             lastAutoSaveData && deepEqualsNoSimulationStatus(
                 lastAutoSaveData.models, savedModelValues)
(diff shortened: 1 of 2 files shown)


sirepo/package_data/static/js/sirepo.js
```diff
@@ -414,35 +414,28 @@ SIREPO.app.factory('appState', function(errorService, fileManager, requestQueue,
...
-                                '; refreshing'
-                            );
-                            refreshSimulationData(resp.simulationData);
-                            errorService.alertText(
-                                "Another browser updated this simulation."
-                                + " This window's state has been refreshed."
-                                + " Please retry your action."
-                            );
+                        if (resp.error) {
+                            if (resp.error === 'invalidSerial') {
```
robnagler, line 418: I fixed the code in #6300 to remove this test here, and instead to [refresh the app](https://github.com/radiasoft/sirepo/pull/6300/files#diff-c1af31ef028c0a9ee6c6e71acb4b0ad646d4136dd06d3a1bc9cdb11c55fbc801R588). It never happens, and when it does, a complete refresh is probably safest given the amount of caching in the app and the fact that simulationData was never used.

Your new code (after #6300 is merged shortly) should simply remove the errorCallback. Easiest thing would be revert this change in your branch so you don't get a conflict.

sirepo/simulation_db.py
```diff
@@ -585,7 +585,11 @@ def _serial_validate(incoming, on_disk):
             PKDict(
                 error="invalidSerial",
                 sim_type=data.simulationType,
-                simulationData=data,
```
robnagler, line 588: Remove this to avoid conflict with #6300.

robnagler (changes requested):

moellep:
OK - I've changed this PR to a draft and will look at it again after #6300

robnagler:
#6300 has been merged.

moellep:
Closing this, will open a new PR.

PR #6352, opened 2023-09-26
Fix #6250 added SIREPO.traceWS and better logging

### LINKED ISSUE #6250: Add SIREPO.traceWS and better logging of ws messages
@e-carlin [writes](https://github.com/radiasoft/sirepo/pull/6092#pullrequestreview-1596418811):
> With websockets and a binary protocol that now doesn't work. What about adding a pkdc to parse_message and websocket_response that allows one to dump the message (with associated uri, seq num, etc)?

Add pkdc like feature called sr_trace or similar to sirepo.js. In the client, you can expand the object as opposed to cluttering a terminal window with all messages between client and server, which are often quite large. sr_trace could get the control regex from pkdebug. We could also set it dynamically in production from the js console.

```js
function sr_trace(...args) {
    if (sr_trace_re) {
        if (sr_trace.caller === null || sr_trace_re.match(sr_trace.caller) {
             // probably want better formatting of sr_trace.caller
             console.log(sr_trace.caller, ...args);
        }
    }
}
```

### LINKED ISSUE #6351: No module named 'sirepo.template.srwpy

robnagler:
Fix #6351 brilliance.py.jinja incorrect srwpy import

### DIFF
--- a/sirepo/package_data/static/js/sirepo.js
+++ b/sirepo/package_data/static/js/sirepo.js
@@ -1,6 +1,7 @@
 'use strict';
 SIREPO.srlog = console.log.bind(console);
 SIREPO.srdbg = console.log.bind(console);
+SIREPO.traceWS = true;
 
 // No timeout for now (https://github.com/radiasoft/sirepo/issues/317)
 SIREPO.http_timeout = 0;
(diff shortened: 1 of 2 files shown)


sirepo/package_data/static/js/sirepo.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 SIREPO.srlog = console.log.bind(console);
 SIREPO.srdbg = console.log.bind(console);
+SIREPO.traceWS = true;
```
moellep, line 4: I don't think we want this to default to true - it junks up the console on dev.

robnagler: Whoops 6f1ddeed712cd546c6ade5d32ed14fdcb873db7d

rorour: Doesn't look like you pushed your changes @robnagler

robnagler: @rorour thanks I had a messed up git config and emacs didn't tell me...

rorour (approved):
Looks good to me

PR #6354, opened 2023-09-26
fix #6353 set EPICS_PVA_ADDR_LIST with broadcast port

### LINKED ISSUE #6353: EPICS_PVA_ADDR_LIST refers to the broadcast port
In the epicsllrf app, you can specify a host as "hostname" or "hostname:port". The port is actually the port used for the PVA udp broadcast port, not the server port. The default PVA broadcast port is 5076 unless specified on the server with EPICS_PVA_BROADCAST_PORT.

So our app should not be setting EPICS_PVA_SERVER_PORT, and should only set the EPICS_PVA_ADDR_LIST to "hostname" or "hostname:port".

moellep:


gurhar1133 (approved):
lgtm

PR #6359, opened 2023-09-27
fix #6334 destroy 3d content when minimized

### LINKED ISSUE #6334: Collapsed vtk reports not drawn if expanded after browser resize

moellep:
- separate radia 3d viewer into radiaViewer and radiaViewerContent so content can be destroyed when minimized
 - added view config to schema to mark 3d views which need special handling when minimized
 - destroy 3d content when panels are minimized


mkeilman:
Funnily, if you look through the commits in the other pr, I initially took roughly this approach with a common vtk view "superclass" ([2e27f72](https://github.com/radiasoft/sirepo/pull/6346/commits/2e27f7233a362252957e367d6ccc87cdb12b0795)). I rejected that approach because the vtk-viewer should determine its own behavior - you should not have to add flags and whatnot in a separate file. (ideally all the older code should use vtk-viewer instead of their own implementations, so we don't have to repeat code...)

I still think the state should be preserved when hidden, and if it requires new api so be it. It's not worth prolonging this fix though.

mkeilman (approved):
lgtm modulo my personal opinions

moellep:
One thing which could help radia would be moving the data out of the viewer directive into a service, similar to cloudmc volumeLoadingService and tallyService. That way moving between tabs or hiding/unhiding a panel wouldn't require a trip back to the server.

PR #6369, opened 2023-09-28
Fix #6367: add useful shell functions

### LINKED ISSUE #6367: add commands for getting mail messages and adding admin

e-carlin:


etc/home-env-funcs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+sirepo_add_admin() {
+    declare u
+    # POSIT: only uids in user and they don't have any specials
+    for u in $(ls ~/src/radiasoft/sirepo/run/user); do
```
robnagler, line 6: I think these functions should be sirepo_dev_add_admin.

`~/src/radiasoft` and `~/mail` should be variables.

etc/home-env-funcs.sh
```diff
@@ -0,0 +1,14 @@
...
+        sirepo roles add "$u" adm
+    done
+}
+
+sirepo_mail() {
+    # xargs trims whitespace
+    grep 'http://' $(ls -dt ~/mail/* | head -n1) | xargs
```
robnagler, line 13: I was curious if perl would be better, and I'm not sure it is:

```perl
perl <<'EOF'
use strict;
chdir($ENV{SIREPO_DEV_MAIL_DIR}) || die;
my($x) = [sort({$b <=> $a} grep(!/\D/, glob('*')))];
if (@$x) {
    open(IN, $x->[0]) || die;
    print((map(m{(http://\S+auth\S+)} ? $1 : (), <IN>))[0], "\n");
}
EOF
```

The shell pipeline definitely keeps things simple. I don't like the xargs comment though. xargs implicitly uses echo as the program. Maybe something like:

```sh
perl -n -e 'm{(http://\S+)} && print("$1\n")' $(ls -dt ~/mail/* | head -n1)
```

You could do the same thing with `sed`.

robnagler (changes requested):

etc/home-env-funcs.sh
e-carlin: 9cf9e4cee

etc/home-env-funcs.sh
e-carlin: Definitely not better to my untrained perl eyes:)

The second option is fine.

9cf9e4cee

etc/home-env-funcs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+_sirepo_dev_mail=~/mail
```
robnagler, line 3: I think these should be globally available and overridable so that we can test the infrastructure:

`export SIREPO_DEV_MAIL_DIR=${SIREPO_DEV_MAIL_DIR:-$HOME/mail}`

Use `$HOME` so it works inside shifter.

etc/home-env-funcs.sh
```diff
@@ -0,0 +1,16 @@
...
+sirepo_dev_add_admin() {
+    declare u
+    # POSIT: only uids in user and they don't have any specials
+    for u in $(ls "$_sirepo_dev_src"/run/user); do
+        sirepo roles add "$u" adm
+    done
+}
+
+sirepo_dev_mail() {
+    perl -n -e 'm{(http://\S+)} && print("$1\n")' $(ls -dt "$_sirepo_dev_mail"/* | head -n1)'"")'
```
robnagler, line 15: Spurious `'"")'`

robnagler (changes requested):
Accommodate https://github.com/biviosoftware/home-env/pull/88#issuecomment-1741402069

robnagler (approved):

PR #6382, opened 2023-10-02
fix #6380 allow for named tmp dir

### LINKED ISSUE #6380: allow for a named tmp_dir (for cached data)
In cloudmc, the static material files need a cache location. Currently this is lib(), but because the files are only transient, they should be kept somewhere in a known place in simulation_db.tmp_dir(). That way the files could be shared across simulations, but removed with normal tmp_file cleanup if space is low.

moellep:


moellep:
@robnagler please take a look - I'm not sure if this should be a call to tmp_dir() or a new api, like cache_dir(). 

cloudmc needs a place for large, static material files which get downloaded and then cached for use in multiple simulations. The cache is per-user for now.

sirepo/simulation_db.py
```diff
@@ -771,31 +771,34 @@ def static_libs():
 
 
 @contextlib.contextmanager
-def tmp_dir(chdir=False, qcall=None):
+def tmp_dir(chdir=False, qcall=None, dirname=None):
```
robnagler, line 774: I think this should be a separate call. It doesn't need to be a contextlib because it isn't going to go away. It could be a lot simpler and clearer to use that way.

robnagler (changes requested):

sirepo/simulation_db.py
moellep: OK - I've added it as a new api: simulation_db.cache_dir(). Please take a look.

sirepo/simulation_db.py
```diff
@@ -114,6 +114,11 @@ def assert_sim_db_file_path(basename):
     return basename
 
 
+def cache_dir(dirname, qcall=None):
+    """Returns a named directory within the user's temporary filespace"""
```
robnagler, line 118: Names and doc:

```
def cache_dir(basename, qcall):
"""Creates a directory in the user's temporary directory.

Unlike `tmp_dir`, this (possibly new) directory will not be deleted automatically.

Args:
    basename (str): sub-directory to use, need not be unique
    qcall (sirepo.quest.API): request state
Returns:
    py.path: path to newly created directory
"""
```

Note that for new APIs, qcall has to be supplied, i.e. this has to be on the api server or the supervisor.

robnagler (changes requested):

robnagler:
This is not going to work as written. The files have to be visible via sim_data or sim_db_file. The assumption in this PR is that they are available via a shared file system.

I also don't think these files should be in tmp. I have a feeling these are persistent to some degree, that is, they need to survive multiple runs (from the issue). I don't think cache is correct either, which implies they are stored elsewhere. We might call it scratch. 

Where do these files come from?

moellep:
The files are only present on the back-end (simulation execution only), and not needed on the front-end or client. It is the place that cloudmc would cache all the material files needed to run the simulations. The material files are large, static, and hundreds could be required to run a simulation. No qcall is needed because a qcall isn't available - I think it gets the user id from an environment variable.

robnagler:
> No qcall is needed because a qcall isn't available - I think it gets the user id from an environment variable.

This is somewhat irrelevant. What we need to know is the file's life span.

moellep:
They could go in lib - which is where they currently get put on sirepo.com, but the files aren't known ahead of time - only when the simulation runs and generates the xml files are the material file names available. On nersc, we would could keep no cache, and always download from github. Either way it needs to come from our server or github, unless there is a way to cache user files on nersc across runs.

robnagler:
Maybe it is ok the way it is. On NERSC, we'll download from github automatically, and it will cache correctly since the user dir structure is the same (the root is different but that's managed automatically). 

I do have a general concern about using simulation_db for this. These files aren't related to the db, that is, they are only related to the local file system, and never stored in the "db". Perhaps we can move this sirepo.sim_run.tmp_dir and .cache_dir. It could use simulation_db.user_dir as the implementation, but it doesn't have to.

moellep:
@robnagler please have another look - I've moved the tmp_dir() and cache_dir() methods to a new sirepo.sim_run module, and updated the callers to simulation_db.tmp_dir()

sirepo/simulation_db.py
```diff
@@ -526,6 +523,33 @@ def read_simulation_json(sim_type, sid, qcall):
     return d
 
 
+def random_id(parent_dir, simulation_type=None):
```
robnagler, line 526: `mkdir_random` is a better name now that it is a public method.

sirepo/sim_data/__init__.py
```diff
@@ -66,10 +66,11 @@ def audit_proprietary_lib_files(qcall, force=False, sim_types=None):
       sim_types (set): Set of sim_types to audit (proprietary_sim_types if None)
     """
     from sirepo import simulation_db
+    import sirepo.sim_run
```
robnagler, line 69: imports like this don't work right. Do this:

from sirepo import simulation_db, sim_run

robnagler (dismissed):

sirepo/simulation_db.py
moellep: changed with 4aaf2e6

sirepo/sim_data/__init__.py
moellep: fixed with 4aaf2e6

It looks like there are many places in the code which do an import from a method in the bad format:
$ trgrep "   import sirepo"

robnagler: I'm not sure this is still a problem but it was a problem at one point. After some testing it seems to import `sirepo` in the local namespace, which is the right thing to do.

PR #6393, opened 2023-10-04
Fix #6315 simulation search

### LINKED ISSUE #6315: Add search field to the simulations page
If an app has > N sims, there should be a search field which shows all sims in a list which match the search name value.

mkeilman:
Notes:
- This adds a search element at the top left of the simulations page
- The element is a directive wrapping the jquery ui [autocomplete](https://api.jqueryui.com/autocomplete/) widget. There is a native html5 [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) element but it does not seem to play well with angular; the widget also looks better
- The widget pops up a list of matching sim names (with folders prepended) as the user types
- When the user selects a sim from the popup list, they get redirected to that sim


moellep:
This works well, Mike. I made a few UI style changes in 18699a5, and if it looks OK, we can merge this.

PR #6405, opened 2023-10-09
Fix #6395 time.sleep documented

### LINKED ISSUE #6395: time.sleep => asyncio.sleep
There are a few places where an asyncio.sleep is more appropriate than a time.sleep in sirepo. One is srw.import_file but there may be others. Review all and document if time.sleep is appropriate.

robnagler:


e-carlin (commented):
For all of the "not in coroutine" why not assert that in code? Seems like that check would be nice to abstract in pkasyncio.

robnagler:
> For all of the "not in coroutine" why not assert that in code?

This is actually [a tricky thing](https://discuss.python.org/t/is-it-possible-to-detect-the-thread-an-event-loop-is-running-in/2892). You might get the event loop, but it is not running in the current thread. Not that we want to write code like this, but it might happen within a framework (Mantid uses threads, for example).  

The cost of simply checking if you are running in an event loop is rather high (exceptions are not cheap). 

I think we should leave this as documentation right now. The purpose of this PR was to audit all time.sleeps. We should avoid them in general. job_cmd should (and will) become async so we can use websockets to communicate with the agent and avoid message copying.

robnagler:
@rorour

rorour (approved):
looks good to me

PR #6407, opened 2023-10-09
Fix #6399 job_supervisor on_message async

### LINKED ISSUE #6399: WebSocketHandler.on_message is synchronous
https://github.com/tornadoweb/tornado/blob/master/docs/releases/v4.5.0.rst#tornadowebsocket says:

> WebSocketHandler.on_message implementations may now be coroutines. New messages will not be processed until the previous on_message coroutine has finished.

https://github.com/tornadoweb/tornado/issues/2532 discusses this. We have to do something like:

> asyncio.ensure_future(self.on_message_async(m_bytes))

robnagler:


sirepo/pkcli/job_supervisor.py
```diff
@@ -96,7 +96,8 @@ def on_close(self):
             pkdlog("error={} {}", e, pkdexc())
 
     async def on_message(self, msg):
-        await _incoming(msg, self)
+        # WebSocketHandler only allows one on_message at a time.
+        asyncio.ensure_future(_incoming(msg, self))
```
e-carlin, line 100: Elsewhere I think we use create_task which I believe is "preferred" https://github.com/python/asyncio/issues/477#issuecomment-268709555. Maybe we abstract in pkasyncio so we always use the same thing? 

Can the async be removed from the def of on_message? Functions with no awaits but marked async look strange to me...

e-carlin (changes requested):

sirepo/pkcli/job_supervisor.py
robnagler: 650825be5. Funny. I read [this](https://github.com/xxleyi/learning_list/issues/120) incorrectly.

robnagler:
@rorour @e-carlin

rorour (approved):
Looks good to me

e-carlin (approved):
What is the impact you are hoping to see from this code? I don't think it is going to change behavior in a meaningful way.

From my reading,  [job_driver.receive](https://github.com/radiasoft/sirepo/blob/master/sirepo/job_driver/__init__.py#L54) is only async in name. It doesn't actually call any functions that await. So, this change won't really make any difference (the code will block regardless). It may be clearer to get rid of async and get rid of create_task so it is clearer to the reader that on_message blocks until whatever it is doing is done.

I suppose that some messages could be handled before the task created by create_task is run. But once that thing runs it is going to run to completion and block.

e-carlin:
The fact that awaits in on_message don't yield is confusing to me. I'm curious how they implemented something like that...

robnagler:
> The fact that awaits in on_message don't yield is confusing to me. I'm curious how they implemented something like that...

All of tornado has to deal with this, because they allow `@gen.coroutine` and `async`, which are not identical. They introspect where they need to. They look at the result of the function call (which is a future or task) and execute it accordingly. [uri_router has to handle this](https://github.com/radiasoft/sirepo/blob/359f584282c32ea3df7c3cbffffea687d0225d1d/sirepo/uri_router.py#L439).

PR #6409, opened 2023-10-09
fix #6303 add accuracy tab to srw beamline wavefront editor

### LINKED ISSUE #6216: remove report retry from client
In sirepo-plotting.js, there is some old code which retries failed requests. Instead, it should display an error message from the server with no retry. It shouldn't show "server error: incomplete request" as it is currently doing.

https://github.com/radiasoft/sirepo/blob/master/sirepo/package_data/static/js/sirepo-plotting.js#L469

### LINKED ISSUE #6303: Add accuracy tab to Initial Wavefront Simulation Grid panel
There is no way to set the "auto-wiggler" method on the SRW beamline initial wavefront. This can be set from the Intensity panel on the source tab, but is unavailable on the Initial Wavefront Simulation Grid. This would be useful for simulating wiggler beamilnes.

x

moellep:
Also fix #6216 removed forceRun from client


sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -437,13 +437,7 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
 
     function initPlot(scope) {
         var interval = null;
-        var requestData = function(forceRunCount) {
```
robnagler, line 440: Looks fine to me. I don't remember why we had to do this...

robnagler (approved):

robnagler:
I'll let you merge

sirepo/package_data/static/js/sirepo-plotting.js
moellep: I think it was a work-around for apps which were trying to plot partially-created datafiles. We've gotten better at handling this on the server, so the client doesn't need the retry feature.

PR #6410, opened 2023-10-09
Fix #6352 remove flask and uwsgi

robnagler:
- Removed db upgrades from 2022 and before


setup.py
```diff
@@ -7,50 +7,44 @@
...
+        "pyIsEmail",
+        "pykern",
+        "pytest-asyncio",
+        "pytz",
+        "requests",
+        "tornado",
+        "user-agents",
+        # only needed for testing (srunit)
+        "websockets",
+        "werkzeug==2.0.3",
```
e-carlin, line 33: Can this be removed too?

sirepo/db_upgrade.py
```diff
@@ -63,143 +66,11 @@ def _20210211_add_flash_proprietary_lib_files(qcall, force=False):
             # Add's the flash proprietary lib files (unpacks flash.tar.gz)
             sirepo.sim_data.audit_proprietary_lib_files(
                 qcall=qcall,
-                force=force,
+                force=True,
                 sim_types=set(("flash",)),
             )
 
 
-def _20210211_upgrade_runner_to_job_db(qcall):
```
mkeilman, line 71: Out of curiosity, what governs (or should govern) when db upgrades are obsolete and can be removed?

mkeilman (approved):
lgtm - and satisfying to ditch 500 lines of code

sirepo/db_upgrade.py
robnagler: This was a judgment call. We don't have any installations using the old runner, that's for sure. The spa session probably is the only one that's a "problem", but it is a delete, not an add, so not critical if it doesn't get done. Flash is proprietary and was technically a duplicate (see the code) so seemed like only one upgrade was necessary.

setup.py
robnagler: Yup 9b57b10df

PR #6422, opened 2023-10-12
fix #6413 epics schema is configurable per simulation instance

### LINKED ISSUE #6413: update epicsllrf to have a configurable pv set
The sirepo epicsllrf app needs to be dynamic and allow changing the schema to match a target control system. The current app's schema matches the simulator two_cav.yml example, but should allow the pv def for a totally different set of controls.

moellep:
Move specific epics configuration into a separate lib file which gets merged with the simulation instance schema.


moellep:
@gurhar1133 please take a look at this. The application now keeps all of the epics-specific PV info in a separate lib file. I created 2 examples, simple and two_cav which run against the two different LLRFSim examples.

gurhar1133 (approved):
looks good

PR #6459, opened 2023-10-20
fix radiasoft/raydata#126: use conda env for chx

rorour:


rorour:
I'm getting `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` when the papermill command in the script runs, and I don't understand why

sirepo/raydata/scan_monitor.py
```diff
@@ -356,18 +356,16 @@ async def _process_analysis_queue():
...
                         for n in d.get_notebooks():
                             p = await asyncio.create_subprocess_exec(
-                                "papermill",
+                                cfg.papermill_script,
                                 str(n.input_f),
                                 str(n.output_f),
-                                *d.get_papermill_args(),
-                                stderr=asyncio.subprocess.STDOUT,
-                                stdout=l,
+                                str(asyncio.subprocess.STDOUT),
```
e-carlin, line 366: I think we want to continue to use the stdin and stdout kwargs that create_subprocess_exec has. That way all output from the subrocess is sent to it. In the way the code is written here only papermill logs will go to the log file. But, things like setting up the conda env (which we'll add) will go to the scan monitor logs. I think that could make debugging confusing and the scan_monitor doesn't need to know about these logs. Something like,
```python
with pkio.save_chdir(d.get_output_dir(), mkdir=True), pkio.open_text(
    "run.log", mode="w"
) as l:
    try:
        for n in d.get_notebooks():
            s = 'execute-analysis.sh' # TODO: def as constant above
            pkjinja.render_resource(_PAPERMILL_SCRIPT, PKDict(input_f=str(n.input_f), ...) output=d.join(s)
            p = await asyncio.create_subprocess_exec(
                s,
                stderr=asyncio.subprocess.STDOUT,
                stdout=l,
            )
```

sirepo/raydata/conda_papermill.sh
e-carlin, line 1: This should be a jinja template and we render it as a shell script with the args as `{{ arg }}` fields.

e-carlin (changes requested):
I didn't run the code but left some comments.

Once you've fixed them lmk and I'll try to debug the json error.

e-carlin:
Now that https://github.com/radiasoft/raydata/pull/129 is merged you can add setting of the conda env to the script for chx. I think that will entail adding the following to the top of the script:

```
export CONDA_PREFIX={{ conda_prefix }}
eval "$($CONDA_PREFIX/bin/conda "shell.bash" "hook" 2> /dev/null)"
conda activate '{{ chx_conda_env }}'
```

`conda_prefix`: This will be configuration. In dev it will be `/home/vagrant/miniconda`
`chx_conda_env`: This will be config too. In dev it will be analysis-2019-3.0.1-chx.

sirepo/raydata/conda_papermill.sh
rorour: 7c7866ddefbd0a7acf2fcf3cbe7a2d8845de81b3

sirepo/raydata/scan_monitor.py
rorour: 7c7866ddefbd0a7acf2fcf3cbe7a2d8845de81b3

sirepo/raydata/scan_monitor.py
```diff
@@ -513,6 +511,13 @@ def _init():
                 pkio.py_path,
                 "root directory for db",
             ),
+            papermill_script=(
```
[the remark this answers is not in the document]
rorour: 7c7866ddefbd0a7acf2fcf3cbe7a2d8845de81b3

rorour:
No longer seeing the json error

sirepo/package_data/raydata-execute-analysis.sh.jinja
e-carlin, line 1: I think we want to add some protection to this script `set -eou pipefail` but it is challenging because the conda activation produces error that can be ignored. Something like this I think should work:
```bash
# No set -eou pipefail before activating the conda env because the
# activation scripts have unbound variables and other errors that can
# be ignored.
if [[ chx == "chx"  ]]; then
    export CONDA_PREFIX=/home/vagrant/miniconda
    eval "$($CONDA_PREFIX/bin/conda "shell.bash" "hook" 2> /dev/null)"
    conda activate 'analysis-2019-3.0.1-chx'
    # Sanity check to make sure the activation was successful.
    if ! conda info --envs | grep -q -E 'analysis-2019-3.0.1-chx\s+\*\s+\/' ; then
        echo 'The active conda env is not what we expect' 2>&1
        conda info --envs
        exit 1
fi
set -eou pipefail
```

sirepo/raydata/scan_monitor.py
```diff
@@ -361,11 +364,21 @@ async def _process_analysis_queue():
...
                     try:
                         for n in d.get_notebooks():
+                            pkjinja.render_resource(
+                                _PAPERMILL_SCRIPT,
+                                PKDict(
+                                    input_f=n.input_f,
+                                    output_f=n.output_f,
+                                    papermill_args=" ".join(d.get_papermill_args()),
+                                    conda_prefix=cfg.conda_prefix,
+                                    chx_conda_env=cfg.chx_conda_env,
```
e-carlin, line 374: Rather than having variables specific to a beamline in this generic code we should move the construction of the script into the analysis_driver for the beamline. That was csx's script won't have any mention of chx and we won't have mention of chx in scan_monitory.py

e-carlin (changes requested):

sirepo/raydata/scan_monitor.py
rorour: > csx's script won't have any mention of chx

Are you thinking that each analysis driver should have its own jinja template? It seems like there will be a lot of repeated code if we start using a conda env for csx as well.

sirepo/package_data/raydata-execute-analysis.sh.jinja
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if [[ {{ catalog_name }} == "chx"  ]]; then
+    export CONDA_PREFIX={{ conda_prefix }}
+    eval "$($CONDA_PREFIX/bin/conda "shell.bash" "hook" 2> /dev/null)"
+    conda activate '{{ chx_conda_env }}'
+fi
+papermill {{ input_f }}  {{ output_f }}  {{ papermill_args }}
```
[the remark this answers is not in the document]
rorour: 175d5d5485f63530a6f576b180c4ca3eca5a3a19

sirepo/raydata/scan_monitor.py
```diff
@@ -361,11 +364,21 @@ async def _process_analysis_queue():
                 ) as l:
                     try:
                         for n in d.get_notebooks():
+                            pkjinja.render_resource(
+                                _PAPERMILL_SCRIPT,
+                                PKDict(
+                                    input_f=n.input_f,
+                                    output_f=n.output_f,
+                                    papermill_args=" ".join(d.get_papermill_args()),
```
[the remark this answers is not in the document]
rorour: 175d5d5485f63530a6f576b180c4ca3eca5a3a19

sirepo/package_data/raydata-execute-analysis.sh.jinja
rorour: 175d5d5485f63530a6f576b180c4ca3eca5a3a19

sirepo/raydata/scan_monitor.py
e-carlin: I am. I agree there will be repeated code. I think there are 2 options. Choose whichever you like more:
1. Separate out for now and we can create an abstraction when the time comes.
2. Create the abstraction now. There shouldn't be mention of words like "chx" in any "general" files (ex scan_monitor.py). Anything specific to a beamline should live in the analysis driver for the beamline (ex the name of the conda env for chx).

rorour: cd9819208e9555b3aef1a844e7a669af3047a86e

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -16,6 +17,10 @@
 
 _ANALYSIS_DRIVERS = PKDict()
 
+_CONDA_PREFIX = "/home/vagrant/miniconda"
```
e-carlin, line 20: This should be config. There is a non-zero chance we will end up mounting their conda prefix and we shouldn't have paths like /home/vagrant hardcoded.

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -78,6 +92,20 @@ def get_run_log(self):
...
+            _PAPERMILL_SCRIPT,
+            PKDict(
+                input_f=input_f,
+                output_f=output_f,
+                papermill_args=" ".join(self.get_papermill_args()),
+                conda_prefix=self.get_conda_prefix(),
+                conda_env=self.get_conda_env(),
+                catalog_name=self.catalog_name,
+            ),
+            output=_PAPERMILL_SCRIPT,
```
e-carlin, line 106: This should be `get_papermill_script_path` we shouldn't rely on the fact that the current working dir is the `output_dir` when we call this. Furthermore, this function should return the output path. That way when we call `render_papermill_script` in the scan_monitor it will tell the caller where the script was written to and we can use that instead of calling `d.get_papermill_script_path` again. I think using the return value ties the code together more and reduces chances for bugs.

sirepo/raydata/analysis_driver/chx.py
```diff
@@ -13,6 +13,9 @@
 
 
 class CHX(sirepo.raydata.analysis_driver.AnalysisDriverBase):
+    def get_conda_env(self):
+        return "analysis-2019-3.0.1-chx"
```
e-carlin, line 17: This name should be from config. Ex
```python
def get_conda_env(self):
  return _cfg.conda_env
```

It is likely that names like this will change as the system matures and environments are updated. Changing configuration is easier than changing code.

sirepo/raydata/scan_monitor.py
```diff
@@ -361,11 +361,13 @@ async def _process_analysis_queue():
...
+                                input_f=n.input_f,
+                                output_f=n.output_f,
+                            )
                             p = await asyncio.create_subprocess_exec(
-                                "papermill",
-                                str(n.input_f),
-                                str(n.output_f),
-                                *d.get_papermill_args(),
+                                "bash",
+                                d.get_papermill_script_path(),
```
e-carlin, line 370: As I noted above, `render_papermill_script` should return the path and that should be used here. You can just call to `render_papermill_script` here so there is no need for a temporary.

sirepo/package_data/raydata-execute-analysis.sh.jinja
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# No set -eou pipefail before activating the conda env because the
+# activation scripts have unbound variables and other errors that can
+# be ignored.
+if [[ -z "{{ conda_env }}" ]]; then
```
e-carlin, line 5: This doesn't work right. For csx it becomes `if [[ -z "None" ]]; then` which evaluates to true.

e-carlin (changes requested):

sirepo/package_data/raydata-execute-analysis.sh.jinja
rorour: 7407ed0e7849051a42a067015819229cd53c406b

sirepo/raydata/scan_monitor.py
rorour: 7407ed0e7849051a42a067015819229cd53c406b

sirepo/raydata/analysis_driver/chx.py
rorour: 7407ed0e7849051a42a067015819229cd53c406b

sirepo/raydata/analysis_driver/__init__.py
rorour: 7407ed0e7849051a42a067015819229cd53c406b

sirepo/raydata/analysis_driver/__init__.py
rorour: 7407ed0e7849051a42a067015819229cd53c406b

sirepo/package_data/raydata-execute-analysis.sh.jinja
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# No set -eou pipefail before activating the conda env because the
+# activation scripts have unbound variables and other errors that can
+# be ignored.
+if [[ -n "{{ conda_env }}" ]]; then
```
e-carlin, line 5: Now that https://github.com/radiasoft/raydata/pull/134 is merged let's enforce that everyone use conda by getting rid of this check.

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -78,6 +88,22 @@ def get_run_log(self):
...
+            _PAPERMILL_SCRIPT,
+            PKDict(
+                input_f=input_f,
+                output_f=output_f,
+                papermill_args=" ".join(self.get_papermill_args()),
+                conda_prefix=_cfg.conda_prefix,
+                conda_env=self.get_conda_env(),
+                catalog_name=self.catalog_name,
+            ),
+            output=self.get_papermill_script_path(),
```
e-carlin, line 102: Store self.get_papermill_script_path a local var, use here, and return at the end. It helps alert the reader that where the file is written to is the path we return.

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -25,6 +29,9 @@ def __init__(self, catalog_name, uid, *args, **kwargs):
     def get_analysis_pdf_paths(self):
         return pkio.walk_tree(self.get_output_dir(), r".*\.pdf$")
 
+    def get_conda_env(self):
+        return ""
```
e-carlin, line 33: This should be something like `raise NotImplementedError("children must implement this")`

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -105,3 +131,10 @@ def init(catalog_names):
             importlib.import_module(f"sirepo.raydata.analysis_driver.{n}"),
             n.upper(),
         )
+
+
+_cfg = pkconfig.init(
+    conda_prefix=pkconfig.RequiredUnlessDev(
```
e-carlin, line 137: I don't think this needs to be RequiredUnlessDev. Currently [we use this path in > dev](https://github.com/radiasoft/raydata/blob/da9a2ed3c038728319129d8ba2ef29442a764fec/container-conf/build.sh#L7).

There is potential that we'll end up using a conda_prefix mounted in the container but for now ~/miniconda is fine.

sirepo/raydata/analysis_driver/chx.py
```diff
@@ -46,9 +42,19 @@ def get_output_dir(self):
...
+            ["run_dose", False],
+            ["username", self._scan_metadata.get_start_field("user")],
+        ]
+
 
 _cfg = pkconfig.init(
     base_dir=pkconfig.Required(
         pkio.py_path, "base directory for notebooks and outputs"
     ),
+    conda_env=pkconfig.RequiredUnlessDev(
```
e-carlin, line 57: No need for RequiredUnlessDev. This is the name of the conda env at BNL.

sirepo/raydata/analysis_driver/csx.py
```diff
@@ -30,4 +33,7 @@ def get_output_dir(self):
     base_dir=pkconfig.Required(
         pkio.py_path, "base directory for notebooks and outputs"
     ),
+    conda_env=pkconfig.RequiredUnlessDev(
```
e-carlin, line 36: No RequiredUnlessDev

e-carlin (changes requested):

sirepo/package_data/raydata-execute-analysis.sh.jinja
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

sirepo/raydata/analysis_driver/__init__.py
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

sirepo/raydata/analysis_driver/__init__.py
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

sirepo/raydata/analysis_driver/__init__.py
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

sirepo/raydata/analysis_driver/chx.py
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

sirepo/raydata/analysis_driver/csx.py
rorour: a4175ab0127a5b2207443ee18cab9d384c401eb9

e-carlin (approved):

PR #6463, opened 2023-10-23
Fix #6454 cloudmc filter types

### LINKED ISSUE #6454: Allow one filter of a given type per tally
Once a filter type has been selected for a tally, remove it from the dropdown menu.

mkeilman:
Notes:
- `multiLevelEditor` for filters contains five instances of the type dropdown; but without modification, `showEnum()` would apply only to the first instance of `optionIndex`
- Accordingly, this has `showEnum()` apply to all the corresponding options in each instance
- `showEnum()` also now takes a parameter to address an optional index for a specific instance

### DIFF
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -1752,7 +1752,7 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
                                 $scope.model[$scope.field],
                                 type(),
                             );
-                        }                            
+                        }
                     }
                 }
                 updateEditor();
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -1719,7 +1719,8 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
             }
 
             function type(index) {
-                return index == null ? $scope.model[$scope.field]._type : $scope.model[`filter${index}`]._type;
+                const f = index == null ? $scope.field : `filter${index}`;
+                return $scope.model[f]._type;
             }
 
             function updateEditor() {
@@ -1727,6 +1728,7 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
                     return;
                 }
                 const inds = SIREPO.UTILS.SirepoUtils.indexArray(5, 1);
+                // can always select 'None'
                 const assignedTypes = inds.map(i => type(i)).filter(x => x !== NONE_TYPE);
                 // remove assigned types
                 ALL_TYPES.forEach(x => {
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -1699,9 +1699,12 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
           </div>
         `,
         controller: function($scope) {
+            const ALL_TYPES = SIREPO.APP_SCHEMA.enum.TallyFilter
+                .map(x => x[SIREPO.ENUM_INDEX_VALUE]);
+            const NONE_TYPE = 'None';
 
             function setView() {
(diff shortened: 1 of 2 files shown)


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -1715,8 +1718,26 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
...
-                return $scope.model[$scope.field]._type;
+            function type(index) {
+                const f = index == null ? $scope.field : `filter${index}`;
+                return $scope.model[f]._type;
+            }
+
+            function updateEditor() {
+                if ($scope.modelName !== 'filter') {
+                    return;
+                }
```
moellep, line 1729: Would it be possible to move the tally.filter specific code out of multiLevelEditor into a view for the tally model? That would keep multiLevelEditor as a general directive, and the filter parts could be contained separately.

Something like this:
```javascript
SIREPO.viewLogic('tallyView', function(panelState, cloudmcService, $scope) {

    function showAvailableFilters() {
        // update filter enum values
    }

    $scope.whenSelected = showAvailableFilters;
    $scope.watchFields = [
        new Array(SIREPO.APP_SCHEMA.constants.maxFilters).fill('').map((_, i) => `tally.filter${i + 1}`),
        showAvailableFilters,
    ];
});
```

mkeilman: I'll do that

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1842,10 +1842,19 @@ SIREPO.app.factory('panelState', function(appState, requestSender, simulationQue
...
+        }
+        // apply to all instances
+        const f = i => i % eType.length === optionIndex;
+        let opt = sel.find('option').filter(f);
+
         if (! opt) {
             // handle case where enum is displayed as a button group rather than a select
-            opt = $(fieldClass(model, field)).find('button')[optionIndex];
+            opt = sel.find('button').filter(f);
         }
```
moellep, line 1858: This part doesn't seem to work the enums displayed as a button group. For an example, run the warpvnd Two-Poles example, then run the Field Calculation on the source tab, and the open the Field Comparison Report dialog, it should hide one of the dimensions since the simulation is 2d:

x

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2048,6 +2049,40 @@ SIREPO.viewLogic('sourceView', function(appState, panelState, $scope) {
...
+        .map(x => x[SIREPO.ENUM_INDEX_VALUE]);
+    const inds = SIREPO.UTILS.indexArray(SIREPO.APP_SCHEMA.constants.maxFilters, 1);
+    const TYPE_NONE = 'None';
+
+    $scope.modelData = appState.models[$scope.modelName];
+
+    function type(index) {
+        return $scope.modelData[`filter${index}`]._type;
+    }
+
```
moellep, line 2064: Here, you don't want to have $scope.modelData, or set it to something when the directive is created. Otherwise it will always be set to the first model edited (in case there are more than one tally in the simulation).

Instead, type() should directly use appState to get the current model values:
```javascript
    function type(index) {
        return appState.models[`filter${index}`]._type;
    }
```

sirepo/package_data/static/js/sirepo.js
mkeilman: [f6a08cd](https://github.com/radiasoft/sirepo/pull/6463/commits/f6a08cdb770b66adc56450af7da774f6d0100a30)

sirepo/package_data/static/js/cloudmc.js
mkeilman: [0baffb8](https://github.com/radiasoft/sirepo/pull/6463/commits/0baffb8cc4966aeb6d6fea035cce4806ec19b0f5)

moellep (approved):

PR #6473, opened 2023-10-27
Fix #6243: omega related sims

### LINKED ISSUE #6243: visiting omega before using other apps will cause sim list call to fail
The omega app currently needs to access the sim_list of other apps, but if they have not been accessed previously (and had examples created), the simulation_list() call will fail due to an undefined qcall.

To reproduce, remove the elegant sim directory, run omega and select elegant as a sim type. The sim_list stateful compute will be called, but will fail.

Somehow a valid qcall needs to be created within the stateful_compute().

gurhar1133:


gurhar1133:
@moellep ping

moellep (approved):

PR #6474, opened 2023-10-29
Fix #6305: create service for allocating global resources.

### LINKED ISSUE #6305: global resource allocation
We need a way to specify global resources and have a service that can manage them.

For example, have a range of ip addresses and ports and dole them out to agents that require them.

e-carlin:
The viz3d app will run an instance of the risivz server and NVIDIA IndeX for each "simulation" a user starts.  Each of those services need a set of resources (ip addresses and ports) to be configured properly. In addition, the resources need to be shared between the services and GUI so the GUI can make requests to the services.

The resources need to be unique across users so we don't run into conflicts. The global resource system creates a mechanism for both the GUI and job agents to request resources.

The resources are allocated per simulation id. This means that within a simulation the GUI and the job agent have the same resources. This allows setting up things like having the GUI connect to NVIDIA IndeX running on a specific port. This also allows a user to have multiple independent simulations running so they can do things like view multiple datasets.

There are still some TODO's left to make the systme fully secure and performant.

In the short-term we need some form of security for the public ports exposed. This will be a middleware layer that takes a request to a port, performs auth checks, and then proxies to the correct service. I plan on investigating a service like configurable-http-proxy (what jupyterhub uses) or possibly traefik or haproxy. They allow dynamic api based configuration of a proxy. The request from the GUI may not go to a specific port but just to sirepo and then be proxied to a specific local port. TBD when I implement.

In the long-term we need a better mechanism for deciding what agents need what resources and when. Currently, if the supervisor is configured with the global resource system
(SIREPO_FEATURE_CONFIG_ENABLE_GLOBAL_RESOURCES=1) then all agents will have the ability to request resources.  There is state that must be stored for each agent to keep track of this. For agents that never request resources this is unnecessary. In addition, we need a way to garbage collect resources. When an agent is terminated its resources should be freed so they can be used by another agent. Currently the viz3d app will have few enough users so we can over-provision resources so that there are enough resources to go around for all possible users.


e-carlin:
@moellep this code is behind a feature flag (`SIREPO_FEATURE_CONFIG_ENABLE_GLOBAL_RESOURCES=1`) so I'm not too worried about it breaking Sirepo outside of viz3d.  But there was some refactoring to sim_db_file so please do a double check that I didn't break anything.

@gurhar1133 there is a companion pr in the rsiviz repo you can use to test the viz3d app.

gurhar1133 (approved):
lgtm and works with rsiviz

moellep (approved):

sirepo/job_supervisor.py
```diff
@@ -388,6 +390,13 @@ async def _receive_api_beginSession(self, req):
             c.destroy(cancel=False)
         return PKDict()
 
+    async def _receive_api_globalResources(self, req):
+        return sirepo.global_resources.for_simulation(
```
robnagler, line 394: pass in req.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2649,6 +2649,9 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, utilit
         );
     };
 
+    self.sendGlobalResources = function(appState, callback, data, errorCb) {
```
robnagler, line 2652: This doesn't read well. It should be requestGlobalResources?

tests/global_resources_test.py
```diff
@@ -0,0 +1,45 @@
...
+    a = _post(fc, "globalResources", d)
+    pkeq(["public_port"], list(a.keys()))
+    pkeq(_PUBLIC_PORT, a.public_port)
+    b = _post(fc, "globalResources", d)
+    pkeq(a, b)
+    c = _post(fc, "statelessCompute", d, method="global_resources")
+    pkeq(a.public_port, c.public_port)
+    e = {"public_port", "ip", "ports"}
+    a = set(c.keys())
+    pkok(e.issubset(a), "epect={} subset of actual={}", e, a)
```
robnagler, line 32: typo

tests/global_resources_test.py
```diff
@@ -0,0 +1,45 @@
...
+        SIREPO_GLOBAL_RESOURCES_PUBLIC_PORTS_MAX=f"{_PUBLIC_PORT + 1}",
+        SIREPO_GLOBAL_RESOURCES_PUBLIC_PORTS_MIN=f"{_PUBLIC_PORT}",
+    )
+
+
+def test_global_resources(fc):
+    from pykern.pkunit import pkeq, pkok
+
+    d = fc.sr_sim_data()
+    a = _post(fc, "globalResources", d)
```
robnagler, line 23: I think we should not send sim_data. Instead we should request global resources for an existing stype and sid. The data should be read from sim_db if need be.

robnagler (commented):
I'm concerned that these new concepts are not the right abstractions. I wrote some of my specific concerns. The phrase "global_resources" is odd. Rather, are these global to the server? To the simulation?

sirepo/feature_config.py
```diff
@@ -151,6 +152,11 @@ def _test(msg):
             set,
             "codes where all users are authorized by default but that authorization can be revoked",
         ),
+        enable_global_resources=(
```
robnagler, line 155: This name is not consistent. Remove enable_.

PR #6478, opened 2023-10-31
fix #5938 run openmc with sbatch

### LINKED ISSUE #5938: run openmc on nersc
openmc was tested on cori in the past outside of Sirepo, and should be able to run on perlmutter.

### LINKED ISSUE #6470: outline dimensions using 1st tally mesh only
The geometry outlines seem to be using the mesh dimensions from the 1st tally only, so if a second tally exists with a different mesh size, the outlines may fail.

For example, this sim: 

https://www.sirepo.com/cloudmc#/geometry/kAzhhTlm

The first tally has a dimension 50x50x1, and the second tally has dimensions 50x50x2

The 2d plot will fail when you try to view all the planes of the t2 tally.

moellep:
Also fixes #6470 compute outlines for all tallies

There is one TODO within job_agent to set a sbatch parameter different for cloudmc vs mpi. I need @robnagler 's help for that.


moellep:
@mkeilman could you review this PR where it fixes #6470 to make sure it doesn't break other things.

sirepo/package_data/static/js/srw.js
```diff
@@ -2548,9 +2548,9 @@ SIREPO.app.directive('simulationStatusPanel', function(appState, beamlineService
                 <div class="col-sm-12" data-simulation-status-timer="simState"></div>
                 <div data-ng-if="simState.showJobSettings()">
                   <div class="form-group form-group-sm">
-                    <div class="col-sm-12" data-model-field="\'jobRunMode\'" data-model-name="simState.model" data-label-size="6" data-field-size="6"></div>
-                    <div data-sbatch-options="simState"></div>
+                    <div data-model-field="\'jobRunMode\'" data-model-name="simState.model" data-label-size="6" data-field-size="6"></div>
```
mkeilman, line 2551: No need to escape the quotes inside a template literal - a bit easier to read without them I think

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -94,14 +94,15 @@ SIREPO.app.factory('cloudmcService', function(appState, panelState) {
     const self = {};
     appState.setAppService(self);
 
-    function findScore(tallies, tally, score) {
-        return findTally(tallies, tally).scores.filter(v => v.score == score).length
+    function findScore(score) {
+        return findTally().scores.filter(v => v.score == score).length
```
mkeilman, line 98: Triple equals recommended `(v => v.score === score)`

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -94,14 +94,15 @@ SIREPO.app.factory('cloudmcService', function(appState, panelState) {
...
             ? score
             : null;
     }
 
-    function findTally(tallies, tally) {
-        return tallies.filter(v => v.name == tally)[0];
+    function findTally() {
+        const a = appState.models.openmcAnimation;
+        return a.tallies.filter(v => v.name == a.tally)[0];
```
mkeilman, line 105: Triple equals recommended

sirepo/template/cloudmc.py
```diff
@@ -197,14 +197,117 @@ def stateless_compute_validate_material_name(data, **kwargs):
...
+        with pkio.open_text(run_dir.join(f"{_PREP_SBATCH_PREFIX}.log"), mode="w") as l:
+            subprocess.run(
+                ["python", f"{_PREP_SBATCH_PREFIX}.py"],
+                check=True,
+                stderr=l,
+                stdout=l,
+            )
+        pkio.write_text(
+            run_dir.join(template_common.PARAMETERS_PYTHON_FILE),
+            template_common.render_jinja(
```
mkeilman, line 214: I don't know whether this gets executed right away or waits for the subprocess. Does it depend on the results?

mkeilman (commented):
Looks like the fix for #6470 looks ok. Other than that, I have just some minor comments

sirepo/template/cloudmc.py
moellep: Yes, subprocess.run() waits for the command to complete before continuing.

sirepo/template/cloudmc.py
```diff
@@ -197,14 +197,117 @@ def stateless_compute_validate_material_name(data, **kwargs):
...
+
+    all_outlines = PKDict()
+    for tally_name, tally_mesh in _get_meshes():
+        tally_ranges = [_center_range(tally_mesh, i) for i in range(3)]
+        # don't include outlines of low resolution dimensions
+        skip_dimensions = PKDict(
+            x=len(tally_ranges[1]) < _MIN_RES or len(tally_ranges[2]) < _MIN_RES,
+            y=len(tally_ranges[0]) < _MIN_RES or len(tally_ranges[2]) < _MIN_RES,
+            z=len(tally_ranges[0]) < _MIN_RES or len(tally_ranges[1]) < _MIN_RES,
+        )
```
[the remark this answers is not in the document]
moellep: updated code to this which shows the relationship between dimension and axis a bit better:
```python
        skip_dimensions = PKDict(
            x=_is_skip_dimension(tally_ranges, 1, 2),
            y=_is_skip_dimension(tally_ranges, 0, 2),
            z=_is_skip_dimension(tally_ranges, 0, 1),
        )
```

moellep:
I address the comments, and also made some additions:
 - each app can control which sbatch and nersc fields are displayed, cloudmc only needs to show sbatchNodes, and nersc project and queue
 - geometry outlines are now included in the summaryData along with the tally results. Otherwise, the get outlines request could fail if the simulation was run on nersc, but the user was no longer logged in there when viewing plots

PR #6497, opened 2023-11-03
fix #6331 preserve zoom unless heatmap axes change

### LINKED ISSUE #6331: paging through heatplot frames should preserve zoom
When a heatplot is loaded, it should compare the axis metadata (label, ranges) and if it is unchanged, it shouldn't reset the axis.scale.domain() value. That way you can page through a plot zoomed into a particular spot and watch it evolve.

moellep:
- also hide 2d axes choices for low resolution dimensions

### DIFF
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -150,7 +150,7 @@ SIREPO.app.factory('cloudmcService', function(appState, panelState) {
     self.findTally = findTally;
 
     self.isGraveyard = volume => {
-        return volume.name && volume.name.toLowerCase() == 'graveyard';
+        return volume.name && volume.name.toLowerCase() === 'graveyard';
     };
 
     self.validateSelectedTally = () => {
@@ -212,7 +212,7 @@ SIREPO.app.controller('GeometryController', function (appState, cloudmcService, 
                 appState.saveChanges('volumes');
             }
         }
-        else if (data.state == 'missing' || data.state == 'canceled') {
+        else if (data.state === 'missing' || data.state === 'canceled') {
             if (self.isGeometrySelected()) {
                 processGeometry();
             }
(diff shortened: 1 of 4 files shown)


mkeilman:
Is there an example I can look at?

moellep:
Yes, if you run the Tokamak Reactor example and view the "t1" tally on the 2D viewer, and then zoom in and switch between frames - the zoom should stay the same across frames:

x

And then switch from the "t1" tally which is in the y plane, to the "t2" tally and notice the slice axis automatically changes to z plane.

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3456,11 +3456,24 @@ SIREPO.app.directive('heatmap', function(appState, layoutService, plotting, util
...
                 select('.sub-title').text(json.subtitle);
+                let c = false;
                 $.each(axes, function(dim, axis) {
-                    axis.values = plotting.linearlySpacedArray(json[dim + '_range'][0], json[dim + '_range'][1], json[dim + '_range'][2]);
+                    const r = axis.values && getRange(axis.values);
+                    axis.values = plotting.linearlySpacedArray(
+                        json[dim + '_range'][0],
+                        json[dim + '_range'][1],
+                        json[dim + '_range'][2],
+                    );
```
mkeilman, line 3466: More concisely: `axis.values = plotting.linearlySpacedArray(...json[dim + '_range'])`

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -793,12 +797,31 @@ SIREPO.app.directive('geometry2d', function(appState, cloudmcService, frameCache
...
+                            delete v[dim];
+                        }
+                    });
+                });
+                SIREPO.GEOMETRY.GeometryUtils.BASIS().forEach(dim => {
+                    const s = ! Object.keys(v).length || dim in v;
+                    panelState.showEnum('tallyReport', 'axis', dim, s);
+                    if (! s && appState.models.tallyReport.axis == dim) {
+                        appState.models.tallyReport.axis = Object.keys(v)[0];
+                    }
```
mkeilman, line 815: Triple equals ===

mkeilman (commented):
Just a couple of suggestions; otherwise good to go

PR #6501, opened 2023-11-08
Fix #5843: copy related omega sims for share

### LINKED ISSUE #5843: share omega sim should copy sub simulations
In the omega app, if you share a simulation link, it should also make a copy of all the referenced elegant/OPAL/genesis sims.

gurhar1133:


moellep (commented):
Two things I found during testing:
 - if one of the simulations in the original is an example "simulation.isExample: true", then that needs to be cleared in the copy, or else the "Delete" will incorrectly be shown as "Discard Changes to Example" in the copied related sim.
 - the lib files needed for the related sim also need to be copied. For example, if the first sim uses a source input file which the target user doesn't have, it needs to be copies over. See the call to lib_files_from_other_user() in server.py api_copyNonSessionSimulation()
 - update the simulation.folder value in all the copied related sims (not the original omega sim) to be "/omega" so it doesn't clutter the other folders with the copies

gurhar1133:
Another look @moellep

moellep:
Looks good!

moellep (approved):

gurhar1133:
@moellep I found a bug where the omega sim copy has two copies in the user who opened the link's omega/simulations dir. 

This happens because
`self._save_new_and_reply(req, data)` 
is called twice. Once before `if hasattr(req.template, "copy_related_sims"):` and once in that if block.

If I remove the first call and then do something like:
```python
if hasattr(req.template, "copy_related_sims"):
    return self._save_new_and_reply(
        req,
        req.template.copy_related_sims(data, qcall=self),
    )
# TODO(robnagler) does not work, supervisor needs to be notified to
# copy the simulation state.
# if hasattr(req.template, 'copy_related_files'):
#     req.template.copy_related_files(data, str(src), str(target))
return self._save_new_and_reply(req, data)
```
it fixes it. Is there any reason why `_save_new_and_reply(` would need to be called before 
```python
sirepo.sim_data.get_class(req.type).lib_files_from_other_user(
            data,
            simulation_db.lib_dir_from_sim_dir(src),
            qcall=self,
        )
```
?? 
If not, I can apply the fix and merge

moellep:
Yes, I think save_new_and_reply can be moved because none of the code depends on ```res```. Also, you can remove the #TODO code about copy_related_files() because that is no longer used anywhere.

sirepo/template/omega.py
```diff
@@ -126,6 +128,33 @@ def background_percent_complete(report, run_dir, is_running):
     )
 
 
+def copy_related_sims(data, qcall=None):
+    for index, sim_obj in enumerate(data.models.simWorkflow.coupledSims):
+        if sim_obj.simulationType and sim_obj.simulationId:
+            p = pkio.py_path(
+                simulation_db.find_global_simulation(
```
robnagler, line 135: You probably want checked=True.

gurhar1133: I have a related issue that I can add `checked=True`. Thank you for pointing out

PR #6509, opened 2023-11-09
Fix #6508: Add config to enable gpu's for docker agents

### LINKED ISSUE #6508: job_driver.docker needs to support gpus
Need to be able to set `--gpus=all`

e-carlin:


sirepo/job_driver/docker.py
```diff
@@ -93,6 +93,7 @@ def init_class(cls, job_supervisor):
                 bool,
                 "mount ~/.pyenv, ~/.local and ~/src for development",
             ),
+            enable_gpus=(False, bool, "apply --gpus=all"),
```
gurhar1133, line 96: `"apply '--gpus=all'"`? Clearer?

gurhar1133 (approved):
Left one comment (might not really matter). Approve assuming that tests pass

sirepo/job_driver/docker.py
robnagler: Make this identical to rsdockerspawner, which is what docker cli expects.

e-carlin: @robnagler can you elaborate on what you'd like to see? [This is what I see](https://github.com/radiasoft/rsdockerspawner/blob/6783877fa9bb526a9da06e3cd6e4700d80e0e618/rsdockerspawner/rsdockerspawner.py#L105) in rsdockerspawner. When you set `--gpus=all` on the command line (as is done here) it sets the DeviceRequests config that rsdockerspawner sets itself.

e-carlin: Thanks b0cbf7352

PR #6522, opened 2023-11-15
fix #6519 fixed bug in flash valid value parsing

### LINKED ISSUE #6519: flash valid values may wrap mid-value
The flash parser needs to handle valid value lists which wrap lines mid value, like this example:

```
op_fillFileType [STRING] ["NOT SPECIFIED"]
        Valid Values: "tops", "ionmix", "ionmix4", "propaceos", "opacplot", "NOT
        SPECIFIED"
```

moellep:
- also fixes a sim_db_file problem when running the LaserSlab example

### DIFF
--- a/sirepo/sim_data/flash.py
+++ b/sirepo/sim_data/flash.py
@@ -243,7 +243,21 @@ class SimData(sirepo.sim_data.SimDataBase):
         make_dir = cls.__run_setup(data, run_dir)
         s = cls.__create_schema(data, make_dir, run_dir)
         cls.__run_make(make_dir)
+        cls.__delete_flash_exe(data)
         cls.__put_sim_files(data, s.flashSchema.enum.SetupDatafiles, make_dir, run_dir)
+
+    @classmethod
+    def __delete_flash_exe(cls, data):
+        """Delete the flash executable if it exists
+
+        The executables are named like _FLASH_EXE_PREFIX_<unique-hash>.
+        They have a hash so we know when we need to recreate the executable
+        based on the user changing parameters.
+
+        sirepo.sim_db_file delete does a glob of the filename to delete. So, we
+        will delete any file that starts with _FLASH_EXE_PREFIX
+        """
+        cls.delete_sim_file(data.models.simulation.simulationId, cls._FLASH_EXE_PREFIX)
 
     @classmethod
     def __extract_problem_files_archive(cls, path):
@@ -275,11 +289,6 @@ class SimData(sirepo.sim_data.SimDataBase):
             .items()
         ):
             p = make_dir.join(c)
-            if b.startswith(cls._FLASH_EXE_PREFIX):
-                cls.delete_sim_file(
-                    data.models.simulation.simulationId,
-                    b.split(cls._FLASH_FILE_NAME_SEP)[0],
-                )
             cls.put_sim_file(data.models.simulation.simulationId, p, b)
             if p.check(link=1):
                 p.copy(run_dir.join(b))
--- a/sirepo/sim_data/flash.py
+++ b/sirepo/sim_data/flash.py
@@ -275,12 +275,11 @@ class SimData(sirepo.sim_data.SimDataBase):
             .items()
         ):
             p = make_dir.join(c)
-            # TODO(pjm): this call broke the LaserSlab example
-            #            files were possibly being deleted after the put occurred
-            # cls.delete_sim_file(
-            #     data.models.simulation.simulationId,
-            #     b.split(cls._FLASH_FILE_NAME_SEP)[0],
-            # )
+            if b.startswith(cls._FLASH_EXE_PREFIX):
+                cls.delete_sim_file(
+                    data.models.simulation.simulationId,
+                    b.split(cls._FLASH_FILE_NAME_SEP)[0],
+                )
             cls.put_sim_file(data.models.simulation.simulationId, p, b)
             if p.check(link=1):
                 p.copy(run_dir.join(b))
--- a/sirepo/sim_data/flash.py
+++ b/sirepo/sim_data/flash.py
@@ -275,10 +275,12 @@ class SimData(sirepo.sim_data.SimDataBase):
             .items()
         ):
             p = make_dir.join(c)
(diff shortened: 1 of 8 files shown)


sirepo/sim_data/flash.py
```diff
@@ -275,10 +275,12 @@ def __put_sim_files(cls, data, datafiles, make_dir, run_dir):
             .items()
         ):
             p = make_dir.join(c)
-            cls.delete_sim_file(
-                data.models.simulation.simulationId,
-                b.split(cls._FLASH_FILE_NAME_SEP)[0],
-            )
+            # TODO(pjm): this call broke the LaserSlab example
```
e-carlin, line 278: @moellep did you see this in dev? I can't recreate it. Wondering if it is an NFS issue on prod.

e-carlin (approved):
The example works for me. 

I think we should figure out what's going on with the delete but other than that this looks good.

sirepo/sim_data/flash.py
moellep: Yes, I was getting this repeatedly on dev. I originally thought the files weren't making it through the sim_db "put", but found that removing the delete and they arrived ok. I think somehow the delete and put calls were being processed out of order, so the delete was happening after the put. I don't know why the delete would be necessary at all as the put would be overwriting the files.

e-carlin: Delete is necessary to clean up the flash exe's. The exe's are named like  flash_exe_uniqueHash. The put file might not have the same uniqueHash (probably won't).  But, looking at the code I think we no longer need to have the hash. It looks like you changed the gui so if any fields are updated that require flash to be compiled again the user is forced to run the setupAnimation. Is that correct? If so, we can get rid of the hash and get rid of the delete.

moellep: I'm not sure if you refresh the browser if it still knows it has been modified, so that may be necessary. We could change the code so it only calls the delete on the flash exe (only if the hash has changed)?

moellep: This is actually a bug somewhere in delete sim_db_file. It looks like only a partial path is being received which causes multiple files to be deleted at once:
```
DELETE flash/KNeZLS2E/al
PUT flash/KNeZLS2E/al-imx-003.cn4
DELETE flash/KNeZLS2E/he
PUT flash/KNeZLS2E/he-imx-005.cn4
DELETE flash/KNeZLS2E/al
PUT flash/KNeZLS2E/al-imx-003.ses
DELETE flash/KNeZLS2E/he
PUT flash/KNeZLS2E/he-imx-005.ses
DELETE flash/KNeZLS2E/amr_runtime_parameters.tpl
PUT flash/KNeZLS2E/amr_runtime_parameters.tpl
DELETE flash/KNeZLS2E/flash_exe
PUT flash/KNeZLS2E/flash_exe-327872c1931944b92c097bedd3218e7a
```
Notice the file names with dashes are being truncated which causes anything that matches it to get deleted. This bug causes Laser Slab to fail when you try to run the simulation becuase some of the *.cn4 are missing.

moellep: I've updated the code so it only deletes the flash executable file. Otherwise files with "-" character would be processed incorrectly.

e-carlin: Aha. We truncate the filenames when we [send the delete request](https://github.com/radiasoft/sirepo/blob/359f584282c32ea3df7c3cbffffea687d0225d1d/sirepo/sim_data/flash.py#L280). And then do a glob [in the delete in the supervisor](https://github.com/radiasoft/sirepo/blob/359f584282c32ea3df7c3cbffffea687d0225d1d/sirepo/sim_db_file.py#L25). So, that behavior needs to be rethought.

Looking at the log you share we put `al-imx-003.cn4` and then call delete on `al*` before doing the put for `al-imx-003.ses`. So, `al-imx-003.cn4` is deleted even though we just put it and need it to run the sim. 

I need to rethink the code. At first glance I think we should only do the glob on the flash exe. It would be nice to not glob at all but we don't have a way of re-creating the hash for the flash_exe so I'm not sure how we'll handle that. Maybe we can hash the whole binary and send that info back and forth to decide if we need to do a recreate/delete.

Another idea is to do the deletes and then do the puts after. That way they won't conflict. That is probably the simplest option.

e-carlin: Sorry GH didn't update and I didn't see your message before sending mine. You solution sounds good.

sirepo/sim_data/flash.py
```diff
@@ -275,12 +275,11 @@ def __put_sim_files(cls, data, datafiles, make_dir, run_dir):
             .items()
         ):
             p = make_dir.join(c)
-            # TODO(pjm): this call broke the LaserSlab example
-            #            files were possibly being deleted after the put occurred
-            # cls.delete_sim_file(
-            #     data.models.simulation.simulationId,
-            #     b.split(cls._FLASH_FILE_NAME_SEP)[0],
-            # )
+            if b.startswith(cls._FLASH_EXE_PREFIX):
```
e-carlin, line 278: What about moving this outside of the loop and making it a named function (maybe even with a docstring)? It is a tricky case that we are trying to delete the exe with a glob that is added for us on the receiving end. I think that deserves a special callout. 


```
def _delete_flash_exe():
   """Delete the flash executable if it exists
    
    The executables are named like _FLASH_EXE_PREFIX_<unique-hash>.
    They have a hash so we know when we need to recreate the executable
    based on the user changing parameters.
    
   sirepo.sim_db_file delete does a glob of the filename to delete. So, we
   will delete any file that starts with _FLASH_EXE_PREFIX
   """"
   cls.delete_sim_file(data.models.simulation.simulationId, cls._FLASH_EXE_PREFIX)

e-carlin:
I left [a style comment](https://github.com/radiasoft/sirepo/pull/6522#discussion_r1397719664) but do with it what you will.

sirepo/sim_data/flash.py
moellep: Yes, sounds good - I made this change with 3ddf958

PR #6529, opened 2023-11-15
Fix #6524: generalize progressbar

### LINKED ISSUE #6524: Generalize progress bar
[related issue](https://github.com/radiasoft/rsiviz/pull/233)

gurhar1133:


sirepo/package_data/static/js/activait.js
```diff
@@ -42,9 +42,7 @@ SIREPO.app.config(() => {
           <input type="text" data-ng-model="model[field]" class="form-control" data-lpignore="true" />
           <span data-ng-show="model.dataOrigin === 'url'" style="font-style: italic; font-size: 80%;">{{ UTILS.formatToThousands(model.bytesLoaded, 1, true) || 0 }} of {{ UTILS.formatToThousands(model.contentLength, 1, true) || 0 }} {{ UTILS.orderOfMagnitude(model.contentLength, true).suffix }}B</span>
           <div class="sr-input-warning"></div>
-          <div data-ng-show="model.contentLength && ! model.bytesLoaded" class="progress">
-            <div class="progress-bar progress-bar-striped active" role="progressbar" aria-valuenow="100" aria-valuemin="0" aria-valuemax="100" style="width:100%"></div>
-          </div>
+          <div data-ng-show="model.contentLength && ! model.bytesLoaded" data-sim-state-progress-bar="" data-default-class="progress-bar-striped active"></div>
```
e-carlin, line 45: I don't love default-class. It is repeated in many places. Looks to me like the "special" cases where we don't pass default-class aren't all that special and instead we could simplify and make each case the same. 

If `visualization.simState.isInitializing()` then set `progress-bar-striped active` otherwise no class.

Do you agree?

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -1842,6 +1842,39 @@ SIREPO.app.directive('simplePanel', function(appState, panelState) {
...
+                    return $scope.defaultClass;
+                }
+                if ($scope.simState.isInitializing()) {
+                    return 'progress-bar-striped active';
+                }
+                return '';
+            };
+
+            $scope.percentComplete = () => {
+                if ($scope.simState && $scope.simState.getPercentComplete) {
```
e-carlin, line 1869: Is there a case where we have simState but no getPercentComplete function?

e-carlin (changes requested):
Left two stylistic comments but other than that looks good. Thanks for cleaning this up.

I've added @moellep as a reviewer. He should have the final say.

sirepo/package_data/static/js/sirepo-components.js
gurhar1133: When you open sirepo.com/srw#/simulations simState is defined, but getPercentComplete is not
Screenshot 2023-11-16 at 1 27 47 PM

e-carlin: Why is the progress bar being rendered on the simulation list page? That seems strange.

gurhar1133: The appFooter in srw.js has a data-download-status and data-download-status has a sim-state-progress-bar

e-carlin: Ah, thanks. `download-status` is a strange directive. It has to [fake out simState](https://github.com/radiasoft/sirepo/blob/19c2e6884f4b13a31b70ba15f90125f5fdcd4b17/sirepo/package_data/static/js/srw.js#L920) and then [sets it later](https://github.com/radiasoft/sirepo/blob/19c2e6884f4b13a31b70ba15f90125f5fdcd4b17/sirepo/package_data/static/js/sirepo-components.js#L3442). We should put the check for simState being valid inside of `download-status` that way we don't have to muddle all code below it with checks.

`<div data-ng-if="angular.equals(simState, {})" class="modal fade" id="sr-download-status" tabindex="-1" role="dialog">`

gurhar1133:
@e-carlin ping

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -1842,6 +1842,38 @@ SIREPO.app.directive('simplePanel', function(appState, panelState) {
...
 
+SIREPO.app.directive('simStateProgressBar', function(appState) {
+    return {
+        restrict: 'A',
+        scope: {
+            simState: '<',
+        },
+        template: `
+            <div class="progress">
+              <div class="progress-bar {{ class() }}" role="progressbar" aria-valuenow="{{ percentComplete() }}" aria-valuemin="0" aria-valuemax="100" data-ng-attr-style="width: {{ percentComplete() || 100 }}%"></div>
```
e-carlin, line 1853: You don't need the `|| 100`. percentComplete always returns a value.

e-carlin (approved):
One minor things but looks good.

PR #6533, opened 2023-11-16
Fix #6532: make chx input and output base_dir separately configurable

### LINKED ISSUE #6532: raydata: make input and output base dir configurable / get rid of RRes...
Raydata is going to write the notebooks out to it's own location that will be configured via the base_dir config. But, we are going to read notebooks in from a different base_dir location. So, we need to be able to configure them separately.

In addition, we aren't going to use the R prefix for RResults and RRespipelines any more. We are going to use the output base_dir configuration instead.

e-carlin:


rorour (approved):
LGTM

PR #6536, opened 2023-11-17
Fix #6525: save subsims on import

### LINKED ISSUE #6525: omega zip export

gurhar1133:
Importing an exported zip now calls copy_related_sims in same way link sharing does


moellep (changes requested):
This implementation works if the export is created and imported on the same system, but wouldn't work across systems (alpha to beta). The export file needs to collect all the related sims and keep them in the zip file, and when imported, it should create the sims, rather than expect the sim id to still be present on the system.

gurhar1133:
> This implementation works if the export is created and imported on the same system, but wouldn't work across systems (alpha to beta). The export file needs to collect all the related sims and keep them in the zip file, and when imported, it should create the sims, rather than expect the sim id to still be present on the system.

Do we have this problem for link sharing too?

moellep:
Link sharing works correctly because it is always on the same system.

gurhar1133:
another look @moellep ?

moellep (approved):

PR #6537, opened 2023-11-20
Fix #6528: migration to deploy latest flash tarball

### LINKED ISSUE #6528: create db_upgrade for new flash

e-carlin:


tests/db_upgrade_data/auth.txt
```diff
@@ -6,7 +6,7 @@ CREATE TABLE db_upgrade_t (
 	PRIMARY KEY (name)
 );
 INSERT INTO db_upgrade_t VALUES('_20230203_drop_spa_session','');
-INSERT INTO db_upgrade_t VALUES('_20230529_deploy_flash_update','');
+INSERT INTO db_upgrade_t VALUES('_20231120_deploy_flash_update','');
```
robnagler, line 9: Something is wrong with this approach, I think. This isn't actually a db upgrade, which is the purpose of the versions in db_upgrade_t. What we want is to promote the tarball in our normal release pipleline if the procedures change. Maybe I'm missing something.

robnagler (changes requested):

tests/db_upgrade_data/auth.txt
robnagler: Yea, I guess that makes sense. It's something we might be able to solve with a symlink instead of a copy of the tarball. This would involve mounting something else inside the containers which is complicated. Leave it as is.

robnagler (approved):

PR #6544, opened 2023-11-21
fix #6438 cache scan table data when navigating between tabs

### LINKED ISSUE #6434: Add scan table search text help
Some help text should be added to the search field explaining fields which are searched and depth of search. For example, mongo text search is on complete words, not partial matches.

### LINKED ISSUE #6438: Cache analysis list results between application tabs
Keep the current analysis list in browser memory, so when you switch to the Queued tab and then back to Run Analysis, you don’t need to wait for the list to reload. This can be done by adding an angularjs service to hold the data.

moellep:
- also fix #6434 add search help text

### DIFF
--- a/sirepo/package_data/static/js/raydata.js
+++ b/sirepo/package_data/static/js/raydata.js
@@ -35,7 +35,7 @@ SIREPO.app.config(() => {
     `;
 });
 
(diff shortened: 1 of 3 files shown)


rorour (approved):
Looks good

PR #6545, opened 2023-11-22
Fix 6541: only run flash db upgrade for users that have the flash role.

e-carlin:
Without this check unnecessary and incorrect work is performed: The flash dir is created for each user in the call to
simulation_db.simulation_lib_dir. Then is removed for users without the flash role in sirepo.sim_data.audit_proprietary_lib_files.


e-carlin:
@robnagler ping

sirepo/auth_db/user.py
```diff
@@ -84,13 +84,23 @@ def is_expired(self, role):
         return r.expiration < sirepo.srtime.utc_now()
 
     def uids_of_paid_users(self):
+        return self.uids_with_roles(sirepo.auth_role.PAID_USER_ROLES)
+
+    def uids_with_role(self, role):
```
robnagler, line 89: I don't think this is useful. If need be, you could check inside uids_with_roles in the for loop:

`for r in ((roles,) if isinstance(roles, str) else roles):`

I wouldn't bother. The user can always pass `[role]`.

sirepo/auth_db/user.py
```diff
@@ -84,13 +84,23 @@ def is_expired(self, role):
...
+        return self.uids_with_roles((role,))
+
+    def uids_with_roles(self, roles):
+        a = sirepo.auth_role.get_all()
```
robnagler, line 93: Gives more info and is shorter:

```py
a = sirepo.auth_role.get_all()
assert not (d := set(roles) - set(a)), f"roles={d} unknown all_roles={a}"
```

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+"""auth_db db upgrade
+
+:copyright: Copyright (c) 2023 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+
+
+_UID_WITH_FLASH_ROLE = "pzHuDps6"
```
robnagler, line 9: document that this needs to be in sync with committed auth.db.

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
...
+    """Test file changes for flash db upgrade.
+
+    1. User with flash role has the flash.rpm removed.
+    2. User w/o the flash role but has a flash dir is unchanged. This
+    is an uncommon case but test whether or not the flash upgrade only
+    happens to users with the flash role.
+
+    """
+    from pykern import pkio, pkunit
+    from pykern.pkunit import pkeq, pkre
```
robnagler, line 32: I've been tending towards pkunit.pkre lately. Not important.

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
...
+    from sirepo import srdb, srunit
+
+    def _check_paths(expect_flash_rpm):
+        r = pkio.sorted_glob(srdb.root().join("**", "flash.rpm"))
+        l = 1 if expect_flash_rpm else 0
+        pkeq(
+            l,
+            len(r),
+            "expect_flash_rpm={}; expecting {} paths found={}",
```
robnagler, line 41: `expect_flash_rpm={}; paths expect={} actual={}`

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
...
+            expect_flash_rpm,
+            l,
+            r,
+        )
+        if expect_flash_rpm:
+            pkre(rf"user/{_UID_WITH_FLASH_ROLE}", str(r[0]))
+        r = pkio.sorted_glob(srdb.root().join("user", "**", "flash"))
+        pkeq(2, len(r), "expecting 2 paths found={}", r)
+
+    pkio.unchecked_remove(srdb.root())
```
robnagler, line 51: Why is this remove necessary? Remove create from setup_module?

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
...
+        pkeq(
+            l,
+            len(r),
+            "expect_flash_rpm={}; expecting {} paths found={}",
+            expect_flash_rpm,
+            l,
+            r,
+        )
+        if expect_flash_rpm:
+            pkre(rf"user/{_UID_WITH_FLASH_ROLE}", str(r[0]))
```
robnagler, line 47: No need for `rf`, just `f`. `r` is only needed to escape literals, not interpolations.

tests/db_upgrade2_test.py
```diff
@@ -0,0 +1,57 @@
...
+        r = pkio.sorted_glob(srdb.root().join("user", "**", "flash"))
+        pkeq(2, len(r), "expecting 2 paths found={}", r)
+
+    pkio.unchecked_remove(srdb.root())
+    pkunit.data_dir().join("db").copy(srdb.root())
+    _check_paths(True)
+    # run db_upgrade
+    with srunit.quest_start():
+        pass
```
robnagler, line 56: I think the test should be inside the with, because the upgrade should be run after the `__enter__`.

robnagler (changes requested):

sirepo/auth_db/user.py
e-carlin: 8e7f059f1

tests/db_upgrade2_test.py
e-carlin: 8e7f059f1

tests/db_upgrade2_test.py
e-carlin: 8e7f059f1

tests/db_upgrade2_test.py
e-carlin: Not necessary. Was copy/paste bug from another test. 8e7f059f1

tests/db_upgrade2_test.py
e-carlin: 6deeb6e48

tests/db_upgrade2_test.py
e-carlin: 6deeb6e48

robnagler (approved):

PR #6550, opened 2023-11-29
fix #6506 run epicsllrf app against hardware board

### LINKED ISSUE #6506: Run epicsllrf app against hardward board
Some items which need to be addressed:
 - convert I and Q values to a complex number and compute magnitude and phase
 - create waveform generator
 - handle new array format for results from pvmonitor

moellep:


sirepo/package_data/template/epicsllrf/lib/epicsConfig-epicsSchema.simple.json
```diff
@@ -1,7 +1,7 @@
 {
     "constants": {
         "epicsModelPrefix": "LLRFSim",
```
gurhar1133, line 3: this file seems to be missing an `epicsModelName` that is resulting in an error in the simple example. The panel in the UI needs this field.

gurhar1133:
The two cav example works. The simple has a small bug. Left a comment about it.

gurhar1133:
@moellep made the mix to get simple example working again

PR #6551, opened 2023-11-29
Reinstate creation of output dir on prod

hyperrealist:
We removed the `mkdir` from prod in a previous [commit](https://github.com/radiasoft/sirepo/blob/a157442375d0f194c3b965a5a5787e87bdbbdf28/sirepo/package_data/raydata-execute-analysis.sh.jinja#L14), however while testing on CHX I ran into an issue where the output dir didn't exist because it is specific to the user and cycle. This is a proposal to reinstate the mkdir in prod to fix this issue.

### DIFF
--- a/sirepo/package_data/raydata-execute-analysis.sh.jinja
+++ b/sirepo/package_data/raydata-execute-analysis.sh.jinja
@@ -11,9 +11,6 @@ if ! conda info --envs | grep -q -E '{{ conda_env }}\s+\*\s+\/' ; then
     conda info --envs
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


e-carlin:
Thanks @hyperrealist! 

We use some common structures around branch names and commits. So, I ended up forking your branch to add them and opening a PR from that branch (https://github.com/radiasoft/sirepo/pull/6554).

We name branches with the issue name and then a short description (ex `6553-create-dir`). And we put `Fix #nnn: description` as the commit message.

We'll get your change in through that PR. Thanks again.

PR #6554, opened 2023-11-29
Fix #6553: create the output dir in all cases.

### LINKED ISSUE #6553: raydata: create output dir
Previously we thought the dir would always exist on prod. After some testing that proved to not be true. In dev, we need to create the dir because there is no outside infrastructure creating dirs. So, just create in all instances.

e-carlin:
After some testing we determined that the dir needs to be created in prod as well as dev. So, just create it in all cases.


rorour (approved):
Looks good to me

Issue #6231, opened 2023-08-15
2d options and layout ideas


moellep:
The openmc-plotter has a lot of good ideas for layout:

 - options in sub panels on the right margin
 - options to color the geometry only, by material, temperature, density
 - colormap min/max values
 - mask zero (no value)
 - tallies in log or linear scale
 - independent alpha settings for geometry and tally
 - zoom over info (material name, density, temperature, tally value)
 
and some awkward things we shouldn't use:
 - entry fields for x/y/z origin, width/height 

![x](https://github.com/radiasoft/sirepo/assets/6516910/8023ceb3-b0b4-48f2-b191-e7a862b3df90)

moellep:
Another useful option with be both an upper and lower threshold value. Currently we have only a lower threshold.

mkeilman:
I think the slice axis control works better as a dropdown than as buttons, which take up a lot of space and look off when only a subset is displayed

PR #6559, opened 2023-11-30
for #6231 add tally settings side panel

moellep:
Moved tally controls to a side panel which auto-updates on changes.

x


moellep:
Yes, I agree the formatting can be improved incrementally.

PR #6575, opened 2023-12-07
Fix #5807 radia user optimizable fields

### LINKED ISSUE #5807: Allow user to specify optimizable fields
Let users dynamically add optimizable fields in addition to those in the schema

mkeilman:
Quite a bit to digest here, but it breaks down into two main pieces:

Optimization:
- Any Float field of a "radiaObject" is considered to be a possible optimization parameter - the fields do not have to be marked or of a special type in the schema. FloatArray fields are not included but may be if time permits
- Only one built-in objective function is available. This should probably be replaced with a file upload with specific constraints - again possible with enough time
- Additionally only one rsopt optimization package is currently available
- Added a new dipole example with optimization ready to try

Scriptable fields:
- Basically a wrapper and extra UI around existing RPN code variables
- This does require typing of fields ("ScriptableField", "ScriptableArray") in the schema
- Adds a nice "code completion" to those fields so the user does not have to remember the names of objects
- Currently there is no guard against deleting objects that are referred to by other objects etc.


moellep:
It would be nice if the "Parameterized C-Bend Dipole" example already had some optimization parameters selected, so you could run a test optimization without having to pick some valid values initially.

moellep:
Sorry - I though the Parameterized C-Bend Dipole was the optimization example, I see it is the Optimizable Dipole example.

moellep:
The prev/play/next buttons can be hidden from the Optimization Path plot by adding  'optimizerAnimation' to the SIREPO.SINGLE_FRAME_ANIMATION value in app.config()

mkeilman:
> It would be nice if the "Parameterized C-Bend Dipole" example already had some optimization parameters selected, so you could run a test optimization without having to pick some valid values initially.

There is a new example for that purpose (optimizableDipole.json). Or did you mean in addition to that?

PR #6579, opened 2023-12-12
Fix #6132 cloudmc energy filter

### LINKED ISSUE #6132: Energy filter breaks plotting

mkeilman:
Notes:
- This adds a double-ended slider to specify the range of energies to sum over when an energy filter is in place
- The slider code ("JRange") duplicates a lot of the planePosition code. The latter is preserved to simplify the PR but they should be consolidated
- Of interest is the method of fetching the sim frame. Instead of adding a bunch of new parameters, the entire energy range object is sent as json. I don't know if this is a suitable solution but it simplified things for me


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2076,14 +2129,45 @@ SIREPO.viewLogic('tallyView', function(appState, panelState, $scope) {
...
+        const isValid = filter.start < filter.stop;
+        rangeFields.forEach(x => {
+            validationService.validateField(
+                'energyFilter',
+                x,
+                'input',
+                isValid,
+                'Start must be less than stop'
+            );
+        });
```
moellep, line 2152: Is there a way to test this error message in the UI? I didn't see a way to get it in an invalid state.

mkeilman: You should get the message by setting the start value to be greater than the stop value:
Screenshot 2023-12-18 at 10 11 45

moellep: Nice, I thought it was somehow associated with the JRange.

Issue #6134, opened 2023-07-19
file upload beyond size limit does not show an error


moellep:
Uploading a file which is beyond the size limit causes a 413 Request Entity Too Large, but the file upload modal remains showing "Please Wait..." forever. It should handle the error and show an appropriate message.

PR #6594, opened 2023-12-13
Fix #6555: upload file error

### LINKED ISSUE #6555: Sirepo upload file modal does not respond to errors
I tried uploading an 800Mb file which is larger than we allow in nginx. nginx responded with a 413 Request Entity Too Large but nothing changed in the modal after this. It continued to look like it was uploading

Screen Shot 2023-11-29 at 2 55 05 PM

In the case of dev you can trigger this by putting an `assert 0` in sirepo.server.api_uploadFile. 

The modal should respond to errors letting the user know there were problems.

gurhar1133:
max_message_bytes is passed to tornado which then closes the socket with reason = 'message too big' which comes through as the reason in the socket.onclose handler on the client. @moellep and I thought just closing the modal (since it cant be exited out of while uploading) and displaying error text might be a sufficient fix here

this addresses [this issue also](https://github.com/radiasoft/sirepo/issues/6134)

### DIFF
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -4699,13 +4699,11 @@ SIREPO.app.service('fileUpload', function(authState, msgRouter) {
                 callback(response.data);
             },
             function(response) {
-                if (response.status) {
-                    callback(
-                        {
-                            error: `File upload failed due to server error (status=${response.status || 'unknown'})`
-                        }
-                    );
-                }
+                callback(
+                    {
+                        error: `File upload failed due to server error (status=${response.status || 'unknown'})`
+                    }
+                );
             },
         );
     };
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -4693,9 +4693,11 @@ SIREPO.app.service('fileUpload', function(msgRouter, errorService) {
             headers: {'Content-Type': undefined}
         }).then(
             function(response) {
+                srdbg('onSuccess response:', response);
                 callback(response.data);
             },
-            function() {
+            function(response) {
+                srdbg('onRejected response:', response);
                 callback({error: 'File Upload Failed'});
             },
         );
--- a/sirepo/package_data/static/js/sirepo-components.js
+++ b/sirepo/package_data/static/js/sirepo-components.js
@@ -1256,7 +1256,7 @@ SIREPO.app.directive('columnEditor', function(appState) {
     };
 });
 
-SIREPO.app.directive('fileUploadDialog', function(appState, fileUpload, panelState, requestSender) {
+SIREPO.app.directive('fileUploadDialog', function(appState, errorService, fileUpload, panelState, requestSender) {
     return {
         restrict: 'A',
         scope: {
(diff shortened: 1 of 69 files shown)


e-carlin (changes requested):
Running rsiviz (removing max_message_bytes configuration in sirepo-dev) I uploaded an 800mb file and saw this 

Screen Shot 2023-12-15 at 9 29 42 AM

I don't know if that is something we can control but I've never seen it before and we should at least know what's going on. 

There is still the problem I outlined in the issue. If any of the code inside of sirepo.server.api_uploadFile raises and unhandled error we are still just stuck in the upload dialog with "please wait...". I think this code needs to be more robust against errors.

sirepo/reply.py
```diff
@@ -610,9 +610,9 @@ def _gen_exception_reply_WWWAuthenticate(self, args):
             401, headers=PKDict({"WWW-Authenticate": 'Basic realm="*"'})
         )
 
-    def _gen_http_exception(self, code, headers=None):
+    def _gen_http_exception(self, status, headers=None):
         return self.from_kwargs(
-            content=_HTTPException(PKDict(code=code, headers=headers))
+            content=_HTTPException(PKDict(status=status, code=status, headers=headers))
```
e-carlin, line 615: I think setting status here is at the wrong level of abstraction. In the http_response case we set the status in _http_error which is at a higher level of abstraction where we know about how code should be converted to status. In the websocket case http status doesn't even really make sense. There are no http status codes when you're sending websocket messages. So, this needs to be moved somewhere else or eliminated. 

To help me understand what path to take can you explain why we had to add status here? I remember doing it previously but now can't find the case where it is needed...

gurhar1133: I just changed that code back. It was a relic of a suggestion we discussed when we were trying to tease apart the problem.

From our slack convo:
" Its confusing who is "wrong" (server vs gui). Maybe we should set status in the server? Or maybe we should read code in the gui? Not sure. The server code is rather tangled with all of the different options of code paths so it is pretty hard to tell. I changed sirepo.reply to look like
```
    def _gen_http_exception(self, status, headers=None):
        return self.from_kwargs(
            content=_HTTPException(PKDict(status=status, code=status, headers=headers))
        )
```
Now it has both status and code in it."

The issue actually seems like it was on the client side. msgRouter in sirepo.js needed to change how it handles httpException. 
Before:
```python
if (n === "httpException") {
    r.status = content.status;
}
```
after:
```python
if (n === "httpException") {
     r.status = content.params.code;
}
```

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -4694,10 +4698,16 @@ SIREPO.app.service('fileUpload', function(msgRouter) {
             function(response) {
                 callback(response.data);
             },
-            function() {
-                //TODO(pjm): error handling
-                srlog('file upload failed');
-            });
+            function(response) {
+                if (response.status) {
```
e-carlin, line 4702: Don't need the if anymore

e-carlin (changes requested):

sirepo/reply.py
e-carlin: Ah, thanks.

e-carlin (approved):

PR #6607, opened 2023-12-18
Fix #4864 cloudmc source normalization

### LINKED ISSUE #4864: source term normalization
This is a place-holder for an issue with more info. Cloudmc was demonstrated to the DAGMC group, and one of their suggestions was adding source-term normalization (multiplying heatplot values by a constant?)

mkeilman:
This adds a field to normalize tally scores to a given particle count; that is if "number of particles per generation" is P and "normalize to" is N, the heatmap tally values are multiplied by (N / P).

Note that almost all scores are "per particle" - the exceptions known at this time are "inverse-velocity" and "decay-rate".


moellep (approved):
Looks good. I made one minor change with 16b3eab to handle the case where old results had been cached by the browser and num_particles was not present.

PR #6612, opened 2023-12-19
fix #6611 frame_arg enums may be treated as intergers

### LINKED ISSUE #6611: srw beamline plots broken
The sim_frame call in template.srw is failing with a re.sub() failure.

moellep:


sirepo/sim_data/__init__.py
```diff
@@ -720,7 +720,10 @@ def _frame_param_to_field(cls, param):
         from json.decoder import JSONDecodeError
 
         try:
-            return pkjson.load_any(param)
+            # floating point numbers will have lost the + sign when
+            # converted from a query string
+            # ex. 1e+40 would result in "1e 40"
+            return pkjson.load_any(re.sub(r"(\d)e (\d)", r"\1e+\2", param))
```
robnagler, line 726: This is problematic. The query args should be converted properly when used in a link.

moellep: The query is coming in with %2B which I think is being double-decoded from %2B to "+" to " ".

moellep: The double %2B decoding is happening here:

first:
https://github.com/radiasoft/sirepo/blob/master/sirepo/uri_router.py#L189

second:
https://github.com/radiasoft/sirepo/blob/master/sirepo/uri_router.py#L494

I don't understand why the "+" is being substituted in the uri there:
```python
parts = re.sub(r"\+", " ", path).split("/")
```

moellep: Apparently added by me in 2018 to fix a problem with findByName:

https://github.com/radiasoft/sirepo/commit/83b2ccba197faaed8019290836aa323aa55a7ff6

robnagler: I think we have to understand why errors like this occur, and fix them at the root. `+` is not a valid escape in the  path. They are only valid for [query](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) and [forms](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1) That substitution in the path is incorrect.

moellep: Yes, I think we originally had example links with "+" in the path. I checked now and the /light links do not have that. I've removed the "+" replacement and also the work-around code which was needed in sim_data for large floats (1e+40) with 5b6e2a9

PR #6621, opened 2023-12-21
fix #6308: raydata rename uid to rduid

### LINKED ISSUE #6308: 20231130 Raydata rename all uid to scan_uid
uid means something in Sirepo (users id). So, can be confusing when reading raydata code for uid to mean something else (the uid of the scan). I think renaming to scan_uid globally will fix this.

rorour:


moellep (approved):
This change seems to work well, but I think we need a upgrade to run on analysis_t to rename the uid column: alter table analysis_t rename column uid to rduid;

@e-carlin do you know where the appropriate spot for the upgrade to be added so it will automatically run at BNL?

moellep:
@e-carlin could you take a look at this PR - I'd like to get it approved before the other raydata PRs because it touches a lot of code and should go before other changes.

e-carlin:
Yes, I'll look first thing tomorrow.

e-carlin (changes requested):
Looks good except for the noted db upgrade. We don't have a place currently for upgrades. But, I think something like this would work.

```python
    # TODO(e-carlin): sort
    @classmethod
    def _db_upgrade(cls):
        def _fixup_running_statuses():
            for x in cls.session.query(cls).filter(
                cls.status.in_(_AnalysisStatus.NON_STOPPED)
            ):
                cls.set_scan_status(
                    PKDict(rduid=x.rduid, catalog_name=x.catalog_name),
                    _AnalysisStatus.ERROR,
                )

        def _rename_uid_column_to_rduid():
            if "rduid" in cls.__table__.columns.keys():
                return
            cls.session.execute(
                sqlalchemy.text(
                    f"ALTER TABLE {cls.__tablename__} RENAME COLUMN uid TO rduid"
                )
            )
            cls.session.commit()

        _rename_uid_column_to_rduid()
        _fixup_running_statuses()

    @classmethod
    def init(cls, db_file):
        global engine
        if engine is None:
            engine = sqlalchemy.create_engine(f"sqlite:///{db_file}")
        cls.metadata.create_all(bind=engine)
        cls.session = sqlalchemy.orm.Session(bind=engine)
        cls._db_upgrade()
```

sirepo/raydata/scan_monitor.py
```diff
@@ -171,6 +162,38 @@ def statuses_for_scans(cls, catalog_name, rduids):
...
+            for x in cls.session.query(cls).filter(
+                cls.status.in_(_AnalysisStatus.NON_STOPPED)
+            ):
+
+                cls.set_scan_status(
+                    PKDict(rduid=x.rduid, catalog_name=x.catalog_name),
+                    _AnalysisStatus.ERROR,
+                )
+
+        def _rduid_in_schema():
```
e-carlin, line 177: Is there a reason to prefer this over the code I shared? The code I shared I grabbed from how we are looking up columns elsewhere in sirepo https://github.com/radiasoft/sirepo/blob/c12aedc0d2d6b60186015dc88091baafb2698503/sirepo/auth_db/__init__.py#L41.

sirepo/raydata/scan_monitor.py
```diff
@@ -171,6 +162,38 @@ def statuses_for_scans(cls, catalog_name, rduids):
...
+            global engine
```
e-carlin, line 178: Doesn't need to be global. You only need to use global if you are assigning to it.

sirepo/raydata/scan_monitor.py
rorour: The code you shared will always be true because it checks the schema defined in the class. we need to check the actual db schema

sirepo/raydata/scan_monitor.py
rorour: 897e04f358c990b4061a172752be22c4121e4e69

sirepo/raydata/scan_monitor.py
e-carlin: Aha, thanks!

e-carlin (approved):

Issue #6191, opened 2023-08-02
raydata: column to add should be pulled from all visible scans


e-carlin:
Currently we select the columns based on the first scan https://github.com/radiasoft/sirepo/blob/a66d8b1337ec8b2d2fd4d07a6da8e9ce7ff6fc5e/sirepo/raydata/scan_monitor.py#L510 

We should show all possible columns for all scans on the page. Even if some scans don't have a specific column.

PR #6623, opened 2023-12-22
fix #6585 scan_id is searchable, added individual field search

### LINKED ISSUE #6420: Add FLASH to sirepo.lib.Importer
Making the FLASH parsers available would be nice for rsopt use on several projects.

### LINKED ISSUE #6581: search for specific field
Search should allow searching a particular field for a value. An additional input field could be added which allows selecting a column from a list and entering a search term for that column. Multiple specific search fields should be allowed.

### LINKED ISSUE #6585: scan_id isn't searchable in mongo
For CHX, it looks like searching on an excact scan_id is not returning results. If that field is not indexed by mongo text search, an additional query arg can be explicitly added (OR scan_id=XXX) if the search term seems like a scan_id (\d+).

moellep:
- fix #6581 and fix #6420 search for a specific field
- fix https://github.com/radiasoft/sirepo/issues/6191 column list refreshed from each scan result

### DIFF
--- a/sirepo/raydata/scan_monitor.py
+++ b/sirepo/raydata/scan_monitor.py
@@ -185,6 +185,34 @@ class _RequestHandler(_JsonPostRequestHandler):
     async def post(self):
         self.write(await self._incoming(PKDict(pkjson.load_any(self.request.body))))
 
+    def _build_query(self, text):
+        nums = []
+        terms = re.findall(r'(".*?")', text)
+        text = re.sub(r'".*?"', "", text)
+        for t in re.split(r"\s+", text.strip()):
+            if t.startswith("-"):
+                terms.append(t)
+            elif re.search(r"^\d+$", t):
+                nums.append(int(t))
+            elif t:
+                terms.append(f'"{t}"')
+        terms = " ".join(terms)
+
+        if len(nums) and len(terms):
+            return {
+                "$and": [
+                    databroker.queries.TextQuery(terms).query,
+                    {
+                        "scan_id": {"$in": nums},
+                    },
+                ],
+            }
+        elif len(nums):
+            return {
+                "scan_id": {"$in": nums},
+            }
+        return databroker.queries.TextQuery(terms).query
+
     def _databroker_search(self, req_data):
         def _search_params(req_data):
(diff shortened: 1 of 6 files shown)


moellep:
@e-carlin and @rorour please give this a test (searching, column selection, pdf download). I refactored the client quite a bit and could have introduced new bugs.

sirepo/raydata/scan_monitor.py
```diff
@@ -252,6 +309,13 @@ def _text_query(search_text):
             s.status = d.get(s.uid, _AnalysisStatus.NONE)
         return l, pc
 
+    def _is_float(self, value):
```
rorour, line 312: This name is a little misleading, I think `_is_float_or_float_string` or `_can_float` / `_can_cast_to_float` would be clearer. I saw the comment above where it's used but I think it's better to move that documentation into in the function name or docstring. I also think a more specific exception (ValueError) would be an improvement.

rorour (changes requested):
- "Search Text" is case insensitive but search text in individual fields are not
- The box for the "Select Search Field" dropdown is very wide. Not sure if it needs to be this wide to accommodate some long field names, but for the longest one I saw (OAV_resolution [um_pixel]) it was still almost twice as wide as it needs to be.
- Functionally, the "Add Column" dropdown is a big improvement, but visually, the button looked a lot cleaner. Making the text bold might look better, and maybe you could also make it less wide (tighter around the "Add Column" text) since the field names don't actually appear in that box when selected.

sirepo/raydata/scan_monitor.py
moellep: I've made some of the changes above with 65cd9d7
- layout changes to the search field selection and add column selector
- renamed _is_float() to _can_cast_to_float() and catch ValueError only

I couldn't easily fix the case sensitivity difference because the standard mongo $text indexed case insensitive while the individual field search can't be changed to case insensitive without changing it to use regular expressions, which I think might slow it down.

The "Add Column" is now bold, but I couldn't modify the margins too much because the "select" input field really wants to size to the maximum text value.

e-carlin (approved):

rorour (approved):

PR #6628, opened 2024-01-05
fix #6069: raydata logging data in the notebook should be viewable by users

### LINKED ISSUE #6069: 20231130 raydata: logging data in the notebook should be viewable by users
Users should be able to view any logging done in the analysis notebooks. Maybe it already goes to the run.log? A user specifically mentioned that they set the log level to warning and would like some way of seeing all of the warnings that were logged when running an analysis.

rorour:


e-carlin (changes requested):
I think there are a couple of problems with this approach:
1. I think not showing "Ending" is a matter of presentation to the user. On disk it would be nice to have the full log. That way if we are ever down in the weeds debugging the system we can go to the log file (not through the gui) and see exactly everything that happened. Then when presenting to the user we can make decisions to remove lines like "Ending". There is a cost to this approach. We will have to do processing each time we present to the user or de-normalize the log. But,  I think it is worth it to have the full log.
2. I think "Ending" is too common of a phrase to filter out every line that contains it. I think something like `^Ending Cell \d+-+$` would give us a better chance at not accidentally filtering out lines we want included. I think better to be too strict than too lax in a case like this.

sirepo/raydata/analysis_driver/__init__.py
```diff
@@ -75,6 +75,8 @@ def get_papermill_args(self):
         ]:
             res.extend(["-p", f"'{n}'", f"'{v}'"])
         res.append("--report-mode")
+        res.append("--log-output")
```
e-carlin, line 78: These make the output way better thank you:)

How about wrapping them up in a single call? I think it is a bit cleaner.

Like
```python
res.extend(('--report-mode', '--log-output', '--no-progress-bar'))
```

e-carlin (approved):

sirepo/raydata/analysis_driver/__init__.py
rorour: 49fa5cc88665cf01d9f4cd3fb82e446484e0f1bd

PR #6636, opened 2024-01-10
Fix #5198 import_file to stateful_compute and generalize sim_db_file

### LINKED ISSUE #5198: importFile to job_api and wrapped in stateleful compute
These operations will block the api gateway.

### LINKED ISSUE #6530: srw stateful_compute_model_list & delete_user_models write db
Not caught by assert_server_side, see #5254.  Necessary, because stateful_compute was changed in #5198 fix to be able to read libfileuri. 

These two operations need to be server side in sim_api.

### LINKED ISSUE #6639: $http service does not serialize FormData
Assumes the data consists of simple strings. fileUpload service could serialize, which would solve the problem for the specific use case [in elegant.js](https://github.com/radiasoft/sirepo/pull/6636/files#diff-3b333f8702dd2df9571306a3856b11af1df0e09253328c52453f889af83bdf80R1048).

### LINKED ISSUE #6647: job_cmd should not cache lib_file_list

### LINKED ISSUE #6666: Failed to retrieve remote data: name '_header_str_to_dict' is not defined. A report was sent to the server.
This is fixed #6636

### LINKED ISSUE #6668: template.activait.state*compute_get_archive_file_list unused
i couldn't find a reference to it.

### LINKED ISSUE #6687: srunit_servers terminate subprocess and then kill
The terminate doesn't always work. This is what's causing the MAX_CASE_RUN_SECS errors in GitHub actions. A fix will have to wait until #6636 is merged.

```diff
         for x in p:
             x.terminate()
-            x.wait()
+            try:
+                x.wait(timeout=2)
+            except subprocess.TimeoutExpired:
+                x.kill()
+                x.wait(timeout=2)
```

robnagler:
- stateful_compute correctly supports access to sim_db
- simplified and improved sirepo.importer
- removed lib_file_uri, since was not working correctly with fastcgi path through job_agent.
  cache was not getting cleared as in the regular case.
- sim_db_file 
    - supports put, get, exists, size, move, copy, etc. for any file in sim_db for
      a single user. More calls can be easily added.
    - sim_db_file restructured to be client/server in one file so cohesive API
    - sim_db_file server is now only path for interacting with db from the agent
- added new import and stateful compute tests
- sim_data
    - Fix #6647 lib_file_write replaces lib_file_write_path and uses sim_db_file client in agent
    - moved template_common.remote_file_to_simulation_lib to lib_file_save_uri
- agent_supervisor_api
    - added write_error to log exception
- javascript
    - requestSender.formatUrl adds simulation_type if not supplied
    - Fix #6668 remove cruft from activait js
- job_cmd
    - type check return values from dispatched routines
- simulation_db
    - moved sim_db_file_uri and sim_db_file_uri_to_path to sim_db_file
- template_common
    - better support in JobCmdFile
- tests
    - consolidated import tests
    - replace conftest.import_req with srunit.template_import_file
    - created srunit_servers to better support server fixture creations
    - added sim_db_file server start to allow direct testing of
      sim_db_file api so can test deviance of the server interface
      directly. Without this, all calls were structured through
      job_agent, and only conformance.
    - moved api_and_supervisor out of conftest
    - Fix #6687 terminate test server subprocess and then kill if process does not stop
- util
    - added yield_to_event_loop to document asyncio.sleep(0)
- activait
    - refactored _pyplot_data_url to be robust (call clf and cla) and modular
    - removed stateless_compute_get_remote_data and remote_data_bytes_loaded as not used
    - simplified sim_data DataReader stucture and fixed bug around use of keys() in constant
- elegant
    - elegant import_file returns `importState=needLattice` instead of
      creating and deleting simulation for ele file. This change makes
      imports stateless until final step.
    - elegant name of import is now the ele file, not the lte
    - Fix #6639 elegant eleData must be serialized to post
- flash
    - stateless_compute_get_remote_data uses lib_file_save_uri
    - renamed stateless_computes to stateful
    - TODO: stateful computes need to use sim_db_file #6675
- omega
    - refactored get_data_file
- radia
    - add sim_data.prepare_import_file_args to allow radia to override so it
    - radia.prepare_import_file_args loads lib file before importing; remove extra step on import
    - radia/geometryReport.py.jinja includes `_put_sim_files` instead of it being in header
    - import of zip `yield_to_event_loop` appropriately
    - Fix #6666 remove _header_str_to_dict, not needed
- srw
    - Fix #6530 all lib_file operations go through sim_db_file
- zgoubi
    - sim_data.zgoubi_lib_files_with_zip unused so removed


moellep (changes requested):
Testing with genesis, I find I can get the importer into an odd state with file existence. 

If I import a genesis file which requires a lib file, it will prompt, but later checks to see if it exists will always fail, even though it is present in genesis/lib.

Or if it does import successfully, and I delete the file and the sim, and then import the sim again, it doesn't prompt for the file even though it doesn't exist.

I think there is a cached file list somewhere which is getting checked rather that the actual available lib files (_cfg.lib_file_list).

Attached is an example genesis test case with a lib file.
[genesis-test-with-lib-files.zip](https://github.com/radiasoft/sirepo/files/13910145/genesis-test-with-lib-files.zip)

robnagler:
Yup https://github.com/radiasoft/sirepo/issues/6647

robnagler:
@moellep @e-carlin The [PR description](https://github.com/radiasoft/sirepo/pull/6636#issue-2073364730) summarizes all changes. There's a lot here. One thing I did not test is the flash changes, because I don't know it. I didn't change the methods, just renamed them.

e-carlin:
I'm testing flash and fixing a couple related bugs.

moellep:
Looks good - here is one bug I've found:

In omega, if you run a workflow, it results in the error:
```method=lib_file_write_path not available in job_agent```

To reproduce, open omega, create a new sim.
Select OPAL, CSR Bend Drift for #1
Select elegant, bunchComp - fourDipoleCSR for #2
Then select "Start New Simulation"

File: sirepo/sim_db_file.py [hunk 1]
```diff
@@ -1,48 +1,378 @@
...
+            pkdlog(
+                "uri={} body={} exception={} stack={}",
+                self.request.path,
+                self.request.body,
+                e,
+                pkdexc(),
+            )
+            try:
+                self.write({state: "error"})
+            except Exception:
```
e-carlin, line 182: Why wouldn't we want to blow up?

moellep:
This appears to be a timing issue - the user beam/undulator list in SRW doesn't seem to be available initially. When you open a SRW sim which is loading plots, and then open the electron beam and change the beam name, it shows the error: 
```
TypeError: Cannot read properties of undefined (reading 'some')
    at updateListFromModel (srw.js?2140:38)
```

To reproduce:
Open the NSLS-II CHX sim, select "discard changes to example", and while it is still loading plots, open the electron beam modal, change the Beam Name and select Save Changes.

Once the plots have loaded, the beam name can be edited normally. I couldn't reproduce this on alpha or prod, so possibly this is related to dev, or the priority of the model list request changed.

sirepo/template/activait.py
```diff
@@ -1184,18 +1162,13 @@ def _dice_coefficient(mask1, mask2):
                 d.append(_dice_coefficient(pair[0], pair[1]))
             return d
 
-        plt.figure(figsize=[10, 10])
+        f = plt.figure(figsize=[10, 10])
```
e-carlin, line 1165: f is unused.

sirepo/template/silas.py
```diff
@@ -45,10 +44,9 @@ def background_percent_complete(report, run_dir, is_running):
 
 def get_data_file(run_dir, model, frame, options):
     if model in ("tempProfileAnimation", "tempHeatMapAnimation"):
-        return PKDict(
-            tempProfileAnimation=_TEMP_PROFILE_FILE,
-            tempHeatMapAnimation=_TEMP_HEATMAP_FILE,
-        )[model]
+        return (
+            _TEMP_PROFILE_FILE if model == "tempProfileAnimation" else TEMP_HEATMAP_FILE
```
e-carlin, line 48: TEMP_HEATMAP_FILE missing leading underscore.

sirepo/uri_router.py
```diff
@@ -213,6 +213,28 @@ def sr_set_log_user(self, log_user):
 
     class _WebSocket(websocket.WebSocketHandler):
         async def on_message(self, msg):
+            import asyncio
```
e-carlin, line 216: Why? We import asyncio in sirepo.util which is imported everywhere so I can't imagine keeping the import local here is helpful.

tests/csv_import_test.py
```diff
@@ -7,10 +12,10 @@ def test_import():
 
     for d in pkunit.case_dirs(group_prefix="conformance"):
         for f in pkio.sorted_glob("*.csv"):
-            _ = sirepo.csv.read_as_number_list(f)
+            _ = csv.read_as_number_list(f)
```
e-carlin, line 15: Why bother with the underscore? And the pass below isn't needed.

e-carlin (approved):
Some minor comments but everything worked well.

robnagler:
> In omega, if you run a workflow, it results in the error:
> method=lib_file_write_path not available in job_agent

Fix 616f145 including a test for the opal-elegant workflow and a single genesis workflow. @moellep if you have a list of coupled simulations, I can add others. I don't know what the output of runSimulation should be except completed. Perhaps the test could generate a report to be more complete.

sirepo/sim_db_file.py
```diff
@@ -1,48 +1,378 @@
...
+
+        Args:
+            sim_type (str): which code
+            sid_or_lib (str): simulation id. if None or `_LIB_DIR`, is library file
+            basename (str): file name with extension
+
+        Returns:
+            str: uri to be passed to SimDbClient functions
+        """
+        from sirepo import simulation_db, template
```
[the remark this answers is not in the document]
robnagler: which problem?

sirepo/sim_db_file.py
robnagler: This is go-style. The None cascades to a place where something can do something about it. The error context is printed correctly (pkdlog logs the line), but the server returns 403.

tests/csv_import_test.py
robnagler: 82da885

sirepo/uri_router.py
robnagler: 82da885

sirepo/template/silas.py
robnagler: 82da885

sirepo/template/activait.py
robnagler: 82da885

sirepo/sim_db_file.py
```diff
@@ -1,48 +1,378 @@
...
-            raise sirepo.tornado.error_forbidden()
-        return sirepo.simulation_db.sim_db_file_uri_to_path(
-            path=m.group(1),
-            expect_uid=u,
-        )
+            return None
+
+    if len(uri) <= 0:
+        # no point in logging anything
+        return
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/sim_db_file.py
```diff
@@ -1,48 +1,378 @@
...
+
+class SimDbUri(str):
+    def __new__(cls, sim_type, slu, basename):
+
+        if isinstance(slu, cls):
+            if basename is not None:
+                raise AssertionError(
+                    f"basename={basename} must be none when uri={slu} supplied"
+                )
+            if sim_type != slu._stype:
```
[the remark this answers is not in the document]
robnagler: 82da885 and _stype (_sim_data) is only being accessed in `__new__`.

sirepo/sim_db_file.py
```diff
@@ -1,48 +1,378 @@
...
+    def __uri_arg_to_path(self, uri):
+        res = _uri_parse(uri, uid=self.__uid, is_arg_uri=True)
+        if res is None:
+            self.send_error(403)
+            return None
+        return res
+
+
+class SimDbUri(str):
+    def __new__(cls, sim_type, slu, basename):
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/sim_db_file.py
```diff
@@ -1,48 +1,378 @@
...
+
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/sim_data/__init__.py
```diff
@@ -611,8 +725,27 @@ def proprietary_code_lib_file_basenames(cls):
         return []
 
     @classmethod
-    def put_sim_file(cls, sim_id, file_path, basename):
-        return cls._put_sim_db_file(file_path, cls._sim_file_uri(sim_id, basename))
+    def put_sim_file(cls, sim_id, src_file_name, dst_basename):
+        """Write a file to the sim repo dir
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/package_data/static/js/sirepo.js
```diff
@@ -2414,18 +2414,23 @@ SIREPO.app.factory('requestSender', function(cookieService, errorService, utilit
         if (! map[n]) {
             throw new Error(`routeName=${n} not found in map=${map._name}`);
         }
-        var r = map[n];
-        var u = r.baseUri ? '/' + r.baseUri : '';
+        let r = map[n];
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/package_data/static/js/radia.js
```diff
@@ -1400,102 +1401,41 @@ SIREPO.app.directive('dmpImportDialog', function(appState, fileManager, fileUplo
...
         controller: function($scope) {
             $scope.inputFile = null;
-            $scope.fileURL = null;
             $scope.isMissingImportFile = function() {
                 return ! $scope.inputFile;
             };
             $scope.fileUploadError = '';
             $scope.isUploading = false;
-            $scope.importDmpFile = function(inputFile) {
+            $scope.importDmpFile = function (inputFile) {
```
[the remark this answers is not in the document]
robnagler: 82da885

sirepo/sim_db_file.py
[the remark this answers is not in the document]
e-carlin: That we have to be so careful with imports. Now there is another file (sim_db_file) where we have to remember to import in a specific way.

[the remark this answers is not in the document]
robnagler: sim_db_file doesn't need to be imported, which is why it is written that way. Template should not have to be a special import, actually. simulation_db is slowly getting refactored so that it doesn't have to be a special import either. Ideally, you would refer to it only via qcall, and in job agent, it would not be referred to at all (just util, const, etc.). 

So I don't think the problem is getting worse, but it is complicated to change interfaces over time as we are doing.

sirepo/sim_db_file.py
robnagler: > I didn't understand your response. I probably just don't understand this code...
>> The None cascades to a place where something can do something about it
> If there is a None being used in a meaningful way here please be explicit about it. I can't tell from looking at the code that there is a meaningful use of None.

I was thinking about different code when I wrote this. The reason for the pass here is that there's something wrong with tornado. Instead I logged the error 11793fdc3920d87ea957f47a361c3502f8e1111a. I don't like just dropping into tornado and hoping it handles the problem. I think we should understand all errors, and translate them to replies ourselves.

This particular code will be very hard to test.

>> The error context is printed correctly (pkdlog logs the line)
> If there is an error with write we don't log (as far as I can tell). That's what I'm curious about. Why would we want to swallow that error?

Apologies for the confusion.

robnagler:
[@moellep writes:](https://github.com/radiasoft/sirepo/pull/6636#issuecomment-1911013215)
> This appears to be a timing issue - the user beam/undulator list in SRW doesn't seem to be available initially. When you open a SRW sim which is loading plots, and then open the electron beam and change the beam name, it shows the error:

#6700 addresses this, not relevant to this PR.

moellep (approved):

Issue #6505, opened 2023-11-09
problem running cloudmc on nersc


moellep:
Running cloudmc on nersc encounters this error:

[Errno 2] No such file or directory: '/pscratch/.../cloudmc/...-...-openmcAnimation/run.log'

Looking in that directory, it looks like it has successfully setup the simulation and run the cloudmc "prep-sbatch.py" successfully, but there is no run.log, or log file with a stack trace of what was executing when the error occurred.

moellep:
@robnagler related to this, you could also review the merged PR which added cloudmc nersc support:
#6478

robnagler:
More data: It failed the first time. I ran the sbatch script that was created, and it ran fine. On the second try via the UI it worked. The only difference I could see in the sacct is:

```diff
134c134
<        "kill_request_user": "<user>",
---
>        "kill_request_user": "",
```

There's no cancel in the agent, sirepo, or supervisor logs. 

I ran it again after reducing the batches to 3 and particles to 1000. Then it failed with:

```txt
slurmstepd: error: *** JOB <snip> ON nid004208 CANCELLED AT 2023-11-13T23:43:16 ***
srun: error: Unable to create step for job <snip>: Job/step already completing or completed
```

There was a run.log, but the UI reported that there was none. However, I could download the run.log from the UI and it was there when I looked in the shell. 

[NERSC doc](https://docs.nersc.gov/jobs/troubleshooting/) says: 

```txt
Possible causes/remedies:
This may be caused by a system issue. Please report to NERSC.
```

I ran it again, which worked.

Not sure what's going on. Just reporting here as a checkpoint.

robnagler:
Submitted to NERSC. I can't see a pattern. The only odd one is that the run.log was not there when the job crashed but was there a few seconds later so that it could be downloaded. That may be a red herring. Let's see what NERSC says.

moellep:
It could be that the template.cloudmc module is trying to read the run.log before it is available, and when it fails it ends up canceling the whole simulation. I'll add check that verifies that the run.log is present before reading.

moellep:
7fddb21 appears to have fixed this on alpha and beta.

PR #6641, opened 2024-01-10
fix #6633 fix missing html type on delete button

### LINKED ISSUE #6633: set button type "button" to avoid enter-key trigger
In cloudmc, if you are editing materials and select the enter key to submit the form, the first delete button gets invoked instead. To avoid this for any non-submit button, set the button type="button":
ex.
```javascript
                        <button type="button" data-ng-click="deleteComponent($index)"
                          class="btn btn-danger btn-xs">
                          <span class="glyphicon glyphicon-remove"></span>
                        </button>
```

moellep:
Also for #6505 guard against missing run.log on nersc

### DIFF
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -1578,7 +1578,7 @@ SIREPO.app.directive('materialComponents', function(appState, panelState) {
                   <td>
                     <div class="sr-button-bar-parent pull-right">
                       <div class="sr-button-bar">
-                        <button data-ng-click="deleteComponent($index)"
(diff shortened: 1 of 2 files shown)


mkeilman (approved):
All good. I can't test the nersc bit, but do we need to restrict any of the file writing (`export_to_xml()` etc) to first rank?

moellep:
For nersc, the normal parameters.py and export_to_xml() get run during code generation, and the only part that actually runs in parallel is the openmc-sbatch.py template:
https://github.com/radiasoft/sirepo/blob/master/sirepo/package_data/template/cloudmc/openmc-sbatch.py.jinja

PR #6646, opened 2024-01-12
Fix #6638: default log parser

### LINKED ISSUE #6638: default log parser
We need a default log parser for better code sharing and maybe a default option

gurhar1133:
@e-carlin any thoughts you have on where I'm going with this would be appreciated


gurhar1133:
@e-carlin ping

sirepo/template/template_common.py
```diff
@@ -83,6 +85,54 @@ def __init__(self, **kwargs):
...
+class LogParser:
+    def __init__(
+        self,
+        run_dir,
+        log_filename,
+        error_patterns=_DEFAULT_ERROR_PATTERNS,
+        multiline_error_pattern=r"^Traceback .*?^\w*Error: (.*?)\n",
+        default_msg="An unknown error occurred",
+    ):
+        self.run_dir = run_dir
```
e-carlin, line 97: We tend to inherit from PKDict to avoid having this repetition:

```python
class LogParser(PKDict):
  def __init__(self, run_dir, log_filename**kwargs):
    super().__init__(run_dir=run_dir, log_filename=log_filename, **kwargs)
    self.pksetdefault(
      error_patterns=_DEFAULT_ERROR_PATTERNS,
       ...
      )

**Review by e-carlin — COMMENTED** [review]
Looks great! I left some comments on ways to improve using the LogParser class but the general structure is good. Thanks for cleaning this code up.

File: sirepo/template/madx.py [hunk 2]
```diff
@@ -160,6 +160,18 @@ def write_files(self, data, source_path, dest_dir):
         return PKDict()
 
 
+class MadxLogParser(template_common.LogParser):
```
e-carlin, line 163: I think this class (and all classes like it) can be private.

sirepo/template/cloudmc.py
```diff
@@ -666,20 +666,15 @@ def _is_sbatch_run_mode(data):
...
-    if res:
-        return res
-    return "An unknown error occurred, check CloudMC log for details"
+    res = template_common.LogParser(
+        run_dir,
+        template_common.RUN_LOG,
+        # ERROR: Cannot tally flux for an individual nuclide.
+        error_patterns=(r"^\s*Error:\s*(.*)$", r"^\s*error:\s*(.*)$"),
+    ).parse_log_file_for_errors()
+    if res == "An unknown error occurred":
```
e-carlin, line 675: Can we just set default_msg instead of doing this?

sirepo/template/template_common.py
```diff
@@ -83,6 +85,48 @@ def __init__(self, **kwargs):
...
+            m = re.search(
+                self.multiline_error_pattern,
+                pkio.read_text(f),
+                re.MULTILINE | re.DOTALL,
+            )
+            if m:
+                e = m.group(1)
+        return e
+
+    def _parse_log(self, file, result):
```
e-carlin, line 121: Don't pass result into this function. I think it is always `""`. It is dangerous to update values passed into functions (hard to know what state you are messing with). So better to avoid when possible.

sirepo/template/template_common.py
```diff
@@ -40,6 +40,8 @@
 #: stderr and stdout
 RUN_LOG = "run.log"
 
+_DEFAULT_ERROR_PATTERNS = (r"Error: (.*)",)
```
e-carlin, line 43: Make this a class variable. It is only relevant for the LogParser class.

sirepo/template/template_common.py
```diff
@@ -83,6 +85,48 @@ def __init__(self, **kwargs):
...
+class LogParser(PKDict):
+    def __init__(self, run_dir, log_filename, **kwargs):
+        super().__init__(run_dir=run_dir, log_filename=log_filename, **kwargs)
+        self.pksetdefault(
+            error_patterns=_DEFAULT_ERROR_PATTERNS,
+            multiline_error_pattern=r"^Traceback .*?^\w*Error: (.*?)\n",
+            default_msg="An unknown error occurred",
+        )
+
+    def parse_log_file_for_errors(self):
```
e-carlin, line 97: I think `log_file` is redundant. Just `parse_for_errors` is sufficient.

sirepo/template/template_common.py
```diff
@@ -83,6 +85,48 @@ def __init__(self, **kwargs):
...
+        p = self.run_dir.join(self.log_filename)
+        if not p.exists():
+            return res
+        with pkio.open_text(p) as f:
+            res = self._parse_log(f, res)
+        if res:
+            return res
+        return self.default_msg
+
+    def parse_multiline(self):
```
e-carlin, line 108: I'm not 100% positive (so please look into the code) but I don't think this needs to be a special case. I think for mpi we can define a `class _MPILogParser` and override `_parse_log` with this. We can set `default_msg` to be `None`. Then I think the only case that would be a change in behavior with the existing implementation is what we return in the log file not existing case. Currently we return `None`. If we use parse_log_file_for_errors() we will return `""`. But, looking at the js code we decide to display an alert based on the truthiness of the message. Both `None` and `""` are falsey so in the end there won't be an actual change in behavior.

e-carlin (changes requested):
Looks good. I left some general comments on the structure. Once those are cleaned up I think it will be best to have Paul review since he knows all the in's and out's of this code.

gurhar1133:
@e-carlin ping

sirepo/template/template_common.py
```diff
@@ -83,6 +83,34 @@ def __init__(self, **kwargs):
         )
 
 
+class LogParser(PKDict):
+    def __init__(self, run_dir, log_filename, **kwargs):
```
e-carlin, line 87: log_filename should be optional and default to RUN_LOG

e-carlin (commented):
One minor change. Other than that lgtm. Please have Paul give the final review and sign off.

e-carlin:
The structure looks good to me. Paul can make sure bugs weren't introduced.

moellep:
I'll keep testing, but here is one case I found:
- In cloudmc, open the Aurora sim, on the Visualization tab change the Run Mode value to "eigenvalue" and run the simulation. You should see an error message "No fission sites banked on MPI rank 0", but it is showing "An unknown error occurred, check CloudMC log for details".

moellep:
Changing the error pattern matcher to IGNORECASE fixes the cloudmc problem, but uncovers another:
x

So, possibly an option could be added to LogParser to ignorecase, or that should be the default option. And an option for stop_at_first_error, or similar. Or the LogParser could ignore duplicate error lines.

robnagler:
I would like to throw in that I am seeing this off of "main": OPAL failed\nAn unknown error occurred. The problem is that this doesn't give any information to debug on the server side. The logs are empty. I have to go to the run dir. Ideally, if there is a problem that is unknown, perhaps the last part of the log should be written to stderr in job_cmd or something that allows a developer or sysadmin to debug.

moellep:
I've added some test logs with a39af57
I think we should add a unit test for this because parsing logs is an important job we can improve.

sirepo/template/opal.py
```diff
@@ -116,6 +116,22 @@ def is_ignore_field(self, field):
         return field == "name" or field == self.IS_DISABLED_FIELD
 
 
+class _OpalLogParser(template_common.LogParser):
+    def _parse_log(self, file_path):
+        res = ""
+        visited = set()
+        with pkio.open_text(file_path) as f:
+            for line in f:
+                if re.search(r"^Error.*?>\s*[\w\"]", line):
```
moellep, line 125: I think the part above should be in the base class - line parsers should just be given a line of input, and return any errors parsed out of it. The superclass should collect the errors and de-dup them.

tests/template/logparser_data/cloudmc1.txt
```diff
@@ -0,0 +1,134 @@
...
+ Preparing distributed cell instances...
+ Reading plot XML file...
+ Maximum neutron transport energy: 20000000 eV for N15
+ Initializing source particles...
+
+ ====================>     K EIGENVALUE SIMULATION     <====================
+
+  Bat./Gen.      k            Average k
+  =========   ========   ====================
+ ERROR: No fission sites banked on MPI rank 0
```
gurhar1133, line 122: @moellep  is there value in keeping the pieces of the log above this line?

moellep: Not in this case - just the "ERROR: No fission sites banked on MPI rank 0" line.

tests/template/logparser_data/opal1.txt
```diff
@@ -0,0 +1,27 @@
...
+OPAL{0}>
+OPAL{0}> The optimiser (former opt-Pilot) is integrated
+OPAL{0}>
+OPAL{0}> Please send cookies, goodies or other motivations (wine and beer ... )
+OPAL{0}> to the OPAL developers opal@lists.psi.ch
+OPAL{0}>
+OPAL{0}> Time: 21:38:11 date: 26/01/2024
+OPAL{0}>
+OPAL{0}> * Reading input stream 'opal.in'
+Error{5}>
```
gurhar1133, line 23: and the lines above this one @moellep ?

tests/template/logparser_data/opal2.txt
```diff
@@ -0,0 +1,52 @@
...
+OPAL{0}> 
+OPAL{0}> The optimiser (former opt-Pilot) is integrated 
+OPAL{0}> 
+OPAL{0}> Please send cookies, goodies or other motivations (wine and beer ... ) 
+OPAL{0}> to the OPAL developers opal@lists.psi.ch
+OPAL{0}> 
+OPAL{0}> Time: 21:42:40 date: 26/01/2024
+OPAL{0}> 
+OPAL{0}> * Reading input stream 'opal.in'
+Error{1}> Error{2}>
```
gurhar1133, line 23: lines above this? @moellep

tests/template/logparser_data/opal3.txt
```diff
@@ -0,0 +1,109 @@
...
+OPAL{0}> 
+OPAL{0}> The optimiser (former opt-Pilot) is integrated 
+OPAL{0}> 
+OPAL{0}> Please send cookies, goodies or other motivations (wine and beer ... ) 
+OPAL{0}> to the OPAL developers opal@lists.psi.ch
+OPAL{0}> 
+OPAL{0}> Time: 22:35:26 date: 26/01/2024
+OPAL{0}> 
+OPAL{0}> * Reading input stream 'opal.in'
+Error{1}> S1: No filename for a field map given. Will assume the default map "1DPROFILE1-DEFAULT".
```
gurhar1133, line 23: above this @moellep ?

tests/template/logparser_data/opal4.txt
```diff
@@ -0,0 +1,55 @@
...
+OPAL{0}> 
+OPAL{0}> The optimiser (former opt-Pilot) is integrated 
+OPAL{0}> 
+OPAL{0}> Please send cookies, goodies or other motivations (wine and beer ... ) 
+OPAL{0}> to the OPAL developers opal@lists.psi.ch
+OPAL{0}> 
+OPAL{0}> Time: 22:40:08 date: 26/01/2024
+OPAL{0}> 
+OPAL{0}> * Reading input stream 'opal.in'
+Error{3}> S1: No filename for a field map given. Will assume the default map "1DPROFILE1-DEFAULT".
```
gurhar1133, line 23: lines above @moellep ?

tests/template/logparser_data/shadow1.txt
```diff
@@ -0,0 +1,291 @@
...
+       46032                   accepted.
+     4100000    rays have been rejected so far.
+       46596                   accepted.
+     4150000    rays have been rejected so far.
+       47170                   accepted.
+                  47500
+     4200000    rays have been rejected so far.
+       47720                   accepted.
+     4250000    rays have been rejected so far.
+       48313                   accepted.
```
gurhar1133, line 194: above lines @moellep ?

moellep: The example log files have the full log - you only want to get the error parts out.

tests/template/logparser_data/opal3.txt
moellep: The existing opal parser has a special case to ignore 1DPROFILE1-DEFAULT

tests/template/logparser_data/shadow1.txt
moellep: The full shadow log should result in one error line: "Compound is not a valid chemical formula and is not present in the NIST compound"

gurhar1133:
can you take another look @moellep ?

sirepo/template/madx.py
```diff
@@ -159,6 +159,18 @@ def write_files(self, data, source_path, dest_dir):
...
+    def _parse_log_line(self, line):
+        res = ""
+        if re.search(r"^\++ (error|warning):", line, re.IGNORECASE):
+            line = re.sub(r"^\++ ", "", line)
+            res += line + "\n"
+        elif re.search(r"^\+.*? fatal:", line, re.IGNORECASE):
+            line = re.sub(r"^.*? ", "", line)
+            res += line + "\n"
+        return res
+
```
moellep, line 172: Looks much better. In cases like the above, you don't need to use the ```res``` variable because you get at most one error from the line. I saw similar code in the opal.py module:
```python
         if re.search(r"^\++ (error|warning):", line, re.IGNORECASE):
            return re.sub(r"^\++ ", "", line)
        if re.search(r"^\+.*? fatal:", line, re.IGNORECASE):
            return re.sub(r"^.*? ", "", line)
        return None
```

moellep (commented):
Looks great. Very nice to have a set of unit tests which can be updated as new test cases come up.

gurhar1133:
I think I addressed most recent comments @moellep

moellep:
Looks good. @e-carlin can you check if your request was implemented and resolve it?

e-carlin (approved):

PR #6667, opened 2024-01-19
Fix #6661 gui updates

### LINKED ISSUE #6661: GUI design updates I
Implement some of the changes previously discussed (see https://github.com/radiasoft/sirepo/tree/gui-experiments)

mkeilman:
A handful of updates (not everything in the gui-experiments branch):
- Bolder dark gray panel titles. There was some discussion of using blue, but in any case they should be readily distinguishable from clickable elements
- "Save Changes" is now "Save"
- "Action buttons" used to move, copy etc. items in a list are a darker blue. Implemented only for radia at the moment; they should live in a directive or class for general use
- Supporting changes in the css including trying out custom properties

Suggestions encouraged!


moellep:
Two things I notice:
- The "Save" button is now too small, smaller than the cancel button. Adding a style like: "min-width: 7em" would help.
- The formatting applied to the panel headings should also be applied to the modal dialog headings.
x

PR #6678, opened 2024-01-24
Fix #4791 add sources to visualization

### LINKED ISSUE #4791: Draw Source(s)
Request from a potential customer:

For a fixed source simulation, with [point,box] sources, It would be nice to have an indication of where the source is relative to the tally visualization. For Example, in the Aurora example, there should be a point labeled with the source to the right. Likewise, for a box source, a wire frame box would be a nice-to-have.

It's more difficult to indicate angle and energy. For a point source, a mono-directional source could have a dashed vector indicating the direction. A polar azimuthal source could show a dashed-line cone. But how do we indicate these things from a box? Or from a user-defined distribution?

I recommend we limit these visualizations to point source and box, and only draw angular information from the center of the box case.

- [ ] box source
- [ ] point source

We'll have to ponder how to display angle and energy without the plot getting too crowded...

mkeilman:
This adds representations of sources and associated particles to the visualization. Notes:
- Point and box sources generate circle(sphere) and rectangle(box) shapes in 2d(2d) respectively. Other shapes may be added later
- `openmc.lib.sample_external_source()` generates a representative sample of 100 source particles (over all sources, not per source) at the start of the simulation. These are _not_ particles used in the actual calculation. Eigenvalue runs can save their source particles, but not a subset of them, while fixed source runs cannot save them at all
- Users can select the number of particles drawn, which for simplicity is the first N of all those generated
- The source particles are displayed are arrows, colored via a colormap according to energy. The lengths are fixed to a reasonable percentage of the maximum mesh extent; however they could be scaled to energy etc.
- In 2d only, photons have dashed lines (vtk has no easy way to do the same)

Various supporting classes are also included


moellep:
Looks good. I'll approve and create a few follow-on issues.

PR #6680, opened 2024-01-24
fix #5826 new drag-and-drop across windows for lattice commands

### LINKED ISSUE #5826: drag command between windows
A nice feature would be to have two simulations open and drag a command between windows to copy the command into the new simulation.

The angular drag-and-drop currently clips the dragged item within the browser window, but this drag and drop example shows it is possible to drag and drop between windows.

https://web.dev/drag-and-drop/

moellep:


sirepo/package_data/static/js/sirepo-command.js
```diff
@@ -286,14 +293,28 @@ SIREPO.app.directive('commandTable', function(appState, commandService, latticeS
...
+                    }
+                    $scope.commands.splice(prevIndex, 1);
                 }
                 $scope.commands.splice(index, 0, data);
+                $scope.selectItem(data);
                 saveCommands();
             };
+
             // expects a negative number to move up, positive to move down
             $scope.moveItem = function(direction, command) {
```
mkeilman, line 319: As long as this directive is being changed, might as well use arrow syntax for existing functions for consistency:
 ```
$scope.moveItem = (direction, command) => {...}
```
etc.

File: sirepo/package_data/static/js/sirepo-components.js [hunk 2]
```diff
@@ -5329,3 +5329,131 @@ SIREPO.app.service('utilities', function($window, $interval, $interpolate) {
...
+        },
+        controller: function($scope, $element) {
+            let dragCount = 0;
+
+            function updateCount(newCount) {
+                dragCount = newCount;
+                if (dragCount == 0) {
+                    $element.removeClass('sr-drag-over');
+                }
+                else if (dragCount == 1) {
```
mkeilman, line 5421: triple equals

mkeilman (commented):
Works well - can even drag to other tabs which is nice. Just some style comments.

There are some UI items that don't hold this up but might be nice:
- Toggle a selected command when clicking on it again
- Unselect a selected command when clicking anywhere
- Provide a trash can icon to drop on for deletion

sirepo/package_data/static/js/sirepo-command.js
moellep: Yes, there are many cases of the old function arg format, but I didn't change it for moveItem() in this case because I didn't modify that code. We could have a separate issue to clean up old function calls across sirepo, sirepo-component, sirepo-lattice and sirepo-command.

PR #6683, opened 2024-01-24
fix #6482: make oauth work with websockets

### LINKED ISSUE #6482: 20231211 Make oauth work with websockets
We had an issue related to flash not working https://github.com/radiasoft/sirepo/issues/6378 . I [fixed that](https://github.com/radiasoft/sirepo/pull/6481) but noticed along the way that oauth only works when websockets are disabled (feature_config.ui_websocket=False). 

My hunch is that the oauth redirect response happens through a websocket message and we try to do Set-Cookie but this doesn't have any effect when operating in websockets. So, we'll need to find a way around this. 

Some possible ideas:
- Augment our websocket messages so they have a cookie field and have the browser update the cookie
- Don't use websockets for the oauth flow

rorour:


rorour:
@robnagler said:
> Easy. There's a regex for this. See sirepo.js `_isAuthUrl`

This doesn't work in this case because we don't hit a sirepo auth url until after flash redirects to sirepo. The flow is /simulation-list -> flash.rochester.edu -> /sim-oauth-flash-authorized and we need the sirepo->flash redirection to use http.

robnagler:
> redirection to use http

This should work already so there must be a bug in the redirection handling. 

- [raise_authorize_redirect raises Redirect](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/oauth.py#L53)
- [_Redirect provides a websocket_content](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/reply.py#L768) that returns httpRedirect
- [handleSRException should execute a globalRedirect](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/package_data/static/js/sirepo.js#L2551)
- [which should update $window.location.href](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/package_data/static/js/sirepo.js#L2534) which should do an http redirect to flash.rochester.edu

The changes in this PR are fragile, because they depend on a lot of things, and don't fix the fundamental issue with httpRedirect srExceptions not redirecting.

rorour:
Maybe the way I worded my comment was confusing. I think that code you linked is working correctly. 

[The oauth flow begins](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/auth/__init__.py#L175) from [the first request made with "flash" as sim type](https://github.com/radiasoft/sirepo/blob/cee3e921903277e5ad751c6b8337573809a3fd78/sirepo/http_request.py#L71), and that decision is made on the front end where I made my changes. I agree that depending on "/simulation-list" is fragile.

robnagler:
Got it, thanks. oauth.raise_authorize_redirect does two things: sets the cookie and redirects. It's the cookie setting that is the problem. That fails under websockets.

I think this should be server-side, that is, when the server needs the client to save a cookie, it should be able to tell the client. I see two choices:
1. Teach the websockets server to send the cookie and the client to save it using cookieService.
2. Add a "need a cookie" exception that causes the websockets code to drop the websocket, reconnect the same request (whatever it is, e.g. /simulation-list) without websockets. 

I think option (1) is better, because there would be fewer redirects. This would client up the client quite a bit.

robnagler:
Let's release this as is, since I think this is holding up pushing websockets to production.

We can create a new issue that will fix the problem properly.

e-carlin:
This doesn't work if a user goes directly to a simulation (ex clicks on a bookmark). It assumes the route is `/simulation-list` which isn't the case if they go straight to a sim.

rorour:
Should we disable websockets for oauth sim types for the time being?

robnagler:
I'm looking at this PR as a stopgap. I think the only real solution is to do the cookie thing. I will work on that. Let's see how I progress.

robnagler:
> Should we disable websockets for oauth sim types for the time being?

That's more work. We don't have that feature in the code. I'll work on the cookie thing. It shouldn't be too hard.

rorour:
> > Should we disable websockets for oauth sim types for the time being?
> 
> That's more work. We don't have that feature in the code. I'll work on the cookie thing. It shouldn't be too hard.

I was thinking something like removing `url === "/simulation-list"` from my changes here

robnagler:
> I was thinking something like removing url === "/simulation-list" from my changes here

Good thought. If you want, do that, and we can push websockets out.

rorour:
@e-carlin ready for review

e-carlin (approved):

PR #6703, opened 2024-01-29
Fix #6642 dont reset epics plots

### LINKED ISSUE #6642: changing generator value shouldn't reset plots
When running epicsllrf against llrfsim, if you change a generator value and submit it, the plots reset the zoom and selected values. It should keep showing the selected value with the current zoom instead.

For example, if I'm looking at a zoomed in plot of just the Forward value:
x

and update the generator, it is currently reseting the plot:
x2

gurhar1133:


gurhar1133:
@moellep ping

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3891,6 +3891,17 @@ SIREPO.app.directive('parameterPlot', function(appState, focusPointService, layo
                     });
             };
 
+            $scope.plotLabels = (plots) => {
```
mkeilman, line 3894: Convention is to omit parens for single parameters: `$scope.plotLabels = plots => {...}`

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3891,6 +3891,17 @@ SIREPO.app.directive('parameterPlot', function(appState, focusPointService, layo
...
+                let res = [];
+                if (plots == null) {
+                    return res;
+                }
+                for (let plot of plots) {
+                    res.push(plot.label);
+                }
+                return res;
```
mkeilman, line 3902: Simpler is `$scope.plotLabels = plots => (plots || []).map(p => p.label);`

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -3527,7 +3527,7 @@ SIREPO.app.directive('parameterPlot', function(appState, focusPointService, layo
             var includeForDomain = [];
             var childPlots = {};
             var scaleFunction;
-            var plotVisibilty = {};
+            var plotVisibility = {};
```
mkeilman, line 3530: Thanks for correcting the misspelling!

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -4088,21 +4099,32 @@ SIREPO.app.directive('parameterPlot', function(appState, focusPointService, layo
...
-                    if (cachedPlotVisibilty(i, $scope.modelName)) {
+                    if (cachedPlotVisibility(i, $scope.modelName)) {
                         setPlotVisible(i, ! isPlotVisible(i));
                     }
                 });
 
-                $scope.$on(
-                    $scope.modelName + '.changed',
-                    () => {
+                function handlePlotSelection() {
```
mkeilman, line 4107: I'm not sure whether it's worth making a function here since it is defined and then immediately invoked, and only in that one place...

sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -4088,21 +4099,32 @@ SIREPO.app.directive('parameterPlot', function(appState, focusPointService, layo
...
+                    const c = $scope.plotLabels(plots);
+                    const p = appState.models[$scope.modelName].selectedPlots || c;
+                    if (JSON.stringify(c) != JSON.stringify(p)) {
```
mkeilman, line 4110: Labels are not necessarily unique so this may not be a valid equality test. Also use `===` wherever possible as a matter of convention, though it doesn't really matter here...

mkeilman (commented):
@moellep is doing the main review but I had a few style comments...

sirepo/package_data/static/js/sirepo-plotting.js
robnagler: When was this convention made? I don't think this is a good idea. It's always good for automated refactoring/mods to have a consistent syntax. This makes searches for functions easy: `r"\w+\s*=\s*([^\)]+)\s*=>"`

mkeilman: I think when we moved to arrow notation a couple of years ago, though I can't find the discussion...

sirepo/package_data/static/js/sirepo-plotting.js
gurhar1133: not unique in the sense that two different y values might have the same label? If so, when would that happen?

mkeilman: One would hope that the labels are always different, but they're arbitrary. This change affects all parameter plots, so better to be cautious I think

sirepo/package_data/static/js/sirepo-plotting.js
robnagler: > Convention is to omit parens for single parameters

I was responding to this specifically.

sirepo/package_data/static/js/sirepo-plotting.js
gurhar1133: Ok, maybe I could use 'field' instead of 'label'. I think i'll have to make some schema changes

sirepo/package_data/static/js/sirepo-plotting.js
mkeilman: I mean we made that decision around the same time we moved to arrows. I don't recall strong preference either way though. You make a good point so let's so that way. iow ignore my comment @gurhar1133 :)

sirepo/package_data/static/js/sirepo-plotting.js
mkeilman: I guess I don't know what plot changes this is supposed to pick up; maybe wait to see what @moellep has to say?

moellep:
I made a bunch of changes with 7ed8aed, and include notes in that commit. The code should be a little clearer now - give it a test to be sure it still works.

gurhar1133:
LGTM @moellep. To work around [#6729](https://github.com/radiasoft/sirepo/issues/6729) I changed sirepo.sim_db_file.in_job_agent() to:
```python
def in_job_agent():
    return False
```

PR #6730, opened 2024-02-05
fix #6358 correct numpy related bug in warpvnd

### LINKED ISSUE #5278: fedora36: warpvnd  EGun example doesn't work
source:
`Simulation Error: Out of range float values are not JSON compliant: 1 animation frames`

Visualization:
Same underlying error but it results in a "report not generated" in the gui

### LINKED ISSUE #6358: warpvnd Two Poles example is broken
Running the warpvnd Two Poles examples on the visualization tab causes the error: "Error: 0".

### LINKED ISSUE #6671: gridded tec 2d impact density report not generated
Simulation Steps 200 (could that cause it?)

```txt
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/job_cmd.py", line 252, in _do_get_simulation_frame
    return template_common.sim_frame_dispatch(
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/template_common.py", line 688, in sim_frame_dispatch
    res = o(frame_args)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/warpvnd.py", line 107, in sim_frame_impactDensityAnimation
    return _extract_impact_density(frame_args.run_dir, frame_args.sim_in)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/warpvnd.py", line 584, in _extract_impact_density
    return _extract_impact_density_2d(run_dir, data)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/warpvnd.py", line 589, in _extract_impact_density_2d
    all_particles = np.load("all-particles.npy")
  File "/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/numpy/lib/npyio.py", line 427, in load
    fid = stack.enter_context(open(os_fspath(file), "rb"))
```

Screen Shot 2024-01-19 at 4 19 21 PM

moellep:
Also fix #6671 fix #5278


sirepo/package_data/template/warpvnd/egun.py.jinja
```diff
@@ -72,11 +72,11 @@ if can_write_files:
...
-                'lost': particles[1]
-            },
-            hf, path='particle'
+    with open("{{ particleFile }}", "wb") as f:
+        pickle.dump(
+            dict(
+                kept=particles[0],
+                lost=particles[1],
+            ),
+            f,
```
mkeilman, line 81: Is there a way to use something other than pickle? It's known to be insecure

mkeilman (commented):
Only one comment around the security concerns of pickle files

sirepo/package_data/template/warpvnd/egun.py.jinja
robnagler: Do not use pickle. If you need density, use msgpack or hdf5 otherwise use json.

robnagler: There's also an allow_pickle=True above. Do not use pickle for this.

moellep: I'll update it to use msgpack. hdf5 doesn't work because the array sizes are all different.

robnagler: Why is the allow_pickle necessary?

moellep: I don't know why that was added - possibly there was an object in the data. I'll see if it can be taken out.

sirepo/template/warpvnd.py
```diff
@@ -689,11 +695,10 @@ def _extract_optimization_results(run_dir, data, args):
 
 def _extract_particle(run_dir, model_name, data, args):
     limit = int(args.renderCount)
-    hf = h5py.File(str(run_dir.join(_PARTICLE_FILE)), "r")
-    d = template_common.h5_to_dict(hf, "particle")
-    kept_electrons = d["kept"]
-    lost_electrons = d["lost"]
-    hf.close()
+    with open(_PARTICLE_FILE, "rb") as f:
```
mkeilman, line 698: I think in general we should convert files for existing simulations when we change formats, if possible, though in practical terms it may not matter here

moellep: I think in this case, just getting the app working again is good enough.

PR #6731, opened 2024-02-05
fix #6697 update srw schema to allow totalSamples == 0

### LINKED ISSUE #6679: optimization percentComplete divide by zero
srw > Young's Double Slit Experiment >  Machine Learning > [enter nothing] Save Changes > Start New Data Generation:

```txt
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/pkcli/job_cmd.py", line 72, in _background_percent_complete
    return template.background_percent_complete(
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/srw.py", line 249, in background_percent_complete
    return _machine_learning_percent_complete(run_dir, res)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/template/srw.py", line 2219, in _machine_learning_percent_complete
    res.percentComplete = 100 * count / dm.exportRsOpt.totalSamples
ZeroDivisionError: division by zero
```

### LINKED ISSUE #6697: srw totalSamples ML causes copy error
If you visit the SRW Machine Learning tab, make no changes, and then return to other tabs, it saves exportRSOpt.totalSamples = 0 which is invalid with the schema, and this will cause the app to crash when "open as copy" is selected.

To reproduce:
Ceate a new SRW sim.
Go to the Beamline tab and add an aperture and a watchpoint.
Go to the Machine Learning tab, then go to the Beamline tab, and back to the Machine Learning tab.
Go to the Simulations list (this will invoke an autoSave).
Select "open as a new copy" on that simulation.

I think "0" should be added as a valid value schema, and when running the rsopt process, it should detect totalSamples == 0 and raise an error which is displayed to the user.

moellep:
Hide ML run panel unless totalSamples is provided, fix #6679


sirepo/package_data/static/json/srw-schema.json
```diff
@@ -674,7 +674,7 @@
             "numWorkers": ["Number of Workers", "Integer", 1],
             "randomSeed": ["Random Seed", "RandomSeed", "", "Explicit seed for random numbers in the range [{{ SIREPO.APP_SCHEMA.model.exportRsOpt.randomSeed[SIREPO.INFO_INDEX_MIN] }}, {{ SIREPO.APP_SCHEMA.model.exportRsOpt.randomSeed[SIREPO.INFO_INDEX_MAX] }}). Leave empty to allow server to decide the seed.", 0, 4294967296],
             "scanType": ["Scan Type", "ScanType", "random", "Grid samples each parameter uniformly. Random takes a random sample for each parameter."],
-            "totalSamples": ["Total Samples", "Integer", 100, "Grid: &lt;number of samples&gt;^&lt;number of params&gt;. Random: &lt;number of samples&gt;", 1],
+            "totalSamples": ["Total Samples", "Integer", 100, "Grid: &lt;number of samples&gt;^&lt;number of params&gt;. Random: &lt;number of samples&gt;", 0],
```
mkeilman, line 677: We should change the default to 0 - new simulations see 100 and the sim panel shows up. Maybe also run `scope.updateTotalSamples()` and save quietly when the rsOptElements directive first loads.

mkeilman (changes requested):
Only real change is to make sure totalSamples is initialized properly, otherwise good

sirepo/package_data/static/json/srw-schema.json
moellep: I update totalSamples default to 0 now

mkeilman (approved):
lgtm
(except for the test failure 😄 )

PR #6751, opened 2024-02-08
Fix #6689 websocket tracks cookies in UI and api server

### LINKED ISSUE #6686: Remove react implementation
Too many security patches being issued by dependabot

### LINKED ISSUE #6689: Websockets set cookie
https://github.com/radiasoft/sirepo/pull/6683#issuecomment-1909065749

### LINKED ISSUE #6727: Nested quests share too much state
The back link (qcall attr) on auth for the nested quest, point to the parent qcall, not the child's qcall. Since this is relied on heavily, these back links should be fixed up in parent_set.

auth probably should be cloned so that the auth state is not modified.

auth_db might need to be shared so we don't open two sessions. This presents the problem with UserDirNotFound, which will be raised and handled (deleting the auth_db entries), and then the caller will also detect the same state. That's unusual, but still might happen. 

This was detected with beginSession (nested quest) which is pre-authorization in _call_api, and then the _call_api goes and checks permissions. The cookie is logged out (auth backlink above), but it is not actually returned to the user yet so a step is skipped and there's a login exception that occurs as opposed to a full reset (redirect to app root) which is the correct behavior.

auth2_test expects a login exception when it should actually get a go back to root first, which is the correct place to be when the logout occurs. The login exception will be handled correctly.

### LINKED ISSUE #6748: websockets do not logout if another window logs out

### LINKED ISSUE #6760: checkCookieRedirect redirects to same localroute causing app to hang
checkCookieRedirect [compares the path literally](https://github.com/radiasoft/sirepo/blob/282a67457c3d9e73c3050b7021888c62d22232e3/sirepo/package_data/static/js/sirepo.js#L2376), and does not take into account uri params. For example, if going from `/simulations` to `/simulations/Examples`, it's actually the same local route so angular doesn't restart the controller even thought the location has changed. The loading spinner runs, but SimulationsController never starts.

This might have something to do with simulations and simulationsFolder having identical configurations except for "/:folderPath?", but I think the it's better if cookieRedirect is coded more carefully. For example, it should use [`$rootScope.$evalAsync`](https://github.com/angular/angular.js/issues/9607#issuecomment-59087310) to do the `$location.path` change. It's also not clear that it should trust a previous route from the same tab. This should only happen if the browser had to reload the app to login.

To reproduce (websockets or not):
- Login with server running no_smtp_mail
- Stay on simulations page
- Copy uri to another tab
- Logout of one tab
- In the other tab, open Examples

robnagler:
- msgRouter, uri_router._WebSocket, srunit._WebSocket track cookies to allow authentication
- web sockets support asyncMsg kind for setCookies
- httponly no longer set on cookies so can be accessed in Javascript
- quest better modularlization
- user_path checked earlier (require_user) so better responses
- _SRequest kinds are subclasses
- Fix #6748 save cookies in srunit and sirepo.js and close websocket if changed
- Fix #6727 all Attr's copied on init_quest_for_child
- Fix #6686 remove special APIs for react
- _call_api.body can be falsey (empty dict), check None
- msgRouter.websocket retries with binary backoff (max 1 minute)
- msgRouter better error handling and better modularlized
- _SReply.destroy more robust against non-existent values
- srunit.assert_http_redirect handles srException correctly
- call_api/_sync accepts body, not data
- refactored refresh code so more modular and less implicit coupling
- factored out `uri` from `requestSender` so `msgRouter` can redirect
- Fix #6760 checkCookieRedirect renamed checkLoginRedirect, and is careful about redirect loops


robnagler:
whoever merges, i'm updating https://github.com/radiasoft/sirepo/pull/6751#issue-2125627359 for the squash merge

robnagler:
@moellep @e-carlin fyi, there are still defects in this PR around multiple browser tabs. I know the problems and will fix tomorrow. #6760 is unrelated to this PR, but obviously ran into it while testing. That confused things quite a bit. I will end up fixing this too, because can't really test without that fix. Single tab login/logout, etc. should work ok.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3471,7 +3481,7 @@ SIREPO.app.directive('splitPanels', function($window) {
     var MIN_TOP_PERCENT = 15;
     var TOP_PAD = 12;
     return {
-        controller: function($scope) {
+                    controller: function($scope) {
```
e-carlin, line 3484: fix formatting

sirepo/reply.py
```diff
@@ -133,6 +147,7 @@ def gen_exception(self, exc):
         # to the server, which will have code to handle this case.
         try:
             if isinstance(exc, sirepo.util.ReplyExc):
+                self.qcall.cookie.has_sentinel()
```
e-carlin, line 150: Not used?

e-carlin (changes requested):
I ran into one error while testing.  To reproduce:
```
rm -rf run/
git checkout master
bash run.sh flash
# visit blast2 simulation and wait for problem files to unpack
# stop server
git checkout 6689-websocket-cookie
sirepo roles delete <uid> flash
bash run.sh flash
# refresh page with blast2 simulation
# log into flash center
# see error on redirect back to sirepo
```

Traceback
```
Feb 12 18:26:08 1588953     0 ../sirepo/uri_router.py:419:_call_api api=simOauthFlashAuthorized exception= stack=Traceback (most recent call last):
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/uri_router.py", line 410, in _call_api
    await getattr(qcall, route.func_name)(**kwargs)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/sim_oauth/flash.py", line 25, in api_simOauthFlashAuthorized
    o, _ = sirepo.oauth.check_authorized_callback(self)
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/oauth.py", line 27, in check_authorized_callback
    assert t
AssertionError
```

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3296,6 +3296,18 @@ SIREPO.app.directive('ldapLogin', function (requestSender) {
...
+        return Object.keys(SIREPO.refreshModalMap).reduce(_refreshTemplate, '');
+    };
+
+    const _refreshTemplate = (rv, id) => {
+        const x = SIREPO.refreshModalMap[id];
+        if (! x) {
+            throw new Error(`not found id=${id} in refreshModalMap`);
+        }
+        return rv + `<div data-confirmation-modal="" data-is-required="true" data-id="${x.modal}" data-title="${x.title}" data-ok-text="Refresh" data-ok-clicked="refreshPage()">${x.msg}. Select <b>Refresh</b> to update this simulation.</div>\n`;
+    };
```
moellep, line 3309: I think a map() rather than reduce() is simpler here:
```javascript
    const _refreshModals = () => {
        return Object.values(SIREPO.refreshModalMap).map(_refreshTemplate).join('');
    };

    const _refreshTemplate = (x) => {
        return `<div data-confirmation-modal="" data-is-required="true" data-id="${x.modal}" data-title="${x.title}" data-ok-text="Refresh" data-ok-clicked="refreshPage()">${x.msg}. Select <b>Refresh</b> to update this simulation.</div>\n`;
    };
```

robnagler: `map().join('')` is exactly the same as `reduce('')` so I think one function call is simpler than two.

sirepo/reply.py
robnagler: [26d97f5](https://github.com/radiasoft/sirepo/pull/6751/commits/26d97f545074f52d3d357a8a3e70521ac279fccf)

sirepo/package_data/static/js/sirepo-components.js
robnagler: [26d97f5](https://github.com/radiasoft/sirepo/pull/6751/commits/26d97f545074f52d3d357a8a3e70521ac279fccf)

sirepo/auth/__init__.py
```diff
@@ -148,6 +146,15 @@ class _Auth(sirepo.quest.Attr):
     METHOD_EMAIL = METHOD_EMAIL
     METHOD_GUEST = METHOD_GUEST
 
+    _QUEST_KEY = "auth"
```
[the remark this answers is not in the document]
robnagler: I didn't doc each one. I added a separate issue https://github.com/radiasoft/sirepo/issues/6774

sirepo/quest.py
```diff
@@ -310,7 +361,17 @@ def _wrapper(*args, **kwargs):
 
 
 class _Bucket(Attr):
-    pass
+    _QUEST_KEY = "_bucket"
```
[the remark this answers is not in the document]
robnagler: [26d97f5](https://github.com/radiasoft/sirepo/pull/6751/commits/26d97f545074f52d3d357a8a3e70521ac279fccf)

sirepo/auth/__init__.py
```diff
@@ -148,6 +146,15 @@ class _Auth(sirepo.quest.Attr):
...
+
+    _INIT_QUEST_FOR_CHILD_KEYS = frozenset(
```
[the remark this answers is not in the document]
robnagler: [26d97f5](https://github.com/radiasoft/sirepo/pull/6751/commits/26d97f545074f52d3d357a8a3e70521ac279fccf)

robnagler:
[5dfb6c4](https://github.com/radiasoft/sirepo/pull/6751/commits/5dfb6c4bc79533b2fc6e2d2c42cb1210c23745e5) fixes the cookie errors (assert t) seen in flash.

sirepo/package_data/static/js/sirepo-components.js
moellep: Rather than keys(), If you call ```Object.values(SIREPO.refreshModalMap).reduce(_refreshTemplate, '')``` then you won't need the error handling in _refreshTemplate()

robnagler: Thanks! [cfc5fc9](https://github.com/radiasoft/sirepo/pull/6751/commits/cfc5fc9bf9751306ddf38f8009d6ad53a523b116)

sirepo/cookie.py
```diff
@@ -9,20 +8,32 @@
...
-_MAX_AGE_SECONDS = 10 * 365 * 24 * 3600
-
 #: Identifies if the cookie has been returned at least once by the client
 _COOKIE_SENTINEL = "srk"
 
 #: Unique, truthy that can be asserted on decrypt
-_COOKIE_SENTINEL_VALUE = "z"
+_COOKIE_SENTINEL_V2 = "2"
+
+#: Old value, which will get rewritten on first request, removing httponlya
```
e-carlin, line 25: typo httponlya

e-carlin (approved):

robnagler:
@moellep I'd like to get this on alpha this week.

moellep (approved):

PR #6759, opened 2024-02-09
Fix #6653: subsims during write params

### LINKED ISSUE #6653: generate subsimulation run files during template.omega write_parameters
Currently the omega app creates a parameters.py file which collects all the sims, modifies them, generates the code and runs each one.

Instead, all the code generation should happen in template.omega write_parameters(). All the code in the generated parameters.py up to and including the prep_run_dir() should happen before, and not included in the parameters.py.

The parameters.py job will be only to run sims and convert data between them.

gurhar1133:


moellep:
I converted this to a draft until we have it working end-to-end with omega.

robnagler:
@moellep what should we do about this PR?

PR #6763, opened 2024-02-12
fix #6761 move reportId to plot internals

### LINKED ISSUE #6761: move reportId input plotting API
The reportId parameter is passed into reportPanel, reportHeading, reportContent, but is really used internally to the plots to manage the report overlay. I think the plots can manage this state, rather than maintaining it outside the panels.

moellep:


sirepo/package_data/static/js/sirepo-plotting.js
```diff
@@ -970,6 +970,8 @@ SIREPO.app.factory('plotting', function(appState, frameCache, panelState, utilit
             };
         },
 
+        randomId: () => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER),
+
```
mkeilman, line 974: Might as well use `SIREPO.UTILS.randomString()`

mkeilman (commented):
One suggestion around the random id, otherwise lgtm

sirepo/package_data/static/js/sirepo-plotting.js
moellep: I added SIREPO.UTILS.randomId() which uses randomString(). The HTML id spec recommends starting ids with a letter.

mkeilman: Cool

mkeilman (approved):

PR #6773, opened 2024-02-14
Fix #6721: improve dice plot

### LINKED ISSUE #6721: improve dice coefficient plot
Currently the Dice Coefficient plot is generated on the server using matplotlib and returned to the client as a single image. Instead this should be converted into a a standard Sirepo plot, possibly a contoured line plot to represent the bars.

x

gurhar1133:
Also found a bug in _histograp_plot(). It returns string data which messes with the y domain of the plots since `'80' < '9' == true` in js


gurhar1133:
@moellep ping

moellep (approved):

PR #6781, opened 2024-02-16
Fix #6777 cloudmc upper threshold

### LINKED ISSUE #6777: Add upper threshold value
From the suggestion in #6231

mkeilman:
Notes:
- Changed single threshold value to a FloatArray
- Added validation of the min/max values
- Added a warning field to the numArray directive


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -544,6 +555,24 @@ SIREPO.app.factory('tallyService', function(appState, cloudmcService, utilities,
...
+            self.cachedSettings[k] = appState.models.openmcAnimation[k];
+        });
+        const t = appState.models.openmcAnimation.thresholds;
+        t.global = [self.minField, self.maxField];
+        // since tallies are counts, compare the lower threshold to 0
+        // instead of the global min
+        t.val[0] = Math.max(t.val[0], 0);
+        t.val[1] = Math.min(t.val[1], t.global[1]);
+    };
+
```
moellep, line 574: When you discard changes to the Tokamak Reactor example and run the simulation, it shows an error about the default threshold:
x

It should probably set the lower/upper to the min/max value when initialized. 

Also, if you switch from the t1 tally to the t2 tally, it resets the theshold to the new limits, but if you return to t1, it is showing a constrained result:
x2

I think we may want to keep track of the selected threshold per tally, or reset it to the tally min/max when switching between tallies.

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -805,11 +834,12 @@ SIREPO.app.directive('geometry2d', function(appState, cloudmcService, frameCache
...
                 );
+                tallyService.updateThresholds();
                 const r =  {
                     aspectRatio: ar,
-                    global_max: tallyService.maxField,
-                    global_min: tallyService.getMinWithThreshold(),
-                    threshold: appState.models.openmcAnimation.threshold,
+                    global_min: appState.models.openmcAnimation.thresholds.global[0],
+                    global_max: appState.models.openmcAnimation.thresholds.global[1],
+                    threshold: appState.models.openmcAnimation.thresholds.val,
```
moellep, line 842: I think it the plot is more meaningful if you match the colorscale to the selected threshold, rather than the global min max:

global colorscale range:
x

threshold colorscale range:
x2

mkeilman: [27f93d3](https://github.com/radiasoft/sirepo/pull/6781/commits/27f93d3d43b6ba6fea5b3898b31a02b219c6f53e)

sirepo/package_data/static/js/cloudmc.js
mkeilman: [8678f98](https://github.com/radiasoft/sirepo/pull/6781/commits/8678f98d491108e1a5c3c4628ef68e23de6c1fc0)

This just moves the lower threshold to 0 if it crosses the upper when changing the data plotted. Tracking every tally/score/aspect combination with its own thresholds seems too complicated to me

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -544,6 +555,28 @@ SIREPO.app.factory('tallyService', function(appState, cloudmcService, utilities,
...
+        t.global = [self.minField, self.maxField];
+        // since tallies are counts, compare the lower threshold to 0
+        // instead of the global min
+        t.val[0] = Math.max(t.val[0], 0);
+        t.val[1] = Math.min(t.val[1], t.global[1]);
+        // if the lower threshold is now too large, reset to 0
+        if (t.val[0] >= t.val[1]) {
+            t.val[0] = 0;
+        }
+    };
```
moellep, line 577: I think anytime a tally or score changes, the threshold should get reset to min/max for that selection. Otherwise selecting a score could unexpectedly show a partial plot.

Fo example, on the Tokamak example, select the absorption score, followed by the flux score without having made any previous threshold changes.
x

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -527,6 +538,32 @@ SIREPO.app.factory('tallyService', function(appState, cloudmcService, utilities,
...
+            return;
+        }
+        updateCache();
+        const t = appState.models.openmcAnimation.thresholds;
+        t.global = [self.minField, self.maxField];
+        // since tallies are counts, compare the lower threshold to 0
+        // instead of the global min. The user can set it to the actual min if desired
+        t.val[0] = 0;
+        t.val[1] = t.global[1];
+    };
```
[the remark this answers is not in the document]
mkeilman: [94d7a93](https://github.com/radiasoft/sirepo/pull/6781/commits/94d7a93eb73ca9f2baa4c72c19d6c436dcd41661)

PR #6788, opened 2024-02-19
fix #6787 cache databroker catalogs

### LINKED ISSUE #6787: cache databroker instances in scan_monitor
This simple call in sirepo.raydata.databroker is incredibly expensive and will degrade performance over time, eventually causing the scan_monitor to become unavailable. Currently it is getting called on each search request, and again for each row in the search results. The scan_monitor should keep the databroker instance and not create it each request.

```python
def catalog(name):
    return databroker.catalog[name]
```

moellep:


sirepo/raydata/databroker.py
```diff
@@ -6,6 +6,8 @@
 from pykern.pkdebug import pkdp
 import databroker
 
+_CACHED_CATALOGS = {}
```
e-carlin, line 9: Any reason not to use a PKDict?

sirepo/raydata/databroker.py
```diff
@@ -39,7 +41,15 @@ def suid(self):
...
 
 def catalog(name):
-    return databroker.catalog[name]
+    # each call to databroker.catalog[name] create a new pymongo.MongoClient
+    # so keep a catalog cache
+
+    # the cached connection could timeout eventually, but the scan_monitor service
+    # is polling for new scans, which should keep it active
+
+    if not name in _CACHED_CATALOGS:
```
e-carlin, line 50: `if name not in ...` ([pep8 style](https://peps.python.org/pep-0008/))

sirepo/raydata/databroker.py
moellep: no reason, other than it wasn't already imported to the module. I've updated it now

PR #6789, opened 2024-02-19
Fix #6779 improve supervisor logging and asynchrony

### LINKED ISSUE #6779: Improve logging in the supervisor to track hangs

robnagler:
- increase purge_free_simulations asynchrony to avoid I/O blocking
   which may be an issue with a few recent failures we have seen.
- Added a bit more logging, especially start/done on purging

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -368,11 +368,11 @@ class DriverBase(PKDict):
 
         All slots are allocated and only freed when the op is
         destroyed. We don't need to recheck the slots, because
-        `destroy_op` cancels this task. `_agent_ready` is state held
+        job_supervisor.destroy_op frees the slots. `_agent_ready` is state held
         outside this op so it needs to be rechecked when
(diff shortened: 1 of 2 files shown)


moellep (approved):
Seems to work fine. 

What it the advantage of putting the purge jobs in a timer within job_supervisor, vs running tasks like that from outside the server (like a command from cron)?

robnagler:
> What it the advantage of putting the purge jobs in a timer within job_supervisor, vs running tasks like that from outside the server (like a command from cron)?

I actually explained that in the email this morning. :) "We designed the supervisor so that locks could be completely avoided, and the coro-thread model is probably enough to ensure that the supervisor doesn't do too much work and remains a switch." 

The job_supervisor assumes it is the only process which looks at the `supervisor-job` directory so we don't need locks when modifying job files and therefore can keep them in memory and write to disk only for persistence. If this were managed by a db, this wouldn't be an issue, but this would complicate the job supervisor, since it would have to retrieve the records of the job and couldn't keep them in memory.

PR #6790, opened 2024-02-19
Fix #6416: no materials export

### LINKED ISSUE #6416: export as zip with no materials causes server error
In cloudmc, if you have a geometry, but haven't assigned materials to the volumes, and try to export-as-zip, it causes a server error.

gurhar1133:


sirepo/template/cloudmc.py
```diff
@@ -469,24 +469,24 @@ def _generate_distribution(dist):
     return _generate_call(t, args)
 
 
-def _generate_materials(data):
+def _generate_materials(data, v):
```
e-carlin, line 472: Use a descriptive variable name (j2_ctx?) when it is a function parameter (here and in the other methods). This will also mean you don't need to rename d to v.

sirepo/package_data/template/cloudmc/parameters.py.jinja
```diff
@@ -1,3 +1,6 @@
+{% if incomplete_data_msg != "" %}
+# Unable to generate sim due to: {{ incomplete_data_msg }}
```
e-carlin, line 2: I think just testing for `if incomplete_data_msg` is fine. A non-empty string in python is True.

What about raising an error? If someone tries to run the file w/o looking at it if it just has the comment it will look like it succeeded but didn't do anything. Raising an error I think will help them more quickly see what's wrong.

e-carlin (changes requested):

sirepo/package_data/template/cloudmc/parameters.py.jinja
gurhar1133: @moellep what do you think about generating an error that will be caught by the parser in the parameters.py? That way _generate_parameters() will run and the export will work, but if the user tries to run the sim it will be clear what is wrong

gurhar1133: @moellep ping

moellep: Yes, I think that makes sense, something like this:
```python
{% if incomplete_data_msg != "" %}
raise AssertionError('{{ incomplete_data_msg }}')
{% else %}
```

moellep: Also, along these lines, I don't think you need to include incomplete_data_msg in the jinja, just return the string directly:
```python
    if v.incomplete_data_msg:
        return v.incomplete_data_msg
    return template_common.render_jinja(
        SIM_TYPE,
        v,
    )
```

moellep (changes requested):

sirepo/template/cloudmc.py
```diff
@@ -546,14 +547,17 @@ def _generate_parameters_file(data, run_dir=None):
...
+    v.sources = _generate_sources(data, v)
     v.sourceFile = _source_filename(data)
     v.maxSampleSourceParticles = SCHEMA.model.openmcAnimation.numSampleSourceParticles[
         5
     ]
-    v.tallies = _generate_tallies(data)
+    v.tallies = _generate_tallies(data, v)
     v.hasGraveyard = _has_graveyard(data)
+    if v.incomplete_data_msg:
+        return f'raise AssertionError("{"Unable to generate sim:" + v.incomplete_data_msg}")'
```
e-carlin, line 560: You don't want to do string concatenation within `{}` in an f-string. I think this is the same but with less quoting.

```
f'raise AssertionError("Unable to generate sim: {v.incomplete_data_msg}")'
```

e-carlin (changes requested):

e-carlin (approved):

moellep (approved):

PR #6791, opened 2024-02-20
Fix #6719 add petdb example

### LINKED ISSUE #6719: Create petdb example for activait
This simulation should be turned into a Sirepo example for activait:
https://www.sirepo.com/activait#/data/se8UEFrU

The input data file is large, and should not be kept in the sirepo repo, but accessed similar to the way the cloudmc app downloads large geometry files for example sims:
https://github.com/radiasoft/sirepo/blob/master/sirepo/template/cloudmc.py#L138

Benbenbenin0:
- Added simulation to template examples
- Created [github repo](https://github.com/radiasoft/sirepo-data-activait) for storing large example files
- Implemented downloading split simulation data from repo then saving combined data locally

NOTE:
- There was an issue with the simulation looking for the sample data before it was downloaded/combined, added a 3 second wait in `activait.js` before loading sample data.
- Looking for suggestions on `stateful_compute_download_remote_lib_file()` in `activait.py,` needs refactor but running low on hours.
- First sirepo PR in a while, likely requires rigorous review.


sirepo/package_data/static/js/activait.js
```diff
@@ -1233,8 +1233,9 @@ SIREPO.app.directive('imagePreviewPanel', function(requestSender) {
                     }
                 );
             };
-
-            loadImageFile();
+            setTimeout(function(){
```
e-carlin, line 1236: You noted this in your commit and this is a major code smell. What if things are slow and it takes longer than 3s? Instead of doing this we need to structure the code so whatever needs to be ready is ready when this is called. Or make this function somehow protect itself. Lmk if you want to meet and figure out how to do that.

sirepo/package_data/static/json/activait-schema.json
```diff
@@ -291,7 +291,9 @@
             "featureRangeMax": ["Feature range maximum", "Integer", 1],
             "inputsScaler": ["Inputs Scaler", "Scaler", "RobustScaler"],
             "outputsScaler": ["Outputs Scaler", "Scaler", "RobustScaler"],
-            "url": ["Data URL", "URL", ""]
+            "url": ["Data URL", "URL", ""],
+            "exampleFile": ["", "String", ""],
```
e-carlin, line 295: I don't really like this structure. It doesn't allow for growth in the future. What happens when there are 3 files? I'd get Paul's thoughts on this. But, one idea is we have a name like "exampleDirectory" there is then a name with that directory in sirepo-data-activait. That dir can have an arbitrary number of files. If there is one file then we just save it. If there is more than one file then we combine them. If ordering matters then give the files names like 00.h5, 01.h5, etc and use sorted glob to order and iterate over them.

e-carlin (changes requested):
Welcome back:) Code worked well but I left some comments on the structure.

gurhar1133:
@moellep and @e-carlin can you guys take a look at this one for me? The main change I made is that there is loading indication that hides the image preview panel and column selectors while the datafile is loading.

Once this is merged I can add some more examples that will download datafiles from remote URL.

gurhar1133:
@moellep @e-carlin ping

sirepo/package_data/static/js/activait.js
```diff
@@ -1388,7 +1418,6 @@ SIREPO.app.directive('imagePreview', function(appState, requestSender, panelStat
                     $scope.colBName = 'Images';
                 }
             }
-
             initFromResponse($scope.imageInfo);
         }
     };
```
moellep, line 1423: I opened the Oxford Pets example on a fresh sirepo database, and saw this error before it loaded the file normally:

```Application Error: TypeError: Cannot read properties of undefined (reading 'setCustomValidity'). ```
```
angular.js:15567 TypeError: Cannot read properties of undefined (reading 'setCustomValidity')
    at $scope.validateNumSelected (activait.js?20240328.184633:1161:19)
    at setModel (activait.js?20240328.184633:1103:24)
    at activait.js?20240328.184633:1175:17
    at m.$eval (angular.js:19393:16)
    at b (angular.js:19535:11)
    at t (angular.js:19793:11)
    at m.$digest (angular.js:19096:11)
    at m.$apply (angular.js:19500:13)
    at angular.js:21601:25
    at Tg.completeTask (angular.js:21231:7)
(anonymous) @ angular.js:15567
```

gurhar1133:
@moellep I think this is good now

gurhar1133:
@e-carlin ping

e-carlin (approved):

PR #6797, opened 2024-02-20
Fix #6783 job runCancel & timeout separated and simplified

### LINKED ISSUE #6783: cancel of an analysis job also cancels compute job
This only happens (currently) on a timeout of the analysis job. 

A cancel of a compute job should not cancel an analysis job. It may be the compute job is canceled, because it hasn't converged, and still has valid data for the analysis job. 

The way job_supervisor.api_runCancel is written, one analysis job cancels other analysis jobs.

Simplify this to only cancel the requested job.

robnagler:
- runCancel only cancels the OP_RUN
- timeouts cancel the timed out op 
- cancel op creation simplified and not connected to a _ComputeJob since it isn't
- job_agent StreamClosedError doesn't print stack trace
- job_agent connect timeout is 10 secs
- job_cancel_test is better name than cancel_delay_test


moellep (approved):

sirepo/job_supervisor.py
```diff
@@ -257,9 +254,44 @@ async def receive(cls, req):
...
+                _supervisor=_Supervisor(),
+                api="cancel_or_timeout",
+                driver=to_cancel.driver,
+                is_destroyed=False,
+                max_run_secs=None,
+                msg=PKDict(opIdsToCancel=[to_cancel.op_id]),
+                op_name=job.OP_CANCEL,
+                uid=to_cancel.driver.uid,
+            )
+
```
moellep, line 273: When I set the SIREPO_JOB_SUPERVISOR_MAX_SECS_IO to a small value, and try to download a file, it fails with a "413 content to large". Is that the expected error, or should it be some kind of http timeout?

robnagler: Thanks for testing. 

698e3a59ef34eb3b73284cff515b89d3be2f0f47 added this to job_api:

```py
                if r.state == sirepo.job.CANCELED:
                    # POSIT: Users can't cancel donwloadDataFile. So canceled means there was a
                    # timeout (max_run_secs exceeded).
                    raise _content_too_large(req)
```

moellep: I'll open a new issue for this - it would be too confusing on the client if it returns "content too large" for a timeout. Even a "resource unavailable" or similar would be less confusing.

PR #6821, opened 2024-02-28
fix #6588 Allow analysis queue to be reordered

### LINKED ISSUE #6587: add Queue add time
Scans should be processed in the order they have been added to the queue. When an item is added, an additional column (add time, or similar) should be saved, and displayed in the UI for sorting.

### LINKED ISSUE #6588: Reorder scan queue
Related to #6587, it should be possible to move items around in the queue, possibly by changing the "queue add time".

moellep:
also fix #6587 show queue order on pending list and allow sorting on queue order 
Scan items may be moved to the front or back of the list


rorour (approved):
Looks great!

PR #6823, opened 2024-02-28
Fix #6819 heatmap offset

### LINKED ISSUE #6819: Heatmap colorbar pointer offset when PLOTTING_HEATPLOT_FULL_PIXEL set
The size of the the image data gets changed by the "pixel size" but the conversion of the mouse point to a heatmap value does not. Thus the value is offset, increasing to a full pixel as you move from the center to the edges. See these examples from elegant with superimposed crosshairs indicating the nominal "center" of a pixel:
PLOTTING_HEATPLOT_FULL_PIXEL true:
image
PLOTTING_HEATPLOT_FULL_PIXEL false (normal operation):
image

mkeilman:
Notes:
- Heatmap indices are calculated from the full length of the arrays if `SIREPO.PLOTTING_HEATPLOT_FULL_PIXEL` is set
- The indices are also floored (rather than rounded) in that case
- Indices renamed to `i, j` from `x, y` to distinguish them from geometry
- Current coding conventions applied (`var` to `const`, `let` etc,)


moellep:
This is a good fix.

PR #6840, opened 2024-03-04
fix #6033 add weight windows options

### LINKED ISSUE #6033: Add weight window parameters to cloudmc
Add support for weight windows. Choose a good default based on the input geometry bounds.

https://docs.openmc.org/en/stable/pythonapi/generated/openmc.WeightWindows.html

### LINKED ISSUE #6829: Add sourceNormalization back to tallySettings
The sourceNormalization field was inadvertently  removed from the tallySettings view and should be added back.

moellep:
also fix #6829 add source normalization field back to UI


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -71,15 +71,20 @@ SIREPO.app.config(() => {
...
         </div>
-        <div data-ng-switch-when="TallyList" data-ng-class="fieldClass">
+        <div data-ng-switch-when="PlotTallyList" data-ng-class="fieldClass">
           <div class="input-group">
             <select class="form-control" data-ng-model="model[field]" data-ng-options="t.name as t.name for t in model.tallies"></select>
           </div>
         </div>
+        <div data-ng-switch-when="SettingsTallyList" data-ng-class="fieldClass">
+          <div class="input-group">
+            <select class="form-control" data-ng-model="model[field]" data-ng-options="t as t.name for t in appState.models.settings.tallies track by t.name"></select>
```
mkeilman, line 81: Either replace a null value with the first tally or add an "empty" option e.g.
```html
<select class="form-control" data-ng-model="model[field]" data-ng-options="t as t.name for t in appState.models.settings.tallies track by t.name">
    <option value="" disabled selected>Select a tally</option>
</select>
```
If the latter, all such dropdowns should have a similar option.

sirepo/package_data/static/json/cloudmc-schema.json
```diff
@@ -263,18 +269,18 @@
     },
     "model": {
         "box": {
-            "_super": [ "_", "model", "spatial"],
+            "_super": ["_", "model", "spatial"],
```
mkeilman, line 272: Thanks for cleaning this up

sirepo/template/cloudmc.py
```diff
@@ -577,6 +584,26 @@ def _generate_range(filter):
...
 
 
+def _generate_run_mode(data, v):
+    r = data.models.openmcAnimation.jobRunMode
+    cores = 1 if r == "sequential" else sirepo.mpi.cfg().cores
+    if v.isPythonSource:
+        cores = 0
+    if data.models.settings.varianceReduction == "weight_windows_tally":
+        if v.isSBATCH:
+            raise AssertionError("Weight Windows are not yet available with sbatch")
```
mkeilman, line 594: This is fine, but I think there should also be equivalent logic on the client side to disable running with sbatch if this setting is in place.

mkeilman (changes requested):
Just a couple of items

sirepo/package_data/static/js/cloudmc.js
moellep: I've added a proper tallyList directive now which will set the value if no item is selected. c866f2a

mkeilman (approved):
good to go

Issue #6825, opened 2024-02-28
Convert activait image previews to reports


moellep:
There are 4 panels in Activait which show image previews. The plots use either stateful-compute or analysis-jobs to get the data. This has 2 drawbacks:
- The jobs can't be canceled when switching between tabs, so if you switch between tabs quickly, you end up waiting for the previous tab's items to complete. Reports automatically cancel when you switch between tabs.
- There is no caching of the results. Reports cache reply values and will use the cache unless the input fields change.

The fix would be to change imagePreviewPanel to a plot type, and add these models:
ImageSampleReport
SegmentSampleAnimation
BestLossImageAnimation
WorstLossImageAnimation

PR #6856, opened 2024-03-06
Fix #6722: compare model outputs

### LINKED ISSUE #6722: Comparing model outputs
It would be very useful to be able to compare results between two different models. If two activait sims share the same dataset, it should be possible to show some side-by-side results for comparison.

This could be implemented as a Model Comparison tab which could show the same input image/contour and the predicted contour from both models.

The Comparison tab could be added for image-to-image datasets initially. It would show a sim selector which populates with other activait sims using the same dataset. (Maybe even multiple comparison sims could be selected).

gurhar1133:
Image to image model comparison. Used pets dataset to test this one

### DIFF
--- a/sirepo/package_data/static/js/activait.js
+++ b/sirepo/package_data/static/js/activait.js
@@ -347,7 +347,9 @@ SIREPO.app.controller('ComparisonController', function (activaitService, appStat
             self.compare = true;
             otherSimId = appState.models.otherSims.otherSims;
             appState.models.dicePlotComparisonAnimation.otherSimId = otherSimId;
+            appState.models.epochComparisonAnimation.otherSimId = otherSimId;
             appState.saveChanges('dicePlotComparisonAnimation');
+            appState.saveChanges('epochComparisonAnimation');
             // appState.saveChanges('otherSims');
         } else {
             srdbg("no selection");
--- a/sirepo/package_data/static/js/activait.js
+++ b/sirepo/package_data/static/js/activait.js
@@ -10,6 +10,10 @@ SIREPO.app.config(() => {
         'epochComparisonAnimation',
         'dicePlotAnimation',
         'dicePlotComparisonAnimation',
+        'bestLossesAnimation',
+        'bestLossesComparisonAnimation',
+        'worstLossesAnimation',
+        'worstLosssComparisonAnimation',
     ];
     SIREPO.PLOTTING_HEATPLOT_FULL_PIXEL = true;
     SIREPO.FILE_UPLOAD_TYPE = {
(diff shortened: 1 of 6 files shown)


sirepo/package_data/static/js/activait.js
```diff
@@ -249,6 +259,14 @@ SIREPO.app.directive('appHeader', function(appState, activaitService) {
                 }
                 return false;
             };
+            $scope.isImageToImage = () => {
+                var info = appState.models.columnInfo;
+                if (info) {
+                    var idx = info.inputOutput.indexOf('output');
+                    return info.shape[idx].slice(1, info.shape[idx].length).length > 1;
+                }
+                return false;
```
moellep, line 268: In the pets example, selecting the checkboxes in the Columns panel is showing a "undefined reading 'slice'" javascript error

sirepo/package_data/static/js/activait.js
```diff
@@ -1091,14 +1147,78 @@ SIREPO.app.directive('columnSelector', function(appState, activaitService, panel
...
+            $scope.load = function(json) {
+                $scope.imageInfo = json.images;
+                $scope.imageInfo.method = "imagePreview";
+            };
+        },
+        link: function link(scope, element) {
+            plotting.linkPlot(scope, element);
+        },
+    };
+});
```
moellep, line 1171: I replaced the "imageViewer" view which uses the old stateless-compute style of loading images with a proper imageSamplesReport. This allows the server to cache the results. This will give you an example for the other items in #6825 which should replace all the image stateless-compute calls with Report or Animation models.

moellep (changes requested):

sirepo/package_data/static/js/activait.js
moellep: I also replace the stateless-compute calls on the regression tab with Animation plots. @gurhar1133, could you update the plots on the comparison tab with similar changes?

sirepo/package_data/static/js/activait.js
```diff
@@ -249,6 +263,14 @@ SIREPO.app.directive('appHeader', function(appState, activaitService) {
                 }
                 return false;
             };
+            $scope.isImageToImage = () => {
+                var info = appState.models.columnInfo;
+                if (info && info.inputOutput.indexOf('output') >= 0) {
+                    var idx = info.inputOutput.indexOf('output');
+                    return info.shape[idx].slice(1, info.shape[idx].length).length > 1;
+                }
```
moellep, line 271: When viewing the "2019 World Happiness" example, I'm seeing a "cannot read properites of undefined" in isImageToImage()

moellep (changes requested):

moellep (approved):

PR #6877, opened 2024-03-12
fix #6876 moved command drag component to badge

### LINKED ISSUE #6876: allow selecting text for lattice commands
With the current drag-n-drop feature, the command text can't be copied and pasted. Instead, only the badge part of the row should be draggable.

moellep:


gurhar1133 (approved):
Looks good. Worked well on safari, chrome and firefox

PR #6883, opened 2024-03-13
Fix #6833: sort tally

### LINKED ISSUE #6833: Sort tally scores by name
After the tally scores have been selected, they should get sorted by category (as they appear in the score list).

For example, the Tokamak Reactor example, the tallies below should have the scores in the same order.
x

gurhar1133:


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -346,6 +346,22 @@ SIREPO.app.controller('VisualizationController', function(appState, authState, c
...
+                if (! v[s.score]) {
+                    r.push(s);
+                    v[s.score] = true;
+                }
+            }
+            t.scores = r.sort((a, b) => a.score.localeCompare(b.score));
+            appState.saveQuietly('settings');
+        }
+    });
+
```
moellep, line 364: I made a few changes to this PR. In order to get the tally score sorting right everywhere (including on the score dropdown), I sort the values before they get saved to the model.
x

I found that if you had a score accidentally duplicated (like two "flux" scores in the list), openmc would fail, so I also added a uniqueness check.

- Use localeCompare() rather than sort() for strings in general so you don't get ascii sorting, and things remain alphabetical regardless of case.

@gurhar1133 give it a test and feel free to make the sorting/unique method better.

gurhar1133: Yeah, good call. I just made it do the sort on load since some of the examples start with unsorted tally lists.

PR #6888, opened 2024-03-14
Fix #6863: slider cleanup

### LINKED ISSUE #6863: Volumes list has chart junk
The volumes list in cloudmc is too noisy currently with repeated 0, 1 ranges:

x

In this case since all the ranges are the same, it should only be shown on the first item, or left off all together, and only show the selected value.

gurhar1133:


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2705,13 +2706,14 @@ SIREPO.app.directive('jRangeSlider', function(appState, panelState) {
             model: '=',
             modelName: '<',
             onChange: '&',
+            sliderIndex: '=',
```
mkeilman, line 2709: sliderIndex is not the best thing to use here; there are likely other conditions that may require hiding the labels. Something like hideMinMax is more flexible

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2801,16 +2803,24 @@ SIREPO.app.directive('jRangeSlider', function(appState, panelState) {
...
                 if (range.space === 'linear') {
-                    return Array.isArray(v) ? v.map(x => $scope.formatFloat(x)) : $scope.formatFloat(v);
+                    return Array.isArray(v) ? v.map(x => formatFloat(x)) : formatFloat(v);
+                }
+                return Array.isArray(v) ? v.map(x => formatFloat(toLog(x, range))) : formatFloat(toLog(v));
+            };
+
+            const formatFloat = val => SIREPO.UTILS.formatFloat(val, 4);
+
+            $scope.formatMinMax = val => {
```
mkeilman, line 2818: We've decided on using parens around single parameters: `const formatFloat = (val) => {...}` etc.

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2801,16 +2803,24 @@ SIREPO.app.directive('jRangeSlider', function(appState, panelState) {
...
+                if ($scope.sliderIndex === 0) {
+                    return formatFloat(val);
                 }
-                return Array.isArray(v) ? v.map(x => $scope.formatFloat(toLog(x, range))) : $scope.formatFloat(toLog(v));
+                return "";
             };
```
mkeilman, line 2823: More concise is
```
$scope.formatMinMax = (val) => $scope.hideMinMax ? '' : formatFloat(val);
```

mkeilman (changes requested):
A couple of suggestions...

mkeilman (approved):
One nitpick, then clear to merge.

gurhar1133:
@moellep take a peak?

moellep (approved):

PR #6891, opened 2024-03-14
Fix #6809 missing terminate call

### LINKED ISSUE #6809: adm_and_own_jobs_test leaves agent running
Even when run successful, the job_agent is running. This can cause interesting confusion in the logs, because the old processes continue to write into the log. The job_agent will try to reconnect to a non-existent supervisor which results in the logs growing (over the course of a day to 1GB). This leaves "holes" in the file which show up as nulls when read. This makes emacs go crazy when parsing the output. This null problem has been around for a long time, and I think this is the main cause so glad this issue is clear.

To be clear, other tests leave the agent running, too.

### LINKED ISSUE #6887: remove pytest cruft since only runs under `pykern test`

robnagler:
Fix #6887 each call to pytest_runtest_protocol is new,
because pykern.pkcli.test runs each test in a spearate process.

### DIFF
--- a/sirepo/srunit_servers.py
+++ b/sirepo/srunit_servers.py
@@ -3,6 +3,7 @@
 :copyright: Copyright (c) 2024 RadiaSoft LLC.  All Rights Reserved.
 :license: http://www.apache.org/licenses/LICENSE-2.0.html
 """
+
 import contextlib
 
 # limit sirepo/pykern global imports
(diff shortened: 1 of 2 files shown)


gurhar1133 (approved):
LGTM. It worked well when I tested with `adm_and_own_jobs_test`. Didn't leave a bunch of job_agent processes around

robnagler:
I will merge this and https://github.com/radiasoft/sirepo/pull/6895 tomorrow. There are a lot of changes, but they should be fairly easy for people to merge. It's necessary to do these quickly so that people's branches can just be merged from master.

PR #6896, opened 2024-03-15
Fix #6368: re process geometry when damcAnimation directory is not there

### LINKED ISSUE #6368: better UI when cloudmc geometry needs rebuilding
Currently if the dagmcAnimation directory has been deleted for a cloudmc simulation, the browser shows an empty geometry, and when the simulation is reloaded, it rebuilds the geometry. Instead, it should automatically start rebuilding the geometry if the directory has been removed.

gurhar1133:


sirepo/package_data/static/js/cloudmc.js
```diff
@@ -245,23 +248,43 @@ SIREPO.app.controller('GeometryController', function (appState, cloudmcService,
...
+            },
+            {
+                method: 'check_animation_dir',
+                args: {
+                    modelName: 'dagmcAnimation',
+                },
+            },
+        );
+    }
+
```
moellep, line 271: I made a few changes:
- don't call initSimulationState() until the run_dir status is known. Otherwise it could make a call to run-status before the stateful compute has finished and think it has volumes to display.
- moved self.isProcessing to a local variable hasGeometry because it isn't referenced outside the controller
- changed stateless compute to stateful compute because it checks state on the server (whether the directory exists)
- Moved the stateful compute argument within an "args" value to keep it out of the top-level argument space.

@gurhar1133 give it a try - I tested in several cases (discard changes, cancel geometry, new simulation)

gurhar1133: Yes, this is good. Should I merge?

PR #6901, opened 2024-03-18
Fix #6867: in use file error

### LINKED ISSUE #6867: delete file should return an error if the file is in use
When you try to delete an InputFile for a simulation, it should validate that the file is not in use in another simulation. There appears to be some logic for this, but it is not working, at least with elegant.

To reproduce:
Open the elegant "bunchComp - fourDipoleCSR" example, edit the zwake element, download the INPUTFILE field to your computer and rename the file to test.sdds. Then open a copy of that simulation and change the zwake input field to a new file created previously, "test.sdds". Run the simulation and check that it completes successfully. Then open the original "bunchComp - fourDipoleCSR" example, edit the zwake element and delete the "test.sdds" file. It should show an error that the file is in use (preferably with a link to the simulation using the file). When you return to the simulation copy and run it, it will fail with a missing datafile error.

gurhar1133:
some calls to _lib_file_basenames for some sims will return full filenames with MODEL-FIELD.fname.ext and some will return fname.ext.

Tested this with SRW, opal, genesis, elegant. LMK if should test with more sims. Also seems to correctly delete unused files.

### DIFF
--- a/sirepo/package_data/static/js/cloudmc.js
+++ b/sirepo/package_data/static/js/cloudmc.js
@@ -323,11 +323,6 @@ SIREPO.app.controller('VisualizationController', function(appState, authState, c
         return `Completed batch: ${self.simState.getFrameCount()}`;
     };
     self.startSimulation = function() {
-        if (appState.applicationState().settings.varianceReduction == 'weight_windows_tally'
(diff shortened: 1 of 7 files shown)


moellep (approved):
Looks good - I'm glad it was such a simple fix

PR #6921, opened 2024-03-20
fix #6350: test all schema-common routes

### LINKED ISSUE #6350: 20231115 Test all schema-common.json routes
We should have the ability to test all routes. I removed errorLogging accidentally, and we don't have a test for it. It's an important route to have a test for. I think we can make sure we have tests for all routes so this is an audit task at first and then we'll create tests for the missing routes.

rorour:
fix https://github.com/radiasoft/sirepo/issues/6350: test all schema-common routes, remove github auth


sirepo/package_data/static/json/schema-common.json
```diff
@@ -559,7 +555,7 @@
         "502": {
             "url": "http-proxy-error.html",
             "msg": "Proxy Error",
-            "route": "proxy-error"
+            "route": "proxyError"
```
robnagler, line 558: A lot of nice catches, thanks.

tests/jupyterhub_test.py
```diff
@@ -0,0 +1,27 @@
...
+def setup_module(module):
+    os.environ.update(
+        SIREPO_FEATURE_CONFIG_PROPRIETARY_SIM_TYPES="jupyterhublogin",
+        SIREPO_AUTH_ROLE_MODERATION_MODERATOR_EMAIL="x@x.x",
+    )
+
+
+def test_check_auth_jupyterhub(fc):
+    fc.sr_login_as_guest()
+    # user doesn't exist
```
robnagler, line 20: This should validate that the jupyter user doesn't exist and does exist by checking the directory in srdb.

tests/jupyterhub_test.py
```diff
@@ -0,0 +1,27 @@
...
+
+def setup_module(module):
+    os.environ.update(
+        SIREPO_FEATURE_CONFIG_PROPRIETARY_SIM_TYPES="jupyterhublogin",
+        SIREPO_AUTH_ROLE_MODERATION_MODERATOR_EMAIL="x@x.x",
+    )
+
+
+def test_check_auth_jupyterhub(fc):
+    fc.sr_login_as_guest()
```
robnagler, line 19: Also would be good to checkAuthJupyterHub without a login.

tests/server_test.py
```diff
@@ -15,6 +15,40 @@ def test_elegant_run_file(fc):
     _get_file(fc, _run_elegant(fc), "downloadRunFile")
 
 
+def test_error_logging(fc):
+    from pykern.pkcollections import PKDict
+    from pykern.pkunit import pkeq
+
+    r = fc.sr_post("errorLogging", PKDict(message="error message"))
```
robnagler, line 22: temporary is unnecessary

tests/server_test.py
```diff
@@ -15,6 +15,40 @@ def test_elegant_run_file(fc):
...
+
+def test_favicon(fc):
+    from pykern.pkcollections import PKDict
+    from pykern.pkunit import pkeq
+    import sirepo.resource
+
+    for a, p in PKDict(favicon="favicon.ico", faviconPng="favicon.png").items():
+        r = fc.sr_get(a)
+        with open(sirepo.resource.static("img", p), "rb") as f:
+            s = f.read()
```
robnagler, line 34: temporary unnecessary

tests/custom_errors_test.py
```diff
@@ -0,0 +1,21 @@
...
+def test_custom_errors(fc):
+    from pykern.pkunit import pkre
+    import sirepo.simulation_db
+    import sirepo.uri
+
+    for k, v in sirepo.simulation_db.SCHEMA_COMMON.customErrors.items():
+        r = fc.sr_get(sirepo.uri.local_route(fc.sr_sim_type, v.route))
+        r.assert_success()
+        with open(sirepo.resource.static("html", v.url), "r") as f:
+            s = f.read()
```
robnagler, line 20: temporary unnecessary

tests/server_test.py
```diff
@@ -15,6 +15,40 @@ def test_elegant_run_file(fc):
...
+    pkeq("ok", r.state)
+
+
+def test_favicon(fc):
+    from pykern.pkcollections import PKDict
+    from pykern.pkunit import pkeq
+    import sirepo.resource
+
+    for a, p in PKDict(favicon="favicon.ico", faviconPng="favicon.png").items():
+        r = fc.sr_get(a)
```
robnagler, line 32: temporary unnecessary.

tests/server_test.py
```diff
@@ -23,6 +57,29 @@ def test_myapp_basic(fc):
...
+
+def test_simulation_schema(fc):
+    from pykern.pkcollections import PKDict
+    from pykern.pkunit import pkexcept
+
+    r = fc.sr_post_form(
+        "simulationSchema",
+        data=PKDict(simulationType=fc.sr_sim_type),
+    )
+
```
robnagler, line 72: blank lines unnecessary

tests/server_test.py
```diff
@@ -23,6 +57,29 @@ def test_myapp_basic(fc):
...
+    for k in ("model", "view"):
+        assert k in r.keys()
```
robnagler, line 74: Maybe check a specific value, e.g.

`pkeq("/simulation-schema", r.route.simulationSchema)`

Testing for keys is probably ok, too, but it's pretty arbitrary, and a one-liner is sufficient for this particular purpose.

tests/server_test.py
```diff
@@ -23,6 +57,29 @@ def test_myapp_basic(fc):
...
+
+    with pkexcept("unexpected status"):
+        fc.sr_post_form(
+            "simulationSchema",
+            data=PKDict(simulationType="xyzzy"),
```
robnagler, line 79: If we ever get a code named after [xyzzy](https://en.wikipedia.org/wiki/Xyzzy_(computing)), all our tests will break. ;-)

tests/server_test.py
```diff
@@ -39,6 +96,15 @@ def test_srw(fc):
         r.assert_http_status(200)
 
 
+def test_srw_light(fc):
+    from pykern.pkunit import pkre
+
+    r = fc.sr_get("srwLight")
+    r.assert_success()
+    pkre("<!DOCTYPE html>", r.data)
```
robnagler, line 104: Make a single regex, and can simply check for the existence of SRWLightGateway. The data-ng-app is unnecessary.

robnagler (changes requested):
Looks great! This was a lot of work, and worth it imho. Minor comments.

tests/jupyterhub_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/jupyterhub_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/custom_errors_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: I'm not really understanding the benefit of that example. Are we not testing to see that it returned a valid schema? I added a check more specific to myapp, let me know what you think. 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/server_test.py
rorour: 1a08a18cb5e9c42888b493a7621bd56f9ac7faea

tests/jupyterhub_test.py
```diff
@@ -16,11 +16,33 @@ def setup_module(module):
...
 def test_check_auth_jupyterhub(fc):
+    import sirepo.srdb
+    from pykern import pkio
+    from pykern.pkunit import pkeq
+
+    def _get_num_jupyterhub_user_dirs():
+        return len(pkio.sorted_glob(sirepo.srdb.root().join("jupyterhub/user/**"))) - 1
+
+    def _jupyter_user_dir_exists(uid):
+        for x in [
```
robnagler, line 27: I think this could be simpler:

`sirepo.srdb.root().join("jupyterhub", "user", uid.lower()).exists(dir=True)`

tests/jupyterhub_test.py
```diff
@@ -16,11 +16,33 @@ def setup_module(module):
...
+                return True
+        return False
+
+    fc.sr_logout()
+    n = _get_num_jupyterhub_user_dirs()
+    fc.sr_get("checkAuthJupyterHub")
+    pkeq(n, _get_num_jupyterhub_user_dirs())
     fc.sr_login_as_guest()
-    # user doesn't exist
+    pkeq(False, _jupyter_user_dir_exists(fc.sr_uid))
```
robnagler, line 40: It's generally [not good style](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html) to test for equals False or True.
`pkok(not _jupyter_user_dir_exists(fc.sr_uid))`

tests/server_test.py
```diff
@@ -60,16 +60,15 @@ def test_simulation_redirect(fc):
...
 
     r = fc.sr_post_form(
         "simulationSchema",
         data=PKDict(simulationType="myapp"),
     )
-    for k in ("model", "view"):
-        assert k in r.keys()
-    for k in ("dog", "heightWeightReport"):
-        assert k in r.model.keys()
+    pkok(isinstance(r.model, PKDict), "expected model to be PKDict")
```
robnagler, line 69: This check is sufficiently tested by the r.model.dog check. It's better to not test too much stuff. If something comes back with r.model.dog as an attribute, then it's likely a schema. It could be data so you could go one level deeper, but that's not likely to be broken. Tests should be as minimal as possible to accomplish the goal.

robnagler (changes requested):

tests/jupyterhub_test.py
rorour: sometimes the jupyter username is prefixed with guest-, so this won't work

tests/jupyterhub_test.py
rorour: [e474deb](https://github.com/radiasoft/sirepo/pull/6921/commits/e474debba2686e7e319531ca53708db538f023f7)

tests/server_test.py
rorour: [e474deb](https://github.com/radiasoft/sirepo/pull/6921/commits/e474debba2686e7e319531ca53708db538f023f7)

tests/jupyterhub_test.py
```diff
@@ -16,11 +16,33 @@ def setup_module(module):
...
+            os.path.basename(d)
+            for d in pkio.sorted_glob(sirepo.srdb.root().join("jupyterhub/user/**"))
+        ]:
+            if uid.lower() in x:
+                return True
+        return False
+
+    fc.sr_logout()
+    n = _get_num_jupyterhub_user_dirs()
+    fc.sr_get("checkAuthJupyterHub")
```
robnagler, line 37: I think we should check that we get back a login redirect.

tests/jupyterhub_test.py
robnagler: We should know when this happens, and we should code for it. At a minimum, this test should work:

`bool(len(pkio.sorted_glob(sirepo.srdb.root().join("jupyterhub", "user", f"*{uid.lower()}*")))`

This is a cleaner condition, because it leverages glob as opposed to working around glob. The `**` in the code is not right, because it shouldn't be recursive. The user directories are at a flat level. uid is random and that's a good enough test on a clean db, which this is.

robnagler (changes requested):
Almost!

tests/jupyterhub_test.py
rorour: [b92ef48](https://github.com/radiasoft/sirepo/pull/6921/commits/b92ef4813f20692833d3b820375d74e338453758)

robnagler (approved):

PR #6929, opened 2024-03-22
Fix #6527 cloudmc 1d energy plot

### LINKED ISSUE #6527: Add 1d plot of a specific energy bin when an energy filter is active
- [x] #6132
"See the notebook at EFilterExample.ipynb for an example."

mkeilman:
This is the 1d energy plot, as far as it goes. Probably needs some cleanup. Notes:
- If a tally has an energy filter, the 2d plot displays crosshairs on the cursor. The coordinates and the heatmap value appear at the top left
- Select a cell with option/alt click (regular click gets called after a mouseup so it selects a cell if you drag the image; control-click pops up a context menu)
- The fields enabling these lives in the tallyReport model, so they can be manually set/unset if needed
- Once a cell is selected a parameter plot for that gets displayed. It shows all the scores but could be tailored to display on the selected score


sirepo/template/cloudmc.py
```diff
@@ -408,6 +401,65 @@ def _dagmc_animation_python(filename):
...
+
+    tally = _get_tally(data.models.settings.tallies, tally_name)
+    mesh = _get_filter(tally, "meshFilter")
+    e = _get_filter(tally, "energyFilter")
+    r = data.models.energyAnimation
+    mean = numpy.reshape(
+        getattr(t, "mean")[:, :, t.get_score_index(r.score)].ravel(),
+        (*mesh.dimension, -1),
+    )
+    # std_dev = getattr(t, "std_dev")[:, :, t.get_score_index(r.score)].ravel()
```
gurhar1133, line 433: comment needed?

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -3048,14 +3072,85 @@ SIREPO.viewLogic('tallySettingsView', function(appState, cloudmcService, panelSt
...
+            $scope.$watch('tallyService.fieldData', (newValue, oldValue) => {
+                if (newValue !== oldValue) {
+                    updateRange();
+                }
+            });
+        },
+    };
+});
+
+SIREPO.app.directive('planePositionSlider', function(appState, panelState, tallyService) {
```
gurhar1133, line 3128: Maybe a shared directive for energyRangeSlider and planePositionSlider? Seems like a lot of similarity

sirepo/package_data/static/js/cloudmc.js
```diff
@@ -2191,54 +2197,82 @@ SIREPO.app.directive('multiLevelEditor', function(appState, panelState) {
...
         },
         link: function(scope, element, attr, ngModel) {
             function validate() {
-                const t = scope.field.val;
-                const hasVals = ! t.some(x => x == null);
+                const t = scope.model[scope.field];
+                if (! t) {
+                    return;
+                }
+                const err = t.some(x => x === null || x === undefined)
```
gurhar1133, line 2260: Do we have a policy about nesting ternary expressions? Personally hard for me to read, but that could just be me.

gurhar1133 (commented):
Left a couple minor style comments. Also, while testing the app I noticed a couple things.

When you add a tally but don't select a score for that tally, the app will let you run the sim, but the sim will error. Maybe this should be prevented? Maybe a separate issue?

Also, when you add a new tally with a score and an energyFilter and run the sim, it will work, but then when you select the tally after it has run the plot panel will show the error: "Tally t2 contains no Mesh". This seems to happen because the meshFilter needs to be added. Maybe adding a meshFilter should be required?

sirepo/package_data/static/js/cloudmc.js
robnagler: I don't think they should be indented like that:

```js
                const err = t.some(x => x === null || x === undefined)
                    ? 'Enter values'
                    : t[0] === 0 && t[1] === 0
                    ? ''
                    : t[0] >= t[1]
                    ? 'Lower limit must be less than upper limit'
                    : '';
```js

Part of the problem is `t`. If `t[0]` and `t[1]` were something like t.lower and t.upper it would be clearer.

rorour (approved):

sirepo/template/cloudmc.py
moellep: I think it is there because eventually the plot will have error bars based on the std_dev values.

sirepo/package_data/static/js/cloudmc.js
moellep: I'll change the indent. I agree that named fields would be better. In this case it would be easier to have individual model fields for the lower and upper limit, ex. energyFilter.lowerLimit and energyFilter.upperLimit.

moellep:
Thanks for testing - I've opened a new issue for the tally missing score/mesh: #7034

PR #6935, opened 2024-03-22
Fix #6933 srunit_servers startup cleaner

### LINKED ISSUE #6933: job_api needs to catch ConnectionRefusedError
```
Mar 22 16:38:30 860349     0 ../../sirepo/job_api.py:178:api_jobSupervisorPing unexpected exception=<class 'ConnectionRefusedError'> exc=[Errno 111] Connection refused stack=Traceback (most recent call last):
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/job_api.py", line 160, in api_jobSupervisorPing
    r = await self._request_api(
  File "/home/vagrant/src/radiasoft/sirepo/sirepo/job_api.py", line 310, in _request_api
    r = await r
ConnectionRefusedError: [Errno 111] Connection refused
```

robnagler:
- api_jobSupervisorPing catch correct exception(s)
- srunit_servers is more exact in what it catches and how it checks


gurhar1133 (approved):
LGTM

I was poking around with whether `HTTPClientError` is to coarse. From the source code it looks like `HTTPStreamClosedError` and `HTTPTimeoutError`. I'm not sure if one of those would be better for this use case however

robnagler:
> I'm not sure if one of those would be better for this use case however

Thanks for looking. I wonder if we want to catch this exception at all. `_request_api` should be the one to catch this error in the first place. I created this for more exploration: https://github.com/radiasoft/sirepo/issues/6937. Feel free to self-assign, if it interests you.

PR #6954, opened 2024-03-29
Fix #6724: topaz example

### LINKED ISSUE #6724: Add Sirepo examples for Topaz and HB2A datasets
Create new Sirepo examples for Topaz and HB2A data, referencing existing ML jupyter notebooks. 

The datasets are large and will need to use the features implemented in #6719.

gurhar1133:


gurhar1133:
@moellep ping

moellep (commented):
Do we ever use the kernel_initializer outside the default keras values? 
Conv2Transpose, Dense --> glorot_uniform
Conv2D --> he_normal

I saw in the pets example, you are using "random_normal" - did you see an example which used this, rather than the "he_normal" default?

gurhar1133:
> Do we ever use the kernel_initializer outside the default keras values? Conv2Transpose, Dense --> glorot_uniform Conv2D --> he_normal
> 
> I saw in the pets example, you are using "random_normal" - did you see an example which used this, rather than the "he_normal" default?

Oh, I think I may have been mistaken about the defaults. Is it he_normal for Conv2D and glorot_uniform for Conv2Transpose, Dense? If so I can adjust things.

sirepo/package_data/static/json/activait-schema.json
```diff
@@ -422,6 +431,7 @@
             "dimensionality": ["Dimensionality", "Integer", 32],
             "layer": ["Layer", "NeuralNetLayer", "Conv2D"],
             "kernel" : ["Kernel Size", "Integer", 3],
+            "kernel_initializer": ["Kernel Initializer", "KernelInitializer", "GlorotUniform"],
```
moellep, line 434: There needs to be a fixup in sim_data.activait to add this recursively to the layers. Running through the neuralNet.layers field and into the concatenate child models, calling update_model_defaults() on each.

To reproduce the problem, go to the master branch, discard changes to the Pet db example and open a new copy. Then switch to the new branch, and open the same example.

moellep:
I looked and you are correct - it looks like glorot_uniform is the keras default for Conv2D, etc. So that is correct. There does need to be a sim_data fixup which I mentioned before, and once that is added, you don't need to add the updates to pets.json because that will happen automatically.

gurhar1133:
Could you take another look when you get a chance @moellep I updated the fixup

moellep (approved):

PR #6960, opened 2024-04-01
6832 preserve score across tally

moellep:
Need to merge these two PRs together before submitting a PR to main.


moellep:
Backing out this PR

PR #6961, opened 2024-04-01
Fix #6870: case insensitive sim search

### LINKED ISSUE #6870: simulation search should be case insensitive
When you type into the search box on the simulations page, it should be case insensitive. For example, searching "undulator" should show the "Undulator Radiation" sim.

gurhar1133:
While working on this also noticed that using keys to navigate the search selections set the input value to [object Object]. Fix for that included in this PR by adding a focus event handler.


gurhar1133:
@moellep  ping

PR #6983, opened 2024-04-05
Fix #6959: import across sims using madx

### LINKED ISSUE #6959: Import madx, opal, elegant native lattices into madx, opal, elegant
The lattice codes should accept input from different codes, translating to madx first, then to the selected code.

For example, the elegant app should access *.ele, *.lte, *.in, *.madx input files. If and opal file is imported (*.in), then it should be translated to madx and then into elegant using the OpalMadxConverter --> ElegantMadxConverter.

gurhar1133:


sirepo/package_data/static/js/opal.js
```diff
@@ -223,9 +223,7 @@ SIREPO.app.directive('appFooter', function() {
         },
         template: `
             <div data-common-footer="nav"></div>
-            <div data-import-dialog="" data-title="Import Opal File" data-description="Select an OPAL .in or .madx file." data-file-formats=".in,.madx,.zip">
-              <div data-import-options=""></div>
-            </div>
+            <div data-elegant-import-dialog=""></div>
         `,
```
moellep, line 227: This PR works well, except for the case of uploading a *.in file to opal which requires additional files. It shows "Missing data files" but doesn't provide a way to import the missing files:

x

To reproduce, open the AWA Photoinjector example, export as zip then edit the Solenoid T7 filenames slightly. Then select "OPAL Source" from the cogwheel and save the AWA-Photoinjector.in file, then edit the *.in file to use the new names for the T7 files. 

Upload the new AWA-Photoinjector.in file, and it should prompt for the T7 files which don't exist on the server.

gurhar1133: I fixed opal missing data files, but found another edge case. 

When you import ATTR-U-and-W.madx into elegant it hits a key error when the SROTATION elements in madx are converted to SROT. Working on it.

gurhar1133:
@moellep another look pls?

moellep (approved):

PR #6985, opened 2024-04-08
Fix #6900: scrollbar for sim folders

### LINKED ISSUE #6900: simulation list folder list should be separately scrollable
If the list of folders on the simulations list is very long, you can select a folder, but not see the simulations in the folder without scrolling back up the page. The folder list should have its own vertical scroll bar.

x

gurhar1133:


sirepo/package_data/static/css/sirepo.css
```diff
@@ -897,6 +897,7 @@ li.sr-dropdown-submenu > a {
   overflow-y: auto;
   border: 1px solid #e7e7e7;
   min-height: 80ex;
+  max-height: calc(100vh - 65px);
   border-radius: 4px;
```
moellep, line 901: I change the css a bit to use a calc() and directly on the sr-sidebar class. 
x

PR #7005, opened 2024-04-12
Fix #6982: default beamline on delete

### LINKED ISSUE #6982: delete beamline should set new default visualization beamline
In opal, if you delete a beamline which is currently used on the Control or Visualization tab, you will get a server error if you try to export the opal.in file, or the export.zip. When a beamline is deleted, it should change the default beamline if the selected on was deleted.

gurhar1133:
Tracks that reference non existent beamline after deletion are updated..


moellep:
I made a change to the code - there was a problem that beamlineView is only instantiated if a person opens the beamline editor (via the pencil icon on the Beamline Editor panel), so that code wasn't getting executed. There was already a check in the VisualizationController if the command_track.line was set, so I added an additional track that the command_track.line references an existing beamline.

PR #7013, opened 2024-04-15
Fix #6972: Fix cancel in sbatch

### LINKED ISSUE #6972: Canceling job on NERSC doesn't remove from sbatch
It's been over a minute, and usually canceling is very fast. I was able to cancel the job manually.

e-carlin:
- In the cancel case we don't have fields like simulationType. We just have opId's to cancel. So, we can't construct userDir. But, we don't need userDir (or runDir, sbatchCores, etc) in the cancel case (or for any non SLOT_OPS). So, only add those fields when not SLOT_OP.
- Add a check inside of the job_agent for op_name to keep job_driver.sbatch and job_agent in sync. job_agent._Cmd's require fields like runDir.
- Add a test for sbatch cancel which wasn't covered previously.

### DIFF
--- a/sirepo/srunit.py
+++ b/sirepo/srunit.py
@@ -153,6 +153,27 @@ class _TestClient:
         if feature_config.cfg().ui_websocket:
             self._websocket = _WebSocket(self)
 
+    def iter_sleep(self, kind, op_desc):
+        import time
+
+        def _setup():
+            rv = PKDict(_ITER_SLEEP[kind])
+            rv.countdown = range(rv.count, -1, -1)
+            return rv
+
+        s = _setup()
+        for i in s.countdown:
+            yield
+            if i > 0:
+                time.sleep(s.sleep_secs)
+        else:
+            pkfail(
+                "timeout secs={} kind={} op_desc={}",
+                s.sleep_secs * s.count,
+                kind,
+                op_desc,
+            )
+
     @contextlib.contextmanager
     def sr_adjust_time(self, days):
         from sirepo import srtime
@@ -324,27 +345,6 @@ class _TestClient:
             raw_response=True,
             **kwargs,
         )
-
-    def sr_iter_sleep(self, kind, op_desc):
-        import time
-
-        def _setup():
-            rv = PKDict(_ITER_SLEEP[kind])
-            rv.countdown = range(rv.count, -1, -1)
-            return rv
-
-        s = _setup()
-        for i in s.countdown:
-            yield
-            if i > 0:
-                time.sleep(s.sleep_secs)
-        else:
-            pkfail(
-                "timeout secs={} kind={} op_desc={}",
-                s.sleep_secs * s.count,
-                kind,
-                op_desc,
-            )
 
     def sr_login_as_guest(self, sim_type=None):
         """Sets up a guest login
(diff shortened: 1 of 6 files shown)


sirepo/srunit.py
```diff
@@ -478,20 +470,28 @@ def sr_sbatch_animation_run(self, sim_name, compute_model, reports, **kwargs):
...
                 password=p,
                 report=compute_model,
-                simulationId=data.models.simulation.simulationId,
-                simulationType=data.simulationType,
+                simulationId=d.models.simulation.simulationId,
+                simulationType=d.simulationType,
                 username=p,
             ),
         )
+        self.sr_sbatch_logged_in = True
```
robnagler, line 494: The post could possibly fail with data.error? Maybe assert_success() would be good to call.

tests/sbatch_test.py
```diff
@@ -13,13 +13,66 @@
...
+        "runSimulation",
+        PKDict(
+            models=d.models,
+            report=compute_model,
+            simulationId=d.models.simulation.simulationId,
+            simulationType=d.simulationType,
+        ),
+        expect_completed=False,
+    )
+
```
robnagler, line 47: no blank

tests/sbatch_test.py
```diff
@@ -13,13 +13,66 @@
...
+            break
+        time.sleep(5)
+        r = fc.sr_post("runStatus", r.nextRequest)
+    else:
+        pkfail("failed to start running in sirepo and slurm reply={}", r)
+    r = fc.sr_post("runCancel", r.nextRequest)
+    for _ in range(5):
+        if _squeue_num_jobs() == 0:
+            break
+        time.sleep(5)
```
robnagler, line 64: I think we should modularize this with a name:

```py
while self.iter_sleep("slurm", "runCancel"):
    do some check

def iter_sleep(self, kind, op_desc):
    def _setup():
        rv = PKDict(_ITER_SLEEP[kind])
        rv.countdown = range(rv.count, -1, -1)
        return rv
        
    s = _setup()
    for i  in s.countdown():
        yield
        if i > 0:
            time.sleep(s.sleep_secs)
    else:
        pkfail("timeout secs={} kind={} op_desc={}", s.total_secs, kind, op_desc)
```

robnagler (changes requested):

sirepo/srunit.py
e-carlin: 5ce403daf

tests/sbatch_test.py
e-carlin: 5ce403daf

tests/sbatch_test.py
e-carlin: 5ce403daf

sirepo/srunit.py
```diff
@@ -317,6 +325,27 @@ def sr_get_root(self, sim_type=None, **kwargs):
             **kwargs,
         )
 
+    def sr_iter_sleep(self, kind, op_desc):
```
robnagler, line 328: This no longer needs to be 'sr_', because it no longer inherits from flask.TestClient.

robnagler (changes requested):

sirepo/srunit.py
e-carlin: d0bb36461

robnagler (approved):

Issue #7026, opened 2024-04-24
implicit security check in Jupyter sim_apis


rorour:
[Both of the APIs in sirepo/sim_api/jupyterhublogin.py begin with a call ](https://github.com/radiasoft/sirepo/blob/57ea8a13deafb49a42f9d3217275fd1126d0737d/sirepo/sim_api/jupyterhublogin.py#L32
) to `self.parse_params()`. These are the only calls in Sirepo to `self.parse_params()` that do not utilize the return value; they are intended to check that the user has the sim_type_jupyterhublogin role. These APIs depend on `Quest.parse_params` calling `http_request.parse_params` which calls `auth.check_sim_type_role` inside of a nested function. There is no documentation on either `parse_params` function indicating that they do a security check. 

At the very least, I want to add a comment above these calls in the Jupyter APIs indicating that these are security checks. Or, just change them to direct calls to `auth.check_sim_type_role`.

But I don't like that this check needs to be added inside each API. It seems like this check should be done at a higher level, like as an APIPerm `require_user_with_role`. What are your thoughts? @robnagler

robnagler:
Yes, this is a real problem and #5026 hints at the general issue. Ideally, we'd always know the sim_type in every request, if there was one, and then could check it in check_api_call. However, we don't have that. It possibly could be added to the header of HTTP requests, and explicitly in the SIM requests. An alternative is to put this in template.import_module, but that's probably too late.

What we need to do is push on quest.Spec so that we can know to call parse_params, parse_post, etc. before the function is called.

Issue #3748, opened 2021-08-06
Jupyter user created with first_5Flast


robnagler:
This user was created a few days ago. The email address and jupyterhub.user_name has an underscore which got mapped by jupyterhub to `_5F`.

robnagler:
There were two users. I edited the db, fixed up dirs, and jupyterhub.

robnagler:
[This code](https://github.com/radiasoft/sirepo/blob/e6555962dae411dec0c288bab27e1da50ed0ae04/sirepo/sim_api/jupyterhublogin.py#L93) is incorrect:

```py
        return re.sub(
            r"\W+",
            _HUB_USER_SEP,
            (github_handle or qcall.auth.logged_in_user_name()).split("@")[0],
        ).lower()
```

`\W` does not include underscore, because `_` is a word character. This would be better compiled:

```py
_INVALID_CHARS = re.compile(r"[^a-z0-9]+")
```

And used:

```py
return _INVALID_CHARS.sub(_HUB_USER_SEP, name.lower())
```

The `github_handle` should be outside the `split` but this is mooted by #7027. 

`logged_in_user_name` is not so great. There should be a function to get the local part from auth_db.email or auth itself.

Issue #293, opened 2016-08-17
Overlapping graph axis labels and title


robnagler:
The axis labels are overlapping with the title in the elegant bunchCompLSC example. Just noting. I'm sure this happens other places:

![screen shot 2016-08-17 at 8 54 31 am](https://cloud.githubusercontent.com/assets/5495179/17741138/48b3df48-6458-11e6-8252-cccbabaa5cc4.png)

PR #7027, opened 2024-04-24
fix #6925: Remove all JupyterHub migrate code

### LINKED ISSUE #6925: Remove all JupyterHub migrate code
https://github.com/radiasoft/sirepo/issues/6350#issuecomment-2013660881

### LINKED ISSUE #7010: add back uwsgi_temp_path to nginx_proxy.conf.jinja
> since we are running nginx as vagrant we still need it to override default paths that are only writable by root

needed for running Jupyter in dev

rorour:
- fix #6925: Remove all JupyterHub migrate code
- fix #7010: add back uwsgi_temp_path to nginx_proxy.conf.jinja
- consolidate jupyterhublogin pkcli tests
- add test for JupyterHub APIs when user does not have role


sirepo/package_data/static/json/jupyterhublogin-schema.json
```diff
@@ -5,5 +5,8 @@
                 "jupyterhublogin.js"
             ]
         }
+    },
+    "appDefaults": {
+        "route": "jupyterhublogin"
```
rorour, line 10: Having trouble figuring out what the default route should be

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -41,47 +41,19 @@ async def api_checkAuthJupyterHub(self):
...
-      2. User signs into sirepo under one@any.com. They migrate their data using
-         GitHub handle y. They sign into sirepo under two@any.com. They choose
-         to migrate GitHub handle y again. We should let them know that they
-         have already migrated.
-      3. one@any.com signs up for jupyter and does not migrate data. They are
-         given the username one. two@any.com signs up for jupyter and they
-         migrate their data, but they have no data to migrate. They have the
-         github handle one and no previous data. They should be alerted that
-         they can't migrate that GitHub handle.
-      4. A new user signs in with foo@any.com and they do not select to
```
robnagler, line 78: still valid and maybe the other too

robnagler (changes requested):

sirepo/package_data/static/json/jupyterhublogin-schema.json
robnagler: Does jupterhublogin work?

sirepo/sim_api/jupyterhublogin.py
rorour: Can you elaborate?

robnagler: I think this is the default case when they had a github handle with a jupyter address. There will be a collision. Maybe rewording the 1-4 text to describe the case for normal collisions.

sirepo/package_data/static/json/jupyterhublogin-schema.json
rorour: It just hangs for me

sirepo/package_data/nginx_proxy.conf.jinja
```diff
@@ -14,6 +14,8 @@ http {
     client_body_temp_path {{ run_dir }}/client_body_temp;
     proxy_temp_path {{ run_dir }}/proxy_temp;
     fastcgi_temp_path {{ run_dir }}/fastcgi_temp;
+    # uwsgi_temp_path needed for running jupyter
```
e-carlin, line 17: I think this comment could be expanded to capture more of what's going on. We have to define all of these paths because we are running nginx as a non-priveleged user. nginx default paths expect write level privs on files only writable by root. So, we have to define our own paths in a place writable by us (vagrant). 

I don't think order matters in this part of the file so you can move all paths to the top of the http block and write a comment that covers all of them.

sirepo/package_data/static/js/jupyterhublogin.js
```diff
@@ -3,31 +3,8 @@
...
-});
-
-SIREPO.app.controller('NameConflictController', function(requestSender, jupyterhubloginService, $route, $scope) {
-    const self = this;
-
-    self.noMigration = function() {
-        jupyterhubloginService.doMigration(false);
-    };
-
-    self.logout = function() {
```
e-carlin, line 25: We still need logout.

tests/jupyterhub_test.py
```diff
@@ -15,6 +15,21 @@ def setup_module(module):
     )
 
 
+def test_api_security_without_role(fc):
+    from sirepo.pkcli import roles
+    import sirepo.auth_role
+
+    x = sirepo.auth_role.for_sim_type("jupyterhublogin")
+    roles.delete(fc.sr_uid, x)
+    assert x not in fc.sr_auth_state().roles, f"{x} role should have been removed"
```
e-carlin, line 24: Use pkunit functions

tests/pkcli/jupyterhublogin_test.py
```diff
@@ -28,3 +27,16 @@ def test_create_new_user(auth_fc):
     pkunit.pkeq(r, sirepo.pkcli.jupyterhublogin.create_user(e, n))
     # create_user is idempotent. Returns user_name if user already exists
     pkunit.pkeq(r, sirepo.pkcli.jupyterhublogin.create_user(e, n))
+
+
+def test_logout(auth_fc):
+    # Clears third party (jupyterhub) cookies
```
e-carlin, line 33: This test needs to check that the cookie is actually cleared.

e-carlin (changes requested):
```
$ cd ~/src/radiasoft/sirepo
$ rm -rf run
$ bash etc/run.sh jupyter
```

That leads me to this page
Screen Shot 2024-04-29 at 9 54 56 AM

I see there was a discussion of default route so maybe this relates to it. Not sure.

sirepo/package_data/nginx_proxy.conf.jinja
rorour: fc537f91744023171f36839b42a1c335bed539b9

sirepo/package_data/static/js/jupyterhublogin.js
rorour: This is part of NameConflictController which is no longer used.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -29,6 +29,7 @@
 class API(sirepo.quest.API):
     @sirepo.quest.Spec("require_user", sim_type=f"SimType const={_SIM_TYPE}")
     async def api_checkAuthJupyterHub(self):
+        # check that user has role
```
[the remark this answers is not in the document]
rorour: this will be going away eventually https://github.com/radiasoft/sirepo/issues/7026. I added a more detailed comment. [c726701](https://github.com/radiasoft/sirepo/pull/7027/commits/c72670133ae55943f058d0447b6b4ff060741d6d)

tests/jupyterhub_test.py
rorour: [a213ff2](https://github.com/radiasoft/sirepo/pull/7027/commits/a213ff25569cff082c5f0a22dd37ec4a7c5b20c6)

sirepo/sim_api/jupyterhublogin.py
rorour: [d3a035f](https://github.com/radiasoft/sirepo/pull/7027/commits/d3a035fba7ef0e4619416db1861c69dc51efe402)

rorour:
related issue noting how these changes work https://github.com/radiasoft/sirepo/issues/7092
@e-carlin is working on the last unadressed comment **https://github.com/radiasoft/sirepo/pull/7027#discussion_r1583378866**

e-carlin:
@rorour https://github.com/radiasoft/sirepo/issues/7098. The flag can be removed and we should check_dir always

tests/pkcli/jupyterhublogin_test.py
rorour: https://github.com/radiasoft/sirepo/issues/7096

rorour:
Ready for review

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +60,21 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
                 qcall=qcall,
-                check_dir=True,
             )
             qcall.auth_db.model("UserRole").add_roles(
                 roles=[sirepo.auth_role.for_sim_type("jupyterhublogin")],
             )
         return PKDict(email=email, jupyterhub_user_name=n)
+
+
+def old_user_dirs():
```
e-carlin, line 70: After looking at this I think something a bit more generic would be good. Call it "uknown_user_dirs" and just output the dirs on a new line but no `rm -rf`. It could be useful in the future. And it is easy enough to add the `rm -rf` from the outside.

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +60,21 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
+    """Outputs commands to delete dirs for Jupyter users that are not in database."""
+    with sirepo.quest.start() as qcall:
+        m = ""
+        r = qcall.auth_db.model("JupyterhubUser").search_all_for_column("user_name")
+        for d in pkio.sorted_glob(
+            sirepo.sim_api.jupyterhublogin.cfg().user_db_root_d.join("*")
+        ):
+            if not os.path.basename(d) in r:
```
e-carlin, line 78: d.basename

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +60,21 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
+                m += f"\nrm -rf {d}"
+        pkdlog(m)
```
e-carlin, line 80: you can just `return m`. This way the function is more flexible. It can be called in python and get a useful result back or on the cli and pkcli does the work of converting the returned value to a printable value.

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +60,21 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
                 roles=[sirepo.auth_role.for_sim_type("jupyterhublogin")],
             )
         return PKDict(email=email, jupyterhub_user_name=n)
+
+
+def old_user_dirs():
+    """Outputs commands to delete dirs for Jupyter users that are not in database."""
+    with sirepo.quest.start() as qcall:
+        m = ""
+        r = qcall.auth_db.model("JupyterhubUser").search_all_for_column("user_name")
```
e-carlin, line 74: I think better to query the db each time. I would be curious which is faster but iterating over r for each user dir is not very performant. Likewise, if we had more users than we do it could be too much to pull them all into memory. 

So, the code below would become
```
if not qcall.auth_db.model("JupyterhubUser").unchecked_search_by(user_name=d.basename)
```

e-carlin (changes requested):

sirepo/pkcli/jupyterhublogin.py
robnagler, line 74: Convert this to a frozenset, because it is inside a loop, and there are NxM checks here otherwise. Indeed, I think search_all_for_column could just return a frozenset. I noticed auth.get_uids calls it, but that function is never called. search_all_for_column is unsorted anyway so it doesn't need to be a list, and it's used in contexts where it is better to be a set. There is one case (pkcli.roles.lists) it should be converted to a list, but that's the atypical case.

robnagler (changes requested):
Looks good some minor changes and suggestions.

tests/jupyterhub_test.py
```diff
@@ -52,3 +68,16 @@ def _jupyter_user_dir_exists(uid):
 
 def test_jupyterhub_redirect(fc):
     fc.sr_get("redirectJupyterHub", redirect=False).assert_http_redirect("jupyterHub")
+
+
+def test_logout(auth_fc):
+    # Clears third party (jupyterhub) cookies
+    # TODO(e-carlin): https://github.com/radiasoft/sirepo/issues/7096
+    # This test should verify that the cookie was actually removed.
+    fc = auth_fc
```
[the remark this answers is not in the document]
rorour: Why is it usually aliased?

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -40,75 +42,41 @@ async def api_checkAuthJupyterHub(self):
...
         check_dir (bool): assert that an existing user does not have a dir with
-                          the same name (before modifying the name to eliminate
-                          conflicts)
+                          the same name
     Returns:
-        user_name (str): The user_name of the new user
+        user_name (str): The user_name of the new or existing user
     """
 
     def __handle_or_name_sanitized():
```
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

tests/template/model_units_test.py
```diff
@@ -1,11 +1,9 @@
 # -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

tests/jupyterhub_test.py
```diff
@@ -52,3 +68,16 @@ def _jupyter_user_dir_exists(uid):
...
+
+    from pykern.pkdebug import pkdp
```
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -123,7 +88,7 @@ def __user_name():
     # POSIT: if two creates happen simultaneously, there may be an existence
     # collision, but the db will be consistent, because this call happens
     # first, before db insert.
-    if check_dir and _user_dir(qcall, u).exists():
+    if _user_dir(qcall, u).exists():
         raise AssertionError(f"existing user dir with same name={u}")
```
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +60,21 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
-                check_dir=True,
             )
             qcall.auth_db.model("UserRole").add_roles(
                 roles=[sirepo.auth_role.for_sim_type("jupyterhublogin")],
             )
         return PKDict(email=email, jupyterhub_user_name=n)
+
+
+def old_user_dirs():
+    """Outputs commands to delete dirs for Jupyter users that are not in database."""
```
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/pkcli/jupyterhublogin.py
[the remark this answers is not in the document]
rorour: [ad6298a](https://github.com/radiasoft/sirepo/pull/7027/commits/ad6298ae14d580069dd27cda09abc4bfad79f545)

sirepo/package_data/static/js/jupyterhublogin.js
```diff
@@ -3,31 +3,13 @@
...
 
-SIREPO.app.controller('JupyterhubMigrateController', function(authState, jupyterhubloginService, requestSender, $scope) {
-    const self = this;
-    self.isLoading = true;
+SIREPO.app.controller('JupyterHubLoginController', function(requestSender) {
+    var self = this;
+
     requestSender.sendRequest(
         'redirectJupyterHub',
         () => {self.isLoading = false;}
```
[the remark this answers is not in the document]
rorour: [df67e71](https://github.com/radiasoft/sirepo/pull/7027/commits/df67e71f66d2eb0a191b02f81d5749152a85aeeb)

sirepo/pkcli/jupyterhublogin.py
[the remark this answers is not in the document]
rorour: > We have some directories with begin with a capital letter that are not users, e.g. USPAS-Instructor. These directories should not be deleted. You'll need to exclude those.

Couldn't valid usernames also begin with a capital letter? I think I either need a complete list to exclude here, or whoever is deleting could check against a list before deleting anything. I took @e-carlin's suggestion of removing the "rm -rf" and just outputting the dir.


> There was (is?) a bug where the user name contains _5F. These should be called out and fixed. This could be an echo in the output pointing out these names.

Can you please provide more context? What bug would it cause here? Maybe this should be its own pkcli to identify these problematic usernames?

sirepo/pkcli/jupyterhublogin.py
rorour: [c4d59a6](https://github.com/radiasoft/sirepo/pull/7027/commits/c4d59a6548a384c05945af051a95c0d1ecb498ea)

rorour: It will need to be converted to a list also for authState. (exception=Object of type frozenset is not JSON serializable) In my opinion this change overcomplicates things.

sirepo/pkcli/jupyterhublogin.py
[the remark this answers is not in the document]
robnagler: > Couldn't valid usernames also begin with a capital letter?

No. However, this is not clear, and there are many defects. api_authEmailLogin [enforces lower case](https://github.com/radiasoft/sirepo/blob/1789d23f51556178466ae83d7dc9e7cb323f80e3/sirepo/auth/email.py#L121), but this is definitely insufficient. `create_user_from_email` should enforce lower case, or better, `auth_db.email.__init__` should override super and force unverified_email and user_name in `fields` to be lower case.  

pkcli.admin and jupyterhublogin should force the emaila argument to lower case.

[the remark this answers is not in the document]
robnagler: > Can you please provide more context? What bug would it cause here? Maybe this should be its own pkcli to identify these problematic usernames?

I think as long as we are cleaning up invalid/old names. We should do that, since that's the name of the function.

#3748 describes the problem. 5F is _ (underscore) in ascii. There was a double encoding going on. I think we fixed it, but it may still be there. We restrict the jupyter users to a specific syntax. I think I just figured out the bug. I'll document in #3748.

sirepo/pkcli/jupyterhublogin.py
robnagler: OK. Don't change search_all, but do use a frozenset to test the condition. This speeds the inside test to a hash lookup.

rorour: [9fc0efd](https://github.com/radiasoft/sirepo/pull/7027/commits/9fc0efd75cc4f7666288b72685a967ed6f474cfe)

sirepo/pkcli/jupyterhublogin.py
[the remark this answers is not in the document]
rorour: [5fcc5e6](https://github.com/radiasoft/sirepo/pull/7027/commits/5fcc5e68653ffaa21bfd7912d81f95c61e304a11)
[eae3cde](https://github.com/radiasoft/sirepo/pull/7027/commits/eae3cdeac182f2aa627c0d92b2abe06a9ce5d86d)

e-carlin (approved):

sirepo/auth/__init__.py
```diff
@@ -270,6 +270,13 @@ def is_premium_user(self):
             role=sirepo.auth_role.ROLE_PAYMENT_PLAN_PREMIUM,
         )
 
+    def local_part_user_name(self):
+        """If user_name is email address, return the local part.
+        Otherwise, just return user_name.
+        """
+
+        return self.logged_in_user_name().split("@")[0]
```
robnagler, line 278: This should assert `@` is in the name by checking there are exactly two parts. The name should reflect that it works for the logged_in_user, e.g. logged_in_user_email_local_part. It's a mouthful, but it describes what is going on. The coupling is a bit problematic. I think we might want instead to have a function logged_in_email and have it assert the email. Then a separate function that splits emails into parts eg in sirepo.util.

sirepo/auth_db/email.py
```diff
@@ -39,6 +39,12 @@ class AuthEmailUser(sirepo.auth_db.UserDbBase):
     token = sqlalchemy.Column(sqlalchemy.String(sirepo.util.TOKEN_SIZE), unique=True)
     expires = sqlalchemy.Column(sqlalchemy.DateTime())
 
+    def __init__(self, *args, **kwargs):
+        for x in ("unverified_email", "user_name"):
+            if kwargs.get(x):
```
robnagler, line 44: I think an explicit null check is more robust here:

```py
if (v := kwargs.get(x)) is not None:
    kwargs[x] = v.lower()
```

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +59,25 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
             )
             qcall.auth_db.model("UserRole").add_roles(
                 roles=[sirepo.auth_role.for_sim_type("jupyterhublogin")],
             )
         return PKDict(email=email, jupyterhub_user_name=n)
+
+
+def unknown_user_dirs():
+    """Returns directory names for Jupyter users that are not in database.
+    Excludes directories that begin with capital letter.
```
robnagler, line 71:
```txt
Returns:
    list: directories for jupyter users that are not in the database.
```

sirepo/pkcli/jupyterhublogin.py
```diff
@@ -58,9 +59,25 @@ def maybe_create_sirepo_user(qcall, email, display_name):
...
+    """
+    with sirepo.quest.start() as qcall:
+        m = ""
```
robnagler, line 74: list is better. Separate presentation from business logic. Strings are generally problematic, especially with full paths that might name special characters or spaces in them. If our goal is to delete the directories, we can transform with Python or b fixing #293.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -21,6 +21,8 @@
 
 _HUB_USER_SEP = "-"
 
+_INVALID_USER_NAME_CHARS = re.compile(r"[^a-z0-9]+")
```
robnagler, line 24: A comment here might be nice, because this includes `-` which is a valid username char but not for this purpose.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -40,76 +44,35 @@ async def api_checkAuthJupyterHub(self):
...
-        return re.sub(
-            r"\W+",
-            _HUB_USER_SEP,
-            # Get the local part of the email. Or in the case of another auth
-            # method (ex github) it won't have an '@' so it will just be their
-            # user name, handle, etc.
-            (github_handle or qcall.auth.logged_in_user_name()).split("@")[0],
-        ).lower()
+    def __name_sanitized():
+        return _INVALID_USER_NAME_CHARS.sub(
```
robnagler, line 70: remove leading and trailing _HUB_USER_SEP characters, and then make sure the string is non-zero length.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -40,76 +44,35 @@ async def api_checkAuthJupyterHub(self):
...
+            _HUB_USER_SEP, qcall.auth.local_part_user_name().lower()
```
robnagler, line 71: local_part_user_name should guarantee lower. A logged in user's email should always be in lower case.

sirepo/sim_api/jupyterhublogin.py
```diff
@@ -40,76 +44,35 @@ async def api_checkAuthJupyterHub(self):
...
-    def __handle_or_name_sanitized():
-        return re.sub(
-            r"\W+",
-            _HUB_USER_SEP,
-            # Get the local part of the email. Or in the case of another auth
-            # method (ex github) it won't have an '@' so it will just be their
-            # user name, handle, etc.
-            (github_handle or qcall.auth.logged_in_user_name()).split("@")[0],
-        ).lower()
+    def __name_sanitized():
```
robnagler, line 69: I don't think this needs to be a double under. That's only necessary when it is in the global scope of a class.

tests/jupyterhub_test.py
```diff
@@ -15,6 +15,22 @@ def setup_module(module):
...
 
 
+def test_api_security_without_role(fc):
+    from pykern.pkunit import pkok
+    from sirepo.pkcli import roles
+    import sirepo.auth_role
+
+    x = sirepo.auth_role.for_sim_type("jupyterhublogin")
+    roles.delete(fc.sr_uid, x)
+    pkok(x not in fc.sr_auth_state().roles, f"{x} role should have been removed")
```
robnagler, line 25: don't use f-strings. allow pkdebug to handle: `"{} role should...", x`

robnagler (changes requested):

sirepo/auth/__init__.py
rorour: That seems unnecessarily complicated for a function that is only called once. I also don't see how this suggested change would handle usernames that aren't email addresses. I just renamed the function to `logged_in_user_name_local_part`. [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/auth_db/email.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/pkcli/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/pkcli/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/sim_api/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/sim_api/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/sim_api/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

sirepo/sim_api/jupyterhublogin.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

tests/jupyterhub_test.py
rorour: [3a9f0d1](https://github.com/radiasoft/sirepo/pull/7027/commits/3a9f0d1399801354cea9e77a661893218636e2dd)

robnagler (approved):

PR #7029, opened 2024-04-25
Fix #7020: read during write hdf5

### LINKED ISSUE #7020: Reading hdf5 file while writes occurring needs to catch multiple exceptions
In the silas app with a sliced laser pulse (ex. 12 slices), if you change the "Number of Mesh Points in Slice" to a larger value, like 128, the plot may fail when the data is requested as the slice writes are occurring. The code is currently catching a BlockingIOError and retrying after a timeout, but in this case it was a regular IOError "Errno 11 Unable to open file (unable to lock file, errno = 11, error message = "Resource temporarily unavailable".

We should add a common hdf5 file reader which handles these cases. Something like adding a method to ```HDF5Util.read_while_writing(path, callback, retries=10, timeout=3)``` which catches BlockingIOError and IOError and retries.

Then this should be used where-ever hdf5 files are read while writes may be occurring. Alternatively, look into hdf5 read locks.

gurhar1133:


sirepo/template/hdf5_util.py
```diff
@@ -88,3 +90,14 @@ def lineplot(self, plot_attrs):
...
+    @contextlib.contextmanager
+    def read_while_writing(self, retries=10, timeout=3):
+        for _ in range(retries):
+            try:
+                with h5py.File(self.filename, "r") as p:
+                    yield p
+                    return
+            except (BlockingIOError, IOError, KeyError) as e:
+                time.sleep(timeout)
+        raise AssertionError(f"{self.filename} is unavailable")
```
moellep, line 103: I think this only handles the not-found case. When I run the test read program by itself, it doesn't retry:
```
python read.py x.h5 check_key
...
KeyError: "Unable to open object (object 'this_key_wont_exist_at_first' doesn't exist)"
...
RuntimeError: generator didn't stop after throw()
```

I think this is related to this:
https://stackoverflow.com/questions/34775099/why-does-contextmanager-throws-a-runtime-error-generator-didnt-stop-after-thro

Instead, you should pass a callback to read_while_writing() which gets executed within the try/except.

tests/template/hdf5_util_test.py
```diff
@@ -0,0 +1,62 @@
...
+import time
+
+
+def test_read_while_writing():
+    def _read_not_written(case_work_dir):
+        p = _proc(case_work_dir, "read.py", "does_not_exist_at_first.h5")
+        time.sleep(10)
+        os.rename(
+            case_work_dir.join("x.h5"),
+            case_work_dir.join("does_not_exist_at_first.h5"),
```
moellep, line 19: Check in a smaller x.h5 file. It doesn't need to be so big.

tests/template/hdf5_util_test.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import shutil
+import subprocess
+import time
+
+
+def test_read_while_writing():
+    def _read_not_written(case_work_dir):
+        p = _proc(case_work_dir, "read.py", "does_not_exist_at_first.h5")
+        time.sleep(10)
```
moellep, line 16: See if the sleeps can be shorted a bit - this will slow up the builds otherwise.

moellep (changes requested):

moellep (approved):

PR #7065, opened 2024-05-14
fix #6991: replace browser default fullscreen feature

### LINKED ISSUE #6991: Replace fullscreen feature with a full width dialog
The fullscreen feature on plots doesn't work well across browsers. Instead, it should take the plot and show it in a modal dialog which fits the whole page.

rorour:
fix #6991


moellep:
For SRW, it seems to be cutting of the plot3d plots, both horizontal and vertical:

x

moellep (approved):
Looks good. @rorour please give it one last test, I made a few changes:
- fixed panel bleed-through on elegant beamline fullscreen
- broadcast sr-close-full-screen from rootScope so close works from other panels
- only adjust margins when in full-screen mode
- hide scrollbars when fullscreen is enabled

rorour:
Thanks, I didn't notice any bugs

PR #7069, opened 2024-05-14
Fix #5989: reflecting planes

### LINKED ISSUE #5989: Reflecting Planes
We should allow users to create an arbitrary number of reflecting planes, just like they can for tallies and sources.

Each plane can be defined by 4 parameters, (A,B,C,D), but we currently only let them define a max of 2 planes and only by (A,B)
See here https://docs.openmc.org/en/latest/pythonapi/generated/openmc.Plane.html?highlight=plane

I can easily imagine cases where users would want to exploit top/bottom symmetries as well as vertical mirror symmetries.  Particularly for fission reactor models.

gurhar1133:
A, B, C, D parameters and arbitrary number of planes


moellep (changes requested):
This PR has many merge conflicts due to the app being renamed. @gurhar1133 could you try to bring the changes over to the main branch

gurhar1133:
@moellep I fixed the merge conflicts, one thing I not so sure about is the layout for the reflective planes tab, so feedback there would be great.

moellep (approved):
@gurhar1133 I made a few layout improvements. Give it a test and then it can be merged.

PR #7070, opened 2024-05-15
fix #7050: add mode to disable users

### LINKED ISSUE #7050: Add mode to disable users
Should also come with a cli to do this. This would not delete the user, but simply prevent the user from accessing anything once logged in.

rorour:
**contains db upgrade**


sirepo/api_auth.py
```diff
@@ -42,6 +43,8 @@ def check_api_call(qcall, func):
             qcall.auth.require_email_user()
         elif expect == a.REQUIRE_ADM:
             qcall.auth.require_adm()
+        elif expect == a.REQUIRE_NOT_DISABLED:
```
robnagler, line 46: REQUIRE_DISABLED will not be an expected by an API. It's a role that is checked in a way similar to the sym_type roles. It's more of an attribute-based access than a role.  It needs to be checked always, whenever a user is required so it needs to be added to require_user whenever the user is logged in.

sirepo/api_perm.py
```diff
@@ -28,6 +28,8 @@ class APIPerm(aenum.Flag):
     REQUIRE_AUTH_BASIC = aenum.auto()
     #: a cookie has to have been returned, which might contain a user
     REQUIRE_COOKIE_SENTINEL = aenum.auto()
+    # require user without role disabled
+    REQUIRE_NOT_DISABLED = aenum.auto()
```
robnagler, line 32: Not necessary. Sorry for steering you incorrectly.

sirepo/auth_role.py
```diff
@@ -10,6 +10,7 @@
 import sirepo.feature_config
 
 ROLE_ADM = "adm"
+ROLE_DISABLED = "disabled"
```
robnagler, line 13: Let's call this ROLE_ACCOUNT_DISABLED so it's a bit clearer what's disabled.

sirepo/server.py
```diff
@@ -474,7 +474,7 @@ def _redirect(req):
         except sirepo.util.SPathNotFound:
             return _not_found(req)
 
-    @sirepo.quest.Spec("require_user", search="SearchSpec")
```
robnagler, line 477: This would be insufficient in any event if they bookmarked a simulation. This security has to be pervasive.

tests/disabled_user_test.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+
+def test_disabled_user(fc):
+    from pykern.pkcollections import PKDict
+    from pykern.pkunit import pkeq, pkok
+    import sirepo.auth_role
+    import sirepo.pkcli.roles
+
+    a = fc.sr_auth_state()
```
robnagler, line 16: This is actually interesting. We don't necessarily want to return this particular role in auth_state, because it gives an attacker a reason for the forbidden, a way to probe the system. Therefore, auth_state will have to filter this out.

tests/disabled_user_test.py
```diff
@@ -0,0 +1,39 @@
...
+    pkok(
+        sirepo.auth_role.ROLE_DISABLED not in fc.sr_auth_state().roles,
+        "{} found in roles",
+        sirepo.auth_role.ROLE_DISABLED,
+    )
+    fc.sr_post(
```
robnagler, line 22: This should test a few APIs, e.g. running a simulation. You'll need to cache the request info for this case so that the request goes straight to the uri.

robnagler (changes requested):

tests/disabled_user_test.py
rorour: With our code being open source and this PR being public it seems like they could figure it out regardless

rorour: I understand that, but someone could still look through our code/conversations and suspect that they have this role. I don't see how filtering out this role from auth_state makes things any more secure.

robnagler: Here's a new thought: when two programmers disagree, there's something wrong with the design. @rorour, you discovered a problem. The original solution to use "negative role" is an incorrect hack. What we would do in a password-based systems is disable the password so they couldn't login any more. Even that does not give a reason or have any audit support. I chose that solution because we have cli's that support this already. Maybe when two programmers disagree, the customer is being lazy (or cheap :). The right solution is more expensive, or is it?

With moderation, we deny and have a person bound to it with a time. This is a similar problem, but it's different. If we were doing true RBAC, we would have a permission to use the system at all. We could have a db upgrade that adds a role "user" or something like that. require_user would check this role, because that's the surrogate. (In BOP, we had such a role as all access, even [visitors got a role and permissions](https://github.com/biviosoftware/perl-Bivio/blob/master/Util/SQL.pm#L962).) 

We could also add then a UserRoleInvite record when we deny the user with a simple command line utility. The login in require_user would be the same, but it would be positive: in order to validate as a user there would have to be a role. auth.login wouldn't allow the user to login either. Perhaps that's giving information that they are denied, but as you note, they will be able to detect when they are denied in any event, and infer something. The key is that we don't return authState with roles unless they are logged in so they don't really get any more information with auth_state.

Thoughts?

rorour: Yes, I think making the role positive rather than negative is a great improvement on this idea.

Would new users have to wait for approval?

robnagler: > Would new users have to wait for approval?

We won't build this yet, but this might be an option. For now, all users will get the role on creation.

rorour:
(code is not ready for re-review)
I think just calling the role "user" is a little confusing, at least it makes function names like "check_user_role" ambiguous. Maybe one of these would be better:
- enabled_user
- allowed_user 
- account_enabled

robnagler:
One of the problems is that we are abusing the notion of "role". User, Visitor, Admin, etc. are roles. sim_type_flash is not a role, it's a capability or permission. Here's a [nice explanation](https://wpmarmite.com/en/glossary/user-role/).

We could add a keyword, e.g. role_user or user_role. The actual role would be "user". Technically, we should have a role called "visitor". The check for cookieless_user is really weird, of course. Another abuse. 

One possibility is registered_user. However this implies there are unregistered users. Maybe an unregistered_user is a visitor.

I think this is an opportunity to reorient ourselves around roles. We can't fix it, but we can at least understand that we are abusing roles and accept true role names.

rorour:
> We could also add then a UserRoleInvite record when we deny the user with a simple command line utility. 

Can you explain more of what you had in mind for this CLI? I was thinking of adding a function `disable_user(uid_or_email)` which handles the UserRoleInvite entry to pkcli/roles, but how would we get the moderator uid?

robnagler:
> how would we get the moderator uid?

An argument to the cli function.

rorour:
What do you think of adding a new table for this? 
UserRoleInvite requires a not null value for column token (part of the moderation request process), and making that nullable requires modifying the table schema [which is not trivial in sqlite](https://stackoverflow.com/questions/4007014/alter-column-in-sqlite).
There is also no "invite" involved in this case, so that suggests this table isn't right for this purpose.

rorour:
@e-carlin and I just had a discussion. Is "token" really adding any security benefit to role moderation?

Maybe we could drop token from UserRoleInvite and still use it for this purpose. I still think a rename would be good in that case.

robnagler:
> Is "token" really adding any security benefit to role moderation?

It wasn't put in there for security afaict. It is used to find the record. It's not used that way in the moderate list when clicking on one of the links (approve, deny, clarify). It's a primary key, not a token so it's misnamed. The token could be switched to (uid, role), which is the actual primary key.

rorour:
@robnagler ping

robnagler:
ya, working through them...

sirepo/db_upgrade.py
```diff
@@ -85,6 +85,22 @@ def _20240507_cloudmc_to_openmc(qcall):
             _migrate_sim_type("cloudmc", "openmc", qcall, u)
 
 
+def _20240524_add_role_user(qcall):
+    if not qcall.auth_db.table_exists("user_role_invite_t"):
+        return
+    qcall.auth_db.execute_sql(
+        """
+INSERT INTO user_role_moderation_t (uid, role, status, moderator_uid, last_updated)
```
robnagler, line 93: Ident this so easier to read. SQL doesn't care about leading whitespace.

sirepo/db_upgrade.py
```diff
@@ -85,6 +85,22 @@ def _20240507_cloudmc_to_openmc(qcall):
...
+SELECT uid, role, status, moderator_uid, last_updated
+FROM user_role_invite_t
+"""
+    )
+    qcall.auth_db.drop_table("user_role_invite_t")
+    for u in qcall.auth_db.all_uids():
```
robnagler, line 99: Since there's direct SQL above. I think it would better and simpler to write some SQL. There is an efficiency and ACID reason to do this. The shorter the transaction, the better.

```sql
INSERT INTO user_role_t (uid, role, expiration) 
VALUES ((select uid from user_registration_t), "user", NULL);
```

"user" could be an injected string, because it is a constant (no chance of sql injection attacks. This is an upgrade, and denormalizing the value makes it simpler to read and avoids the question of sql injection attack.

sirepo/pkcli/roles.py
```diff
@@ -46,13 +46,30 @@ def delete_roles(*args):
     delete(*args)
 
 
+def disable_user(moderator_uid, uid_or_email):
```
robnagler, line 49: This is a bit tricky. I would say put the uid_or_email first (like the other functions in roles). This was related to a discussion we had about C++ arguments in a class switching in an API and causing @moellep confusion. There's also the general idea that the first argument is what is the subject and the other args are modifiers.

The second argument is just `moderator` (not moderator uid). It would be convenient for admins to not have to remember their uid. extract the uid lookup code in _parse_args so that it can be called separately.

sirepo/pkcli/roles.py
```diff
@@ -46,13 +46,30 @@ def delete_roles(*args):
...
+    """Remove role user
+    Args:
```
robnagler, line 51: Formatting isn't right. Needs to be indented.

sirepo/pkcli/roles.py
```diff
@@ -64,7 +81,7 @@ def list_roles(*args):
 # TODO(e-carlin): This only works for email auth or using a uid
 # doesn't work for other auth methods
 @contextlib.contextmanager
-def _parse_args(uid_or_email, roles):
+def _parse_args(uid_or_email, roles=[]):
```
robnagler, line 84: This is a Python anachronism. `roles` is statically bound to a specific `[]` instance. Since `[]` is mutable, it is dangerous to do that. You could use `None` or `()` here.

tests/adm_and_own_jobs_test.py
```diff
@@ -31,19 +31,22 @@ def test_adm_jobs_forbidden(auth_fc):
...
 
     def _op(fc, sim_type):
         with srunit.quest_start() as qcall:
             qcall.auth_db.model("UserRole").delete_all_for_column_by_values(
                 "uid",
                 [fc.sr_uid],
             )
-        r = fc.sr_post(
+        sirepo.pkcli.roles.add(fc.sr_uid, sirepo.auth_role.ROLE_USER)
+
```
robnagler, line 44: remove blank

tests/adm_and_own_jobs_test.py
```diff
@@ -31,19 +31,22 @@ def test_adm_jobs_forbidden(auth_fc):
...
+    import sirepo.pkcli.roles
 
     def _op(fc, sim_type):
         with srunit.quest_start() as qcall:
             qcall.auth_db.model("UserRole").delete_all_for_column_by_values(
                 "uid",
                 [fc.sr_uid],
             )
-        r = fc.sr_post(
+        sirepo.pkcli.roles.add(fc.sr_uid, sirepo.auth_role.ROLE_USER)
```
robnagler, line 43: The delete_all_for_column_by_values is incorrect. We want to only delete the admin role.

tests/adm_and_own_jobs_test.py
```diff
@@ -113,10 +116,13 @@ def _login_as_user(user):
 
     def _make_user_adm(uid):
         from sirepo.pkcli import roles
+        import sirepo.auth_role
 
         roles.add(uid, auth_role.ROLE_ADM)
         with srunit.quest_start() as qcall:
-            r = qcall.auth_db.model("UserRole").search_all_for_column("uid")
+            r = qcall.auth_db.model("UserRole").search_all_for_column(
+                "uid", role=sirepo.auth_role.ROLE_USER
```
robnagler, line 124: This should be ROLE_ADM. The test doesn't make sense the way it was written or with ROLE_USER. The check is for all ROLE_ADM rows.

tests/auth_role_moderation_test.py
```diff
@@ -27,7 +27,8 @@ def test_moderation(auth_fc):
     auth_fc.sr_sim_type_set("myapp")
     with srunit.quest_start() as qcall:
         qcall.auth_db.model("UserRole").delete_all()
```
robnagler, line 29: There's a pattern here. If you find yourself exposing assumptions in code, e.g. "all users must have ROLE_USER", then something is wrong. The delete_all calls (in all cases) was from "lazy" coding. Since tests always create new databases, this works, but it makes an assumption about the roles table that it shouldn't.

In this case, what's wanted is to delete the sym_type role and moderation record for myapp for this particular user.

tests/pkcli/admin_test.py
```diff
@@ -35,7 +35,7 @@ def test_delete_user():
     with srunit.quest_start() as qcall:
         _is_empty_table(qcall.auth_db.model("JupyterhubUser"))
```
robnagler, line 36: _is_empty_table is the incorrect test. Check for this user in the table(s). The assumption that there is only one user in the db is unnecessary.

tests/pkcli/roles_test.py
```diff
@@ -55,5 +55,6 @@ def test_flash_list_role_by_email(auth_fc):
     pkunit.data_dir().join("db").copy(srdb.root())
     auth_fc.sr_email_login(e, sim_type="flash")
     roles.add(e, *r)
+    r.append("user")
     pkunit.pkeq(r, roles.list(e))
```
robnagler, line 59: Again this makes an incorrect assumption. The test should only make sure that there's an entry for sim_type flash.

tests/role_user_test.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+"""Test auth_role.USER
+
+:copyright: Copyright (c) 2024 RadiaSoft LLC.  All Rights Reserved.
+:license: http://www.apache.org/licenses/LICENSE-2.0.html
+"""
+import pytest
+
+
+def test_role_user(fc):
```
robnagler, line 10: test_disable_user.

tests/role_user_test.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
robnagler, line 1: delete

tests/role_user_test.py
```diff
@@ -0,0 +1,40 @@
...
+"""Test auth_role.USER
```
robnagler, line 2: This tests pkcli.admin.disable_user. Make pkcli.admin2_test. pkcli.admin_test does too much to the db and doesn't use `fc` (that is it doesn't start a server).

tests/supervisor_purge_free_sims_test.py
```diff
@@ -53,8 +53,10 @@ def _make_user_premium(uid):
             auth_role.ROLE_PAYMENT_PLAN_PREMIUM,
         )
         with srunit.quest_start() as qcall:
-            r = qcall.auth_db.model("UserRole").search_all_for_column("uid")
-        pkunit.pkeq(r, [uid], "expecting one premium user with same id")
+            r = qcall.auth_db.model("UserRole").search_all_for_column(
```
robnagler, line 56: Excellent!

robnagler (changes requested):

sirepo/auth/__init__.py
```diff
@@ -511,6 +522,7 @@ def require_user(self):
         s = self._qcall_bound_state()
         u = self._qcall_bound_user()
         if u:
+            self.check_role_user()
```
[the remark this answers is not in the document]
rorour: [793c0b5](https://github.com/radiasoft/sirepo/pull/7070/commits/793c0b547ac000e96d88c0a70af24f34c7ad011f)

sirepo/auth/__init__.py
```diff
@@ -184,6 +184,15 @@ def check_sim_type_role(self, sim_type, force_sim_type_required_for_api=False):
             auth_role_moderation.raise_control_for_user(self.qcall, u, r)
         raise sirepo.util.Forbidden(f"uid={u} does not have access to sim_type={t}")
 
+    def check_role_user(self):
```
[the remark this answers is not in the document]
rorour: [793c0b5](https://github.com/radiasoft/sirepo/pull/7070/commits/793c0b547ac000e96d88c0a70af24f34c7ad011f)

sirepo/db_upgrade.py
rorour: [793c0b5](https://github.com/radiasoft/sirepo/pull/7070/commits/793c0b547ac000e96d88c0a70af24f34c7ad011f)

tests/adm_and_own_jobs_test.py
rorour: [793c0b5](https://github.com/radiasoft/sirepo/pull/7070/commits/793c0b547ac000e96d88c0a70af24f34c7ad011f)

sirepo/auth/__init__.py
```diff
@@ -528,14 +540,14 @@ def require_user(self):
                 e = "invalid"
                 self.reset_state()
                 p = PKDict(reload_js=True)
-            e = "auth_method={} is {}, forcing login: uid=".format(m, e, u)
```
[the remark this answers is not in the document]
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

sirepo/pkcli/roles.py
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

sirepo/pkcli/roles.py
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

tests/role_user_test.py
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

tests/role_user_test.py
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

tests/role_user_test.py
rorour: [609c03c](https://github.com/radiasoft/sirepo/pull/7070/commits/609c03c32717dcd27067fe26bc7c2bbb4ee16dc1)

sirepo/auth/__init__.py
```diff
@@ -690,8 +702,7 @@ def _create_roles_for_new_user(self, method):
     def _create_user(self, module, want_login):
         u = simulation_db.user_create()
         if want_login:
-            self._login_user(module, u)
-            self._create_roles_for_new_user(module.AUTH_METHOD)
+            self._login_user(module, u, want_new_roles=True)
```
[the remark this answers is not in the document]
rorour: I'm having a hard time understanding the second half of your comment. Is this what you mean?

```python3
    def _create_user(self, module, want_login):
        def _create_roles_for_new_user(self, method, uid):
            r = sirepo.auth_role.for_new_user(is_guest=method == METHOD_GUEST)
            if r:
                self.qcall.auth_db.model("UserRole").add_roles(roles=r, uid=uid)

        u = simulation_db.user_create()
        if want_login:
            self._login_user(module, u)
        self._create_roles_for_new_user(module.AUTH_METHOD, uid=u)
        return u
```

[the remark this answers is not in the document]
robnagler:
```py
def _create_user(self, module, want_login):
     u = simulation_db.user_create()
     r = sirepo.auth_role.for_new_user(is_guest=method == METHOD_GUEST)
     if r:
         self.qcall.auth_db.model("UserRole").add_roles(roles=r, uid=uid)
     if want_login:
         self._login_user(module, u)
     return u
```

The want_login has to go at the end because there's check_role_user which needs to be added above.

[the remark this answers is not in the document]
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/auth/__init__.py
```diff
@@ -729,6 +740,9 @@ def _login_user(self, module, uid):
                 s = _STATE_COMPLETE_REGISTRATION
         self.qcall.cookie.set_value(_COOKIE_STATE, s)
         self._set_log_user()
+        if want_new_roles:
+            self._create_roles_for_new_user(module.AUTH_METHOD)
+        self.check_role_user()
```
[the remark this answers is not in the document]
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/auth_role_moderation.py
```diff
@@ -56,18 +56,21 @@ def _send_moderation_status_email(info):
...
-        i = self.auth_db.model("UserRoleInvite").unchecked_search_by(
-            token=req.req_data.token
+        i = self.auth_db.model("UserRoleModeration").unchecked_search_by(
+            uid=req.req_data.uid,
+            role=req.req_data.role,
         )
         if not i:
-            pkdlog(f"No record in UserRoleInvite for token={req.req_data.token}")
+            pkdlog(
+                f"No record in UserRoleModeration for uid={req.req_data.uid} role={req.req_data.role}"
```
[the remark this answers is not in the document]
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/auth_role_moderation.py
```diff
@@ -80,7 +83,7 @@ def _set_moderation_status(info):
             status=req.req_data.status,
             moderator_uid=self.auth.logged_in_user(),
         )
-        pkdlog("status={} uid={} role={} token={}", p.status, i.uid, i.role, i.token)
+        pkdlog("status={} uid={} role={}", p.status, i.uid, i.role)
         with self.auth.logged_in_user_set(uid=i.uid, method=self.auth.METHOD_EMAIL):
```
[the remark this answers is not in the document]
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/auth_role_moderation.py
```diff
@@ -174,7 +178,7 @@ def _datetime_to_str(rows):
 
 
 def raise_control_for_user(qcall, uid, role):
-    s = qcall.auth_db.model("UserRoleInvite").get_status(role=role)
+    s = qcall.auth_db.model("UserRoleModeration").get_status(role=role)
```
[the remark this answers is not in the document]
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/db_upgrade.py
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

sirepo/pkcli/roles.py
rorour: [634efd8](https://github.com/radiasoft/sirepo/pull/7070/commits/634efd840a1338c81fced52736fe4cfde315bb63)

tests/adm_and_own_jobs_test.py
rorour: [4508278](https://github.com/radiasoft/sirepo/pull/7070/commits/45082783727593c14d65c391f7a7e8c66412753f)

sirepo/db_upgrade.py
```diff
@@ -96,6 +96,10 @@ def _20240524_add_role_user(qcall):
         """
     )
     qcall.auth_db.drop_table("user_role_invite_t")
+    qcall.auth_db.execute_sql(
+        f"INSERT INTO user_role_t (uid, role, expiration)"
+        + f"VALUES ((select uid from user_registration_t), {sirepo.auth_role.ROLE_USER}, NULL);"
```
robnagler: Does this work? I think ROLE_USER needs to be quoted:

`f"VALUES ((select uid from user_registration_t), '{sirepo.auth_role.ROLE_USER}', NULL);"`

robnagler (changes requested):
One fix required.

sirepo/db_upgrade.py
rorour: That is an old commit.

PR #7076, opened 2024-05-16
Fix #7075: remove refresh_user

### LINKED ISSUE #7075: sirepo.jupyterhub remove refresh_user
In the default case refresh_user is always True. We implemented refresh_user but I don't think it is actually necessary. In the case where a user loses sim_type_jupyterhublogin role (and their container is removed) then authenticate is called again. 

The refresh_user code is failing on calls to update the hub activity which results in additional calls and gums up our logs.

e-carlin:
refresh_user isn't needed for our use. When a user's authentication status changes we remove the role and remove their server. If they try to login again the authenticate() code is called.

refresh_user was failing during notifying the hub of activity after some minutes of uptime. This resulted in many unnecessary requests and log messages.


rorour (approved):
lgtm

e-carlin:
ping @robnagler

robnagler (approved):

PR #7089, opened 2024-05-24
fix #6477: pkcli setup_dev should make mock .tar.gz file for flash instead of text file

### LINKED ISSUE #6477: pkcli setup_dev should make mock .tar.gz file for flash instead of text file

rorour:


sirepo/pkcli/setup_dev.py
```diff
@@ -37,25 +37,29 @@ def _proprietary_codes():
...
         u = f"{cfg.proprietary_code_uri}/{s}-dev.tar.gz"
         try:
-            urllib.request.urlretrieve(u, r)
+            urllib.request.urlretrieve(u, z)
         except urllib.error.URLError as e:
             if not isinstance(e.reason, FileNotFoundError):
                 raise
-            pkdlog("uri={} not found; mocking empty file={}", u, r)
+            pkdlog("uri={} not found; mocking empty file={}", u, z)
+            t = d.join("foo.txt")
```
e-carlin, line 60: I might call this README instead of foo. That way someone might be prompted to read the file and see our message.

sirepo/pkcli/setup_dev.py
```diff
@@ -37,25 +37,29 @@ def _proprietary_codes():
...
             pkio.write_text(
-                r,
+                t,
                 "mocked by sirepo.pkcli.setup_dev",
             )
+            subprocess.run(["tar", "-czf", z, t, "--remove-files"])
```
e-carlin, line 65: -czf could all be expanded to their `--` form. Our general policy is that outside of the terminal the long args increase readability. In the terminal short args decrease readability but the trade off is worth it so we don't have to type quite as much. 

I'd use `check_call` instead of `run`. That way if the command fails an exception is raised in python.

e-carlin (changes requested):

sirepo/pkcli/setup_dev.py
rorour: [2837f23](https://github.com/radiasoft/sirepo/pull/7089/commits/2837f23b92744ad1c7a55d4f2a0d7086d48c8f20)

sirepo/pkcli/setup_dev.py
rorour: [2837f23](https://github.com/radiasoft/sirepo/pull/7089/commits/2837f23b92744ad1c7a55d4f2a0d7086d48c8f20)

e-carlin (approved):

PR #7103, opened 2024-06-04
fix #6394: Add support for srunit logging in uri_router

### LINKED ISSUE #6394: Add support for srunit logging in uri_router
Instead of setting UserAgent. srunit can set a value in the ws header that uri_router prints out in dev mode. @rorour look in `srunit._TestClient._requests_op` for where the user agent is set.

### LINKED ISSUE #6565: log api, sid and uid early in request
`uri_router._call_api` should log the sid, uid, etc. This is done for websockets already so may be able to just move the code.

rorour:
fix #6394: Add support for srunit logging in uri_router
fix #6565: log api early in request

### DIFF
--- a/sirepo/srunit.py
+++ b/sirepo/srunit.py
@@ -701,6 +701,8 @@ class _TestClient:
         from pykern.pkdebug import pkdlog, pkdexc, pkdc, pkdp
         from sirepo import const
 
+        if "headers" not in kwargs:
+            kwargs.headers = PKDict()
         kwargs.headers.setdefault(
             "User-Agent",
             f"{const.SRUNIT_USER_AGENT} {pykern.pkinspect.caller()}",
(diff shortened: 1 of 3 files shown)


sirepo/uri_router.py
```diff
@@ -169,6 +169,10 @@ class _HTTPRequest(web.RequestHandler):
         async def _route(self):
             p = sirepo.uri.decode_to_str(self.request.path)
             e, r, k = _path_to_route(p[1:])
+            pkdlog(
+                "start path={}",
+                p,
+            )
```
rorour, line 175: I think this is too early to log uid. Websockets only log uid at end of request and http already does this as well.

sirepo/srunit.py
```diff
@@ -940,12 +940,10 @@ def _combine_req(encoded_uri):
 
         def _must_be_http(uri):
             # POSIT: /auth- match like sirepo.js msgRouter and https?:
-            # for browser click on email msg. If there are headers,
+            # for browser click on email msg. If there is Authorization header,
```
robnagler: This approach does not work, because websockets don't get http headers so reverted this when I did other changes.

robnagler (commented):
This was more than I could explain. I hope what I did is clear. Please review.

rorour (commented):
looks good to me

PR #7108, opened 2024-06-04
fix #6756: Check for smart quotes in js

### LINKED ISSUE #6756: Check for smart quotes in js
Javascript fails silently when “smart quotes” are used instead of "straight quotes". test.sh could check for the presence of smart quotes.

rorour:
AngularJS sometimes gives an error depending on where the quotes are. To recreate a case where it fails silently, change a data-ng-switch-when in AppFieldEditors to use smart quotes, ex:
`data-ng-switch-when="SimArray"` -> `data-ng-switch-when=“SimArray”`.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -13,7 +13,8 @@ _err() {
 
 _js() {
     local jsfiles=( sirepo/package_data/static/js/*.js )
-    _no_prints '\s(srdbg|console.log|console.trace)\(' "${jsfiles[@]}"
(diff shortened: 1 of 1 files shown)


e-carlin (approved):
I wonder if there is a way to test generally for smart quotes? Seems like PERL has some facilities for it https://stackoverflow.com/a/11380608  but not sure how widespread or correct it is. Anyways, this is good enough for now and if we find ourselves changing this code again we should find something more general.

PR #7126, opened 2024-06-13
Fix #7125: adopt modern language features

### LINKED ISSUE #7125: adpot modern language features
Noticed this tornado change https://github.com/tornadoweb/tornado/pull/3403 . They're using `pyupgrade` to automate the changes. Going to test it out on our code.

e-carlin:
Ran `pyupgrade --py39-plus $(find . -name '*.py')`


e-carlin:
Tossing this out to see if people care. I think the consistency is nice but not sure it is worth the possible introduction of subtle bugs.

robnagler:
We are generating data in files. It removed a blank line from a generated template, but didn't modify the source, because the source may be partially in the template generation code and partially in the template template.

e-carlin:
Ya. After looking more closely at the results I think this tool doesn't work as well as I would hope. For example, [this fairly simple](https://github.com/radiasoft/sirepo/blob/1e5244ccba7b2483f94e42601aacfc2c942d0ca4/sirepo/template/hellweg.py#L477) format it doesn't correct. 

I'm going to close for now. These changes aren't of great importance and it is a decent amount of work to verify the changes (and things that should've been changed).

PR #7129, opened 2024-06-18
Fix #6988: plot csv download

### LINKED ISSUE #6988: add csv plot option to all parameter plots
Currently elegant has a CSV Data File download option. This would be useful for other apps as well, in particular opal.

gurhar1133:


sirepo/template/template_common.py
```diff
@@ -609,7 +638,14 @@ def parameter_plot(x, plots, model, plot_fields=None, plot_colors=None):
...
     if plot_fields:
         res.update(plot_fields)
-    return res
+    columns_dict = PKDict()
+    for i, plot in enumerate(res.plots):
+        c = plot.get("col_name", False)
+        if not c:
+            c = plot.get("name", f"col{i + 1}")
+        columns_dict[c] = plot.points
+    pandas.DataFrame(columns_dict).to_csv(f"{model.plotName}.csv", index=False)
```
moellep, line 647: I uncovered a tricky problem related to browser caching:

When the CSV is created at the time the plot is generated, it can become stale if the client is caching the plot data. We don't have caching implemented yet for web-sockets, but will add it on the client before it is released to production.

For example, enable http communication by turning off the env flag SIREPO_FEATURE_CONFIG_UI_WEBSOCKET. Then visit a visualization plot in elegant. Change the parameters for a plot, and it will request the plot and the csv file will get generated. Then change the plot parameters back to the original, and it will make a new request, but the browser cache will get used, and no call will be made to the server. At that point if you select "download as csv", it will end up with the parameters from the previous plot.

PR #7133, opened 2024-06-19
Fix #6061: Allow configuration of default columns on a per beamline basis

### LINKED ISSUE #6061: raydata: better ordering of results columns
scan_id and plan_name should be included in the default results columns. The ordering should be:

run analysis, view log, status, scan id, plan name, start stop, suid

e-carlin:
In the schema one can set constants.defaultColumns.<beamline>.<list-of-columns> to configure additional default columns for the beamline. These are added to the always on defaults of start, stop, and suid.


sirepo/raydata/scan_monitor.py
```diff
@@ -657,12 +676,15 @@ def _scan_info(
...
-
+    for c in itertools.chain(
+        _default_columns(req_data.catalogName).keys(),
+        req_data.get("selectedColumns", []),
+    ):
+        d[c] = (
+            getattr(m, c)()
+            if c in _METADATA_COLUMNS
+            else m.get_start_field(c, unchecked=True)
+        )
```
rorour, line 687: Personally I find the old way much clearer

rorour (approved):

sirepo/raydata/scan_monitor.py
e-carlin: I agree. The problem is that now _default_columns contains columns from metadata and start doc so we need a way to handle that. Any suggestions on how to improve?

rorour: Ah, that makes sense. I think it's good enough then.

PR #7135, opened 2024-06-21
fix #7117 zpre simulation now runs with material temperature

### LINKED ISSUE #7117: ZPRE example failing with nuclear data error
The openmc ZPRE example is failing with the error:
```Nuclear data library does not contain cross sections for He3 at or near 911.150000 K.```

moellep:
- Added additional material library choices: ENDFB-8.0-NNDC and FENDL-3.1d
- Use temperature.method "interpolation"
- Update zpre example to use the ENDFB-8.0-NNDC which has multiple temperatures


gurhar1133 (approved):
lgtm

PR #7140, opened 2024-06-26
fix #6624: raydata fix column sorting

### LINKED ISSUE #6624: start/stop column sorting is confusing
Both start and stop columns are sorting using the single "time" field. When sorting by "start", it looks like it is actually sorting by the "stop" value. If only the "stop" time is sortable, we should disable sorting on the "start" column. Below shows an example, where "start" is selected for sorting, but "stop" is the value which is actually in sorted order.

raydata

rorour:


rorour:
@moellep ping

moellep (approved):

PR #7146, opened 2024-06-27
Fix #7142: secure communication between agent and scan_monitor

### LINKED ISSUE #7142: raydata: secure communication between agent and scan monitor
Started in https://github.com/radiasoft/sirepo/pull/5416 but it is so old master has diverged too far to easily reconcile. Going to reimplement.

e-carlin:
Create sirepo.http_util to handle auth headers.


sirepo/agent_supervisor_api.py
```diff
@@ -71,11 +50,7 @@ def request(method, uri, token, data=None, json=None):
         json=json,
         data=data,
         verify=sirepo.job.cfg().verify_tls,
-        headers=PKDict(
-            {
-                sirepo.util.AUTH_HEADER: _AUTH_HEADER_PREFIX + token,
-            }
-        ),
+        headers=sirepo.http_util.auth_header(token),
```
robnagler, line 53: This should be encapsulated in sirepo.tornado.AuthHeaderRequestHandler, since it is the one interpreting the requests. It's find to also have it in sirepo.http_util, but from the perspective of this module, it should be in the superclass.

sirepo/raydata/scan_monitor.py
```diff
@@ -254,8 +255,15 @@ async def _incoming(self, body):
         return getattr(self, "_request_" + body.method)(body.data.get("args"))
 
     async def post(self):
+        self._rs_authenticate()
         self.write(await self._incoming(PKDict(pkjson.load_any(self.request.body))))
 
+    def _rs_authenticate(self):
+        t = super()._rs_authenticate()
```
robnagler, line 262: `*args, **kwargs`

sirepo/http_util.py
```diff
@@ -0,0 +1,48 @@
...
+
+#: Http auth header name
+_AUTH_HEADER = "Authorization"
+
+#: http auth header scheme bearer
+_AUTH_HEADER_SCHEME_BEARER = "Bearer"
+
+#: Regex to test format of auth header and extract token
+_AUTH_HEADER_RE = re.compile(
+    _AUTH_HEADER_SCHEME_BEARER + r"\s(" + sirepo.job.UNIQUE_KEY_CHARS_RE + ")",
```
robnagler, line 20: The `\s` should be `\s+` (see [rfc6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1)). unique_key should be moved to sirepo.util. The connection to job is confusing here.

TODO: create_token should be aligned with unique_key. They use two different sizes, but that's too complicated (involves db upgrade). create_token should be fixed to check `value`.

sirepo/http_util.py
```diff
@@ -0,0 +1,48 @@
...
+    re.IGNORECASE,
+)
+
+
+def auth_header(token):
+    """Construct auth header object.
```
robnagler, line 26:
```txt
Construct RFC6750 auth header compatible with `requests`
```

sirepo/http_util.py
```diff
@@ -0,0 +1,48 @@
...
+    Args:
+      token (str): Secret to be included in the header
+    Returns
+      PKDict: auth header
+    """
+    return PKDict({_AUTH_HEADER: f"{_AUTH_HEADER_SCHEME_BEARER} {token}"})
+
+
+def parse_auth_header(headers):
+    """Validate and retrieve authentication header.
```
robnagler, line 37: `Parase and retrieve RFC6750 bearer token

sirepo/http_util.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Args:
+      headers (object): Object containing a `get` method to retrieve headers by name.
+    Returns:
+      bool|str: False if parseable auth header not found otherwise token from auth header.
```
robnagler, line 42: `str: bearer token from header or None if invalid syntax`

sirepo/http_util.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    if not (h := headers.get(_AUTH_HEADER)):
+        return False
```
robnagler, line 45: None is better than False.

sirepo/raydata/scan_monitor.py
```diff
@@ -254,8 +255,15 @@ async def _incoming(self, body):
         return getattr(self, "_request_" + body.method)(body.data.get("args"))
 
     async def post(self):
+        self._rs_authenticate()
```
robnagler, line 258: I wonder if AuthHeaderRequestHandler.post shouldn't always authenticate, but maybe that breaks other uses.

sirepo/raydata/scan_monitor.py
```diff
@@ -254,8 +255,15 @@ async def _incoming(self, body):
...
         self.write(await self._incoming(PKDict(pkjson.load_any(self.request.body))))
 
+    def _rs_authenticate(self):
+        t = super()._rs_authenticate()
+        if t == sirepo.feature_config.for_sim_type("raydata").scan_monitor_api_secret:
```
robnagler, line 263: SIM_TYPE?

robnagler (changes requested):

sirepo/raydata/scan_monitor.py
e-carlin: I think that could work. Every use eventually calls _rs_authenticate but then do their own handling to validate that the token is what they expect it to be. 

Can you describe / give pseudocode for what you have in mind? Should we use the pattern we have for server api's with deny by default, enforce decoration of all routes, and use an init_api like structure to check?

sirepo/agent_supervisor_api.py
e-carlin: Should the same be done for raydata?

https://github.com/radiasoft/sirepo/pull/7146/files#diff-a23ec8b07403deef35f70ca014733d2a3b9e582c40cdc3b7e862a04201e72616R87 

Should it be a class method that creates it?

sirepo/raydata/scan_monitor.py
e-carlin: 390567377

sirepo/http_util.py
e-carlin: 390567377

> create_token should be fixed to check value

What do you mean by this?

sirepo/http_util.py
e-carlin: 390567377

sirepo/http_util.py
e-carlin: 390567377

sirepo/http_util.py
e-carlin: 390567377

sirepo/http_util.py
e-carlin: 390567377

sirepo/raydata/scan_monitor.py
e-carlin: I'm not sure where that would come from. Just set a variable on the module?

sirepo/agent_supervisor_api.py
robnagler: I think we should put in the class for encapsulation purposes. It could also be in sirepo.http_util for the actual work for things that aren't subclasses.

sirepo/http_util.py
robnagler: NVM. Confusion on my part. value must be passed to util.create_token. I thought it could be None in some cases, because there's AuthUserEmail.create_token() which takes no args

sirepo/raydata/scan_monitor.py
robnagler: Yes, "raydata" occurs in three separate places. It seems like this should be a constant in the module.

sirepo/raydata/scan_monitor.py
robnagler: I think it should simply be tornado.AuthRequestHandler and it should have two methods: get and post. get might always raise Forbidden by default. post would look like:

```py
def post(self):
    self.__authenticate()
    self._rs_post()
```

Basically using a dispatch style so the default is always to authenticate. The subclass would implement the _rs methods. __authenticate would call _rs_authenticate and pass the header (token) to _rs_authenticate. If the subclass didn't implement either _rs* method, this class would raise a server error.

sirepo/agent_supervisor_api.py
robnagler: I think we should put in the class for encapsulation purposes. It could also be in sirepo.http_util for the actual work for things that aren't subclasses.

robnagler (changes requested):

sirepo/raydata/scan_monitor.py
e-carlin: 322a67240

sirepo/raydata/scan_monitor.py
e-carlin: 014b4042b . Still trying to learn when you like to use const's:)

sirepo/agent_supervisor_api.py
e-carlin: 322a67240

sirepo/raydata/scan_monitor.py
robnagler: I guess this is a taste thing. I think in this case, we have a model to name SIM_TYPE so that's why I recommended it. One case I often like consts is when there's no assertion on the string.

robnagler (approved):

PR #7163, opened 2024-07-08
fix #6070: hierarchical sorting

### LINKED ISSUE #6070: raydata: hierarchical sorting
Users should be able to select multiple fields to sort by (ex by scan_id and status).

rorour:


rorour:
@moellep This is a bit hacky. Interested to see your feedback. Map seemed like the best structure in JavaScript to contain the sortColumns, but required encoding/decoding for requests

sirepo/package_data/static/js/raydata.js
```diff
@@ -540,6 +540,30 @@ SIREPO.app.directive('scansTable', function() {
                 }
             }
 
+            function decodeSortColumns(encoded) {
+                let columns = new Map();
+                encoded.split(';').forEach((e) => {
```
e-carlin, line 545: What about storing the encoded value using a 2d array? That way we don't have to with commas and semicolons. And we get things like true/false for free.

e-carlin (commented):
Haven't reviewed fully but one comment on using a 2d array instead of a string for the encoded values.

moellep (changes requested):
I can get the "FieldPath cannot be constructed with empty string" error by selecting "Clear Sort" and letting the UI refresh without selecting a new sort field.

For the sorting behavior, I think it is confusing to select a column and now have it become the primary sort. For example, if I'm viewing the list and it is already sorted by "start", and I select "acquire period", it doesn't change the sorting. It does do that if I select "Clear Sort" first and then select "acquire period" but I don't think that is very intuitive.

Instead, you could remove the "Clear Sort" button and keep track of the last 3 sort columns, applied in order. For example, I view the page initially, and it is sorted by "start" by default. Then I select "acquire period" and it sorts by "acquire period", "start". Then I select "sequence id" and it sorts by "sequence id", "acquire period", "start". If you select a forth column, then only the last three selected columns would be used for sorting.

rorour:
> I'm getting errors when I try to perform certain sorts. For example, select csx, add fccd_intersection, nanop_setup, detectors. The clear sort, sort fccd_intersection, sort nano_setup, sort detectors, see error.

This seems to be some kind of restriction with Mongo sort. 

https://www.mongodb.com/community/forums/t/sorting-on-multiple-fields-that-may-or-may-not-be-arrays-cannot-sort-with-keys-that-are-parallel-arrays/251689

https://jira.mongodb.org/browse/SERVER-826

Any suggestions on how to handle this? I guess we could check the type of data in each column and only allow one array type column to be added to sort.

e-carlin:
Interesting. For now let's just not allow sorting of any array field and see if anyone asks for it.

rorour:
Ready for re-review. I went back and forth on whether to remove the "Clear Sort" button but I did remove it.

sirepo/package_data/static/js/raydata.js
```diff
@@ -809,15 +817,37 @@ SIREPO.app.directive('scansTable', function() {
...
 
             $scope.columnIsSortable = (column) => {
-                return column !== 'stop';
+                if (['stop'].includes(column)) {
+                    return false;
+                } else if ($scope.scans.length > 0 && typeof $scope.scans[0][column] === 'object') {
+                    return false;
+                } else {
+                    return true;
+                }
```
moellep, line 826: It looks like the "sortable" value is getting confused when I sort on a field.

For example, I'm viewing CHX with the columns (sequence id, T_sample_, acquire period, T_yoke, detectors, samples). When I visit the page, it is sorted by start. Then I select "acquire period" and it sort that column, but I can no longer select the column again to reverse the sorting. The other columns are also unsortable at this point, except for "start" and "sample".

rorour: [b98f2e9](https://github.com/radiasoft/sirepo/pull/7163/commits/b98f2e99939da3aafba92716dbfd92e31ed4d63a)

rorour:
@moellep @e-carlin ping

moellep:
I find I can get the sorting in an odd state, where selecting "acquire period" doesn't always sort the same. If I open the CHX page, and sort a few times on "acquire period", it behaves as expected, where empty fields either appear at the top or the bottom of the sort.

But if I open a fresh CHX page and then sort on "T_sample_", "acquire period", "T_yoke", and then back to "acquire period", it always keeps the empty rows at the top of the sorted list even if I keep selecting "acquire period".

x

rorour:
Thanks for catching that. When reversing the order of a column that was already being sorted on, it was remaining at the same sort priority. It makes more sense to make that column become priority again. Should be fixed.

rorour:
Taking screenshots for the report got me thinking about whether it would be nice to have some visual indicator of sort order. Currently a user has to remember which column they clicked most recently, or study the column values. 

I think color gradient of the arrows would work, with the most saturated color being the top priority. Or just an indicator on the top priority column. On the other hand, sort order isn't preserved in appState so maybe it's unlikely that a user would forget which column they clicked last.
Any thoughts?

moellep:
Yes, I like the color saturation idea - show the top priority darker than the sub priorities.

moellep (approved):

sirepo/package_data/static/js/raydata.js
```diff
@@ -677,13 +682,27 @@ SIREPO.app.directive('scansTable', function() {
...
+                    }
+                });
+                return r;
+            };
+
+            $scope.arrowStyle = column => {
+                let r = '';
+                scanArgs.sortColumns.forEach((sortField, i) => {
+                    if (sortField[0] === column) {
+                        let a = i === 0 ? 1 : (-0.3 * i) + 0.8;
```
e-carlin, line 701: i === 0 deserves a comment. I think it is because sortColumns is ordered from highest to lowest priority but it would be good to let the reader know that.

I think we might want an assert that sortColumns.length <= 3. `(-0.3 * i) + 0.8` goes negative at index 3 and above which I don't think is valid. Not 100% about this.

sirepo/package_data/static/js/raydata.js
```diff
@@ -809,15 +828,39 @@ SIREPO.app.directive('scansTable', function() {
             };
 
             $scope.columnIsSortable = (column) => {
-                return column !== 'stop';
+                if (['stop'].includes(column)) {
```
e-carlin, line 831: return false/ true from an if/else statement is a code smell. Almost always you can instead return the result of the test.

` return ! (column === 'stop' || $scope.scans.length > 0 && $scope.scans[0][column] !== null && typeof $scope.scans[0][column] === 'object');`

sirepo/package_data/static/js/raydata.js
```diff
@@ -809,15 +828,39 @@ SIREPO.app.directive('scansTable', function() {
...
+                        scanArgs.sortColumns.splice(i, 1);
+                        scanArgs.sortColumns.unshift(sortField);
+                    }
+                });
+                if (! inList) {
+                    if ($scope.analysisStatus !== 'allStatuses') {
+                        scanArgs.sortColumns = [[column, true]];
+                    } else {
+                        scanArgs.sortColumns.unshift([column, true]);
+                        if (scanArgs.sortColumns.length > 3) {
```
e-carlin, line 859: 3 should be captured in a const somewhere globally. It is important.

e-carlin (changes requested):
Worked well. Some final style comments.

sirepo/package_data/static/js/raydata.js
```diff
@@ -677,13 +682,27 @@ SIREPO.app.directive('scansTable', function() {
...
+                            [`glyphicon-arrow-${sortField[1] ? 'up' : 'down'}`]: true,
+                        };
+                    }
+                });
+                return r;
+            };
+
+            $scope.arrowStyle = column => {
+                let r = '';
+                scanArgs.sortColumns.forEach((sortField, i) => {
```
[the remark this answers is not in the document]
rorour: [7322d19](https://github.com/radiasoft/sirepo/pull/7163/commits/7322d198d8db767467b2dfee55a74b6b9b0cba27)

sirepo/package_data/static/js/raydata.js
rorour: if `a` goes negative, the arrow just won't show, so the behavior is the same. [7322d19](https://github.com/radiasoft/sirepo/pull/7163/commits/7322d198d8db767467b2dfee55a74b6b9b0cba27)

sirepo/package_data/static/js/raydata.js
rorour: I honestly preferred the way I had it, it was easier to read and understand the different cases, and therefore probably easier to change in the future. [7322d19](https://github.com/radiasoft/sirepo/pull/7163/commits/7322d198d8db767467b2dfee55a74b6b9b0cba27)

sirepo/package_data/static/js/raydata.js
rorour: [7322d19](https://github.com/radiasoft/sirepo/pull/7163/commits/7322d198d8db767467b2dfee55a74b6b9b0cba27)

sirepo/package_data/static/js/raydata.js
e-carlin: It is an interesting matter of style. Our style tends to favor less code over more in general. Fewer locals, omitting else, etc. I would put this in that category. I find following the braces harder to read.

Some linters complain about code with boolean tests that return the result of the test
https://pylint.readthedocs.io/en/stable/user_guide/messages/refactor/simplifiable-if-expression.html

e-carlin (approved):

PR #7168, opened 2024-07-09
fix #7156: reformat raydata detailed status panel contents

### LINKED ISSUE #7156: raydata: scan status use friendly datetime instead of unix timestamp
Currently the scan details shows timestamps as a unix timestamp. They should be a human readable timestamp.

rorour:


e-carlin (changes requested):
Are there instructions somewhere on how to create the detailedStatusFile and get it in the right place so the GUI picks it up? Maybe the notebook can be modified to generate this file for us (for now) so we don't have to copy files around?

rorour:
added in https://github.com/radiasoft/raydata/pull/158

e-carlin (approved):

PR #7178, opened 2024-07-16
Fix #7169: Allow upload of .stp files

### LINKED ISSUE #7169: cloudmc: step file to h5m
Users should be able to upload step files and we can then convert them to h5m so they can be used by openmc.

To start with use well formed step files that don't require modification to be usable by openmc (ex they are already watertight). A follow on PR will improve the conversion so poorly formed files can be fixed.

e-carlin:
.stp files will be converted to .h5m as part of dagmAnimation.

Currently there is a basic sanity check (check_watertight) that the conversion went ok. In the future we will do more to verify (and help fix) conversions that had trouble.


e-carlin:
@moellep you can test with the box.stp file added to test data.

https://github.com/radiasoft/sirepo/pull/7178/files#diff-0c3834b09578f93afbeca9e80db77fe315f490e76719f500f3112415fd8cc8e1

sirepo/template/openmc.py
```diff
@@ -46,7 +46,8 @@ def background_percent_complete(report, run_dir, is_running):
 def get_data_file(run_dir, model, frame, options):
     sim_in = simulation_db.read_json(run_dir.join(template_common.INPUT_BASE_NAME))
     if model == "geometry3DReport":
-        return _SIM_DATA.dagmc_filename(sim_in)
+        d, s = _SIM_DATA.dagmc_and_maybe_step_filename(sim_in)
+        return s or d
```
moellep, line 50: In this case, I think it should always download the h5m file because the panel that provides the download is "DAGMC Geometry".

sirepo/pkcli/openmc.py
```diff
@@ -38,6 +38,37 @@ def extract_dagmc(dagmc_filename):
...
+    a = CAD_to_OpenMC.assembly.Assembly([input_step_filename])
+    a.import_stp_files()
+    a.merge_all()
+    a.solids_to_h5m(backend="stl", h5m_filename=output_dagmc_filename)
+    o = subprocess.check_output(("check_watertight", output_dagmc_filename))
+    r = re.findall(rb"(\(0%\) unsealed)", o)
+    if len(r) != 2:
+        raise AssertionError(
+            f"output_dagmc_filename={output_dagmc_filename} not watertight: {o}"
+        )
```
moellep, line 69: A TODO() here would be to show the error message in the UI.

sirepo/template/openmc.py
```diff
@@ -235,8 +241,8 @@ def validate_file(file_type, path):
     import h5py
 
     if file_type == "geometryInput-dagmcFile":
-        if not h5py.is_hdf5(path):
-            return "dagmcFile must be valid hdf5 file"
+        if path.ext == ".h5m" and not h5py.is_hdf5(path):
+            return "File is not a valid hdf5 file."
```
moellep, line 245: Another TODO() here would be to check that the stp file is valid.

moellep (approved):

PR #7182, opened 2024-07-18
fix #6866 client side cache when running with web sockets

### LINKED ISSUE #6866: cache sim_frame results on client when using websockets
Previous sim_frames were cached in the browser's http cache. With websockets that isn't available, so calls to frameCache will need to keep the data locally and then clear on "modelsUnloaded".

Also consider storing the result on the server and not recalculating unless a frame_arg has changed from the prior request.

moellep:
Also includes some improvements for appState.deepCopy() and openmc save calls.


sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,155 @@ SIREPO.app.service('validationService', function(utilities) {
 
 });
 
+SIREPO.app.factory('srCache', function(appState, $rootScope) {
+
```
moellep, line 1548: @e-carlin and @robnagler srCache is the client side cache for web sockets, using indexedDB

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+    // - Keeps MRU date on item access and deletes expired records at startup
+
+    const self = {};
+    const STORE = 'db';
+    const FRAME = 'frame';
+    // 30 days until expired
+    const EXPIRY_TIME = 30 * 24 * 60 * 60 * 1000;
+    let db = null;
+
+    function deleteKeys(keys) {
```
robnagler, line 1561: I've been using arrow style with const style.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        if (keys.length) {
```
robnagler, line 1562: early return?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+                }
+            }
+        }
+    }
+
+    function invokeCallback(callback, value) {
+        $rootScope.$applyAsync(() => callback(value));
+    }
+
+    function key(prefix, value) {
```
robnagler, line 1576: objectKey?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+                return db.transaction(STORE, mode).objectStore(STORE);
+            }
+        }
+        catch (e) {
+        }
+        return null;
+    }
 
-SIREPO.app.factory('frameCache', function(appState, panelState, requestSender, authState, $interval, $rootScope, $timeout) {
+    function open() {
```
robnagler, line 1591: Reading this is hard. I would prefer _open or _init. I know we don't have a style for this, but "open" looks like a primitive.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+            }
+        };
+        c.onerror = () => {
+            invokeCallback(callback, null);
+        };
+    };
+
+    self.saveFrame = (frameId, modelName, data) => {
+        const o = objectStore('readwrite');
+        if (o) {
```
robnagler, line 1674: early return?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        c.onsuccess = (event) => {
+            const d = event.target.result;
+            invokeCallback(callback, d);
+            if (d) {
+                // update MRU (most recently used)
+                self.saveFrame(frameId, modelName, d);
+            }
+        };
+        c.onerror = () => {
+            invokeCallback(callback, null);
```
robnagler, line 1668: do we want to log the error even for debugging purposes? Do we want an errorCallback?

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+                }
+                c.continue();
+            }
+            else {
+                deleteKeys(expired);
+            }
+        };
+    }
+
+    self.clearFrames = (simId, modelName) => {
```
robnagler, line 1632: Having frames and generic "object" seems odd to me. This is srFrameCache or something like that.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        };
+        c.onerror = () => {
+            invokeCallback(callback, null);
+        };
+    };
+
+    self.saveFrame = (frameId, modelName, data) => {
+        const o = objectStore('readwrite');
+        if (o) {
+            data._srcache_mru = new Date().getTime();
```
robnagler, line 1675: I think "mru" should be updateTime, which is a term used elsewhere in the code. The index is updateTime as well.  It's not actually MRU, and instead it has a fixed cutoff.

Maybe over optimization, but creating a new Date every time seems unnecessary.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        return prefix + ':' + value;
+    }
+
+    function objectStore(mode) {
+        try {
+            if (db) {
+                return db.transaction(STORE, mode).objectStore(STORE);
+            }
+        }
+        catch (e) {
```
robnagler, line 1586: At least a comment here.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,154 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        $rootScope.$applyAsync(() => callback(value));
+    }
+
+    function key(prefix, value) {
+        return prefix + ':' + value;
+    }
+
+    function objectStore(mode) {
+        try {
+            if (db) {
```
robnagler, line 1582: I think there's a race condition here on startup. It may be that db is possible, but onsuccess hasn't been called yet.

robnagler (changes requested):
I think the naming needs to be more consistent.

sirepo/package_data/static/js/sirepo.js
moellep: I changed it from the generic open() to initializeDatabase() which is more descriptive.

sirepo/package_data/static/js/sirepo.js
moellep: I've updated the early return calls with a callback:
```
        withObjectStore('readwrite', (o) => {
            // ...
        })
```

sirepo/package_data/static/js/sirepo.js
moellep: Yes, it could be that onsuccess() call is pending, but a call to getFrame would return a cache-miss. I think that is better than having to wait for the browser to sort out the database if there is a problem. In practice the db seems to initialize very quickly.

sirepo/package_data/static/js/sirepo.js
moellep: I don't think we want to log every getFrame() call if it fails. Possibly the first failure. The caller can't really do any additional handling other than treating it as a cache-miss.

sirepo/package_data/static/js/sirepo.js
moellep: @robnagler I've made some improvements with 497424e

robnagler (approved):

e-carlin (approved):
I left a couple comments but overall looks good and worked well.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,157 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        };
+    };
+
+    const invokeCallback = (callback, value) => {
+        $rootScope.$applyAsync(() => callback(value));
+    };
+
+    const objectKey = (prefix, value) => prefix + ':' + value;
+
+    const objectStore = (mode) => {
```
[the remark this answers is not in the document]
moellep: In past projects we've used an "unsafe" prefix for functions which might return null, so it could be ```unsafeObjectStore```. I'm not fond of the "maybe" prefix. In pykern, "unchecked" isn't documenting the return value, but that the function won't throw an exception on failure.

I couldn't find any good references for how to mark a method name which might return null with javascript. Javascript seems to be "unsafe" by default - most of the library functions like getElementById() or getSelection() can return null.

The function could be renamed ```getObjectStore``` and documentation added to the function stating that it will return null if the object store is not available.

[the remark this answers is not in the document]
e-carlin: > In pykern, "unchecked" isn't documenting the return value, but that the function won't throw an exception on failure.

In sirepo it documents the return value (ex [unchecked_get_user](https://github.com/radiasoft/sirepo/blob/7a5b5d08b04fb6175d76fd454bfdacb7c34afdd4/sirepo/auth/__init__.py#L581). 

> getObjectStore

get or unsafe (or unchecked) are all good with me.

sirepo/package_data/static/js/sirepo.js
```diff
@@ -1556,14 +1544,157 @@ SIREPO.app.service('validationService', function(utilities) {
...
+        if (! window.indexedDB || ! SIREPO.authState.uiWebSocket) {
+            return;
+        }
+        const r = window.indexedDB.open('srCache', 2);
+        r.onsuccess = (event) => {
+            db = event.target.result;
+            removeOldRecords();
+        };
+        r.onupgradeneeded = (event) => {
+            db = event.target.result;
```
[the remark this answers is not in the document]
moellep: Yes, the onupgradeneeded handler shouldn't be setting ```db```. It turns out the onsuccess() gets called as well when onupgradedneeded completes sucessfully, so that is the only spot that needs to set the db. I've updated the code for this.

```If the onupgradeneeded event exits successfully, the onsuccess handler of the open database request will then be triggered.```

sirepo/package_data/static/js/sirepo.js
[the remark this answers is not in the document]
moellep: Yes, it looks like there are examples in sirepo of unchecked for null returns. Also ```unchecked_search_by```. 

For now I've changed it to ```getObjectStore``` and added the doc.

PR #7199, opened 2024-07-31
fix #7153: raydata enable/disable auto processing of scans from the GUI

### LINKED ISSUE #7153: raydata: enable/disable auto processing of scans from the GUI
Currently this is a server configuration parameter. We should be able to turn it on / off from the GUI

rorour:
depends on https://github.com/radiasoft/pykern/pull/498


sirepo/raydata/scan_monitor.py
```diff
@@ -592,15 +613,32 @@ def _get_detailed_status(catalog_name, rduid):
...
-        return asyncio.create_task(_poll_catalog_for_scans(catalog_name))
-
-    if not cfg.automatic_analysis:
-        return
     for c in cfg.catalog_names:
-        _CATALOG_MONITOR_TASKS[c] = _monitor_catalog(c)
-    await asyncio.gather(*_CATALOG_MONITOR_TASKS.values())
+        _AUTOMATIC_ANALYSIS_HANDLER_TASKS.append(
+            pkasyncio.create_task(_handle_automatic_analysis(c))
+        )
```
moellep, line 619: I'm getting an error ```module 'pykern.pkasyncio' has no attribute 'create_task'``` when running the scan monitor on this branch - is ```pkasyncio.create_task()``` a new method, or should this be ```asyncio.create_task()```?

rorour: Whoops. It's just a wrapper over asyncio.create_task, but here is the branch with the change. https://github.com/radiasoft/pykern/pull/498

moellep (approved):
This PR will need to wait for pykern pkasyncio-changes PR to be merged.

@e-carlin could you check the logic in scan_monitory.py

e-carlin:
I tried to leave a review in words. But, I was struggling to get my points across. So, I reviewed by changing the code. Please take a look fcd6b63be. We can setup a meeting to discuss if you want

Some general thoughts:
- Reviewing the code there was something that just seemed a bit too complicated to me. I had to jump back and forth between `_AUTOMATIC_ANALYSIS_HANDLER_TASKS` and `_CHANGE_AUTOMATIC_ANALYSIS` to see what was going on. There just seemed to be too much logic for what amounts to one question. Should this task be doing work (analysis) or not?
- I noticed that [this function](https://github.com/radiasoft/sirepo/blob/085e45491c6b79a3d9c31e6c89d9e77954e2e06d/sirepo/raydata/scan_monitor.py#L615) no longer awaited anything but just created some tasks. So, there was no reason to pass it to [run](https://github.com/radiasoft/sirepo/blob/085e45491c6b79a3d9c31e6c89d9e77954e2e06d/sirepo/raydata/scan_monitor.py#L822). asyncio has some real foot guns around tasks that no one awaits. I'm still not 100% sure myself how it all works. But, basically I think if you create a task you should await it. If you don't I think exceptions might not be propagated correctly. Again, not 100% sure myself how it all works but having tasks that weren't awaited was odd to me. 
- [This function](https://github.com/radiasoft/sirepo/blob/085e45491c6b79a3d9c31e6c89d9e77954e2e06d/sirepo/raydata/scan_monitor.py#L632) was a bit of a smell. You just pkdel from _CATALOG_MONITOR_TASKS but the task is never explicitly canceled. Maybe the garbage collector cancels it? Not sure. Cancel is really really tricky. Something we wish we had handled differently in the job_supervisor. But going forward my general rule is to avoid cancel in the form of python CanceledError like the plague. Instead, do what you were sort of doing and signal to tasks that they should stop whatever they are doing (and maybe "cancel" themselves by returning). That's what my code does. It uses [`_WANT_AUTOMATIC_ANALYSIS_FOR_CATALOG`](https://github.com/radiasoft/sirepo/pull/7199/files#diff-57acfe7ca49b49bcc9479f33d06a6548fbef69101ca41595325540233f4035a8R63). This is similar to your `_CHANGE_AUTOMATIC_ANALYSIS` but instead of signaling a task that signals a task like in your code. I just have the lowest level task itself look at the value and do the right thing. I avoid the cancel problem entirely but just having it sit in a loop when it shouldn't be doing any analyzing. 

Hope my explanation helps some...

rorour:
This is a great simplification. I started to lose the forest for the trees with this one.

e-carlin:
I'm glad it was helpful.

I wanted to try and figure out again what the rules are around create_task and awaits. I came across [this issue](https://github.com/python/cpython/issues/104091). This async python stuff is tricky and even the creators of it don't have it all nailed down...

PR #7224, opened 2024-08-23
Fix #7222: Handle 'proxy-for' header

### LINKED ISSUE #7222: Support proxy-for
We need to handle this in sirepo.request. Tornado handles X-Forwarded-For and X-Real-IP. We'll need to set remote_ip specially if this header is set.

e-carlin:
Some proxys set the remote ip in this header.


robnagler (approved):
It would be ideal if we could subclass to fix this, but I see how complicated it is.

PR #7225, opened 2024-08-23
Fix #7223: Support sending to smtp server without authentication.

### LINKED ISSUE #7223: smtp without auth
https://github.com/radiasoft/sirepo/blob/5cf751cf32102a612f1a184f9ad30ccf3e05b973/sirepo/smtp.py#L71 but no auth. We are deploying at a customer that has an smtp server that doesn't require auth.

e-carlin:


sirepo/smtp.py
```diff
@@ -94,14 +95,14 @@ def _init():
         _cfg.server = "not " + _DEV_SMTP_SERVER
         _SEND = _send_directly
         return
-    _SEND = _send_with_auth
+    _SEND = _send_to_configured_smtp_server
```
robnagler, line 98: `_send_via_relay_server`. Config probably should be relay_server, but leave as is.

sirepo/smtp.py
```diff
@@ -94,14 +95,14 @@ def _init():
...
     if pkconfig.in_dev_mode():
         if _cfg.server is None:
             _cfg.server = _DEV_SMTP_SERVER
         return
-    if _cfg.server is None or _cfg.user is None or _cfg.password is None:
+    if _cfg.server is None:
```
robnagler, line 103: It should probably be an error if user is set without password. Certainly password set without user.  Catch that at config time.

robnagler (commented):
Remove `__future__` and `coding`

sirepo/smtp.py
e-carlin: 8f9c96785

sirepo/smtp.py
e-carlin: 8f9c96785

PR #7265, opened 2024-09-19
fix #7152 new "canvas" app for lattice code comparisons

### LINKED ISSUE #7152: create a general purpose lattice app
Create a lattice building application which is independent of a specific code. Allow code generation, simulation and output comparison between elegant, impactx and madx.

moellep:
- adds h5 input beam file support to elegant and mad-x
 - includes an alpha version of the impactx app


e-carlin (approved):
I clicked around the apps that were changed and all looked good.

One thing I saw but I'm not sure I understand the app entirely: If I disable one of the codes in the canvas app I still see their output in the reports. Ex. disable madx but a madx phase space plot is still shown.

PR #7321, opened 2024-10-18
fix #7179 use weight windows file in new simulation

### LINKED ISSUE #1755: warpvnd: resize browser doesn't update 3d axis labels
When viewing the 3D particle trace, if you resize the browser window, the axis labels aren't shown with the new plot width. For example, if you widen the browser window, the x axis labels no longer span the whole plot. This is not a high priority issue.

Screen Shot 2019-07-09 at 3 48 06 PM

### LINKED ISSUE #6325: axis is swapping direction depending on orientation
When looking at the Tokamak Reactor 3D geometry in the z direction, the Y axis swaps direction when the plot is slightly rotated. I would expect the axis to stay constant with the 0 values fixed.

x

x2

### LINKED ISSUE #7179: Run sim with weight windows input file
In openmc, you could run a variance reduction sim, then save the weights file and then use that as input to a larger production run.

There may be an example of running from a weights file here:
https://github.com/fusion-energy/neutronics-workshop/tree/main/tasks/task_14_variance_reduction

The UI needs to allow using a weights file generated from a previous run (possibly saved to the user's lib file list). When a weights file is selected, the variance reduction tab should not be shown.

moellep:
also fix #1755 resizing browser window now correctly updates 3d plot axes
also fix #6325 incorrect axis swap


moellep:
@e-carlin the change to the openmc app was to allow the weight-windows output file to be saved after running a variance reduction sim. There is also an option to save the weight-windows file to the user's lib directory and use that file when running a new simulation. It does this by saving the weight-windows file as a sim_file after the simulation completes (which I think brings it back from the remote (sbatch) to the local server). That sim_file is then available to copy into the lib dir if the user requests it. I tested on dev with a local sbatch and it seems to work ok.

To test it, use the Tokamak Reactor example, then set the Variance Reduction parameters on the Visualization page:

x

Then run the sim, and the weight windows download link and button will appear on the simulation status:

x2

When you select the "Apply Weight Windows" it will show a confirmation page, and then copy the sim file to the lib and set up the simulation for running with that lib file.

sirepo/template/openmc.py
```diff
@@ -673,6 +704,22 @@ def _generate_run_mode(data, v):
...
 
 
+def _generate_save_weight_windows(data):
+    if data.models.settings.varianceReduction in (
+        "weight_windows_mesh",
+        "weight_windows_tally",
+    ):
+        return f"""
+import sirepo.sim_data
+sirepo.sim_data.get_class("{ SIM_TYPE }").put_sim_file(
```
e-carlin, line 714: I think it might make sense to create a put_lib_file interface. The file is only read from the sim dir just to copy it to the lib dir.

e-carlin (approved):
Works well. Left two comments about the structure.

sirepo/template/openmc.py
```diff
@@ -212,6 +220,21 @@ def stateful_compute_download_remote_lib_file(data, **kwargs):
     return PKDict()
 
 
+def stateless_compute_save_weight_windows_file_to_lib(data, **kwargs):
```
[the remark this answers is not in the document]
moellep: updated with e5b7689

sirepo/template/openmc.py
moellep: In this case, it is an option to copy the sim file to the lib directory. The user may not want to use the computed file for other simulations.

PR #7329, opened 2024-10-21
fix #7326 set multiElectronAnimation.watchpointId for examples

### LINKED ISSUE #7326: srw multielectronanimation needs to default watchpointid
When running a partially coherent animation, template srw needs to set the multElectronAnimation.watchPointId if it is not set, probably to the last watchpoint. Otherwise the "Intensity After" value in the title is blank.

moellep:
This backs out the change to animation_test.test_srw() and makes sure all examples have a valid multiElectronAnimation.watchpointId set.


robnagler (approved):
Thanks!

PR #7348, opened 2024-11-05
Fix #7344 require premium for sbatch

### LINKED ISSUE #7344: Add require_premium
UI should check in advance for APIs which are premium and issue a message.

robnagler:


e-carlin (approved):
LGTM

PR #7389, opened 2024-12-03
fix #7388 add security.txt

### LINKED ISSUE #7388: Add security.txt conforms to RFC9116
```txt
Contact: mailto:support@radiasoft.net
Expires: 2025-03-14T00:00:00.000Z
```

Expires is just some time in the future. Make +1y

moellep:
Added dynamic security.txt file API and generate with pkcli.static_files.

Move robots.txt generated file to static because nginx looks for it there.

Removed caching of robots.txt in server because it is referred to using the output from pkcli.static_files on nginx.

Removed copy and verification of favicons to root because nginx refers to the files in static/img already.


sirepo/pkcli/static_files.py
```diff
@@ -28,29 +26,19 @@ def gen(target_dir):
 class _Gen(PKDict):
     def __init__(self, target_dir):
         self.tgt = pykern.pkio.py_path(target_dir)
-        self.count = PKDict(root=0)
         for r, s in sirepo.resource.static_files():
             self._copy(r, s)
-            self._maybe_root(r, s)
-
-        self._verify()
```
robnagler, line 36: Why did you remove this stuff?

moellep: Move robots.txt generated file to static because nginx looks for it there.
Removed caching of robots.txt in server because it is referred to using the output from pkcli.static_files on nginx.
Removed copy and verification of favicons to root because nginx refers to the files in static/img already.

sirepo/server.py
```diff
@@ -432,6 +432,20 @@ async def api_saveSimulationData(self):
             ),
         )
 
+    @sirepo.quest.Spec("allow_visitor")
+    async def api_securityTxt(self):
+        d = sirepo.srtime.utc_now()
+        d = d.replace(year=d.year + 1, hour=0, minute=0, second=0, microsecond=0)
```
robnagler, line 438: This doesn't work right. On 12/31/2024, the server will return 1/1/2025. The server will then return expiry times that are in the past, especially if there is no prod release around 12/25, which is likely. Is there a reason to force to a specfic date/time? If so, set it to +2 years.

PR #7404, opened 2024-12-20
Fix #6914 job_agent recover existing sbatch jobs

### LINKED ISSUE #6914: Add slurm job id to the logs and ideally the job db

### LINKED ISSUE #7308: Remove SIREPO_FEATURE_CONFIG_UI_WEBSOCKET=0 test case
Web sockets are now enabled on all servers, so we could speed up the automated tests by removing the SIREPO_FEATURE_CONFIG_UI_WEBSOCKET=0 case.

### LINKED ISSUE #7385: Remove supervisor `_run` task
Working #6916 after a bit of a break. I realized that the `_run` task is unnecessary, and overcomplicates the process. [`_run`](https://github.com/radiasoft/sirepo/blob/60310d60eb862116bf4a80fada4bab19b437a06a/sirepo/job_supervisor.py#L951) sends the op, and then loops on [`reply_get`](https://github.com/radiasoft/sirepo/blob/60310d60eb862116bf4a80fada4bab19b437a06a/sirepo/job_supervisor.py#L1210). This is how you would write a normal threaded application: receive a message and then pass it off to a worker thread. However, that's now how you should write an asyncio application: callbacks should handle the operation immediately, because all other tasks block anyway. There is not priority scheduling or multi-processing going on. This is very hard for me to get my head around. In asyncio, it's always better to have fewer tasks than more when they are related to a single operation.

[`reply_put`](https://github.com/radiasoft/sirepo/blob/60310d60eb862116bf4a80fada4bab19b437a06a/sirepo/job_supervisor.py#L1219) should look at the response type to decide what to do. In the case of [`_send_with_single_reply`](https://github.com/radiasoft/sirepo/blob/60310d60eb862116bf4a80fada4bab19b437a06a/sirepo/job_supervisor.py#L1042) (all cases except cancel and run), it should put in a queue. In the case of run and cancel, it should simply hand off the response to a handler. Cancel does nothing with the reply. Run will process a status.

Run status is fully asynchronous so it will update the status of the job. If the job is not in the cache, the job should be loaded and updated, because the agent is valid (job_driver validates agents). 

This would allow a simpler model for run status in the agent, which would monitor all jobs it was asked to monitor (at any time). If it is already monitoring, it sends back a reply with the current status (which it already caches in job_cmd). If it is not monitoring, it would start monitoring. This would make reattaching quite simple: request run_status on the job. No need to start a task (and worry about deadlock) as run status replies are idempotent.

robnagler:
- Fix #7308 ui_websocket default is True and removed False case from test.sh
- Fix #7385 job_supervisor run returns immediately and is not a task
- job_supervisor run_status_op pends until run or status watcher complete
- run_status_update is new op that is sent asynchronously from agent to supervisor
- job_agent separate out logic for run/state; reconnects to sbatch job
- job_cmd restructured and more error handling
- job_cmd centralized dispatch in _process_msg
- job_cmd._do_compute more robust and supports separate run/status
- job documents more ops and statuses
- Added max_procs=4 to test.sh to parallelize tests
- Fixed global state checks (mpiexec) to allow parallel test execution
- Increased timeouts to allow for delays during parallel test execution
- Improve arg validation in simulation_db.json_filename
- sbatchLoginService commented out invalid state transitions
- SIREPO.srlog includes time

### DIFF
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -110,11 +110,7 @@ class DriverBase(PKDict):
                 await self.kill()
                 self._websocket_ready_timeout_cancel()
                 self._websocket_ready.clear()
-                w = self._websocket
-                self._websocket = None
-                if w:
-                    # Will not call websocket_on_close()
-                    w.sr_close()
+                self._websocket_close()
                 e = f"job_driver.free_resources caller={caller}"
                 for o in list(self._prepared_sends.values()):
                     o.destroy(internal_error=e)
@@ -201,6 +197,13 @@ class DriverBase(PKDict):
                 # If one kill fails still try to kill the rest
                 pkdlog("error={} stack={}", e, pkdexc())
 
+    def _websocket_close(self):
+        w = self._websocket
+        self._websocket = None
+        if w:
+            # Will not call websocket_on_close()
+            w.sr_close()
+
     def websocket_on_close(self):
         pkdlog("{}", self)
         self._websocket = None
@@ -281,6 +284,21 @@ class DriverBase(PKDict):
         return True
 
     def _agent_receive(self, msg):
+        def _default_unbound(msg):
+            """Received msg unbound to op"""
+            if j := msg.content.get("computeJid"):
+                # SECURITY: assert agent can access to this uid
+                if job.split_jid(j).uid == self.uid:
+                    job_supervisor.agent_receive(msg.content)
+                else:
+                    pkdlog(
+                        "{} jid={} not for this uid={}; msg={}", self, j, self.uid, msg
+                    )
+            else:
+                pkdlog(
+                    "{} missing computeJid, ignoring protocol error; msg={}", self, msg
+                )
+
         def _error(content):
             if "error" in content:
                 pkdlog("{} agent error msg={}", self, c)
@@ -288,51 +306,63 @@ class DriverBase(PKDict):
             pkdlog("{} no 'reply' in msg={}", self, c)
             return "invalid message from job_agent"
 
-        c = msg.content
-        i = c.get("opId")
-        if (
-            "opName" not in c
-            or c.opName == job.OP_ERROR
-            or ("reply" in c and c.reply.get("state") == job.ERROR)
-        ):
-            # Log all errors
-            pkdlog("{} error msg={}", self, c)
-        else:
-            pkdlog("{} opName={} o={:.4}", self, c.opName, i)
-        if i:
-            if "reply" not in c:
-                c.reply = PKDict(state=job.ERROR, error=_error(c))
-            if i in self._prepared_sends:
+        def _log(content, op_id):
+            if (
+                "opName" not in content
+                or content.opName == job.OP_ERROR
+                or ("reply" in content and content.reply.get("state") == job.ERROR)
+            ):
+                # Log all errors, even without op_id
+                pkdlog("{} error msg={}", self, content)
+            else:
+                pkdlog("{} opName={} o={:.4}", self, content.opName, op_id)
+
+        def _reply(content, op_id):
+            if "reply" not in content:
+                # A protocol error but pass the state on
+                content.reply = PKDict(state=job.ERROR, error=_error(content))
+            if op_id in self._prepared_sends:
                 # SECURITY: only ops known to this driver can be replied to
-                self._prepared_sends[i].reply_put(c.opName, c.reply)
+                self._prepared_sends[i].reply_put(content.reply)
             else:
                 pkdlog(
-                    "{} not in prepared_sends opName={} o={:.4} content={}",
+                    "{} op not in prepared_sends opName={} o={:.4} content={}",
                     self,
-                    c.opName,
-                    i,
-                    c,
+                    content.opName,
+                    op_id,
+                    content,
                 )
+
+        c = msg.content
+        i = c.get("opId")
+        _log(c, i)
+        if i:
+            _reply(c, i)
         else:
-            # TODO(robnagler) probably fine but maybe better to validate
-            getattr(self, "_agent_receive_" + c.opName, self._agent_receive_supervisor)(
-                msg
-            )
+            getattr(self, "_agent_receive_" + c.opName, _default_unbound)(msg)
 
     def _agent_receive_alive(self, msg):
         """Receive an ALIVE message from our agent
 
         Save the websocket and register self with the websocket
         """
-        self._websocket_ready_timeout_cancel()
-        if self._websocket:
+
+        def _ignore():
             if self._websocket != msg.handler:
-                self._websocket = None
-                raise AssertionError(
-                    pkdformat("unexpected incoming msg.content={}", msg.content)
-                )
-        else:
-            self._websocket = msg.handler
+                pkdlog("{} reconnected to new websocket, closing old", self)
+                self._websocket_close()
+                return False
+            if self._websocket_ready.is_set():
+                # extra alive message is fine
+                return True
+            # TODO(robnagler) does this happen?
+            pkdlog("{} websocket already set but not ready", self)
+            return False
+
+        self._websocket_ready_timeout_cancel()
+        if self._websocket and _ignore():
+            return
+        self._websocket = msg.handler
         self._websocket_ready.set()
         self._websocket.sr_driver_set(self)
         self._start_idle_timeout()
@@ -340,17 +370,6 @@ class DriverBase(PKDict):
     def _agent_receive_job_cmd_stderr(self, msg):
         """Log stderr from job_cmd"""
         pkdlog("{} stderr from job_cmd msg={}", self, msg.get("content"))
-
-    def _agent_receive_supervisor(self, msg):
-        """Received an error not bound to an op"""
-        if j := msg.content.get("computeJid"):
-            # SECURITY: assert agent can access to this uid
-            if job.split_jid(j).uid == self.uid:
-                job_supervisor.agent_receive(msg.content)
-            else:
-                pkdlog("{} jid={} not for this uid={}; msg={}", self, j, self.uid, msg)
-        else:
-            pkdlog("{} msg={}", self, msg)
 
     async def _agent_start(self, op):
         if self._websocket_ready_timeout:
--- a/sirepo/job_driver/__init__.py
+++ b/sirepo/job_driver/__init__.py
@@ -175,12 +175,21 @@ class DriverBase(PKDict):
         pkdlog("unknown agent, sending kill; msg={}", msg)
         try:
             msg.handler.write_message(PKDict(opName=job.OP_KILL))
+        except tornado.websocket.WebSocketClosedError:
+            pkdlog("websocket closed {} from unknown agent", self)
         except Exception as e:
             pkdlog("error={} stack={}", e, pkdexc())
 
     def send(self, op):
         pkdlog("{} {} runDir={}", self, op, op.msg.get("runDir"))
-        self._websocket.write_message(pkjson.dump_bytes(op.msg))
+        try:
+            self._websocket.write_message(pkjson.dump_bytes(op.msg))
+            return True
+        except tornado.websocket.WebSocketClosedError:
+            pkdlog("websocket closed op={}", op)
+        except Exception as e:
+            pkdlog("error={} op={} stack={}", e, op, pkdexc())
+        return False
 
     @classmethod
     async def terminate(cls):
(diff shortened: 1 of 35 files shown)


robnagler:
GitHub Actions speed up with max_procs=4 is 3x (8 vs 25 minutes). The docker pull, pip install, fmt, etc. take 2.5 minutes so the speed up is actually linear. I'm going to add SIREPO_MPI_CORES=2, because I think this will test the code better.

robnagler:
@e-carlin I'm still testing. Good to get started on the review now, though.

robnagler:
@e-carlin ready for a review. Tests pass, and seems to work on NERSC. I've done a lot more testing of NERSC than local. I didn't test docker, but I don't think I modified that.

sirepo/job_driver/__init__.py
```diff
@@ -308,16 +327,30 @@ def _agent_receive_alive(self, msg):
...
         self._start_idle_timeout()
 
-    def _agent_receive_error(self, msg):
-        # TODO(robnagler) what does this mean? Just a way of logging? Document this.
-        pkdlog("{} msg={}", self, msg)
+    def _agent_receive_job_cmd_stderr(self, msg):
+        """Log stderr from job_cmd"""
+        pkdlog("{} stderr from job_cmd msg={}", self, msg.get("content"))
+
+    def _agent_receive_supervisor(self, msg):
```
e-carlin, line 344: Can you explain this name? From my reading something like "_agent_receive_unknown_op" is maybe a bit clearer. I think more of a doc would be helpful too. One has to really know our system to know what's going on. I'm struggling to imagine a scenario where we want to act on something without an opId.

sirepo/job_driver/__init__.py
```diff
@@ -308,16 +327,30 @@ def _agent_receive_alive(self, msg):
         self._websocket_ready_timeout_cancel()
         if self._websocket:
             if self._websocket != msg.handler:
-                raise AssertionError(pkdformat("incoming msg.content={}", msg.content))
+                self._websocket = None
```
e-carlin, line 330: This is the second place you've added this. Maybe we add a with that cleans up websocket?

e-carlin (changes requested):
I'm working my way through reviewing. Probably another day. I left some initial comments.

Some quirks I noticed
- If I'm running a sim (doesn't need to be under sbatch) and I kill -9 it from the terminal I the GUI reports it as canceled. Seems like it should be error
- If I kill -9 an agent (again doesn't need to be under sbatch) then the gui continues to report "running: awaiting output". Even after refresh.

sirepo/job_driver/__init__.py
```diff
@@ -194,6 +203,7 @@ async def terminate(cls):
 
     def websocket_on_close(self):
         pkdlog("{}", self)
+        self._websocket = None
```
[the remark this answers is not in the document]
robnagler: The next line starts free resources, which is async. The _websocket is invalid at the point of websocket_on_close so it has to be invalided in case other functions try to use it.

robnagler:
Added to https://github.com/radiasoft/sirepo/issues/7406#issuecomment-2564392776.

sirepo/job_driver/__init__.py
robnagler: A "with" is not possible. This is a callback from tornado. I refactored in 6e0f1ee, lmk.

sirepo/job_driver/__init__.py
robnagler: refactored and changed semantics in 6e0f1ee

sirepo/job_supervisor.py
```diff
@@ -1209,27 +1342,27 @@ async def prepare_send(self):
...
-
         if (r := await self._reply_q.get()) is None:
             pkdlog("{} no reply", self)
             return None
         self._reply_q.task_done()
         return r
 
-    def reply_put(self, reply):
-        self._reply_q.put_nowait(reply)
+    def reply_put(self, reply_op_name, reply_content):
```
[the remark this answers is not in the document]
robnagler: 6e0f1ee

e-carlin (changes requested):
Two errors while running simulations:

- openmc > aurora > wait for volume extraction > visualization > vagrant cluster > login > start > error: [No such file or directory]: open('/var/tmp/vagrant/sirepo/user/ZSLW4c4Y/openmc/ZSLW4c4Y-VqEsWQZE-openmcAnimation/in.json', 'r')
- flash > blast2 > run setup and compile > visualization > vagrant cluster > login > start > error: /home/vagrant/.pyenv/versions/py3/bin/python: can't open file '/var/tmp/vagrant/sirepo/user/6NiqqZff/flash/6NiqqZff-ykM8ISjL-animation/parameters.py': [Errno 2] No such file or directory

sirepo/job_supervisor.py
```diff
@@ -861,31 +956,98 @@ async def _receive_api_runSimulation(self, req, recursing=False):
...
+                job.OP_RUN,
+                req,
+                is_run_status_op=True,
+                jobCmd=job.CMD_COMPUTE,
+                nextRequestSeconds=self.db.nextRequestSeconds,
+            )
         except Exception as e:
-            if o:
-                o.destroy(internal_error=f"_receive_api_runSimulation exception={e}")
+            if self.db.dbUpdateTime == d:
```
e-carlin, line 1016: I'm concerned about this code. If a job needs to wait on another simulation to complete (op_slot.alloc or cpu_slot.alloc) the dbUpdateTime is updated (_db_write_file is called when setting job situation). If it then raises at some point in the send I think we want the status to be set to job.ERROR but it won't be because the conditional is not True. I don't exactly know what causes a raise in send so maybe it is unlikely enough to not be worth covering. I think the job will be "missing" which is probably ok. 

A larger concern is that the code could be changed in some way in the future to where the dbupdatetime is modified during the normal course of events. It is always modified during _db_update which seems like a change someone could easily introduce. I don't think it will be obvious to the person changing the code that this case is now broken. Not sure how to protect other than possibly using some other field besides dbUpdateTime for the check.

sirepo/pkcli/job_agent.py
```diff
@@ -408,11 +399,21 @@ async def _fastcgi_op(self, msg):
             )
             # last thing, because of await: start fastcgi process
             await self._cmd(m, send_reply=False)
+            if not self._fastcgi_msg_q:
+                return self.format_op(
+                    msg,
+                    job.ERROR,
+                    PKDict(state=job.ERROR, error="fastcgi process got an error"),
```
e-carlin, line 406: **PKDict? format_op only has 2 positional args.

sirepo/pkcli/job_agent.py
```diff
@@ -1011,6 +888,517 @@ async def _read_stream(self):
...
+                return "FAILED"
+            # sacct outputs state for each part of the job (shifter, external, etc.) so be pessimistic.
+            rv = set()
+            for l in re.split(r"\s+", p.stdout):
+                if len(l) and not l.startswith("-") and l != "State":
+                    rv.add(l)
+            if len(rv) == 1:
+                # Normal case
+                return next(iter(rv))
+            elif len(rv) > 1 and "CANCELLED" in rv:
```
e-carlin, line 1303: this can just be an if

sirepo/pkcli/job_agent.py
```diff
@@ -1011,6 +888,517 @@ async def _read_stream(self):
...
+                    "{} sacct error exit={} sbatch={} stderr={} stdout={}",
+                    self,
+                    p.returncode,
+                    self._sbatch_status.sbatch_id,
+                    p.stderr,
+                    p.stdout,
+                )
+                if "disabled" in p.stderr:
+                    # Only in dev: saccount is not configured and job not running, assume canceled
+                    return "CANCELLED"
```
e-carlin, line 1292: One `L` in canceled.

sirepo/pkcli/job_agent.py
```diff
@@ -230,27 +259,43 @@ async def loop(self):
...
+            if c.jid == jid:
+                c.destroy()
+
     async def send(self, msg):
         if not self._websocket:
             return False
         try:
-            await self._websocket.write_message(msg)
+            if not isinstance(msg, _OpMsg):
+                raise AssertionError("expected _OpMsg type={} msg={}", type(msg), msg)
```
e-carlin, line 288: I think you want an f-string or pkdformat

e-carlin (changes requested):
I've reviewed everything. Just a few more comments.

The code works well. There are a lot of changes and a lot of cases so I'm sure there are some I didn't exercise.

sirepo/job_supervisor.py
robnagler: The larger concern is real, but it's the nature of distributed systems: collisions happen. I don't know a way around it.

In this particular case, it's important to realize that this is runSimulation which has control of the runDir, which is locked [until the reply happens](https://github.com/radiasoft/sirepo/blob/3d0638ee0836e6d568c51e1fb48fd29f1ec95f01/sirepo/job_supervisor.py#L1128). Anything that needs run_dir, will have to wait for the reply (or timeout). The only other op that touches the db is cancel, which takes priority over any errors, that is, if it modifies the db to set the state to canceled, then it's update is the correct one, and runSimulation should not update.

This is implicit coupling. The explicit coupling would be to have a separate db object (or process) that would manage the transaction which would only complete if there wasn't a simultaneous update. This is exactly what ACID does: ensures integrity of updates. The dbUpdateTime is a substitute for a row lock, which is what Postgres uses. We could, of course, introduce a lock on the db object. Locking is fraught if added after the fact. A deadlock is harder to debug than a row stomp. With a row stomp, you can see a sequence of events in a log and possibly figure out what's going on. With a deadlock, especially with asyncio with no ability to interrogate the state, it's very tricky to figure out what's going on. The tendency in the last few decades has been to drop locking and instead to blow the transaction if there's a collision. 

Again, the code is problematic, but a larger refactoring is required: separate out the db and separate out the messaging. The job_supervisor could also be a state machine.

sirepo/pkcli/job_agent.py
robnagler: Two L's in this case, because it is the status value from SLURM.

sirepo/pkcli/job_agent.py
robnagler: d0649f2ee8e3ca5d0a94eabf9ba7ec8772cb40d9

sirepo/pkcli/job_agent.py
robnagler: d0649f2ee8e3ca5d0a94eabf9ba7ec8772cb40d9

sirepo/pkcli/job_agent.py
robnagler: d0649f2ee8e3ca5d0a94eabf9ba7ec8772cb40d9

robnagler:
> I've reviewed everything. Just a few more comments.

Thank you. I know it was a lot and very complicated.

> The code works well. There are a lot of changes and a lot of cases so I'm sure there are some I didn't exercise.

I appreciate the testing.

e-carlin (changes requested):
I ran into the same openmc error

https://github.com/radiasoft/sirepo/pull/7404#pullrequestreview-2525972862

robnagler:
Fix #7404 had to write in.json. Refactored that code. openmc works now. I didn't test flash.

e-carlin (approved):
Openmc and flash work.

I can't create a reproducible example but this has happened to me 3 times: The supervisor can get in a state where it no longer responds to SIGTERM. 

```
~$ ps uww | grep job_supervisor
vagrant   426434  6.5  2.0 596624 164596 pts/2   Sl+  16:41   0:32 /home/vagrant/.pyenv/versions/3.9.15/envs/py3/bin/python /home/vagrant/.pyenv/versions/py3/bin/sirepo job_supervisor
~$ kill -SIGTERM 426434
~$ ps uww | grep job_supervisor
vagrant   426434  6.5  2.0 596624 164596 pts/2   Sl+  16:41   0:33 /home/vagrant/.pyenv/versions/3.9.15/envs/py3/bin/python /home/vagrant/.pyenv/versions/py3/bin/sirepo job_supervisor
```

When I send the SIGTERM to the supervisor it logs
```
Jan 06 16:49:33 426434     0 sirepo/job_driver/local.py:87:kill LocalDriver(a=foCR k=sequential u=d2dF []) pid=427224
```

The closest I can get to a reproducible example is it seem to only happen when the supervisor is signaled while a job is running.

robnagler:
created https://github.com/radiasoft/sirepo/issues/7416

robnagler:
Should we merge?

e-carlin:
Good with me!

PR #7428, opened 2025-01-20
fix #7279 full support for PNNL material database

### LINKED ISSUE #7279: expand PNNL materials
We currently have a limited number of "standard materials" available.
Untitled

This should be expanded to include a search mechanism for all the the PNNL materials:
https://www.pnnl.gov/main/publications/external/technical_reports/PNNL-15870Rev2.pdf

If possible, use this library:
https://github.com/fusion-energy/neutronics_material_maker/

moellep:
Added a type-ahead lookup list option when defining materials. 

A condensed hdf5 file can be created from the full PNNL json file using pkcli.openmc.create_standard_materials_file()

The hdf5 file is stored in the cloudmc data repo sirepo-data-cloudmc.

During geometry initialization, the PNNL cache is checked and the file is retrieved from the data repo if needed.


e-carlin (changes requested):
How do I run this? I tried opening the Aurora example and changing the material for copper to "Steel, Stainless 202". I then ran the sim and got this error
```
Could not find nuclide C12 in the nuclear data library.
```

moellep:
Good point - the PNNL references nuclides which aren't present in the older ENDFB-7.0 source. I've updated the examples to use the more recent version ENDFB-8.0. You should be able to "discard changes to example" and then use the material lookup.

e-carlin (approved):
lgtm

Issue #6535, opened 2023-11-17
don't save NaN value to sirepo data


moellep:
In this SRW example (running on dev), the beamline report will produce a bad result for the final plot. The client is then storing NaN values in the min/max IntensityLimit fields, which will then crash when subsequent simulation saves ares attempted.

https://www.sirepo.com/srw#/source/IzlnEjjw

This would be a possible work-around, in srw.js:
```javascript
    if (m && 'useIntensityLimits' in m && m.useIntensityLimits == '0' && ! isNaN(info.fieldIntensityRange[0])) {
```

robnagler:
@moellep is this what the customer is seeing?

PR #7434, opened 2025-01-28
fix #7433 ensure computed intensity limit return a valid number

### LINKED ISSUE #7433: Intensity limit calculation can return NaN values
The intensity limit calculation in Sirepo can return NaN values which may be added to the simulation data on the client. When the client tries to save the new data, a server error occurs because the json can't be serialized. No error message is shown in the UI indicating the save failed.

moellep:
- updated autoSave to display an alert if an error occurs during save
 - protect against NaN intensity ranges on the client (for cached plot data) and on the server


moellep:
This PR also fixes #6535

PR #7451, opened 2025-02-12
fix #7449 reset jobRunMode if not present in application

### LINKED ISSUE #7449: Exported production sims on dev may show "login to undefined"
If you have a sim running on nersc and export to a dev system, the sim won't allow running with a "login to undefined".

moellep:


sirepo/job_supervisor.py
```diff
@@ -1078,6 +1078,9 @@ async def _receive_api_statelessCompute(self, req):
         return await self._send_op_analysis(req, "stateless_compute")
 
     def _is_sbatch_login_ok(self, req):
+        r = req.content.get("jobRunMode", self.db.jobRunMode)
+        if r not in sirepo.simulation_db.JOB_RUN_MODE_MAP:
```
e-carlin, line 1082: We have a similar check in [job_api.py](https://github.com/radiasoft/sirepo/blob/c35b955781b998dd8073b33803e1e3da69293269/sirepo/job_api.py#L367). Can the fix be put there?

e-carlin (changes requested):

sirepo/job_supervisor.py
moellep: I moved that code to an assertion in job_api. It shouldn't get triggered because the fix on the client makes sure it doesn't call sbatch related methods if it is not enabled.

e-carlin (approved):

PR #7456, opened 2025-02-17
Fix #7455: reword email login text

### LINKED ISSUE #7455: Registration message is too subtle
At the top of the form, have it say something like:

`Enter your institutional email address. Gmail.com, Yahoo.com, Outlook.com, etc. will be rejected.`

At the bottom:

`When you click continue, we'll send an authorization link to your inbox.`

If registering, we should check for gmail/yahoo/outlook addresses and immediately reject them. The list should be configurable.

@bruhwiler do you agree?

e-carlin:
Screenshot 2025-02-17 at 11 29 03 AM


bruhwiler:
Looks good to me.

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3621,7 +3621,10 @@ SIREPO.app.directive('emailLogin', function(requestSender, errorService) {
             $scope.login = function() {
                 var e = $scope.data.email;
                 errorService.alertText('');
-                if (! ( e && e.match(/^.+@.+\..+$/) )) {
+                if (
+                  ! (e && e.match(/^.+@.+\..+$/)) ||
+                  e.match(/@(gmail|yahoo|outlook)\.com$/i)
```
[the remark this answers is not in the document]
e-carlin: I'm not sure what you meant by "add parens to the first regexp" but I did this 5e7af2f33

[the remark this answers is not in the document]
e-carlin: https://github.com/radiasoft/sirepo/pull/7456/commits/589d8d9529c337d73b88c5beec4c0740e9aa6b33

e-carlin:

sirepo/package_data/static/js/sirepo-components.js
```diff
@@ -3600,13 +3600,22 @@ SIREPO.app.directive('emailLogin', function(requestSender, errorService) {
             </div>
         `,
         controller: function($scope) {
+            function showInvalidEmail() {
+                $scope.showWarning = true;
+                $scope.warningText = 'Email address is invalid. Please update and resubmit.';
```
robnagler, line 3605: I think, in this case, we should say "please use your institutional email, we do not allow ${domain} addresses". This could be generated server side.

sirepo/auth/email.py
```diff
@@ -173,3 +185,14 @@ def avatar_uri(qcall, model, size):
         hashlib.md5(pykern.pkcompat.to_bytes(model.user_name)).hexdigest(),
         size,
     )
+
+
+def init_apis(*args, **kwargs):
+    global _cfg
+    _cfg = pkconfig.init(
+        deny_access_domains=(
+            {"gmail.com", "yahoo.com", "outlook.com"},
```
robnagler, line 194: I don't think we should have any addresses by default. Also, `set` is sufficient for the parser. It always creates a `frozenset`.

sirepo/auth/email.py
```diff
@@ -147,6 +152,13 @@ def _send_smtp(user_data, uri):
         m = self.auth_db.model(UserModel)
         u = m.unchecked_search_by(unverified_email=email)
         if not u:
+            # POSIT: Email has been validated by _parse_email so simple split is ok
+            if email.split("@")[1] in _cfg.deny_access_domains:
```
robnagler, line 156: Make the conditional check a function that returns the email: `n.new(unverified_email=_assert_allow(email))`. 

(My pedantic self would say that deny_access_domains should be checked for empty, but since this is a low cost split, I will only comment parenthetically. :)

robnagler (changes requested):

sirepo/package_data/static/js/sirepo-components.js
e-carlin: 34cc6ad44

sirepo/auth/email.py
e-carlin: 34cc6ad44

sirepo/auth/email.py
e-carlin: 34cc6ad44

PR #7493, opened 2025-03-21
Fix #7487 remove predefined.json and simplify code

### LINKED ISSUE #7487: srw predefined.json & model_defaults_test and the general problem
modeL_defaults_test failed in a parallel test because predefined.json didn't exist. This probably was because the file gets deleted by another test. It's using resource_path which goes to sirepo/package_data.

The general problem is that predefined.json is an awkward file. It's dynamic and static. It gets built with the package and dynamically created in sim_data. There's tones of special code to handles this one file.

It takes less than a second to create on my VM. @moellep could we solve the problem differently?

robnagler:
@moellep when I removed create_predefined, I noticed the only thing that was used was beams. I suspect you ran into problems in job cmd and add the following, is that correct?

```py
_USER_MODEL_LIST_FILENAME = PKDict(
    electronBeam="_user_beam_list.json",
    tabulatedUndulator="_user_undulator_list.json",
)
```


moellep (approved):
I tested it and it looks good. I think the _USER_MODEL_LIST_FILENAME predates job cmd.

PR #7525, opened 2025-04-10
Fix #7524 Try resolving by email if uid doesn't work

### LINKED ISSUE #7524: In pkcli roles, lookup by 'email' if uid lookup fails
With ldap authentication, the 'email' field may not actually be a well-formed email (ie just a username without an '@example.com'). Looking up by the email field if a uid lookup fails can be handy for such situations.

himanshugoel2797:


robnagler:
See #7528

PR #7528, opened 2025-04-14
Fix #7522 Switch to ldap3 for auth.ldap

### LINKED ISSUE #7522: Make LDAP auth more flexible
As discussed in Slack, we use LDAP authentication in our internal Sirepo install. Our systems expect a different DN format, with an empty suffix and a different prefix, so we make the prefix configurable similarly to the suffix. 
The python-ldap package is also not documented in as much detail as the ldap3 package and the latter provides wheels, thus not requiring openldap-devel to be installed via the system package manager.

robnagler:


robnagler:
@himanshugoel2797 please try this branch. You will need to set SIREPO_AUTH_LDAP_EMAIL_DOMAIN to bnl.gov or something that creates a valid email address.  #7525 was only necessary, because #7523 put invalid email addresses in the db. The test passes now. It was faster for me to make a new PR.

himanshugoel2797:
@robnagler it seems to work well! I can log in (Duo pushes from the LDAP server included), and can assign roles by using "username@bnl.gov", as expected.

robnagler:
Thanks for testing @himanshugoel2797. This is on alpha. It will be pushed possibly this week.

PR #7548, opened 2025-04-24
fix #7545 no moderation required for non-email authentication methods

### LINKED ISSUE #7545: Make moderation configurable
Moderation should be a configurable feature. If it is not enabled, then all incoming users would be assumed to be authorized.

moellep:
Turning off SIREPO_FEATURE_CONFIG_USER_MODERATION will stop moderation requests.


moellep:
This PR should be reworked, removing the feature_config flag, and only enabling the user moderation workflow if email auth is used.

moellep:
I've updated the PR now so the moderation information is only collected for email authentication.

Also, on dev a new user automatically gets assigned all non-admin roles, regardless of authentication method used. Previously, this was only for the "guest" authentication on dev.

PR #7586, opened 2025-06-28
fix #7585 add feature_config is_moderated

### LINKED ISSUE #7585: isModerated as feature_config

moellep:
@robnagler I'm not sure if this is the best way to share the code between auth.login and auth_role_moderation.


robnagler (approved):
Needs more work, but this will work for now.

PR #7587, opened 2025-06-28
Fix #7579 cortex_sql_db and cortex_xlsx

### LINKED ISSUE #7579: cortex spreadsheet parser & sql db

robnagler:


sirepo/template/cortex_sql_db.py
```diff
@@ -0,0 +1,86 @@
...
+_BASE = "cortex.sqlite3"
+
+
+def insert_material(parsed, qcall=None):
+    def _values(cols, vals):
+        return PKDict((c, vals[c]) for c in cols if c in vals)
+
+    with _session(qcall=qcall) as s:
+        i = s.insert(
+            "material",
```
moellep, line 24: The insert may fail with an IntegrityError ```UNIQUE constraint failed: material.material_name``` if the same xlsx file is uploaded twice. It should return a better error message to the client in this case.

sirepo/template/cortex_sql_db.py
```diff
@@ -0,0 +1,86 @@
...
+                is_homogenized_hcpb=b,
+                is_homogenized_wcll=b,
+                is_neutron_source_dt=b,
+                is_plasma_facing=b,
+                neutron_wall_loading="str 32",
+            ),
+            material_component=PKDict(
+                material_component_id="primary_id 2",
+                material_id="primary_id",
+                material_component_name="str 8 unique",
```
moellep, line 80: material_component_name should be "unique" across the database, but only to that material_id.

If you upload a spreadsheet which shares components with another material, you get the error ```UNIQUE constraint failed: material_component.material_component_name```

sirepo/template/cortex_sql_db.py
```diff
@@ -0,0 +1,86 @@
...
+                ),
+            )
+
+
+@contextlib.contextmanager
+def _session(qcall):
+    s = sirepo.sim_data.get_class("cortex")
+    if w := (not (p := s.lib_file_exists(_BASE, qcall=qcall))):
+        p = pykern.pkio.py_path(_BASE)
+
```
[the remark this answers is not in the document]
robnagler: 926956a

sirepo/template/cortex_sql_db.py
robnagler: 926956a

moellep (approved):

PR #7639, opened 2025-08-20
fix #7638 implement Material View

### LINKED ISSUE #7638: preview material fields
After import, it should be possible to review the material values.

moellep:


sirepo/template/cortex.py
```diff
@@ -19,6 +21,18 @@ def stateful_compute_cortex_db(data, **kwargs):
         return PKDict(
             api_result=sirepo.template.cortex_sql_db.list_materials(),
         )
+    if data.args.api_name == "delete_material":
```
robnagler, line 24: Not thrilled with the use of stateful_compute. I think its redundant to say "cortex_db" and maybe call it "sql_db" so we can differentiate from the sirepo db (cortex_db is vague in that way).  Also, duplicating dispatch is problematic. Why not stateful_compute_material_detail, stateful_compute_delete_material, and stateful_compute_list_materials? This would save the dispatch logic and rely on errors handled in template_common.

sirepo/template/cortex_sql_db.py
```diff
@@ -20,6 +20,32 @@ class Error(RuntimeError):
     pass
 
 
+def delete_material(material_id):
+    """Cascade delete all rows for a material"""
+
+    def _delete(session, table_name, field, value):
+        session.delete(table_name, PKDict({field: value}))
```
robnagler, line 27: add uid for security.

sirepo/template/cortex_sql_db.py
```diff
@@ -178,3 +229,18 @@ def _optional(v):
             ),
         ),
     )
+
+
+@contextlib.contextmanager
+def _session(qcall):
+    s = sirepo.sim_data.get_class("cortex")
```
robnagler, line 236: Probably should just use SIM_DATA.

robnagler (changes requested):

sirepo/template/cortex_sql_db.py
moellep: This is tricky because the uid can't come from the client. And on the server it is running in a job_cmd, so there isn't a qcall to get the user info.

sirepo/template/cortex_sql_db.py
moellep: We could pass _SIM_DATA from the template.cortex to cortex_sql_db, but it would need to be on each call.

sirepo/template/cortex_sql_db.py
moellep: simulation_db.user_path() pass to uid_from_dir_name()

sirepo/template/cortex_sql_db.py
moellep: use same header as template.cortex to create _SIM_DATA

PR #7642, opened 2025-08-24
Fix #7641 sim_api.cortex

### LINKED ISSUE #7641: cortex db service
Implement as sim_api so that the uid happens in the api server. Use a thread to call do the work.

robnagler:
Moved all the code to sim_api, because it should be happening in the api server, not in the supervisor or the agent. The work involved is minimal. The database is still bound to sqlite, but can be changes. Needs changes to the export_tea to make that happen. uid is added to queries. Admin query can be added easily with new structure. Removed cruft in the msgrouter.js that was only necessary to support HTTP which Vue doesn't support.


robnagler:
@moellep ready for review.

PR #7956, opened 2026-07-09
fix #7802 retry with current cookies if first request is cookie redirect

### LINKED ISSUE #7802: direct link to sirepo/jupyter results in missing cookies
If you haven't visited sirepo before and go directly to sirepo/jupyter, the browser shows a missing cookies page.

Going directly to sirepo/jupyterhublogin works as expected.

moellep:


robnagler (approved):
Went through the logic and seems ok.

moellep:
It may fail oddly if a client has cookies disabled, but this is hard to test on dev. I'll try this on alpha. It is more important that the first visit works for uncookied users than the case where all cookies are disabled.

robnagler:
The only thing to worry about is a loop. I don't think that would happen, but good to test.

